PR #3, opened 2021-05-12
Update to use TPP fork of Circulation-Web

jonathangreen:
## Description

Update the code in the circulation manager to use the TPP fork of circulation-web. This fork has been pushed to NPM and renamed to circulation-admin. So the changes in this PR reflect both pulling in the new NPM package and renaming the files to reflect the new name of the repo.

PR #4, opened 2021-05-12
Update docker build tag order

jonathangreen:
## Description

When container images have multiple tags, github displays the one with the highest priority. This means that currently the `sha-####` tag is always the one displayed. This is a helpful tag, but I think others should be displayed more prominently. 

`crazy-max/ghaction-docker-meta` was moved into the docker org and renamed to `docker/metadata-action`. This updates to use the latest version from the docker org.

We change the tag order by adding the priority argument to our list of tags. It configures the tags to have the following priority:
- `major.minor`
- `major`
- `branch name`
- `sha-####`

You can see this in action by taking a look at the [image built for this PR](https://github.com/orgs/ThePalaceProject/packages/container/circ-webapp/2495969?tag=sha-b29b042) vs [the one built for PR #3](https://github.com/orgs/ThePalaceProject/packages/container/circ-webapp/2496118?tag=circulation-admin).

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -139,14 +139,14 @@ jobs:
 
       - name: Generate tags for circ-exec image
         id: exec-tags
-        uses: crazy-max/ghaction-docker-meta@v2
+        uses: docker/metadata-action@v3
         with:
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Looks good. Also, "docker/" > "crazy-max".

PR #5, opened 2021-05-12
Update to use tpp fork of core.

jonathangreen:
## Description

Update to use the TPP fork of the core submodule instead of the one from LYRASIS.


jonathangreen:
Closing this, since #3 ended up including it.

PR #7, opened 2021-05-12
Deploy docs to github pages.

jonathangreen:
## Description

New action to build docs, and deploy to gh-pages if we are on `main`.


jonathangreen:
#110 obviates the need for this PR

PR #8, opened 2021-05-13
Handle double escaped entity references.

tdilauro:
## Description

- Adds an extra XML entity dereference for contributor (author, etc.) values in Bibliotheca `item` response data and
- Adds a new test of contributor parsing that includes content containing special characters and an entity reference.

## Motivation and Context

`bibliotheca_format_sweep` is not running to completion.
- `ValueError: Cannot find sort name for a contributor with no display name!`

Items returned by the Bibliotheca CloudLibrary `items` API may have contributors with two-level deep entity references, so `<Authors>Raji Codell, Esmé</Authors>` may be transmitted as `<Authors>Raji Codell, Esm&amp;#233;</Authors>`, rather than `<Authors>Raji Codell, Esm&#233;</Authors>`. This resulted in incorrect parsing of contributor fields, which, in turn lead to:
- invalid contributor entries (existing ones will need remediation) and
- the aforementioned crashes of `bibliotheca_circulation_sweep`.

## How Has This Been Tested?

In addition to the regular tests, ran `bibliotheca_circulation_sweep` against an existing test collection and verified that:
- escaped strings were processed properly and
- the script was able to run to completion.

## Checklist:

### DIFF
--- a/tests/test_bibliotheca.py
+++ b/tests/test_bibliotheca.py
@@ -1246,6 +1246,16 @@ class TestItemListParser(BibliothecaAPITest):
             "Baum, Frank L. (Frank Lyell)")
         assert "Baum, Frank L." == author.sort_name
 
+        # Contributors may have two levels of entity reference escaping,
+        # one of which will have already been handled by the initial parse.
+        # So, we'll test zero and one escapings here.
+        authors = list(ItemListParser.contributors_from_string(
+            u'Raji Codell, Esmé; Raji Codell, Esm&#233;'))
+        author_names = [a.sort_name for a in authors]
+        assert len(authors) == 2
+        assert len(set(author_names)) == 1
+        assert all(u'Raji Codell, Esmé' == name for name in author_names)
+
         # It's possible to specify some role other than AUTHOR_ROLE.
         narrators = list(
             ItemListParser.contributors_from_string(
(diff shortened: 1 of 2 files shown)


tests/test_bibliotheca.py
```diff
@@ -1246,6 +1246,16 @@ def test_contributors_for_string(cls):
             "Baum, Frank L. (Frank Lyell)")
         assert "Baum, Frank L." == author.sort_name
 
+        # Contributors may have two levels of entity reference escaping,
+        # one of which will have already been handled by the initial parse.
+        # So, we'll test zero and one escapings here.
+        authors = list(ItemListParser.contributors_from_string(
+            u'Raji Codell, Esmé; Raji Codell, Esm&#233;'))
```
tdilauro, line 1253: In normal operation, `ItemListParser.contributors_from_string` receives Unicode strings in py2, regardless of what the surround tests imply.

jonathangreen (approved):
Looks good to me. Without looking at it deeply, it seems like a problem with the Bibliotheca API, but it makes sense to handle it here.

tdilauro:
Right. And while the extra de-ref isn't idempotent, per se, it does converge. One of the author names in the test is the already converged value to demonstrate that. So, if this does get fixed in the API, it should still be safe.

PR #9, opened 2021-05-18
Allow docker builds to be disabled.

jonathangreen:
## Description

Align the docker builds with the ones from https://github.com/ThePalaceProject/library-registry


tdilauro (approved):
Looks good!

PR #11, opened 2021-05-19
Python3

jonathangreen:
## Description

Pulling all the upstream Python 3 commits into a branch for testing. The submodule commit is updated to point to [python3 branch](https://github.com/ThePalaceProject/circulation-core/tree/python3 ) in circulation-core.


tdilauro:
I haven't actually reviewed this yet, but yes! Please and thank you! 🚀

PR #12, opened 2021-05-19
ODL collections can configure primary identifier preference

tdilauro:
## Description

Adds primary ID settings from `OPDSImporter` (currently only one) to ODL collections. 

Depends on [circulation-core PR #9](https://github.com/lyrasis/simplye-server-core/pull/9).

## Motivation and Context

`ODLImporter` is a subclass of `OPDSImporter`, but does not inherit its configuration settings because ODL configuration settings are managed by `ODLAPI`.

## How Has This Been Tested?

Ran core and circulation tests. Verified via admin API that the optional setting appears for ODL collections and that the correct value appears in the database when selected.

## Checklist:

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -38,6 +38,7 @@ from core.model import (
     LicensePool,
     Loan,
     MediaTypes,
+    Representation,
     RightsStatus,
     Session,
     get_one,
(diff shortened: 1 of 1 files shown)


jonathangreen:
@tdilauro this PR is getting pretty stale. Is it still needed or can it be closed?

PR #13, opened 2021-05-20
Add S3 analytics provider

vbessonov:
## Description

NOTE: Migrated from https://github.com/lyrasis/simplye-circulation/pull/16.

This PR adds an S3 analytics provider.

## Checklist:


jonathangreen:
@dbernstein @tdilauro since it looks like this PR was replaced by #422 can are we good to close this?

tdilauro:
@jonathangreen I believe that is correct and this code is old and would be hard to sync.

PR #16, opened 2021-05-27
Remove old RBDigital integration code.

jonathangreen:
## Description

Remove RBDigital integration because all their collections have been moved over to Overdrive, so this code is no longer used.

Depends on: 
https://github.com/ThePalaceProject/circulation-core/pull/10

🚨  Before this is merged, there needs to be a database migration written to remove all of the RBD content and collections that this update will orphan. 

## Motivation and Context

The RBD code is no longer used, so its worth removing from the CM.

## How Has This Been Tested?

Run tests against it, and tried out the admin interface.


vbessonov (approved):

tdilauro (approved):
Looks good.

migration/20210622-remove-rbd.sql
```diff
@@ -0,0 +1,44 @@
+BEGIN;
```
tdilauro, line 1: This seems like it should work, but the convention for CM and core thus far has been to wrap transactions inside a DO block as a dollar-quoted literal, like:
```
DO $$
  BEGIN
    ...
  END;
$$;

jonathangreen: Interesting. I didn't notice that was the convention before. 

I've never seen a SQL `DO` statement before TBH. It must be something PG specific.

jonathangreen: Neat, the postgres `DO` block is an SQL extension that lets you execute a function in a procedural language. [Docs are here](https://www.postgresql.org/docs/9.6/sql-do.html). Good to know for the future that you can actually write a script to make some updates. Thanks for pointing that out! 

Looked through the code that runs the updates and it looks like it automatically wraps everything in a transaction anyway, so the transaction I had as part of this script was redundant. I removed it so the update is following the convention the CM was using.

tdilauro (approved):
This looks good.

PR #17, opened 2021-05-27
Modify tox config

jonathangreen:
## Description

Change the ports and container names in the tox config to allow both the circulation and core to run tox on the same machine at the same time. 

Same changes implemented in core:
https://github.com/ThePalaceProject/circulation-core/pull/11

Also make a couple more small changes to the default tox config:
- Pass the CI environment variable though from tox into the testing process. Pytest looks to see if this environment variable is set and doesn't truncate error output if it is. Github actions sets the CI env variable by default.
- Remove the `-v` flag from the test output, since we don't need verbose output, it just makes the test logs in github actions hard to read.

## Motivation and Context

This is helpful in development, so you can be running both sets of tests simultaneously.

## How Has This Been Tested?

Ran tox locally against circ and core at the same time.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,23 +5,23 @@ skipsdist = true
 [testenv]
 deps = -r requirements-dev.txt
 commands_pre =
-    docker: docker exec es elasticsearch-plugin -s install analysis-icu
-    docker: docker restart es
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
LGTM! 🎷

PR #18, opened 2021-06-15
[WIP] Add quick workarounds for the issues found in the Columbia's ONIX files

vbessonov:
## Description

This PR contains quick and dirty hacks to fix the issues found in the Columbia's ONIX files. I'm not willing to merge it 

## Checklist:

### DIFF
--- a/api/onix.py
+++ b/api/onix.py
@@ -222,7 +222,10 @@ class ONIXExtractor(object):
             publishing_date = parser.text_of_optional_subtag(record, 'publishingdetail/publishingdate/b306')
             issued = None
             if publishing_date:
-                issued = datetime.datetime.strptime(publishing_date, "%Y%m%d")
+                try:
+                    issued = datetime.datetime.strptime(publishing_date, "%Y%m%d")
+                except ValueError:
+                    issued = datetime.datetime.strptime(publishing_date, "%Y")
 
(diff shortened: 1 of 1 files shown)


api/onix.py
```diff
@@ -222,7 +222,10 @@ def parse(cls, file, data_source_name, default_medium=None):
             publishing_date = parser.text_of_optional_subtag(record, 'publishingdetail/publishingdate/b306')
             issued = None
             if publishing_date:
-                issued = datetime.datetime.strptime(publishing_date, "%Y%m%d")
+                try:
+                    issued = datetime.datetime.strptime(publishing_date, "%Y%m%d")
+                except ValueError:
+                    issued = datetime.datetime.strptime(publishing_date, "%Y")
```
tdilauro, line 228: Since ONIX has a variety of supported date formats and we're already being somewhat imprecise, perhaps we could use a more generalized approach that will handle more possibilities:
```
from dateutil.parser import isoparse
...
publishing_date = parser.text_of_optional_subtag(record, 'publishingdetail/publishingdate/b306')
issued = isoparse(publishing_date) if publishing_date else None

tdilauro (approved):
This looks good and should not break parsing of data that we have successfully processed in the past.

PR #15, opened 2021-05-21
Pull in admin UI from CDN

jonathangreen:
## Description

ADR describing these changes here:
https://www.notion.so/lyrasis/Distribute-admin-UI-with-jsdelivr-417a5accc475480788e90fb74b9bbe18

## How Has This Been Tested?

I tested it with and without the files locally running `python app.py`. Once the docker containers build from this PR I will test that it works in a "production" build.

PR #20, opened 2021-07-09
Sync upstream changes

tdilauro:
## Description

- Pulls in upstream changes from NYPL's python3 branch.
- Minor updates to README for Palace (much more to do here).
- Update to [release 0.0.3](https://github.com/ThePalaceProject/circulation-admin/releases/tag/0.0.3) of `circulation-admin`.

## How Has This Been Tested?

- Manual testing of database migration.
- Spot testing of admin interface (not exhaustive).
- Ran all tests.

## Checklist:


api/admin/package-lock.json
```diff
@@ -1,14 +1,13 @@
 {
   "name": "circulation",
```
jonathangreen, line 2: We should be able to totally drop this file. We don't need a `package-lock.json` checked in here now that we are pulling the admin interface in from a CDN.

api/admin/package.json
```diff
@@ -12,6 +12,6 @@
     "npm": ">=7.0.0"
```
jonathangreen, line 12: This file can be dropped as well.

api/firstbook.py
```diff
@@ -94,12 +94,13 @@ def remote_pin_test(self, barcode, pin):
                 str(e),
                 self.NAME
             )
+        content = response.content.decode("utf8")
```
jonathangreen, line 97: Its really an upstream issue, so I'm fine merging as is, but I think we should really be using `response.text` here instead of just assuming `utf8`.

https://docs.python-requests.org/en/master/user/quickstart/#response-content

api/bibliotheca.py
```diff
@@ -1144,6 +1145,10 @@ class DateResponseParser(BibliothecaParser):
 
     def process_all(self, string):
         parser = etree.XMLParser()
+        # If the data is an HTTP response, it is a bytestring and
+        # must be converted before it is parsed.
+        if isinstance(string, bytes):
+            string = string.decode("utf-8")
```
jonathangreen, line 1151: This makes me a bit queasy, but again its an upstream issue, so we can merge as is, then deal with it later. I think we should be figuring out where this is being called with a http response and instead using `response.text` instead of just doing a blind decode here.

https://docs.python-requests.org/en/master/user/quickstart/#response-content

api/firstbook2.py
```diff
@@ -100,12 +100,13 @@ def remote_pin_test(self, barcode, pin):
                 str(e),
                 self.NAME
             )
+        content = response.content.decode("utf8")
```
jonathangreen, line 103: Same `response.text` comment.

api/millenium_patron.py
```diff
@@ -433,6 +433,8 @@ def patron_dump_to_patrondata(self, current_identifier, content):
 
     def _extract_text_nodes(self, content):
         """Parse the HTML representations sent by the Millenium Patron API."""
+        if isinstance(content, bytes):
+            content = content.decode("utf8")
```
jonathangreen, line 437: Same `response.text` comment 😱

jonathangreen (commented):
This looks good to me. Only change I would like to see before merging is to drop the `package*.json` files. The other bits response encoding are minor and can be left for future work.

api/admin/package-lock.json
tdilauro: I can't do that yet because the [PR for pulling from CDN](https://github.com/ThePalaceProject/circulation/pull/15) has not yet been merged.

api/bibliotheca.py
tdilauro: I agree that these will need to be addressed eventually. You and I discussed this (in a discussion on the related `circulation-core` PR, I think) and we decided that, in order to make merging from upstream easier, we would not try to address these piecemeal in the short tem unless we ran into a specific problem.

api/millenium_patron.py
tdilauro: I feel your pain.

api/admin/package-lock.json
jonathangreen: 🤦 My bad. I was sure that went in already. I guess I was thinking of the CDN PR for library-registry.

api/admin/package.json
jonathangreen: also a 🤦

api/bibliotheca.py
jonathangreen: Agreed. Was just making notes as I went through the PR. Don't think any of the encoding issues need to be dealt with here.

jonathangreen (approved):
Looks good 👍🏻

PR #21, opened 2021-07-12
Fix checkout of unlimited-access books from collections without a CirculationAPI implementation

vbessonov:
## Description

This PR fixes checkout of unlimited-access books from collections without a CirculationAPI implementation.

## Motivation and Context

Circulation Manager cannot checkout and fulfill unlimited-access books that belong to collections with CirculationAPI implementation. This PR fixes this issue.

## Checklist:

- [] I have updated the documentation accordingly.


api/circulation.py
```diff
@@ -928,7 +933,9 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, part=None, fulfi
             if licensepool.self_hosted:
                 fulfillment = self._try_to_sign_fulfillment_link(licensepool, fulfillment)
         else:
-            api = self.api_for_license_pool(licensepool)
+            if not api:
+                raise NoLicenses()
```
tdilauro, line 937: `NoLicenses` makes sense in the loan/borrow context, but it doesn't seem a good fit for fulfillment. Is `NoLicenses` handled anywhere in `Circulation.fulfillment` call stacks?

vbessonov: Good catch! There are a [couple of different exceptions](https://github.com/ThePalaceProject/circulation/blob/main/api/controller.py#L1639) supported by Circulation Manager:
```python
        except DeliveryMechanismConflict, e:
            return DELIVERY_CONFLICT.detailed(e.message)
        except NoActiveLoan, e:
            return NO_ACTIVE_LOAN.detailed(
                    _('Can\'t fulfill loan because you have no active loan for this book.'),
                    status_code=e.status_code
            )
        except CannotFulfill, e:
            return CANNOT_FULFILL.with_debug(
                unicode(e), status_code=e.status_code
            )
        except FormatNotAvailable, e:
            return NO_ACCEPTABLE_FORMAT.with_debug(
                unicode(e), status_code=e.status_code
            )
        except DeliveryMechanismError, e:
            return BAD_DELIVERY_MECHANISM.with_debug(
                unicode(e), status_code=e.status_code
            )
```

I think we can use `NoActiveLoan`. What do you think?

tdilauro: I don't think we should raise `NoActiveLoan` unless there really is not one. Otherwise, it would be confusing to the patron when they can see a loan on their bookshelf, but get a message that says otherwise. Presumably we were able to get to `Circulation.fulfill` because there is a loan. If we have a loan and are unable to fulfill, it would probably be better to fall back to `CannotFulfill`.

tdilauro (approved):
Looks good to me! 🚀 🎸

PR #23, opened 2021-08-10
Bugfix/onix corporate contributor

tdilauro:
## Description

- Adds support for contributors with corporate names.
- Generalizes pub date parsing support to any ISO 8601 date.

## Motivation and Context

- Some imports were failing because they had corporate contributors.
- Generalized ISO 8601 parsing simplifies date handling and should cover more cases until we introduce more format specific handling.

## How Has This Been Tested?

- Added new tests for contributors with corporate names.
- Ran existing tests.
- Successfully imported the previously failing content.

## Checklist:

- [N/A] I have updated the documentation accordingly.

### DIFF
--- a/api/onix.py
+++ b/api/onix.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 
+import dateutil.parser
 from enum import Enum
 from lxml import etree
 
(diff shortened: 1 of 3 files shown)


vbessonov (approved):
It looks good!

PR #24, opened 2021-08-12
Force DirectoryImport._locate_file read books and covers as binary files

vbessonov:
## Description

This PR adds changes forcing `DirectoryImport._locate_file` read books and covers as binary files.

## Motivation and Context

Currently `DirectoryImport._locate_file` tries to open and decode books and covers as UTF-8 text files which is wrong.

## Checklist:


jonathangreen (approved):
Looks good to me

PR #27, opened 2021-08-23
Make sure Onix ISO8601 dates containe timezone information.

jonathangreen:
## Description

Tests on the `main` branch are failing because merging `python3` and `main` brought in two sets of changes. 

1. https://github.com/ThePalaceProject/circulation/pull/23
    Which changed the ONIX parser to use `dateutil.parser.isoparse`
2. https://github.com/NYPL-Simplified/circulation/pull/1589
    Which changed all the dates in the `python3` branch to be timezone aware

## Motivation and Context

The tests are failing because we are trying to compare a timezone aware date with a non-timezone aware date in the ONIX parser tests. This PR updates the ONIX parser to default to UTC if no timezone info is given, with a warning that his is happening.

### DIFF
--- a/api/onix.py
+++ b/api/onix.py
@@ -1,4 +1,5 @@
 import logging
+import pytz
 
 import dateutil.parser
 from enum import Enum
@@ -224,6 +225,12 @@ class ONIXExtractor(object):
             issued = None
             if publishing_date:
                 issued = dateutil.parser.isoparse(publishing_date)
+                if issued.tzinfo is None:
+                    cls._logger.warning(
+                        "Publishing date {} does not contain timezone information. Assuming UTC."
+                        .format(publishing_date)
+                    )
+                    issued = issued.replace(tzinfo=pytz.UTC)
 
             identifier_tags = parser._xpath(record, 'productidentifier')
             identifiers = []
--- a/api/onix.py
+++ b/api/onix.py
@@ -1,5 +1,4 @@
 import logging
-import pytz
 
 import dateutil.parser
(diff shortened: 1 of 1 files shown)


jonathangreen:
@vbessonov This is my take on how we should fix these test failures.

api/onix.py
```diff
@@ -224,6 +225,12 @@ def parse(cls, file, data_source_name, default_medium=None):
             issued = None
             if publishing_date:
                 issued = dateutil.parser.isoparse(publishing_date)
+                if issued.tzinfo is None:
+                    cls._logger.warning(
+                        "Publishing date {} does not contain timezone information. Assuming UTC."
+                        .format(publishing_date)
+                    )
+                    issued = issued.replace(tzinfo=pytz.UTC)
```
vbessonov, line 233: This looks good but I would suggest to use [`datetime_helpers.to_utc`](https://github.com/ThePalaceProject/circulation-core/blob/main/util/datetime_helpers.py#L30) to convert datetime to UTC.

vbessonov (changes requested):

vbessonov (approved):
This looks good to me.

api/onix.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import pytz
```
tdilauro, line 2: Is this import still needed, now that you've switched to the `to_utc` helper?

tdilauro (approved):
Looks good. There may be a unneeded import, for which I left a separate comment.

PR #28, opened 2021-08-23
Update upstream sync action

jonathangreen:
## Description

Update the code sync workflow to use a python script located in the [CI-Scripts](https://github.com/ThePalaceProject/ci-scripts) repo that we can share across all the code we would like to sync in our organization.

Using an external python script lets us more gracefully handle possible problems with the sync, like the local or remote branches not existing.

This updates the sync action in this repo to match https://github.com/ThePalaceProject/library-registry/pull/22

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -10,28 +10,30 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
🎺

PR #34, opened 2021-09-02
Make ProQuestOPDS2Importer's and ODL2Importer's parser argument optional

vbessonov:
## Description

This PR fixes the following error occurring when there is a ProQuest\ODL 2.x integration set up in Circulation Manager:
```
{"host": "d9cdc81fe392", "app": "simplified", "name": "root", "level": "INFO", "filename": "config.py", "message": "Connecting to database: postgres://demo:***@se-test-db-alpha.lyrtech.org/demo_simplye", "timestamp": "2021-09-02T13:08:55.305443+00:00"}
  File "/var/www/circulation/env/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/var/www/circulation/env/lib/python3.6/site-packages/flask/app.py", line 1945, in full_dispatch_request
    self.try_trigger_before_first_request_functions()
  File "/var/www/circulation/env/lib/python3.6/site-packages/flask/app.py", line 1993, in try_trigger_before_first_request_functions
    func()
  File "/var/www/circulation/api/routes.py", line 42, in initialize_circulation_manager
    app.manager = CirculationManager(app._db)
  File "/var/www/circulation/api/controller.py", line 165, in __init__
    self.load_settings()
  File "/var/www/circulation/api/controller.py", line 258, in load_settings
    library, self.analytics
  File "/var/www/circulation/api/controller.py", line 401, in setup_circulation
    return cls(self._db, library, analytics)
  File "/var/www/circulation/api/circulation.py", line 412, in __init__
    api = api_map[collection.protocol](_db, collection)
TypeError: __init__() missing 1 required positional argument: 'parser'
```

## Checklist:

### DIFF
--- a/api/proquest/importer.py
+++ b/api/proquest/importer.py
@@ -10,6 +10,7 @@ import webpub_manifest_parser.opds2.ast as opds2_ast
 from flask_babel import lazy_gettext as _
 from requests import HTTPError
 from sqlalchemy import or_
+from webpub_manifest_parser.opds2 import OPDS2FeedParserFactory
 from webpub_manifest_parser.utils import encode
 
 from api.circulation import BaseCirculationAPI, FulfillmentInfo, LoanInfo
@@ -182,7 +183,7 @@ class ProQuestOPDS2Importer(OPDS2Importer, BaseCirculationAPI, HasExternalIntegr
         self,
         db,
         collection,
-        parser,
+        parser=None,
         data_source_name=None,
         identifier_mapping=None,
         http_get=None,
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
Looks good. Fixed issue on demo.

api/proquest/importer.py
```diff
@@ -230,7 +231,7 @@ def __init__(
         super(ProQuestOPDS2Importer, self).__init__(
             db,
             collection,
-            parser,
+            parser if parser else RWPMManifestParser(OPDS2FeedParserFactory()),
```
tdilauro, line 234: It would be good to have a test that ensures `super` called with correct value when parser=None.

tdilauro (commented):
Suggesting a couple of tests to ensure that the correct default (or at least a non-`None` one) is chosen when `None` is passed in.

This is especially helpful when the import is accidentally removed. Otherwise, the `else` won't be used in the tests and we might only notice an error in production.

One alternative would be to code `parser=RWPMManifestParser(OPDS2FeedParserFactory())` in the __init__ parameters, which would also ensure an error during tests if any of the classes were not present.

PR #37, opened 2021-09-08
Add a configuration setting allowing to restrict supported identifier types

vbessonov:
## Description

This PR adds a configuration setting allowing to restrict supported identifier types for ODL 2.x and ProQuest collections:
image

Moved the configuration to its own class so other importers may be able to use it. The importer logic is partly in a `IgnoreIdentifierMixin` and partly in the `OPDS2Importer`. This is  primarily because the `ignored_identifiers` MAY be only part of skipped identifiers and more logic may be added later that may be OPDS2 specific.

NOTE: For release notes, this changes the behaviour of the importers, they will now import all types by default where before they were only importing ISBNs. This is the behaviour we want, but it is good to note in the release notes.

## How Has This Been Tested?

Additional tests. Manually tested in my multiple developer environments and in a deployed environment.

## Checklist:


mdellabitta:
Just was able to do a successful harvest of DPLA's OPDS2 feed with this branch!

RishiDiwanTT:
@tdilauro @jonathangreen Although all tests pass now, I would like to have some real world data to test against.
A feed that has the previously unsupported identifiers.

Apart from the that, the screenshot in the PR description does not match anything I see while trying to create a new collection, nor does any change in the 2 PRs indicate that this additional setting should be present anywhere. Which worries me

jonathangreen:
You should be able to import this feed locally in your CM: 
http://bookshelf-feed-demo.us-east-1.elasticbeanstalk.com/v1/publications

It contains `urn:uuid:` identifiers that would have not been supported before.

jonathangreen:
@RishiDiwanTT when you configured the collection where you able to configure the supported identifier types?

RishiDiwanTT:
Yes 
image

However I would like to turn off an ID type and try, I couldn't figure out which type `urn:uuid` is in the code (or google).
What value should I remove?

api/odl2.py
```diff
@@ -49,6 +53,21 @@ class ODL2Importer(OPDS2Importer, HasExternalIntegration):
 
     NAME = ODL2API.NAME
 
+    FEEDBOOKS_AUDIO = "{0}; protection={1}".format(
```
jonathangreen, line 56: This seems unrelated to the code here. Did this perhaps come in accidentally?

core/opds2_import.py
```diff
@@ -100,11 +114,80 @@ def parse_manifest(self, manifest):
         return result
 
 
-class OPDS2Importer(OPDSImporter):
+class OPDS2ImporterConfiguration(ConfigurationGrouping):
+    """Contains configuration settings of ProQuestOPDS2Importer."""
```
jonathangreen, line 118: Comment seems incorrect. I don't think these settings are scoped to the proquest importer.

core/opds2_import.py
```diff
@@ -763,6 +853,45 @@ def _find_formats_in_non_open_access_acquisition_links(
 
         return formats
 
+    @contextmanager
+    def _get_configuration(
+        self, db: sqlalchemy.orm.session.Session
+    ) -> OPDS2ImporterConfiguration:
+        """Return the configuration object.
+        :param db: Database session
+        :type db: sqlalchemy.orm.session.Session
```
jonathangreen, line 862: Should replace these doc comment types with type annotations.

core/opds2_import.py
```diff
@@ -763,6 +853,45 @@ def _find_formats_in_non_open_access_acquisition_links(
...
+        """
+        with self._configuration_factory.create(
+            self._configuration_storage, db, OPDS2ImporterConfiguration
+        ) as configuration:
+            yield configuration
+
+    def _get_supported_identifier_types(self) -> Set[int]:
+        """Return a set of supported identifier types.
+        :return: Set of supported identifier types
+        :rtype: Set[int]
```
jonathangreen, line 874: Would be good to remove this since there is a type annotation for it.

core/opds2_import.py
```diff
@@ -763,6 +853,45 @@ def _find_formats_in_non_open_access_acquisition_links(
...
+        return self._supported_identifier_types
+
+    def external_integration(
+        self, db: sqlalchemy.orm.session.Session
+    ) -> ExternalIntegration:
+        """Return an external integration associated with this object.
+        :param db: Database session
+        :type db: sqlalchemy.orm.session.Session
+        :return: External integration associated with this object
+        :rtype: core.model.configuration.ExternalIntegration
```
jonathangreen, line 891: Same comment about the type annotations.

core/opds_import.py
```diff
@@ -2069,14 +2069,17 @@ def feed_contains_new_data(self, feed):
...
+        for raw_identifier, remote_updated in last_update_dates:
 
-            identifier = self._parse_identifier(identifier)
+            identifier = self._parse_identifier(raw_identifier)
             if not identifier:
                 # Maybe this is new, maybe not, but we can't associate
                 # the information with an Identifier, so we can't do
                 # anything about it.
-                self.log.info("Ignoring %s because unable to turn into an Identifier.")
+                if raw_identifier:
```
jonathangreen, line 2079: We should probably still log something here if we don't have a raw identifier, so there is some evidence in the logs that the identifier is missing.

jonathangreen (changes requested):
Looks good other then the few minor comments that I made. Going to do some testing / importing in a test CM now to see how it works. Will add some more comments once that testing is done.

jonathangreen (changes requested):

jonathangreen:
I was able to set this up, and import the bookshelf feed with UUID identifiers. If I enable URI identifiers it works, and if I disable them, the UUIDs aren't imported. So I believe this is working. Just needs some of the other comments here cleaned up.

core/opds_import.py
RishiDiwanTT: Had the same thought, however this is what the change was in the PR, I can change it to this

api/odl2.py
RishiDiwanTT: Yes it seems they are using LICENSE_FORMAT directly from the ODLImporter class. I will remove both

core/opds2_import.py
RishiDiwanTT: Actually saw this during the merge and forgot about it later. Fixed.

api/odl2.py
```diff
@@ -16,13 +16,17 @@
...
+    OPDS2ImporterConfiguration,
+    OPDS2ImportMonitor,
+    RWPMManifestParser,
+)
 from core.util import first_or_default
 from core.util.datetime_helpers import to_utc
 
 
-class ODL2APIConfiguration(ConnectionConfiguration):
+class ODL2APIConfiguration(OPDS2ImporterConfiguration):
```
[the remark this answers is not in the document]
RishiDiwanTT: Made it multi-inheritance, option is back in the admin panel now

core/model/constants.py
```diff
@@ -195,6 +196,31 @@ class IdentifierConstants(object):
...
+    OCLC_NUMBER = IdentifierConstants.OCLC_NUMBER
+    OPEN_LIBRARY_ID = IdentifierConstants.OPEN_LIBRARY_ID
+    BIBLIOCOMMONS_ID = IdentifierConstants.BIBLIOCOMMONS_ID
+    URI = IdentifierConstants.URI
+    DOI = IdentifierConstants.DOI
+    UPC = IdentifierConstants.UPC
+    BIBBLIO_CONTENT_ITEM_ID = IdentifierConstants.BIBBLIO_CONTENT_ITEM_ID
+    ENKI_ID = IdentifierConstants.ENKI_ID
+    SUDOC_CALL_NUMBER = IdentifierConstants.SUDOC_CALL_NUMBER
+    PROQUEST_ID = IdentifierConstants.PROQUEST_ID
```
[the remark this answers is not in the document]
RishiDiwanTT: Not a fan of this myself, but kept it as the the original PR has this change.
I cannot find anyplace in the code where it is important for this to be in an Enum form, unless the numbers are stored in the DB and correlated somewhere else

api/proquest/importer.py
```diff
@@ -728,10 +728,8 @@ def fulfill(
 
     def external_integration(self, db):
         """Return an external integration associated with this object.
-
```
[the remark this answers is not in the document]
RishiDiwanTT: Yes, I expected the same and had the same confusion about this (here https://github.com/ThePalaceProject/circulation/pull/37#issuecomment-1029006602)

Let me look into this, now that its confirmed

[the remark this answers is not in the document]
RishiDiwanTT: Looks like the merge from main overwrote this (or the conflict resolution did).
Could you point me in the direction of a proquest feed I can test with

core/opds2_import.py
```diff
@@ -191,7 +279,9 @@ def _is_identifier_allowed(self, identifier: Identifier) -> bool:
         :return: Boolean value indicating whether CM can import the identifier
         :rtype: bool
         """
-        return identifier.type == Identifier.ISBN
+        supported_identifier_types = self._get_supported_identifier_types()
```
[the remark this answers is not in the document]
RishiDiwanTT: This will change the behaviour of current ODL2 and OPDS2 collections too. Is that desired?

core/opds_import.py
jonathangreen: Yes. I think this would be good to change.

core/model/constants.py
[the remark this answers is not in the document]
jonathangreen: I don't believe they are. I think the enum is here to make it easier to iterate over all the identifier values that we have defined.

core/opds2_import.py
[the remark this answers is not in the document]
jonathangreen: Yes. I think this is the desired change. I updated to PR description to reflect this and will resolve this comment.

api/proquest/importer.py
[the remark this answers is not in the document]
jonathangreen: You can ask Tara in slack for the feed we configure, but the proquest feed is a bit hard to work with because they keep it behind a firewall only accessible to our production servers. If you want to modify the code here as best you can to make sure these changes work with proquest, I will test with proquest feed on server before merging.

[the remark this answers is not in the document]
RishiDiwanTT: Changed, but untested

RishiDiwanTT:
> Where if we were only storing the type of identifier that we are blocking, then we wouldn't need to do this database migration.

We are in agreement, I am going ahead with changing this to a "Ignore Identifier type" change

tests/core/test_opds2_import.py
```diff
@@ -71,36 +78,52 @@ def _get_work_by_identifier(works, identifier):
...
+    MOBY_DICK_IDENTIFIER = "urn:isbn:978-3-16-148410-0"
+    HUCKLEBERRY_FINN_IDENTIFIER = "http://example.org/huckleberry-finn"
+
     def sample_opds(self, filename, file_type="r"):
         base_path = os.path.split(__file__)[0]
         resource_path = os.path.join(base_path, "files", "opds2")
         return open(os.path.join(resource_path, filename)).read()
 
+    def setup_method(self):
+        super(TestOPDS2Importer, self).setup_method()
```
tdilauro, line 90: Minor: We can simplify to `super().setup_method()` here? But we can clean these up with `pyupgrade` linter.

RishiDiwanTT: I can change this quick, no probkems

RishiDiwanTT: This is actually everywhere, I'd rather leave it to the pyupgrade tool

tdilauro:
I've tested this locally and will deploy to `gorgon` for pre-merge testing in LYR Reads library collections.

tdilauro:
Did some testing in the AWS environment and this seems to work properly.

tdilauro (approved):
Looks good!

tdilauro:
Fixed the configuration screen image in the description.

PR #40, opened 2021-09-14
Fix logo "data" URL creation.

tdilauro:
## Description

- Fixes `data` URLs for logo images so that they don't contain bytes literal notation.
- Adds fixture for test image.
- Split image data URL calculation into helper method.
- Parameterize logo max dimension.

## Motivation and Context
 
The move to Python 3 resulted in logo data URL values with embedded bytes literals (`b'...'`) like
- `data:image/png;base64,b'{base64_image_data}'`

rather than the expected values that look like
- `data:image/png;base64,{base64_image_data}`

Converting the Base64 encoded image data to Unicode strings eliminates the bytes literal notation.

## How Has This Been Tested?

- Manually verified that the resulting logo data URLs render properly in the client apps.
- Updated tests for Python 3.
- New test for data URL helper.
- Ensure that test fixture image properties are appropriate for logo-related tests.

## Checklist:

- [ N/A] I have updated the documentation accordingly.

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -1,6 +1,7 @@
 import base64
 from io import BytesIO
 import json
+from typing import Any, Dict, Optional
 import uuid
 
 import flask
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
Looks good to me.

api/admin/controller/library_settings.py
```diff
@@ -337,18 +331,37 @@ def list_setting(self, setting, json_objects=False):
...
 
-    def image_setting(self, setting):
-        """Retrieve an uploaded image for the given setting from the current HTTP request."""
+    @staticmethod
+    def _data_url_for_image(image: Image, _format="PNG") -> Optional[str]:
+        """Produce the `data` URL for the setting's uploaded image file.
+
+        :param image: A Pillow Image object.
+        :param _format: A valid Pillow image format.
+        :return: The `data` URL if an image file was uploaded, else None.
```
jonathangreen, line 340: This might be a copy paste error from below? This doesn't look like it will ever return `None` to me.

jonathangreen (approved):
Minor comment about a possibly confusing function documentation. Looks good to me otherwise.

api/admin/controller/library_settings.py
tdilauro: Absolutely! Thanks!

api/admin/controller/library_settings.py
```diff
@@ -337,18 +331,37 @@ def list_setting(self, setting, json_objects=False):
 
         return json.dumps([_f for _f in value if _f])
 
-    def image_setting(self, setting):
-        """Retrieve an uploaded image for the given setting from the current HTTP request."""
+    @staticmethod
+    def _data_url_for_image(image: Image, _format="PNG") -> Optional[str]:
```
jonathangreen, line 335: I think to match the type hints to the comments and function behaviour this one should be `-> str`

api/admin/controller/library_settings.py
```diff
@@ -337,18 +331,37 @@ def list_setting(self, setting, json_objects=False):
...
+        :param image: A Pillow Image.
+        :param _format: A valid Pillow image format.
+        :return: The `data` URL for the image.
+        """
+        buffer = BytesIO()
+        image.save(buffer, format=_format)
+        b64 = base64.b64encode(buffer.getvalue())
+        return "data:image/png;base64,%s" % b64.decode('utf-8')
+
+    def image_setting(self, setting: Dict[str, Any], max_dimension=Configuration.LOGO_MAX_DIMENSION) -> str:
```
jonathangreen, line 347: And this one should be `-> Optional[str]`

tdilauro:
Tests are randomly failing due to timeouts. `circulation-core` tests have not failed so far, but `circulation` tests have failed on one of py3.6 and p3.6. I have successfully run them for all versions locally and I restarted them here one more time.

PR #41, opened 2021-09-24
Fix admin static package asset path.

tdilauro:
## Description

Fixes static package path for local copy of admin UI.

This problem affected only those test environments in which the `circulation-admin` client was locally installed or linked.

## Motivation and Context

The local package URL allows package assets to pulled from local static files on the CM server, rather than pulling from a CDN, as is normal.

## How Has This Been Tested?

Manually tested to verify that the assets were retrievable. Adjusted existing tests for the correct path.

## Checklist:


jonathangreen (approved):
Looks good to me

PR #44, opened 2021-10-01
[WIP] Add support for DeMarque LCP content

vbessonov:
## Checklist:


jonathangreen:
Does the hint url need to be associated with each license? Could it instead be associated with a particular patron?

vbessonov:
> Does the hint url need to be associated with each license? Could it instead be associated with a particular patron?

@jonathangreen, yes it can be. `LCPController.get_lcp_passphrase` doesn't even use license ID, it's only used as a loan's external identifier to get a library but `Patron` objects also have links to a library.

jonathangreen:
This is looking good to me. @vbessonov can you resolve the merge conflicts here. I'd like to get this rolled out on Gorgon for testing next week.

PR #45, opened 2021-10-10
Remove random sort option

vbessonov:
## Description

To remove `random` sort option from CM's feed and exclude it from sorting options available in the client apps, it's enough to update the library's configuration and switch it off:
image
1. Open the library's configuration settings.
2. Remove `Random` from `Allow patrons to sort by` list.
3. Select other option than `Random` in the `Default Sort by` list.

However, if it's required to completely remove random sorting from Circulation Manager, the changes in this PR will come in handy.

NOTE: This PR depends on https://github.com/ThePalaceProject/circulation-core/pull/28.

This PR contains a new migration for removing `random` sort option from Circulation Manager:
* It checks the library's default sorting option and, if it's `random`, it replaces it with the default one.
* It checks the library's other available sorting options and excludes `random` from that list.

## Checklist:

### DIFF
--- a/migration/20211009-remove-random-sort-option.py
+++ b/migration/20211009-remove-random-sort-option.py
@@ -5,13 +5,13 @@ import os
 import sys
 
 from contextlib2 import closing
-from migartion_scripts import RandomSortOptionRemover
 
 bin_dir = os.path.split(__file__)[0]
 package_dir = os.path.join(bin_dir, "..")
 sys.path.append(os.path.abspath(package_dir))
 
 from core.model import production_session
+from migartion_scripts import RandomSortOptionRemover
 
 with closing(production_session()) as db:
     remover = RandomSortOptionRemover()


vbessonov:
> This looks good to me. The only concern I have is the new `migartion_scripts.py` file. I like that you tested the migration here, and this file enables easier testing of migration scripts, but I am worried that it makes it harder to drop old migration scripts in the future.
> 
> Right now the repo has every migration script, but I would like to start to drop old migration scripts in the future, so we would only keep migration scripts around for one minor version, then be able to drop them. That way we don't have to maintain a large number of old migration scripts over time.
> 
> Is there another way we could accomplish testing this migration script?

The migration itself is in `migration/20211009-remove-random-sort-option.py` file, so we can easily drop it and, if it's required, we can just remove `RandomSortOptionRemover` class from `migration_scripts.py` and `test_migration_scripts.py`. 
Also, I think we can put everything, `RandomSortOptionRemover` and `TestRandomSortOptionRemover` and the migration, into `migration/20211009-remove-random-sort-option.py`, but that would make it look a bit nasty.

jonathangreen:
The scripts do some import manipulation in order to get the import path setup correctly in order to run migrations. You can test if its working by either running `core/bin/migrate_database` directly, or starting the docker container built from this branch.

vbessonov:
> The scripts do some import manipulation in order to get the import path setup correctly in order to run migrations. You can test if its working by either running `core/bin/migrate_database` directly, or starting the docker container built from this branch.

@jonathangreen, it should be fixed now. The import order in the migration file was incorrect.

PR #54, opened 2021-10-29
Update timeout method to stop hung tests

jonathangreen:
## Description

Update `tox.ini` with pytest [config options](https://docs.pytest.org/en/latest/reference/reference.html#configuration-options). This lets us centralize out pytest config. 

Update `pytest-timeout` to use the [`thread` method](https://pypi.org/project/pytest-timeout/) for timing out the test suite. It seems that the signal method is not working, since [this run](https://github.com/ThePalaceProject/circulation/runs/4047051977?check_suite_focus=true#step:6:173) spent 50 minutes on a single test and never timed out. 

Hopefully the thread method will actually timeout the tests and help us get an understanding of what is failing when the tests hang.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ commands_pre =
     docker: docker restart es-circ
     python -m textblob.download_corpora
 commands =
-    pytest {posargs:--timeout=600 --disable-warnings tests}
+    pytest {posargs:--disable-warnings}
 passenv = SIMPLIFIED_* CI
 setenv =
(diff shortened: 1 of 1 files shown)


jonathangreen:
Going to merge this one. Saw another timeout that didn't trigger `pytest-timeout` today here: 
https://github.com/ThePalaceProject/circulation/runs/4069271317

jonathangreen:
Hopefully we can catch one with this and see whats causing the issue.

PR #57, opened 2021-11-01
Make sure migration script output is logged.

jonathangreen:
## Description

Currently when database migrations are run on container startup, it isn't logged to a file that we can use to check the migration status. This changes the script to write the migration changes out to a log file.

## Motivation and Context

Sometimes we run migrations by starting a new version of the container, rather then running the migrate play. It would be nice to have migration logs in the same place, either way the migration gets run.


tdilauro (approved):
Looks good! 🎸

tdilauro:
The test failure seems unrelated to the change in this PR. Did something else change?

PR #59, opened 2021-11-03
Update RSA dependancy

jonathangreen:
## Description

Update to a newer version of the RSA module that does not have a security advisory.


tdilauro (approved):
🚀

PR #60, opened 2021-11-03
Create codeql-analysis.yml

jonathangreen:
## Description

Setup github automated code security analysis.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,52 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+  schedule:
+    - cron: '35 12 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
```
tdilauro, line 18: I don't think these are strictly required for public repos, like `circulation` and `circulation-core`, but I don't think there's any harm in having them.

tdilauro (approved):
LGTM! 🎸

PR #61, opened 2021-11-03
Lint with black and isort 

jonathangreen:
## Description

Similar to https://github.com/ThePalaceProject/circulation-core/pull/22 this PR adds a linting step to our CI pipeline. The bulk of this PR is automatic reformatting done by isort and black.

This PR also removes old migrations. Some of them did not parse in Python 3 which was causing Black to fail in its linting. Seemed easier to remove them then make the changes to make them work in Python 3. I don't think we need to carry around these old migration files anyway.

## Motivation and Context

Bring circulation into line with these PRs:
https://github.com/ThePalaceProject/circulation-core/pull/22
https://github.com/ThePalaceProject/library-registry/pull/23


tdilauro (approved):
LGTM!

Caveat: I took a good look at the new GH workflow, the updated `tox.ini`, and a sampling of the linted files. I spot checked in another sample of files.

PR #63, opened 2021-11-08
Make LCP's Passphrase hint and Passphrase hint URL configuration settings mandatory and add default values

vbessonov:
## Description

This PR depends on https://github.com/ThePalaceProject/circulation-core/pull/35.

This PR contains quite a lot of different changes:
1. The most important change is making `Passphrase hint` and Passphrase hint URL` configuration settings mandatory and adding meaningful default values.
2. All other changes are mostly related with changes from https://github.com/ThePalaceProject/circulation-core/pull/35 and fixing some failing tests.

## Checklist:

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -120,7 +120,7 @@ class ODLAPIConfiguration(ConfigurationGrouping):
         ),
         type=ConfigurationAttributeType.TEXT,
         required=True,
-        default=DEFAULT_PASSPHRASE_HINT
+        default=DEFAULT_PASSPHRASE_HINT,
     )
 
     passphrase_hint_url = ConfigurationMetadata(
@@ -132,7 +132,7 @@ class ODLAPIConfiguration(ConfigurationGrouping):
         type=ConfigurationAttributeType.TEXT,
         required=True,
         default=DEFAULT_PASSPHRASE_HINT_URL,
-        format="url"
+        format="url",
     )
 
     encryption_algorithm = ConfigurationMetadata(
--- a/api/odl.py
+++ b/api/odl.py
@@ -1,6 +1,5 @@
 import binascii
 import datetime
-import hashlib
 import json
 import logging
 import uuid
@@ -65,7 +64,8 @@ from .shared_collection import BaseSharedCollectionAPI
 class ODLAPIConfiguration(ConfigurationGrouping):
     """Contains LCP License Server's settings"""
 
-    DEFAULT_PASSPHRASE_HINT = "Look in the Palace app."
+    DEFAULT_PASSPHRASE_HINT = "View the help page for more information."
+    DEFAULT_PASSPHRASE_HINT_URL = "https://lyrasis.zendesk.com/"
     DEFAULT_ENCRYPTION_ALGORITHM = HashingAlgorithm.SHA256.value
 
     feed_url = ConfigurationMetadata(
@@ -119,8 +119,8 @@ class ODLAPIConfiguration(ConfigurationGrouping):
             "Hint displayed to the user when opening an LCP protected publication."
         ),
         type=ConfigurationAttributeType.TEXT,
-        required=False,
-        default=DEFAULT_PASSPHRASE_HINT,
+        required=True,
+        default=DEFAULT_PASSPHRASE_HINT
     )
 
     passphrase_hint_url = ConfigurationMetadata(
@@ -130,8 +130,9 @@ class ODLAPIConfiguration(ConfigurationGrouping):
             "Hint URL available to the user when opening an LCP protected publication."
         ),
         type=ConfigurationAttributeType.TEXT,
-        required=False,
-        format="url",
+        required=True,
+        default=DEFAULT_PASSPHRASE_HINT_URL,
+        format="url"
     )
 
     encryption_algorithm = ConfigurationMetadata(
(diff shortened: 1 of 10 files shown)


api/odl.py
```diff
@@ -125,7 +115,7 @@ class ODLAPIConfiguration(ConfigurationGrouping):
         label=_("Passphrase hint"),
         description=_("Hint displayed to the user when opening an LCP protected publication."),
         type=ConfigurationAttributeType.TEXT,
-        required=False,
+        required=True,
```
jonathangreen, line 118: For already configured servers we need to handle the case where this configuration setting isn't already set. As you mentioned in notion, this can be handled by a database migration. I don't like the idea of adding a database migration every time we add a new setting though. 

What I've seen in the past, in code that isn't using your `ConfigurationMetadata` helpers, was a check done in the constructor to see if the setting wasn't set, and if so give the default value. 

For example:
- SIP2 
  - https://github.com/ThePalaceProject/circulation/blob/main/api/sip/client.py#L271
  - https://github.com/ThePalaceProject/circulation/blob/main/api/sip/client.py#L265-L266
- Millenium 
  - https://github.com/ThePalaceProject/circulation/blob/main/api/millenium_patron.py#L157-L158

Perhaps this could be handled by the `ConfigurationMetadata` code, so that is returns the default value for settings that are not in the database.

jonathangreen (changes requested):

api/odl.py
jonathangreen: Continuing the discussion from the weekly sync call this morning @vbessonov 

Here is the test I did:
- Check out commit 8ec48c7108d8453871d20c8d2a183c211681a62b (just before anything for LCP goes in)
- Create a new ODL2 collection.
- Update to the code in this branch.
- Load the CM admin UI screen for the ODL2 collection

This is what is displayed:
Screen Shot 2021-11-08 at 1 03 06 PM

So the admin UI displays the default values for the values not in the database yet.

api/odl2.py
```diff
@@ -174,7 +174,11 @@ def _extract_publication_metadata(self, feed, publication, data_source_name):
             skipped_license_formats = configuration.skipped_license_formats
 
             if skipped_license_formats:
-                skipped_license_formats = set(json.loads(skipped_license_formats))
+                skipped_license_formats = (
+                    set(json.loads(skipped_license_formats))
+                    if isinstance(skipped_license_formats, str)
+                    else set(skipped_license_formats)
```
tdilauro, line 180: This implies that we have this config might be stored in the database in different formats. Is there any reason why we wouldn't want to enforce a single format for these? Alternatively, could we have a configuration getter for these ensures that that it always comes out the same way?

Wrapping the `set` around the entire ternary makes it a little more obvious that it's a similar issue as in `api/saml/provider.py` +88 to +92:
```
set(
  json.loads(skipped_license_formats))
  if isinstance(skipped_license_formats, str)
  else skipped_license_formats
)
```

api/saml/provider.py
```diff
@@ -85,7 +85,11 @@ def __init__(self, library, integration, analytics=None):
                 configuration.patron_id_use_name_id
             )
             self._patron_id_attributes = (
-                json.loads(configuration.patron_id_attributes)
+                (
+                    json.loads(configuration.patron_id_attributes)
+                    if isinstance(configuration.patron_id_attributes, str)
+                    else configuration.patron_id_attributes
+                )
```
tdilauro, line 92: Similar to comment on `api/odl2.py` lines +178 to +180.

tests/test_odl.py
```diff
@@ -16,7 +16,7 @@
     ODLHoldReaper,
     ODLImporter,
     SharedODLAPI,
-    SharedODLImporter,
+    SharedODLImporter, ODLAPIConfiguration,
```
tdilauro, line 19: Minor: import style.

tdilauro (commented):
This generally looks good, but a couple of issues related to how some config is stored in the database and one minor import organization issue.

tests/test_odl.py
vbessonov: Fixed the import.

api/saml/provider.py
vbessonov: I added additional changes to https://github.com/ThePalaceProject/circulation-core/pull/35 in order to correctly parse JSON-encoded `LIST` and `MENU` items directly in `ConfigurationMetadata.__get___` and removed the code above.

api/odl2.py
vbessonov: I added additional changes to https://github.com/ThePalaceProject/circulation-core/pull/35 in order to correctly parse JSON-encoded `LIST` and `MENU` items directly in `ConfigurationMetadata.__get___` and removed the code above.

vbessonov:
@tdilauro, @jonathangreen, I'd like to merge https://github.com/ThePalaceProject/circulation-core/pull/35 first and then resolve the conflicts in this PR (there are a lot of them).

tdilauro (approved):
Looks good! 🚀 

Really nice to see some of the workarounds go away.

tdilauro:
> @tdilauro, @jonathangreen, I'd like to merge [ThePalaceProject/circulation-core#35](https://github.com/ThePalaceProject/circulation-core/pull/35) first and then resolve the conflicts in this PR (there are a lot of them).

@vbessonov I'm not sure where the conflicts are, but you may kill off a bunch of them with `tox -e isort-reformat,black-reformat`.

jonathangreen (approved):

PR #64, opened 2021-11-15
Properly handle invalid default patron values for self-tests

tdilauro:
## Description

- Ensures that problem details and other non-Patron objects are not misinterpreted as test patrons during self-tests.
- Add new tests.

## Motivation and Context

Self-tests displayed unhelpful errors when attempting to get and emit properties of a patron from a problem detail.

## How Has This Been Tested?

- Added new tests and ran existing tests.
- Because of the firewall challenges of testing ILS integrations in local environments, we will need to do additional testing from a test deployment.

## Checklist:

### DIFF
--- a/api/selftest.py
+++ b/api/selftest.py
@@ -16,26 +16,25 @@ from .circulation import CirculationAPI
 from .feedbooks import FeedbooksImportMonitor, FeedbooksOPDSImporter
 
 
-class _NoValidLibrarySelfTestPatron(Exception):
-    """Exception raised when no valid self-test patron found for library.
-
-    Attributes:
-        message -- primary error message.
-        detail (optional) -- additional explanation of the error
-    """
-
-    def __init__(self, message: str, detail: str = None):
-        super().__init__(message, detail)
-        self.message = message
-        self.detail = detail
-
-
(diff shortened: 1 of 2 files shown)


api/selftest.py
```diff
@@ -1,35 +1,54 @@
...
 from core.selftest import HasSelfTests as CoreHasSelfTests
 from core.selftest import SelfTestResult
+from core.util.problem_detail import ProblemDetail
 
 from .authenticator import LibraryAuthenticator
 from .circulation import CirculationAPI
 from .feedbooks import FeedbooksImportMonitor, FeedbooksOPDSImporter
 
 
+class _NoValidLibrarySelfTestPatron(Exception):
```
jonathangreen, line 19: Should this be a subclass of [`BaseError`](https://github.com/ThePalaceProject/circulation-core/blob/3e4f3395409bf18557d7123fb070a321301330c1/exceptions.py#L1)? Might be good to keep all of our exception having a common base class if we can.

If its meant to be used only within `HasSelfTests` should it be defined in that class?

tdilauro:
Moved exception, method, and function into `HasSelfTests`,

jonathangreen (approved):
🚀

PR #68, opened 2021-11-23
Fix the DPLA audiobook download issue

vbessonov:
## Motivation and Context

This PR depends on https://github.com/ThePalaceProject/circulation-core/pull/40.

This PR fixes the DPLA audiobook download issue.

## Checklist:

I tested the changes I made locally:
1. Imported the DPLA ODL+OPDS2 feed.
2. Set up a basic authentication provider.
3. Successfully downloaded `Secret Language of Sisters`  audiobook.
4. Successfully downloaded `They Both Die at the End` Adobe DRM-protected ebook.
5. Successfully downloaded LCP-protected book `Light From Uncommon Stars` (it has both, Adobe and LCP versions).

Below is the screenshot showing the CM's response when I downloaded an Feedbooks DRM-protected audiobook:
image

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -2,7 +2,6 @@ import binascii
 import datetime
 import json
 import logging
-import re
 import uuid
 from io import StringIO
 from typing import Callable, Dict, List, Optional, Tuple, Union
@@ -20,7 +19,7 @@ from uritemplate import URITemplate
 from core import util
 from core.analytics import Analytics
 from core.lcp.credential import LCPCredentialFactory
-from core.metadata_layer import CirculationData, FormatData, LicenseData, TimestampData
+from core.metadata_layer import FormatData, LicenseData, TimestampData
 from core.model import (
     Collection,
     ConfigurationSetting,
@@ -50,7 +49,7 @@ from core.model.configuration import (
     HasExternalIntegration,
 )
 from core.model.licensing import LicenseStatus
-from core.monitor import CollectionMonitor, IdentifierSweepMonitor
+from core.monitor import CollectionMonitor
 from core.opds_import import OPDSImporter, OPDSImportMonitor, OPDSXMLParser
 from core.testing import DatabaseTest, MockRequestsResponse
 from core.util.datetime_helpers import to_utc, utc_now
--- a/api/odl.py
+++ b/api/odl.py
@@ -543,87 +543,46 @@ class ODLAPI(BaseCirculationAPI, BaseSharedCollectionAPI, HasExternalIntegration
     @staticmethod
     def _find_content_link_and_type(
         links: List[Dict],
-        delivery_mechanism: Union[
-            Optional[str], Optional[LicensePoolDeliveryMechanism]
-        ],
-        ignore_drm_scheme: bool = False,
-    ) -> Tuple[str, str]:
+        drm_scheme: Optional[str],
+    ) -> Tuple[Optional[str], Optional[str]]:
         """Find a content link with the type information corresponding to the selected delivery mechanism.
 
         :param links: List of dict-like objects containing information about available links in the LCP license file
-        :param delivery_mechanism: Selected delivery mechanism
+        :param drm_scheme: Selected delivery mechanism DRM scheme
 
         :return: Two-tuple containing a content link and content type
         """
-        content_link = None
-        content_type = None
-
+        candidates = []
         for link in links:
             # Depending on the format being served, the crucial information
             # may be in 'manifest' or in 'license'.
-            if link.get("rel") in ("manifest", "license"):
-                candidate_content_link = link.get("href")
-                candidate_content_type = link.get("type")
+            if link.get("rel") not in ("manifest", "license"):
+                continue
+            href = link.get("href")
+            type = link.get("type")
 
-                if not delivery_mechanism:
-                    # If we don't have a LicensePoolDeliveryMechanism instance,
-                    # we can't really decide whether the link has the correct content and DRM type,
-                    # so we take the first one.
-                    content_link = candidate_content_link
-                    content_type = candidate_content_type
-                    break
-                elif isinstance(delivery_mechanism, str):
-                    # If delivery mechanism is a string,
-                    # then we suppose that it contains the DRM type in the case of the DRM-protected content
-                    # and the content type in the case of OA content.
-                    if delivery_mechanism == candidate_content_type:
-                        content_link = candidate_content_link
-                        content_type = candidate_content_type
-                        break
-                elif isinstance(delivery_mechanism, LicensePoolDeliveryMechanism):
-                    # If we have a LicensePoolDeliveryMechanism instance,
-                    # then we use it to find a link with the correct content and DRM types.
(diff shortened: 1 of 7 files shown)


api/odl.py
```diff
@@ -539,36 +540,58 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
...
+                == MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE
+            )
+        else:
+            return False
 
-        links = doc.get("links", [])
+    def _find_content_link_and_type(
+        self,
+        links: List[Dict],
+        delivery_mechanism: Union[
```
jonathangreen, line 568: Instead of passing this as a union and doing type checking on it, could you just pass it as two string parameters `drm_scheme` and `content_type`, so we don't need to check the type that gets handed in in both `_find_content_link_and_type` and `_is_audiobook`?

jonathangreen:
Tested this on iOS and Android on the demo server and it does fix the issues with Palace Marketplace content. 👍🏻

jonathangreen:
One more comment, as I'm reviewing this, would it be better to address this in the import process, so that the correct delivery mechanisms are stored in the database, instead of addressing it in the fulfill process?

vbessonov:
> One more comment, as I'm reviewing this, would it be better to address this in the import process, so that the correct delivery mechanisms are stored in the database, instead of addressing it in the fulfill process?

I also would like to get that sorted during the import process. There is already a [piece of code](https://github.com/ThePalaceProject/circulation/blob/main/api/odl2.py#L197) migrated from ODL 1.x assigning correct DRM schemes to audiobooks delivery mechanisms.

The problem is that ODL 2.x feed doesn't specify a DRM type:
```
            "licenses": [
                {
                    "metadata": {
                        "identifier": "urn:uuid:c07e1bce-fcba-4f12-99b3-6f43baddb052",
                        "format": [
                            "application/audiobook+json",
                            "text/html"
                        ],
                        "created": "2021-10-04T16:21:27+02:00",
                        "source": "http://www.cantook.net/",
                        "price": {
                            "currency": "USD",
                            "value": 59.99
                        },
                        "terms": {
                            "expires": "2023-10-04",
                            "concurrency": 1,
                            "length": 5097600
                        },
                        "order": {
                            "name": "Tester Requests 3",
                            "identifier": "202110-262-013793",
                            "href": "https://market.feedbooks.com/carts/13793"
                        }
                    },
```

It means that during the import we don't know the DRM type, we get it only during fulfilment when we see it in the LCP document:
```
        {
            "href": "https://license.feedbooks.net/loan/audiobook/manifest/?loan_uuid=e3801188-4baa-11ec-8e47-0242ac180002",
            "type": "application/audiobook+json; protection=http://www.feedbooks.com/audiobooks/access-restriction",
            "rel": "manifest"
        },
```

jonathangreen:
Its not set in the feed, but it is set in the license info document that we look at during the import process. 

```
 "identifier": "urn:uuid:cbe62386-0b0e-47a2-9e4b-f6561bf3e5e9",
 "status": "available",
 "created": "2021-11-20T02:20:17Z",
 "format": "application/audiobook+json; protection=http://www.feedbooks.com/audiobooks/access-restriction"
```

jonathangreen:
So we could use that to set the delivery mechanism during import, instead of during fulfillment.

vbessonov:
> Its not set in the feed, but it is set in the license info document that we look at during the import process.
> 
> ```
>  "identifier": "urn:uuid:cbe62386-0b0e-47a2-9e4b-f6561bf3e5e9",
>  "status": "available",
>  "created": "2021-11-20T02:20:17Z",
>  "format": "application/audiobook+json; protection=http://www.feedbooks.com/audiobooks/access-restriction"
> ```

According to the current version of the ODL spec, `format` is not mandatory and MAY be present in the license info document. But, yes, I can add additional code to fetch it if it exists.

jonathangreen:
Understood its not mandatory in the license info document. It does seem nicer to me to handle this case in the import process rather then fulfillment, especially since we have the license info document loaded anyway and its handled during import in the ODL 1.x code.

vbessonov:
@jonathangreen, I pushed new code to extract audiobooks' content type and DRM scheme and use it to create a correct delivery mechanism.

api/odl.py
```diff
@@ -539,36 +540,36 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
...
         content_type = None
+
         for link in links:
             # Depending on the format being served, the crucial information
             # may be in 'manifest' or in 'license'.
             if link.get("rel") in ("manifest", "license"):
                 candidate_content_link = link.get("href")
                 candidate_content_type = link.get("type")
 
+                if ignore_drm_scheme:
```
jonathangreen, line 568: Now that we are correctly setting the delivery_mechanisms on import, is this code still needed?

api/odl.py
```diff
@@ -597,6 +598,45 @@ def _fulfill(
...
+        links = doc.get("links", [])
+        content_link, content_type = self._find_content_link_and_type(
+            links, delivery_mechanism
+        )
+
+        if (
+            not content_link
+            and not content_type
+            and licensepool.presentation_edition.medium == Edition.AUDIO_MEDIUM
+        ):
```
jonathangreen, line 632: Now that we are correctly setting the delivery mechanisms, is this condition still needed?

api/odl.py
vbessonov: I think it's still required. According to the spec, `format` is not mandatory and **MAY** be present in the License Info Document, so it means that we won't be always available to fetch all the information we need on import.

api/odl.py
vbessonov: I think that this piece of code is still required in the case when `format` is not present in the License Info Document.

api/odl.py
jonathangreen: Its available right now in the DeMarque case, and we are doing this special case for them. So I think we are okay to remove this and handle it during import.

api/odl.py
```diff
@@ -539,29 +540,24 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
...
-
-        expires = doc.get("potential_rights", {}).get("end")
-        expires = dateutil.parser.parse(expires)
-
-        links = doc.get("links", [])
+    @staticmethod
+    def _find_content_link_and_type(
+        links: List[Dict],
+        delivery_mechanism: Union[
+            Optional[str], Optional[LicensePoolDeliveryMechanism]
```
jonathangreen, line 547: Do we ever get a `str` passed into this function now? It looks like its only ever called with `LicensePoolDeliveryMechanism`.

jonathangreen: Seems like the type hint on `_fulfill` is wrong then, it looks like `_find_content_link_and_type` is only called from there, and its type hint is `Optional[LicensePoolDeliveryMechanism]`. Thats what made me think we weren't expecting a string to get to `_find_content_link_and_type`.

I guess this type hint was wrong before, so we can leave it be.

jonathangreen (approved):

jonathangreen (approved):

PR #69, opened 2021-11-28
Fix the ProQuest empty page issue

vbessonov:
## Description

This PR fixes the ProQuest empty page issue. It adds new logic that stops CM from downloading the feed when it receives an empty page.

## Motivation and Context

ProQuest used to throw a 404 or 500 error when we tried to request a non-existent page, but this behavior recently had changed. Instead of throwing an error, they return empty pages, making CM infinitely loop through them.

## Checklist:

### DIFF
--- a/api/proquest/client.py
+++ b/api/proquest/client.py
@@ -1,5 +1,6 @@
 import logging
 from contextlib import contextmanager
+from typing import Dict
 
 import requests
 from flask_babel import lazy_gettext as _
@@ -12,6 +13,7 @@ from core.model.configuration import (
     ConfigurationFactory,
     ConfigurationGrouping,
     ConfigurationMetadata,
+    ConfigurationOption,
     ConfigurationStorage,
 )
 from core.util import is_session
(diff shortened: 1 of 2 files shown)


api/proquest/client.py
```diff
@@ -74,6 +76,26 @@ class ProQuestAPIClientConfiguration(ConfigurationGrouping):
         required=False,
     )
 
+    stop_on_empty_page = ConfigurationMetadata(
```
jonathangreen, line 79: What cases would we ever want to not set this setting? Do we need this as a configuration setting? This functionality seems like the kind of thing that our implementation team doesn't need to know about. Adding this is a setting will just be confusing to them.

api/proquest/client.py
```diff
@@ -459,6 +481,39 @@ def _download_feed_page(self, configuration, page, hits_per_page):
...
+    def _is_feed_page_empty_or_incorrect(feed: Dict) -> bool:
+        """Determine whether the feed page is correctly structured and contains any publications.
+
+        ProQuest used to throw a 404 or 500 error when we tried to request a non-existent page,
+        but this behavior recently had changed. Instead of throwing an error, they
+        return empty pages, making CM infinitely loop through them.
+
+        NOTE: This method doesn't use a proper OPDS 2.x parser because it'd slow down downloading a big feed.
+        Instead, it uses a number of heuristics we got working with the ProQuest feed.
+        If the structure of the feed changes, this method will stop working and the download process will fail.
```
tdilauro, line 494: Within each page of the feed (most importantly, the first page), we get `.opdsFeed.metadata.numberOfItems`. Could we use that value, along with the current number of items per page to calculate the correct number of pages to fetch?

vbessonov: I wouldn't trust the ProQuest pagination mechanism, it doesn't work correctly.

api/proquest/client.py
vbessonov: This functionality is a workaround that we might want to get rid of in the future, if ProQuest fixes the pagination issue.

jonathangreen: Lets remove the config setting, and have this be the default behavior. If we come across blank pages in the middle of the feed, we can reevaluate.

api/proquest/client.py
vbessonov: Jonathan Wyrembelski from ProQuest proposed the same but I'd like to make sure that feed's metadata is correct and do it in a separate PR.

api/proquest/client.py
vbessonov: I removed the configuration setting.

api/proquest/client.py
tdilauro: I won't hold up this PR for this.

jonathangreen (approved):

PR #70, opened 2021-11-28
Do not run ProQuest bookkeeping mechanism if the downloading process hasn't finished successfully

vbessonov:
## Description

This PR fixes the error resulting in the disappeared ProQuest books. It adds additional code to ensure that the ProQuest bookkeeping mechanism (the one responsible for hiding items that are no longer present in the ProQuest feed) doesn't run if the downloading process hasn't finished successfully.

## Motivation and Context

The current version of the ProQuest integration runs the bookkeeper cleaning the stale items all the time, even in the case when the downloading process failed. But when the downloading process fails, there is no list of available ProQuest items, and the bookkeeper considers all the items present in the CM's database to be stale and hides all of them.

## Checklist:


jonathangreen (approved):
LGTM

tdilauro (approved):
Looks good! 🎸

PR #71, opened 2021-11-30
Feature/aws xray

jonathangreen:
## Description

Adds a new optional middleware that enables AWS XRay on requests to the Palace app. 

## Motivation and Context

Should help us troubleshoot performance issues within the app.

## How Has This Been Tested?

I tested in a local environment with the xray daemon running in a docker container. 

This is the command I used to run the xray daemon: 
```
docker run -it --name xray -e AWS_REGION=us-west-2 --rm -p 2000:2000/udp -v ~/.aws/:/home/xray/.aws/:ro -e AWS_PROFILE=simplye amazon/aws-xray-daemon
```

Then I tested the app like this: 
```
PALACE_XRAY=true SIMPLIFIED_PRODUCTION_DATABASE="postgres://simplified:test@localhost:5431/circ" python app.py
```

This gave xray traces from the local app run.

### DIFF
--- a/api/util/xray.py
+++ b/api/util/xray.py
@@ -2,11 +2,13 @@ import logging
 import os
 from typing import Optional
 
-from aws_xray_sdk.core import AWSXRayRecorder, xray_recorder, patch as xray_patch
-from aws_xray_sdk.ext.httplib import add_ignored as httplib_add_ignored
+from aws_xray_sdk.core import AWSXRayRecorder
+from aws_xray_sdk.core import patch as xray_patch
+from aws_xray_sdk.core import xray_recorder
 from aws_xray_sdk.core.models.segment import Segment
 from aws_xray_sdk.ext.flask.middleware import XRayMiddleware
-from flask import Flask, request, Response, session
+from aws_xray_sdk.ext.httplib import add_ignored as httplib_add_ignored
+from flask import Flask, Response, request, session
 
 from core.config import Configuration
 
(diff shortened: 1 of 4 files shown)


api/util/xray.py
```diff
@@ -0,0 +1,98 @@
...
+        # Add library shortname
+        if hasattr(request, "library"):
+            segment.put_annotation("library", str(request.library.short_name))
+
+        # Add patron data
+        if hasattr(request, "patron"):
+            segment.set_user(str(request.patron.authorization_identifier))
+            segment.put_annotation(
+                "barcode", str(request.patron.authorization_identifier)
+            )
```
tdilauro, line 92: How does this square with privacy? Might this limit our ability to activate X-Ray in production? If so, could we use a separate environment variable to enable this part?

api/util/xray.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Add admin UI username
+        if "admin_email" in session:
+            segment.set_user(session["admin_email"])
```
tdilauro, line 96: Possible privacy issue here, as well, I guess. It's a little less concerning, since it's not dealing with content.

tdilauro (approved):
Makes sense to me. I need to get up to speed on X-Ray, but I think having this could be very helpful.

A couple of potential privacy concerns, but they wouldn't need to be addressed until/unless we are using in production.

PR #43, opened 2021-09-28
Update to admin UI v0.0.5 for list/lane manager fixes.

tdilauro:
## Description

Update to admin UI v0.0.5 to prevent list and lanes manager columns collapsing when other columns contain long names/titles.

## Motivation and Context

This is affecting the custom list manager for several of our libraries.

## How Has This Been Tested?

Manual testing by modifying values in the browser developer tools.

## Checklist:

- N/A I have updated the documentation accordingly.
- [X] All new and existing tests passed.

PR #77, opened 2021-12-09
Adopt Analytics singleton

tdilauro:
## Description

- Adopts `Analytics` singleton class from [circulation-core #43](https://github.com/ThePalaceProject/circulation-core/pull/43).
- Fixes a test to reload the `Analytics` configuration. 

## Motivation and Context

Improve `Analytics` load performance by using a shared singleton instance.

## How Has This Been Tested?

- Manually testing.
- Existing tests pass (one needed a small nudge).

## Checklist:

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -212,7 +212,7 @@ class CirculationManager(object):
         interface.
         """
         LogConfiguration.initialize(self._db)
-        self.analytics = Analytics(self._db)
+        self.analytics = Analytics(self._db, refresh=True)
         self.auth = Authenticator(self._db, self.analytics)
 
         self.setup_external_search()
--- a/api/google_analytics_provider.py
+++ b/api/google_analytics_provider.py
@@ -151,4 +151,5 @@ class GoogleAnalyticsProvider(object):
         response = HTTP.post_with_timeout(url, params)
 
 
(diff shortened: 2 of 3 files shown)


api/controller.py
```diff
@@ -212,7 +212,7 @@ def load_settings(self):
         interface.
         """
         LogConfiguration.initialize(self._db)
-        self.analytics = Analytics(self._db)
+        self.analytics = Analytics(self._db, refresh=True)
```
vbessonov, line 215: Why do we need to set `refresh` to `True` here? Won't it cause multiple initializations?

tdilauro: I expect this get run only when we are intentionally reloading. From the logs I produced previously, the biggest impacts were coming from loading here:
```
...from_license_source, /Users/Shared/src/tpp/circulation/core/metadata_layer.py, 112: 
... analytics=Analytics(_db) ...
```
and here
```
ODL2API.__init__, /Users/Shared/src/tpp/circulation/api/odl.py, 221: 
self.analytics = Analytics(_db)
```

Having `refresh=True` here allows us to reload when the configuration changes. We can change that if we decide that this is not desirable.

jonathangreen (approved):
🎸🤘

PR #79, opened 2021-12-10
Authenticator performance

jonathangreen:
## Description

Pull changes from https://github.com/ThePalaceProject/circulation-core/pull/44 into circulation.

## Motivation and Context

- Makes authenticator loading faster by warming the cache with the needed integration settings.
- Makes `setup_adobe_vendor_id` use less costly by short-circuiting a loop as soon as conditions are satisfied.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -3,7 +3,7 @@ import os
 from functools import update_wrapper, wraps
 
 import flask
-from flask import Response, make_response, redirect, request
+from flask import g, Response, make_response, redirect, request
 from flask_cors.core import get_cors_options, set_cors_headers
 from werkzeug.exceptions import HTTPException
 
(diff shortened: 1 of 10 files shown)


api/routes.py
```diff
@@ -3,7 +3,7 @@
 from functools import update_wrapper, wraps
 
 import flask
-from flask import Response, make_response, redirect, request
+from flask import Response, g, make_response, redirect, request
```
tdilauro, line 6: `g` doesn't appear to be used anymore.

`redirect`, which was already present, also appears not to be used.

tdilauro:
I added a commit for reducing the `setup_adobe_vendor_id` run from `load_settings`. The elapsed time from the `pyinstrument`ed data was 4.2 seconds and in my own local baselines it hovered around 4 seconds. I was able to get this down to on the order of 0.75s in my environment by splitting up the per-library configuration loops and short circuiting the check that used `setup_adobe_vendor_id`. There are still improvements that can and should be made under the hood here, but this seemed like a good start. Attempts at pre-caching the values did not yield any consistent benefit and did not justify the extra complexity.

I also added a commit to pick up an elapsed time logging context manager that was added to `core`.

tdilauro (approved):
LGTM! 🚀 

A couple minor things to keep the code clean, but nothing that would prevent merging.

api/routes.py
```diff
@@ -48,6 +48,15 @@ def get_locale():
     return request.accept_languages.best_match(languages)
 
 
+@app.after_request
+def print_cache(response):
+    if hasattr(app, "_db") and HasFullTableCache.CACHE_ATTRIBUTE in app._db.info:
```
vbessonov, line 53: Can it add additional overhead? Can we add an additional check to run it only if `DEBUG` level was turned on?

vbessonov (approved):

PR #80, opened 2021-12-10
Add profilers

jonathangreen:
## Description

Some helper functions for optionally profiling our code as it is run. 

## Motivation and Context

These functions were helpful while debugging the issues with CT. I think they were helpful enough that its worth adding them to the main codebase.

### DIFF
--- a/api/util/profilers.py
+++ b/api/util/profilers.py
@@ -34,7 +34,7 @@ class PalacePyInstrumentProfiler(PalaceProfiler):
     ENVIRONMENT_VARIABLE = "PALACE_PYINSTRUMENT"
 
     @classmethod
-    def configure(cls, app):
+    def configure(cls, app: Flask):
         profile_dir = cls.create_profile_dir()
         if profile_dir is None:
             # We are not configured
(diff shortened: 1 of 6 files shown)


tdilauro (commented):
I took a quick look and it looks good at a glance. It would be good to update the README with at least a cursory description of the facilities and how to enable/configure them.

api/util/profilers.py
```diff
@@ -0,0 +1,95 @@
...
+
+class PalacePyInstrumentProfiler(PalaceProfiler):
+    ENVIRONMENT_VARIABLE = "PALACE_PYINSTRUMENT"
+
+    @classmethod
+    def configure(cls, app: Flask):
+        profile_dir = cls.create_profile_dir()
+        if profile_dir is None:
+            # We are not configured
+            return
```
vbessonov, line 41: Does it make sense to log a warning or error?

api/util/profilers.py
```diff
@@ -0,0 +1,95 @@
...
+
+class PalaceCProfileProfiler(PalaceProfiler):
+    ENVIRONMENT_VARIABLE = "PALACE_CPROFILE"
+
+    @classmethod
+    def configure(cls, app: Flask):
+        profile_dir = cls.create_profile_dir()
+        if profile_dir is None:
+            # We are not configured
+            return
```
vbessonov, line 73: Does it make sense to log a warning or error?

vbessonov (approved):

PR #81, opened 2021-12-14
Order admin/libraries endpoint by Library Name

jonathangreen:
## Description

Small change to make sure that the `admin/libraries` endpoint is ordered by Library name. This got dropped during the optimization work in #79. 

## Motivation and Context

The Libraries in the register library screen used to be ordered by Library name. This stopped being the case after the performance optimizations that were recently done. This restores that functionality.

## How Has This Been Tested?

Load local admin interface and make sure the libraries were ordered by name.


tdilauro (approved):
🚀

PR #87, opened 2021-12-21
Feature/merge core

jonathangreen:
## Description

This PR pulls the git history from the `circulation-core` repository into this repository, melding the two together.

🚨 Because of this this pr *should not* be squash merged, so that we don't lose the history from the `circulation-core`.

## Motivation and Context

Since `circulation-core` is not used anywhere else in Palace, it feels unnecessary to have a sub-module that we need to keep up to date and sync with `circulation`. One solution to this problem would be to pull core out as a submodule, and package it as a library.  But given that core isn't used anywhere else, and its logic is tightly coupled with circulation, it seems more desirable to roll it directly into the circulation repo.

This was done here, by merging the two unrelated repos together. This will preserve the history of the files in core, and make it possible (but more difficult for sure) to pull in changes from both circulation and core upstream.

Other then pulling in the `circulation-core` repo, this PR adds one commit that makes the following changes:
- Updates the `.github/workflows/sync.yml` file to sync both `server_core` and `circulation` as different branches to this repo, so we can use those branches in merge in changes from upstream. It also removes the `core/.github/workflows/sync.yml` file.
- Updates `tox.ini` and remove `core/tox.ini`. The new updated `tox.ini` defines two different environments: `circulation` and `core`. These environments run the `circulation` and `core`. For example you can test circulation with the command: `tox -e py38-circulation-docker` and core with `tox -e py38-core-docker`.
- Updates `.github/workflows/test-build.yml` to use the new environments defined in `tox.ini` and remove `core/.github/workflows/test.yml`.
- Update `docker/simplified_app.sh` to remove the part of the build dealing with git submodules.

Once this is merged in, the intent it to archive the https://github.com/ThePalaceProject/circulation-core repo.

## How Has This Been Tested?

All the tests run, and I've built a docker container and verified I can get to the admin interface.


tdilauro (approved):
LGTM! 🤖 🧑‍🦲 🚀 

Caveat: I reviewed key commits and otherwise sampled a bunch of the moved files for patterns of changes.

PR #88, opened 2021-12-22
Use poetry to manage our dependancies

jonathangreen:
## Description

Updates this repo to bring it into line with what we are doing in library registry by using poetry to manage our python dependencies.

While moving the dependencies over I also took a pass through them and removed a number of outdated ones that were no longer used, or that were transitive dependencies added specifically to lock their versions when we were on Python 2.7. 

I went though and updated the documentation for these changes, as well as adding a markdown linter, and doing a general cleanup of the README file.

## Motivation and Context

Try to use the same dependency management system in all of our Python projects.

## How Has This Been Tested?

Ran unit tests locally. Once this branch builds a docker container, I will do some testing with the container to make sure it looks like things are working well there.

## Checklist:

- [x ] I have updated the documentation accordingly.


tdilauro (approved):
Looks good! 🚀 🎄 

Some comments on the README, but nothing major.

.github/actions/poetry/action.yml
```diff
@@ -0,0 +1,19 @@
...
+outputs:
+  version:
+    description: "Installed version"
+    value: ${{ steps.poetry-version.outputs.version }}
+
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        curl -sSL https://install.python-poetry.org | python
```
vbessonov, line 13: I would suggest to lock the `poetry` version:
```
curl -sSL https://install.python-poetry.org | python3 - --version 1.2.0
curl -sSL https://install.python-poetry.org | POETRY_VERSION=1.2.0 python3 -
```
In the past we had many problems with `poetry` breaking after an update.

docker/simplified_app.sh
```diff
@@ -53,12 +58,17 @@ SIMPLIFIED_ENVIRONMENT=/var/www/circulation/environment.sh
 echo "if [[ -f $SIMPLIFIED_ENVIRONMENT ]]; then \
       source $SIMPLIFIED_ENVIRONMENT; fi" >> env/bin/activate
 
+# Install Poetry
+curl -sSL https://install.python-poetry.org | POETRY_HOME="/opt/poetry" python3
```
vbessonov, line 62: In the future it might worth it to use multi-stage build to save the target image size by getting `poetry` out of it.

PR #90, opened 2022-01-03
Clean up most pytest warnings

tdilauro:
## Description

This branch resolves a bunch of warnings seen when running tests. Some highlights:
- Use logging's `warning`, rather than deprecated `warn`.
- Use SQLAlchemy's `LargeBinary`, instead of deprecated `Binary`.
- Don't start non-test class names with "Test".
- Use raw strings for regex patterns with escape sequences.
- Use "postgresql" as DB dialect, instead of deprecated "postgres".
- Fix some broken tests with inert assertions.
- Be more explicit in some falsy tests.
- Warning messages enabled by default.

## Motivation and Context

- Resolve deprecated approaches and move to current ones.
- Reduce cognitive load from unhelpful messages when running tests.

## How Has This Been Tested?

Fixed broken tests. Ran all tests.

## Checklist:

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1896,7 +1896,7 @@ class SettingsController(AdminCirculationManagerController):
             message = _("Exception getting self-test results for %s %s: %s")
             error_message = str(e)
             args = (self.type, item.name, error_message)
-            logging.warn(message, *args, exc_info=error_message)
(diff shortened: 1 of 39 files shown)


jonathangreen:
Since we're not throwing a billion warnings now, maybe we can change the tox.ini to remove the `--disable-warnings` arg.

jonathangreen (approved):
🚀

PR #104, opened 2022-01-04
Tune codeql config

jonathangreen:
## Description

Drop in a CodeQL config file to exclude some of the rules that are on by default. This should reduce the number of false positives we are getting from this scanning.


tdilauro (approved):
LGTM!

PR #110, opened 2022-01-04
Remove sphinx documentation configuration.

jonathangreen:
## Description

Our Sphinx dependency is badly out of date, and the workflow for building docs isn't even targeting the correct branch. Since we aren't using these docs anyway, instead of updating our builds to correctly create documentation just 🔥 the config for now. It shouldn't be too hard to add it back in if we want to generate documentation in the future.


tdilauro (approved):
LGTM! 🎺

PR #111, opened 2022-01-04
Mark lanes test that occasionally fails

jonathangreen:
## Description

As I've been running CI repeatedly for dependencies, I'm seeing one of the lanes tests failing randomly. Its often enough that its causing failures, but not every run. This PR marks it with @pytest.mark.xfail which will display the test as failed in the pytest output, but not cause the whole test run to fail if the test fails. 

This is just a temp fix, the real fix here is to figure out why the test is occasionally failing. I added a todo to note that we should actually fix the test.

## Motivation and Context

Motivated by a deep desire to not have out CI randomly fail.


tdilauro (approved):
👍🏽

PR #132, opened 2022-01-06
Fix OPDS tests that occasionally fail 

jonathangreen:
## Description

Trying to knock off tests that I see occasionally failing. 

## Motivation and Context

These tests do fail now and then because XML serialization isn't guaranteed to be ordered unless you canonicalize it.

## How Has This Been Tested?

Ran tests, did not see them fail.


tdilauro (approved):
🥇

PR #139, opened 2022-01-12
Fix Marketplace Audiobooks

jonathangreen:
## Description

Return a valid content type when we are getting manifests for Marketplace audiobooks.

## Motivation and Context

I don't really love the fix here, but I think it will solve the issue we are seeing with marketplace audiobooks.

## How Has This Been Tested?

It largely hasn't, once CI passes, we can test out the built container with the app and see if it works.

## Checklist

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -559,12 +559,6 @@ class ODLAPI(BaseCirculationAPI, BaseSharedCollectionAPI, HasExternalIntegration
                 continue
             href = link.get("href")
             type = link.get("type")
-
-            # For DeMarque audiobook content, we need to translate the type property
-            # to reflect what we have stored in our delivery mechanisms.
-            if type in ODLImporter.LICENSE_FORMATS:
-                type = ODLImporter.LICENSE_FORMATS[type][ODLImporter.DRM_SCHEME]
-
             candidates.append((href, type))
 
         if len(candidates) == 0:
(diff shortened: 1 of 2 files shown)


tdilauro:
Latest test failures looked like they were caused by GitHub Actions services, so I re-ran them. They all pass now.

tdilauro:
I did some local testing with this PR.

- Before PR returns an invalid `content-type`:
    
    ```
    http://www.feedbooks.com/audiobooks/access-restriction
    ```
    
    - Example...
        
        ```bash
        curl -sSL -u "${TIM_VLC_BASICAUTH}" "https://gorgon.tpp-qa.lyrasistechnology.org/lyrasis-reads/works/33180/fulfill/22" -i
        ```
        
        ```
        HTTP/1.1 200 OK
        Date: Thu, 13 Jan 2022 14:06:19 GMT
        Content-Type: http://www.feedbooks.com/audiobooks/access-restriction
        Content-Length: 13225
        ...
        Server: nginx/1.14.0 (Ubuntu)
        ```
        
- After PR returns a valid `content-type`:
    
    ```
    application/audiobook+json; protection=http://www.feedbooks.com/audiobooks/access-restriction
    ```
    
    - Example:
        
        ```bash
        curl -sSL -u "${TIM_VLC_BASICAUTH}" "[http://host.docker.internal:6500/lyrasis-reads/works/33180/fulfill/22](http://host.docker.internal:6500/lyrasis-reads/works/33180/fulfill/22)" -i
        ```
        
        ```
        HTTP/1.0 200 OK
        Content-Length: 13225
        Content-Type: application/audiobook+json; protection=http://www.feedbooks.com/audiobooks/access-restriction
        ...
        Server: Werkzeug/1.0.1 Python/3.8.6
        ```
        

The `content-type` is not what we indicate on the fulfillment link in the feed, which is:

```
application/audiobook+json;profile="http://www.feedbooks.com/audiobooks/access-restriction"
```

However, I believe this is the same behavior we saw in the Library Simplified CMs.

PR #142, opened 2022-01-12
Feature/update docker build

aseefahmed:
## Description
Currently the docker build for our circulation manager backend needs some updates. Right now it builds pulling in the code from git within the docker build, it should instead be using the locally available files.

## Motivation and Context
- To avoid cloding duplicate codebase in every jobs

## Checklist


docker/Dockerfile.exec
```diff
@@ -24,14 +24,13 @@ ENV SIMPLIFIED_SCRIPT_NAME ""
 
 # Copy over all Palace build files for this image
 COPY . /ls_build
-
-RUN /bin/bash -c "/ls_build/simplified_app.sh ${repo} ${version} \
-      && /ls_build/logrotate.sh \
+RUN /bin/bash -c "/ls_build/docker/simplified_app.sh ${repo} ${version} \
```
jonathangreen, line 27: Do `${repo}` and `${version}` do anything here now?

.github/workflows/test-build.yml
```diff
@@ -259,7 +256,7 @@ jobs:
       - name: Push circ-webapp image
         uses: docker/build-push-action@v2
         with:
-          context: ./docker
+          context: .
```
jonathangreen, line 259: Do the build args below still have to be passed into this action?

docker/Dockerfile.exec
```diff
@@ -24,14 +24,13 @@ ENV SIMPLIFIED_SCRIPT_NAME ""
...
+      && /ls_build/docker/logrotate.sh \
       && rm -rf /ls_build && /bd_build/cleanup.sh"
 
 VOLUME /var/log
 WORKDIR /home/simplified/circulation/bin
-
+RUN echo ${SIMPLIFIED_DB_TASK}
```
jonathangreen, line 33: Is this line needed?

docker/Dockerfile.webapp
```diff
@@ -10,10 +10,10 @@ ENV LANG=C.UTF-8
 # Copy over all Palace build files for this image
 COPY . /ls_build
```
jonathangreen, line 11: Instead of adding docker to the path everywhere, we could change this `COPY` command to be:
`COPY ./docker /lsbuild`

docker/simplified_app.sh
```diff
@@ -40,10 +41,9 @@ export LANG=C.UTF-8
 useradd -ms /bin/bash -U simplified
 
 # Get the proper version of the codebase.
-mkdir /var/www && cd /var/www
-git clone https://github.com/${repo}.git circulation
-chown simplified:simplified circulation
-cd circulation
+mkdir -p /var/www/circulation && rsync -rtv /ls_build/ /var/www/circulation
```
jonathangreen, line 44: instead of doing an rsync here, couldn't we copy the files directly into this path: 
COPY . /var/www/circulation

jonathangreen (changes requested):

jonathangreen:
One more comment here, can you update the README file, since we no longer need the repo and version arguments.

aseefahmed:
> 

I will do once tested.

docker/Dockerfile.webapp
aseefahmed: i tested. it will not work. it should copy all the files including './docker'.

.github/workflows/test-build.yml
```diff
@@ -74,6 +74,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          clean: false
```
jonathangreen, line 78: Why is this needed?

docker/Dockerfile.webapp
jonathangreen: What happens in this case? Why won't it work?

jonathangreen (changes requested):
A couple more comments here. I did test locally, and the image is working for me. Should be good to merge after these changes.

.github/workflows/test-build.yml
```diff
@@ -130,9 +129,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Docker Buildx
```
jonathangreen, line 133: Why remove this action?

jonathangreen (changes requested):

docker/Dockerfile.webapp
aseefahmed: the whole project needs to be copied into /lsbuild, not just the docker directory

.github/workflows/test-build.yml
aseefahmed: actually, its not needed, i will remove those lines.

docker/Dockerfile.webapp
jonathangreen: I don't understand why this is the case. We did not copy the whole project into lsbuild before, why is it needed now?

aseefahmed: ok. i will do double check if i miss anything

.github/workflows/test-build.yml
aseefahmed: you right, it shouldn't be removed. i'll undo.

jonathangreen:
@aseefahmed I pushed a commit here to not copy everything into the `ls_build` folder. If you can fix those other issues, then I will test again. I think this is close to ready to merge.

docker/Dockerfile.webapp
aseefahmed: seems working now.

jonathangreen (approved):
Tested this locally and it looks good

PR #143, opened 2022-01-13
Don't add timezone to date that already has one.

tdilauro:
## Description

Fixes a crash in `overdrive_new_titles` monitor script caused by trying to add a timezone to a `datetime` that already has one.

## Motivation and Context

The monitor's `start` timestamp originates in a `timestamp with time zone` field in the `timestamps` table, so should be either `None` or a `datetime` with a time zone. It does not need to be converted.

## How Has This Been Tested?

- Manually ran the script on a local instance on which it had previously failed.
- Tested on a QA server.

## Checklist

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -5,7 +5,6 @@ import urllib.parse
 
 import dateutil
 import flask
-import pytz
 from flask_babel import lazy_gettext as _
 
 from core.analytics import Analytics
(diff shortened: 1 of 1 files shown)


api/overdrive.py
```diff
@@ -1364,10 +1363,6 @@ def should_stop(self, start, api_description, is_changed):
             self.log.error("Got invalid date: %s", date_added)
             return False
 
-        # The time stored in the database is UTC, but it's stored
-        # without any time zone information. Add that information so
-        # we can compare it against the date we got from Overdrive.
-        start = pytz.utc.localize(start)
```
jonathangreen, line 1370: Do we know that in all cases this has timezone info attached? If not maybe we should use this helper that handles the case without timezone and the case with timezone: 
https://github.com/ThePalaceProject/circulation/blob/main/core/util/datetime_helpers.py#L35-L49

On the other hand this is much more clear if we know we always have a timezone here.

jonathangreen (approved):

api/overdrive.py
tdilauro: Good question. I verified that `start` originates in a `timestamp with time zone` field in the `timestamps` table, so should be either `None` or a `datetime` with a time zone.

PR #151, opened 2022-01-20
Drop Python 3.6 support

jonathangreen:
## Description

This PR drops Python 3.6 support from circulation. I'm leaving it in draft for now, since we need to start building our container images using a newer version of Python before it can be merged.

## Motivation and Context

Python 3.6 is [end-of-life as of 2021-12-23](https://devguide.python.org/devcycle/#end-of-life-branches). Packages are starting to drop support for it, and we should as well.

## How Has This Been Tested?

- Ran unit tests
- Read over Readme to find references to Python 3.6

## Checklist

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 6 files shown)


jonathangreen:
Going to mark this as ready to review, since our docker images are being built with Python 3.8 now. 🎉

tdilauro (approved):
Looks good! 🎸🤘🏽

PR #152, opened 2022-01-20
Feature/upgrade pyjwt

RishiDiwanTT:
## Description

Upgraded PyJWT to 2.3.0

## Motivation and Context

PyJWT was quite outdated (1.4.2)

## How Has This Been Tested?

Ran all unit tests (only api module uses jwt though)

## Checklist

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -4,7 +4,10 @@ import datetime
 import json
 import logging
 import sys
+import time
 import uuid
+
+from jwt.exceptions import InvalidIssuedAtError
 
 import flask
 import jwt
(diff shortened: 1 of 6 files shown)


api/adobe_vendor_id.py
```diff
@@ -892,9 +897,15 @@ def decode(self, authdata):
...
         # First, decode the authdata without checking the signature.
         decoded = jwt.decode(
-            authdata, algorithm=self.ALGORITHM, options=dict(verify_signature=False)
+            authdata,
+            algorithms=[self.ALGORITHM],
+            options=dict(verify_signature=False, verify_exp=True),
         )
 
+        # Fail future JWTs as per requirements, pyJWT stopped doing this, so doing it manually
+        if "iat" in decoded and decoded["iat"] > time.time():
```
jonathangreen, line 906: I think we need to make sure that `time.time()` is giving us a unix timestamp. It does on most systems, but its not guaranteed. It looks like the [pyjwt library uses](https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L133)
```python
now = timegm(datetime.now(tz=timezone.utc).utctimetuple())
``` 
to get the now value it uses in comparisons.

jonathangreen (changes requested):
Looks good to me other then the minor change I noted to how we get the timestamp for comparison. Also going to tag @tdilauro for a review, since he knows the registry code better then I do.

jonathangreen (approved):

jonathangreen:
Looks good to me

PR #155, opened 2022-01-21
Update common docker image

aseefahmed:
## Description

Right now, the three docker images that we produce do not share any layers. They are all built independently. Since these images are largely the same, we can build them using a multistage build, where each image shares a common base image. 

## Motivation and Context

Update Dockerfile.exec, Dockerfile.scripts and Dockerfile.webapp to use a multistage build where they share a common base image, and integrate this new build into our github action CI so that we are pushing images based on this new build process.

## How Has This Been Tested?

Multistage docker build that is the base image for each of our containers


docker/Dockerfile.base
```diff
@@ -0,0 +1,8 @@
+FROM phusion/baseimage:focal-1.0.0
```
jonathangreen, line 1: Instead of adding a new file `Dockerfile.base`, I would like to see all the files combined into a single Dockerfile, with different named images.

For example you could change this line to:
```
FROM phusion/baseimage:focal-1.0.0 as base
```

Then later in the docker file use
```
from base as webapp
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,8 @@
...
+
+# Copy over all Palace build files for this image
+COPY ./docker /ls_build
```
jonathangreen, line 4: After this layer, you should be able to do a lot of the setup that doesn't require the source code, so that we have some layer caching.

docker/Dockerfile.exec
```diff
@@ -12,20 +12,14 @@ RUN go get -v github.com/readium/readium-lcp-server/lcpencrypt
 ## Final image
 ###############################################################################
 
-FROM phusion/baseimage:focal-1.0.0
+FROM ghcr.io/REPOSITORY_OWNER/docker-base:TAG
```
jonathangreen, line 15: Once you combine the dockerfiles, you don't have to template any names like this.

.github/workflows/test-build.yml
```diff
@@ -61,10 +61,60 @@ jobs:
         env:
           MODULE: ${{ matrix.module }}
 
+  build-base-image:
```
jonathangreen, line 64: Since the build time should be fairly fast once the docker files are compressed into a single one, you should be able to define a single job here: `build-docker` and build and push all the containers within that job.

.github/workflows/test-build.yml
```diff
@@ -61,10 +61,60 @@ jobs:
...
+        id: base-tags
+        uses: docker/metadata-action@v3
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/docker-base
+          tags: |
+            type=semver,pattern={{major}}.{{minor}},priority=10
+            type=semver,pattern={{version}},priority=20
+            type=ref,event=branch,priority=30
+            type=sha,priority=40
+      - name: Build & push base docker image
```
jonathangreen, line 103: I don't think we need to push the base image, as we won't use it directly anywhere.

.github/workflows/test-build.yml
```diff
@@ -97,20 +147,25 @@ jobs:
             type=semver,pattern={{version}},priority=20
             type=ref,event=branch,priority=30
             type=sha,priority=40
-
+      - name: Update base image
+        run: |
+          echo ${GITHUB_SHA::7}
+          repo_owner=$(echo "$GITHUB_REPOSITORY_OWNER" | awk '{print tolower($0)}')
+          sed -i.bak "s/REPOSITORY_OWNER/${repo_owner}/" docker/Dockerfile.exec
+          sed -i.bak "s/TAG/sha-${GITHUB_SHA::7}/" docker/Dockerfile.exec
```
jonathangreen, line 155: This can be removed using a single file

.github/workflows/test-build.yml
```diff
@@ -97,20 +147,25 @@ jobs:
...
       - name: Build & push circ-exec image
         uses: docker/build-push-action@v2
```
jonathangreen, line 157: Can choose which build stage in the single docker file to build here

jonathangreen (changes requested):

docker/Dockerfile.base
jonathangreen: This will require splitting the work done in `simplified_app.sh` into two scripts, one that does setup not requiring source code, and one that requires the application source code. That way we can copy the docker setup, do the general OS setup (installing packages etc), then have a second script that actually sets up the app using our source code.

.github/workflows/test-build.yml
```diff
@@ -61,10 +61,60 @@ jobs:
...
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Cache Docker layer
```
jonathangreen, line 86: Can we use either the [Cache Backend API](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api) or [Registry Cache](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#registry-cache) here instead?

aseefahmed:
> 

Noted Jonathon. I'll update that accordingly soon.

aseefahmed:
> 

@jonathangreen i found these sequences even more feasible :

Set up Docker
Generate tags
Create version file
Build scripts, then test this image, 
Build webapp, then test this image
Push exec image
Push scripts image
Push webapp image

Any thoughts?

jonathangreen:
That sequence makes sense to me.

aseefahmed:
> That sequence makes sense to me.

its done now. can you please have a quick review @jonathangreen

docker/Dockerfile
```diff
@@ -12,23 +12,57 @@ RUN go get -v github.com/readium/readium-lcp-server/lcpencrypt
 ## Final image
```
jonathangreen, line 12: This comment looks out of date, its not the final image being built here, its the base image. Can you add comments like this for the other images in this dockerfile?

Something like: 
```
###############################################################################
# Base image
###############################################################################
```

.github/workflows/test-build.yml
```diff
@@ -98,56 +106,41 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - name: Generate tags for circ-webapp image
+        id: webapp-tags
+        uses: docker/metadata-action@v3
         with:
-          persist-credentials: false
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+          images: ghcr.io/${{ github.repository_owner }}/circ-webapp
```
jonathangreen, line 124: Since these are only used when we are pushing images, can you move the tag generation for each image to right before we push the image?

docker/Dockerfile
```diff
@@ -12,23 +12,57 @@ RUN go get -v github.com/readium/readium-lcp-server/lcpencrypt
...
+
+VOLUME /var/log
+WORKDIR /home/simplified/circulation/bin
+
+CMD ["/sbin/my_init"]
+
+FROM base as webapp
+
+ENV SIMPLIFIED_DB_TASK "ignore"
+ENV LANG=C.UTF-8
```
jonathangreen, line 58: This line can be removed

docker/Dockerfile.base
jonathangreen, line 1: This file isn't necessary anymore

docker/README.md
```diff
@@ -23,6 +23,18 @@ $ docker run --name webapp -d \
     ghcr.io/thepalaceproject/circ-webapp:main
 ```
 
+If the database and ElasticSearch(ES) are running in containers, use the --link option to let the webapp docker container
```
jonathangreen, line 26: Need to update the build commands in the README file to reflect the new Dockerfile. They are referencing Dockerfile.webapp etc.

jonathangreen (changes requested):
This is generally looking good. I requested a couple more changes, then I think we can merge this. We are still not doing a good job with layer caching, but I will add an additional ticket to that for notion, and we can work on that after this is merged.

docker/README.md
aseefahmed: DOEN

jonathangreen (approved):

PR #157, opened 2022-01-21
Remove clever auth

jonathangreen:
## Description

My ongoing quest to remove code from the app that we are not actively using. Decided to remove Clever auth, which is the only implementation of patron oauth.

## Motivation and Context

I'd like to reduce the amount of code in the CM that we are not running in our hosting environment. To make it a bit easier to restructure the code that is existing. 

## How Has This Been Tested?

- Ran unit tests

## Checklist


tests/api/test_authenticator.py
```diff
@@ -1080,15 +845,13 @@ def test_authenticated_patron_unsupported_mechanism(self):
...
         basic = MockBasicAuthenticationProvider(self._default_library, integration)
-        oauth = MockOAuthAuthenticationProvider(self._default_library, "oauth1")
 
         # We can pull the password out of a Basic Auth credential
         # if a Basic Auth authentication provider is configured.
         authenticator = LibraryAuthenticator(
             _db=self._db,
             library=self._default_library,
             basic_auth_provider=basic,
-            oauth_providers=[oauth],
```
tdilauro, line 1091: Do we still need the bearer token signing secret on the next line if we don't have any SAML providers and we no longer have any OAuth providers?

Same below on lines 1103-1104 and possibly in other places.

tdilauro (commented):
This looks good. One question regarding the need for a signing secret in some tests.

jonathangreen:
I'll close this since its old and stale. If we want to get rid of this code we can revive it.

jonathangreen:
Back from the dead

jonathangreen:
I've been working through some of the authentication code, and I think we should pull this auth out. Even if we eventually add it back in if its needed. Whats here is pretty broken, and its hard to refactor to make not broken without api docs / guidance.

I fixed the code review comments you left from last time @tdilauro if you wouldn't mind taking another look.

tdilauro (approved):
Looks good!

I'm approving now, but I think [these problem details](https://github.com/ThePalaceProject/circulation/blob/77a24f9d1464269072c8132cd00fcc63b848b965/api/problem_details.py#L246-L267) can probably go, as well:
```python
INVALID_OAUTH_CALLBACK_PARAMETERS = pd(
    "http://librarysimplified.org/terms/problem/invalid-oauth-callback-parameters",
    status_code=400,
    title=_("Invalid OAuth callback parameters."),
    detail=_(
        "The OAuth callback must contain a code and a state parameter with the OAuth provider name."
    ),
)

UNKNOWN_OAUTH_PROVIDER = pd(
    "http://librarysimplified.org/terms/problem/unknown-oauth-provider",
    status_code=400,
    title=_("Unknown OAuth provider."),
    detail=_("The specified OAuth provider name isn't one of the known providers."),
)

INVALID_OAUTH_BEARER_TOKEN = pd(
    "http://librarysimplified.org/terms/problem/credentials-invalid",
    status_code=400,
    title=_("Invalid OAuth bearer token."),
    detail=_("The provided OAuth bearer token couldn't be verified."),
)
```

PR #161, opened 2022-01-24
Update uwsgi configuration

jonathangreen:
## Description

Add an additional configuration option to UWSGI `die-on-term` that causes UWSGI to respect `sigterm` that happens on docker shutdown.

This PR also updates the UWSGI configuration settings included in the container to match what we are setting in the playbook. Once this is merged, we can drop our custom uwsgi configuration in the playbook.

## Motivation and Context

In the current configuration, this causes a reload, so when we shutdown the docker container, UWSGI reloads, docker waits for it to stop, then eventually times out and forcibly shuts it down. This will speed up our docker shutdown process.

Also centralize the management of the uwsgi.ini file to this repo, instead of having it overridden in the playbook.

## How Has This Been Tested?

Started, stopped the docker container. Looked at logs. Made sure that UWSGI is restarting as expected instead of timing out as it was before.

## Checklist

### DIFF
--- a/docker/services/uwsgi.d/60_processes.ini
+++ b/docker/services/uwsgi.d/60_processes.ini
@@ -1,0 +1,4 @@
+[uwsgi]
+# Default number of processes and threads for the app
+processes = 3
+threads = 2
--- a/docker/services/nginx.conf
+++ b/docker/services/nginx.conf
@@ -9,6 +9,6 @@ server {
     location @circulation {
         include uwsgi_params;
         uwsgi_read_timeout 120;
-        uwsgi_pass unix:/var/www/circulation/uwsgi.sock;
+        uwsgi_pass unix:/var/run/uwsgi/uwsgi.sock;
     }
 }
--- a/docker/services/uwsgi.d/40_log.ini
+++ b/docker/services/uwsgi.d/40_log.ini
@@ -1,0 +1,4 @@
+[uwsgi]
+# location of log files
+logto = /var/log/uwsgi/uwsgi.log
+log-format = %(var.HTTP_X_FORWARDED_FOR) (%(addr)) - - [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)" host_hdr=%(host) req_time_elapsed=%(msecs)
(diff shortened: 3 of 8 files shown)


tdilauro (approved):
Looks good! 🚀

docker/services/uwsgi.d/60_processes.ini
```diff
@@ -0,0 +1,4 @@
+[uwsgi]
+# Default number of processes and threads for the app
+processes = 3
+threads = 2
```
RishiDiwanTT, line 4: This looks like a low number, if we are API/DB intensive 10~15 can run. Although the real number comes out with iterative testing

docker/services/uwsgi.d/60_processes.ini
```diff
@@ -0,0 +1,4 @@
+[uwsgi]
+# Default number of processes and threads for the app
+processes = 3
```
RishiDiwanTT, line 3: Should this be something like (no of cores - [some number of free cores])
https://uwsgi-docs.readthedocs.io/en/latest/Changelog-1.9.20.html#simple-math-in-configuration-files

docker/services/uwsgi.d/60_processes.ini
RishiDiwanTT: We've also not set enable-threads=true, It used to be false by default

jonathangreen: Agreed, these will need to be adjusted as we run the service. I'd like to be able to use the [uwsgi cheaper subsystem](https://uwsgi-docs.readthedocs.io/en/latest/Cheaper.html) to dynamically scale the number of worker processes that we are running, but the large amount of work we do when bringing a new worker online when we create a new [CirculationManager](https://github.com/ThePalaceProject/circulation/blob/881d3b5815b18721212b7d4eb7f14956efac597b/api/controller.py#L130-L147) object means it takes a long time to start a new worker, which limits our options.

docker/services/uwsgi.d/60_processes.ini
jonathangreen: We run many instances of the circulation manager codebase on a single server, since each circulation manager can only handle a subset of the libraries we run currently.

docker/services/uwsgi.d/60_processes.ini
jonathangreen: > We've also not set enable-threads=true, It used to be false by default

When uwsgi is starts with this configuration i see this in the logs: `python threads support enabled`. I believe that means we don't need to set `enable-threads`.

PR #163, opened 2022-01-25
Upgrade SQLAlchemy

io7m:
## Description

This implements an (incomplete) set of changes in order to upgrade to SQLAlchemy 1.4.29. There's an outstanding issue that I'll explain further down. Briefly, the changes are:

* The core/util/__init.py__'s fast_query_count() method has been updated to remove a six year old performance hack which appears to be incorrect in the 1.4 SQLAlchemy branch. This method is only used in one place in the code outside of the test suite, but there may be performance implications there as the method is no longer "fast"; it's just a standard SELECT count()
query now.

* Adjusts some tests to use explicit byte strings rather than plain strings. The newer versions of the packages we're using seem to have gotten stricter with this, and will result in exceptions being raised somewhere inside SQLAlchemy such as: `TypeError: can't escape str to binary`

* The `_distinct` property has been removed and replaced with the `_distinct_on` property, so some tests that were using the old property were broken. Additionally, the SQLCompiler API no longer has a `compile` method; the constructor does the work instead.
See: sqlalchemy/sqlalchemy#4643

* Switch to using 'select()' instead of the old-style Select class.

The remaining issue may require some architectural work to fix. Essentially, the old 1.3 API discouraged (but allowed) listeners added with `@validates` and annotations such as `@event.listens_for(Work.license_pools, "append")` to cause collections to be loaded during event processing. The 1.4 API does not allow this anymore, and you'll get an assertion failure if you try to do it.
The good news is that it seems like this only happens for code in the test suite and, even then, only seems to affect four tests. A good example of this is this test: 

https://github.com/ThePalaceProject/circulation/blob/main/tests/core/test_lane.py#L3260

The `pool.work = work` assignment will, indirectly, trigger this code:

https://github.com/ThePalaceProject/circulation/blob/main/core/model/listeners.py#L190

This will cause this code to be executed:

https://github.com/ThePalaceProject/circulation/blob/90ad5afea73fa8b998b5ffbcbe10426a7153e508/core/model/work.py#L1217

... and in the process, the `_reset_coverage` method will cause a collection of coverage records to be loaded by the ORM. This is specifically disallowed by the 1.4 API, because it could potentially lead to an infinitely-recursing disaster of collections being loaded in response to events.

The error you'll actually see is this line inside SQLAlchemy:

https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/orm/attributes.py#L1416

There's some discussion on it here:

https://github.com/sqlalchemy/sqlalchemy/discussions/7605

As far as I can make out, this problem is limited to the test suite; code isn't normally operating such that coverage records haven't already been loaded.

We _might_ be able to circumvent this by just updating coverage records using the Core API instead of the ORM.
We _could_ also disable the ~four~ six affected tests.
Ideally, updating external indices wouldn't happen this way. We might want to move to some form of asynchronous work queue.

Unfortunately, this is more than I know how to do right now on this codebase!

## Motivation and Context

This may help with some deadlock and performance issues.

## How Has This Been Tested?

I've executed all of the docker tests locally. I've maintained a list of which tests should fail in each commit.

## Checklist


jonathangreen:
I think I'm going to close this one 😢, since we need some architectural work to do this upgrade, and this PR is fairly far out of date at this point. 

Will leave the branch here, so we can reference it when we make the needed changes.

PR #164, opened 2022-01-25
Feature/fix syslog warnings

aseefahmed:
## Description

When starting a container after the update to focal we are getting warnings from Syslog-NG about the configuration file format. Update this format to stop warnings from occurring.

## How Has This Been Tested?

container that starts without this syslog-ng warning


aseefahmed:
@jonathangreen to review

jonathangreen:
This works, but looking upstream, at the https://github.com/phusion/baseimage-docker repo, it seems that this is fixed by this PR https://github.com/phusion/baseimage-docker/pull/591. Which was released in the `focal-1.1.0` image. You should just be able to update the image we use to `phusion/baseimage:focal-1.0.0` to get these changes.

aseefahmed:
> This works, but looking upstream, at the https://github.com/phusion/baseimage-docker repo, it seems that this is fixed by this PR [phusion/baseimage-docker#591](https://github.com/phusion/baseimage-docker/pull/591). Which was released in the `focal-1.1.0` image. You should just be able to update the image we use to `phusion/baseimage:focal-1.0.0` to get these changes.

Yeah that would be better. i will make another commit shortly.

jonathangreen (approved):

PR #165, opened 2022-01-26
Feature/docker arm versions

aseefahmed:
## Description

Right now we only build our docker images for a single platform. We should also build ARM versions of our docker images for use with M1 Macbooks and AWS ARM instances.

## Motivation and Context

Add the additional parameters to our docker builds to build and push images for ARM platforms as well as our current builds.

## How Has This Been Tested?

PR with images building for ARM as well as x86


aseefahmed:
@jonathangreen to review

jonathangreen:
This builds and pushes an arm version of the image, but it doesn't do any tests on it. Can you also run the tests scripts on the arm versions of the image?

jonathangreen (changes requested):
This builds and pushes an arm version of the image, but it doesn't do any tests on it. Can you also run the tests scripts on the arm versions of the image?

aseefahmed:
> This builds and pushes an arm version of the image, but it doesn't do any tests on it. Can you also run the tests scripts on the arm versions of the image?

will have a try

aseefahmed:
@jonathangreen to review

.github/workflows/test-build.yml
```diff
@@ -98,32 +101,85 @@ jobs:
       - name: Create version file
         run: git rev-parse --short HEAD > .version
 
-      - name: Build circ-scripts image
+      - name: Build circ-scripts image (linux/amd64)
```
jonathangreen, line 104: Instead of copy pasting these steps you should be able to use a build matrix here.

jonathangreen:
```yaml
    strategy:
      fail-fast: false
      matrix:
        platform: ["linux/amd64", "linux/arm64"]
```

jonathangreen (changes requested):

aseefahmed:
@jonathangreen to review

jonathangreen (approved):

PR #167, opened 2022-01-27
Upgrade to Pillow 9.0.0

io7m:
## Description

This upgrades to Pillow 9.0.0. This required updating one test image
for a failing test, as the rendered text evidently has ever-so-slightly
different spacing when rendered by the new Pillow. This also reduced
the tolerance for the RMS check in the affected test, because it was
set unnecessarily high due to a mistake in the code ("Authpr" →
"Author").

I've updated the test to save "before" and "after" images to temporary
files, in case this breaks again. The images are removed if the test
completes successfully, but left intact if it doesn't.

## Motivation and Context

https://www.notion.so/lyrasis/Update-Pillow-0a2173702e9f4631b6ae9b3f69441145

## How Has This Been Tested?

Ran the docker tests locally.

## Checklist


io7m:
Looks like reducing the RMS check was too much. I'll set it back to `10`.

jonathangreen (approved):
Looks good to me!

tests/api/admin/controller/test_work_editor.py
```diff
@@ -1007,6 +1032,11 @@ def test_process_cover_image(self):
         )
         assert root_mean_square < 10
 
+        # Remove temporary files if we've gotten this far. Assertion failures should leave
+        # the files intact for manual inspection.
+        for f in [tmpfile_before_0, tmpfile_before_1, tmpfile_after_0, tmpfile_after_1]:
```
RishiDiwanTT, line 1037: A failing test case would fail cleanup, maybe have this in a teardown function or in exception handling

io7m: That was intentional; if the test fails, we want those images left in the filesystem so that they can be inspected visually.

tests/api/admin/controller/test_work_editor.py
```diff
@@ -1003,7 +1003,7 @@ def test_process_cover_image(self):
             )
             / len(image_histogram)
         )
-        assert root_mean_square < 0.01
+        assert root_mean_square < 10
```
tdilauro, line 1006: @io7m Would you add a comment here about how you arrived at this RMS threshold? Also, it's used a couple of times, so it could be helpful to add a descriptive variable and put the explanation on that.

PR #170, opened 2022-01-28
Update Dockerfile

aseefahmed:
## Description

Right now, docker chacing is not doing as expected. We need to find a way to improve docker caching layers and minimize build time. 

## Motivation and Context

Almost the entire `simplified_app.sh` file does not actually require source code. It does an apt-get update, and installs some packages. Then it installs python poetry. All of this is reasonably time consuming and does not need source code. 

Then we pull in the project dependancies. This also does not require source code, but this does require the the poetry files (`pyproject.toml` and `poetry.lock`). 

Then the final step requires our source code. If we split the docker build process into layers like this, then we can more effectively use layer caching. 

The proposal here is to split base into 3 different layers:

- Setup not requiring any of our code
- Setup requiring our dependancies files
- Setup requiring our code

## How Has This Been Tested?

Containers have been run and tested


aseefahmed:
@jonathangreen  to review

docker/Dockerfile
```diff
@@ -14,12 +14,17 @@ RUN go get -v github.com/readium/readium-lcp-server/lcpencrypt
 
 FROM phusion/baseimage:focal-1.1.0 As base
 
+COPY ./docker/simplified_app.sh /ls_build/
```
jonathangreen, line 17: The files in the docker folder don't change very often. You should be able to still do 

```dockerfile
COPY ./docker /ls_build
RUN /bin/bash -c "/ls_build/simplified_app.sh \
      && /ls_build/logrotate.sh"
``` 

You can remove the parts of `simplified_app.sh` that depend on `poetry.lock` & `pyproject.toml` and put them into a different script.

docker/Dockerfile
```diff
@@ -14,12 +14,17 @@ RUN go get -v github.com/readium/readium-lcp-server/lcpencrypt
...
+COPY --chown=1000:1000 poetry.lock pyproject.toml /var/www/circulation/
```
jonathangreen, line 18: Then after this line you can run your new script that installs the dependencies using poetry

jonathangreen (changes requested):

aseefahmed:
@jonathangreen to review. it seems not need to create a seperate script for poetry if we use 

```dockerfile
COPY ./docker /ls_build
COPY --chown=1000:1000 poetry.lock pyproject.toml /var/www/circulation/
```
`

jonathangreen:
We don't need a seperate script for poetry, but if we are trying to make our layer caches as efficient as we can, then it makes sense to have a layer with: 
- The main os + updates + packages we need installed
- The python dependencies we need installed
- The app installed

Since the OS packages don't change very often, the python dependencies change more often, but not that often, and the code changes every time.

aseefahmed:
how is that now @jonathangreen 

```dockerfile
COPY ./docker /ls_build
COPY --chown=1000:1000 poetry.lock pyproject.toml /var/www/circulation/

RUN /bin/bash -c "/ls_build/simplified_app.sh"

RUN /bin/bash -c "/ls_build/install_python_packages.sh \
      && /ls_build/logrotate.sh"

COPY --chown=1000:1000 . /var/www/circulation
```

aseefahmed:
@jonathangreen to reivew

jonathangreen (approved):

PR #171, opened 2022-01-28
Add permissions to workflows

jonathangreen:
## Description

When Dependabot pushes branches, or automatically merges PRs, it currently doesn't have permissions to create docker images, causing builds to fail.

## Motivation and Context

Dependabot is by default given only read permissions on repositories, but it respects the repos `permissions` set in the workflow [[1]](https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/). In order to be able to push docker images it needs `packages: write` permission. 

You can set explicit workflow permissions by setting a permissions key in the workflows [[2]](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/). Setting these should actually be more secure then our previous configuration, as the defaults are quite permissive. 

I was worried that this might give too much permissions to PRs from forks, but we don't run the docker-build workflow on forks and from the blog post: 
> Pull requests from public forks are still considered a special case and will receive a read token regardless of these settings.

## How Has This Been Tested?

Since this is scoped to our actions workflows, the CI passing on this PR should give confidence this is working.

## Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,6 +7,8 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -8,6 +8,9 @@ on:
 jobs:
   sync_with_nypl:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      actions: write
 
     steps:
       - name: Checkout repo to sync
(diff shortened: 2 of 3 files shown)


tdilauro (approved):
This makes sense to me.

One question is whether we should test this with a more restrictive default Workflow Permission set. The concern I have is that, because we have a more permissive Workflow Permissions default in both [this repo](https://github.com/ThePalaceProject/circulation/settings/actions) and [the org](https://github.com/organizations/ThePalaceProject/settings/actions), the workflows might succeed now, but fail later if/when we change one of those.

jonathangreen:
My understanding from https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions is that once you set a permission in a workflow, you override all the defaults, and don't get any permissions other then the ones you specify. So changing org level permissions shouldn't change the perms in this repo.

PR #172, opened 2022-01-30
Add support for WAYFless URLs

vbessonov:
## Description

This PR adds support for WAYFless URLs.

## Checklist


tdilauro (commented):
@vbessonov I found a potential source of the test failures. I'd especially like to confirm your intentions for the return from `SAMLWAYFlessAcquisitionLinkProcessor.fulfil` in some cases.

api/saml/wayfless.py
```diff
@@ -0,0 +1,179 @@
...
+        self, db: sqlalchemy.orm.session.Session
+    ) -> ExternalIntegration:
+        """Return an ExternalIntegration object associated with the collection with a WAYFless url.
+
+        :param db: SQLAlchemy session
+        :return: ExternalIntegration object
+        """
+        return get_one(db, ExternalIntegration, id=self._external_integration_id)
+
+    def fulfil(
```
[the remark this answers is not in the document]
vbessonov: Thanks for noticing it. I found quite a lot of similar typos in the code and tried to fix all of them.

api/saml/wayfless.py
```diff
@@ -0,0 +1,179 @@
...
+                    urllib.parse.quote(fulfillment.content_link, safe=""),
+                )
+
+                self._logger.debug(
+                    f"Old acquisition link {fulfillment.content_link} has been transformed to {acquisition_link}"
+                )
+
+                fulfillment.content_link = acquisition_link
+
+                return fulfillment
```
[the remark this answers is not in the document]
vbessonov: There are the following cases:
1. No `acquisition_link_template`. In this case CM must return the existing `FulfillmentInfo` object.
2. `acquisition_link_template` exists but there is no `saml_credential`. In this case I think we need to raise an error: we expect the patron to be authenticated using SAML but there is no `Credential` object in the database. It probably means that there is some configuration error and either SAML Auth Provider or collection's `acquisition_link_template` are not configured properly. Returning the existing `FulfillmentInfo` object will not work because content provider expects a WAYFless URL.
3. `acquisition_link_template`, `saml_credential` are present but there is no `saml_subject.idp`. In this case we need to raise an error because we cannot form the WAYFless URL without IdP's entityID. Returning the existing `FulfillmentInfo` object will again not work because the content provider expects a WAYFless URL.

io7m:
I just pushed a commit that reorganizes at least some of the system into a mixin-based system. It's not expected to be usable in this form yet, and there are test failures. Thought I'd get the work so far pushed so that I can go back to working on LCP.

io7m:
This appears to be ready to review.

tdilauro:
@io7m Merging #180 added some conflicts. ☹️

io7m:
This is even more ready for review than before.

api/odl2.py
```diff
@@ -16,18 +16,13 @@
...
-)
-from core.opds_import import ConnectionConfiguration
+from core.opds2_import import OPDS2Importer, OPDS2ImportMonitor, RWPMManifestParser
+from core.opds_import import OPDSImporterConfiguration
 from core.util import first_or_default
 from core.util.datetime_helpers import to_utc
 
 
-class ODL2APIConfiguration(OPDS2ImporterConfiguration, ConnectionConfiguration):
+class ODL2APIConfiguration(OPDSImporterConfiguration):
```
RishiDiwanTT, line 25: This change looks a little suspect, but I don't understand the whats actually going on here

io7m: Currently trying to get the IDE to show me `git blame` information. I don't _think_ that change was one of mine, but it seems harmless. There seems to have been a general move to unify OPDS and OPDS2 configuration information, and ODL is compatible with both OPDS 1.2 and OPDS 2.0.

RishiDiwanTT: Don't we lose connection retry counts without ConnectionConfiguration?

RishiDiwanTT: Ok I see this is added to OPDSImporter

RishiDiwanTT:
I can see no issues with the code  :+1:

tdilauro:
I am going to merge this PR with a couple of caveats:
- This PR contains changes to configuration settings support that are needed elsewhere, so there is pressure to merge this code. This part of the PR appears to work properly.
- I am not 100% sure that the WAYFless URL portion of this PR works. I was unable to complete testing with the [SAML testbed](https://github.com/ThePalaceProject/circulation-saml-test), which is not working properly (confirmed by @RishiDiwanTT) and needs an overhaul. However, whether it works or not, it does not appear to negatively impact integrations that do not need WAYFless support, so the need to get the configuration benefits of this PR outweigh concerns about whether this part works. We can fix this later, if needed.
- There are errors in the testing instructions in this PR's "How Has This Been Tested?" section. We will need to update these instructions as we endeavor to complete testing outside the bounds of this PR.

RishiDiwanTT (approved):

PR #176, opened 2022-02-03
Bugfix/subject with missing name and identifier

tdilauro:
## Description

Log an error for some errors during subject classification in `Metadata.apply(), rather than treating them as fatal errors. Specifically, a subject with no name and no identifier will result in NO new classification and a logged error, rather than an exception and potential crash.

## Motivation and Context

Overdrive Format Sweep sweep crashes when a subject is missing both name and identifier.

Notion issue: [overdrive_format_sweep crashes](https://www.notion.so/lyrasis/overdrive_format_sweep-crashes-e89a69a907ec4704ade3871744a470b0)

## How Has This Been Tested?

- Added test for subject missing both `name` and `identifier`.
- All new and existing tests ran successfully.

## Checklist

### DIFF
--- a/core/metadata_layer.py
+++ b/core/metadata_layer.py
@@ -1931,14 +1931,19 @@ class Metadata(MetaToModelUtility):
 
         # Apply all new subjects to the identifier.
         for subject in list(new_subjects.values()):
-            identifier.classify(
-                data_source,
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
Looks good to me

PR #180, opened 2022-02-04
Add MyPy type checking to our CI process

jonathangreen:
## Description

Add [Mypy](http://mypy-lang.org/) type checking to our CI process.

## Motivation and Context

As we are adding type annotation in our code base, it is nice to use them not just for understanding, but to actually type check our code. This shouldn't add any runtime overhead, only when we are actually running the type checker. This also prevents us from adding misleading annotations by mistake, since the annotations we add will actually be checked.

This PR takes [the approach recommended in the mypy docs](https://mypy.readthedocs.io/en/stable/existing_code.html) for existing code bases.

> If your codebase is large, pick a subset of your codebase (say, 5,000 to 50,000 lines) and run mypy only on this subset at first, without any annotations. This shouldn’t take more than a day or two to implement, so you start enjoying benefits soon.

The files we are running mypy on are defined in `pyproject.toml`. This PR runs the type checker on: 
- `core`
- `tests/core`

It added annotations where necessary to get the checker to run, and corrected annotations where they were incorrect.

In some of the code, I converted the older python 2 style doc comments with types to be actual type annotations, and removed the old comments, so they wouldn't get out of sync with the annotations. I used `doc484` and `com2ann` to do the conversion, then did some hand correction of the annotations.

In the `core/model/hassessioncache.py` I modified the type annotations to better reflect whats going on in the code. Originally the type annotations were using a `protocol`, but this code is actually better suited to be typed as a generic. 

## Documentation

I added some documentation on mypy to the notion wiki: 
https://www.notion.so/lyrasis/Mypy-d97f188926294d4c89ceff91ce9556bf

It covers some of the gotchas that I encountered and workarounds used when preparing this PR. 

## How Has This Been Tested?

Testing has been confined to running the type checker and running the unit tests.


io7m:
This brings me joy. :rocket:

tdilauro (approved):
LGTM! ✔️

PR #183, opened 2022-02-09
Remove references to six

jonathangreen:
## Description

Even though we don't have the `six` package in our dependencies, it appears we import and call a few functions from six in our code 😅 .

## Motivation and Context

Remove unnecessary references to `six` still present in the codebase.

## How Has This Been Tested?

Ran unit tests.


tdilauro (approved):
🎸🤘🏽

PR #195, opened 2022-03-01
Feaure/in memory batching

RishiDiwanTT:
## Description
Notion: https://www.notion.so/lyrasis/Implement-proposed-solution-for-search_index_refresh-085989c0bd504f3f8e65eca8993066b7

* Remove unbounded counts on every job loop
* Remove the pressure of JSON serializing from the clster, and adds it to the application server

## Motivation and Context

The search_index_refresh job runs every 6 minutes on the Production systems.
This is a necessary script (maybe not with frequency) and puts a lot of CPU pressure on the RDS cluster

The refactoring here reduces the CPU usage of the script.
What could not be done is changing the `SELECT works JOIN workcoveragerecords` into a `SELECT workcoveragerecords JOIN works` query
This is because the job depends on the fact that there may be works WITHOUT a corresponding workcoveragerecords entry, which would imply a work needing coverage
This can be changed but will need an architectural change of how workcoveragerecords are created

## How Has This Been Tested?

All unit tests are passing
New unit tests are written
Real world data was used to compare results manually

## Checklist

### DIFF
--- a/core/model/work.py
+++ b/core/model/work.py
@@ -19,10 +19,9 @@ from sqlalchemy import (
     String,
     Unicode,
 )
-from sqlalchemy.dialects import postgresql
 from sqlalchemy.dialects.postgresql import INT4RANGE
 from sqlalchemy.ext.associationproxy import association_proxy
-from sqlalchemy.orm import contains_eager, joinedload, relationship, lazyload
+from sqlalchemy.orm import contains_eager, joinedload, relationship
 from sqlalchemy.orm.session import Session
 from sqlalchemy.sql.expression import and_, case, join, literal_column, or_, select
 from sqlalchemy.sql.functions import func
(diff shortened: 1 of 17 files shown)


tests/core/test_external_search.py
```diff
@@ -4484,3 +4578,38 @@ def bulk(self, docs, **kwargs):
         assert work == record.obj
         assert True == record.transient
         assert "There was an error!" == record.exception
+
+
+class DBStatementCounter(object):
```
io7m, line 4583: Apparently we don't need to subclass `object` anymore in Python 3.

```
class DBStatementCounter:
```

RishiDiwanTT: This is true, this is stackoverflow code for counting statements, let me rectify
https://stackoverflow.com/a/33691585

core/model/work.py
```diff
@@ -1428,6 +1432,282 @@ def assign_appeals(self, character, language, setting, story, cutoff=0.20):
     # that Elasticsearch can parse as a date.
     ELASTICSEARCH_TIME_FORMAT = 'YYYY-MM-DD"T"HH24:MI:SS"."MS'
 
+    @classmethod
```
io7m, line 1435: Was this code translated from somewhere?

RishiDiwanTT: Yessir, to_search_documents() was an in-SQL query that did A LOT within the SQL cluster (query + aggregate + jsonify)
We've moved this to an "in-app" version, once approved I will rename THIS method to to_search_documents()
Methid is below in the same file 
https://github.com/ThePalaceProject/circulation/blob/532b7c79c9f1698b2e002a4daa8603c292334517/core/model/work.py#L1431

io7m: Got it, thanks! I wanted to give this a quick read through to see if there were any obvious errors in translation. I'm guessing probably not given that the tests pass.

io7m: Looks correct. :+1:

io7m (approved):

tdilauro:
@RishiDiwanTT CI not passing. Looks like you need to merge the `main` branch to pick up the container fix for `lcpencrypt`.

tdilauro:
@RishiDiwanTT I merged in `main` to pick up changes and force CI. Would you take a quick look to make sure that everything seems okay? If so, go ahead and merge.

RishiDiwanTT:
@tdilauro thats weird because I had rebased main and pushed yesterday, not sure what went wrong

PR #202, opened 2022-03-07
Feature/recursive equivalents coverage

RishiDiwanTT:
## Description
A cache for maintaining the Recursive Equivalents of identifiers in the system.

## Motivation and Context
Currently the recursives are re-computed everytime they are required. Which can be fairly exhaustive, especially with the WorkCoverageRecord workflow (search-index-update)
To save on the CPU cycles of the DB cluster, we pre-compute the recursive equivalents in an isolated job, which can then be used by all other scripts/APIs

The script does the following
- Creates an EquivalencyCoverageRecord for all Equivalencies in the system
- This runs the recursive equivalency stored procedure on each identifier of the records

Eg. Identifiers = [1, 2, 3, 4, 5]
Equivalencies = [(1, 2), (2, 3), (2, 4)]
This will generate the following table

| Parent ID | Identifier ID |
| :---            | :---              |
| 1 | 1 |
| 1 | 2 |
| 1 | 3 |
| 1 | 4 |
| 2 | 2 |
| 2 | 1 |
| 2 | 3 |
| 2 | 4 |
| 3 | 3 |
| 3 | 1 |
| ... | ... |
| 5 | 5 |

Querying the table by `parent_id` will give us the chain for the parent Eg. Querying for `parent_id=1` will give us `1,2,3,4`, where as querying for `parent_id=5` will simply give us `5`

## How Has This Been Tested?
Unit tests have been written for both happy path and edge cases
Manual testing has been done by reseting the coverages
## Checklist

### DIFF
--- a/bin/equivalent_identifiers_refresh
+++ b/bin/equivalent_identifiers_refresh
@@ -1,0 +1,14 @@
+#!/usr/bin/env python
+"""Re-index Equivalent identifiers
+"""
+import os
+import sys
+
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..")
+sys.path.append(os.path.abspath(package_dir))
+
+from core.equivalents_coverage import EquivalentIdentifiersCoverageProvider
+from core.scripts import RunCoverageProviderScript
+
+RunCoverageProviderScript(EquivalentIdentifiersCoverageProvider).run()
--- a/core/coverage.py
+++ b/core/coverage.py
@@ -6,6 +6,8 @@ from sqlalchemy.orm import Load
 from sqlalchemy.orm.session import Session
 from sqlalchemy.sql.functions import func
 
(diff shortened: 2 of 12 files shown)


RishiDiwanTT:
TODO: Add to crontab

io7m (commented):
I've no particular criticisms of this code, but I think we should at least have type annotations on all methods. Could you add them?

Personally, I'd like to get us to the point where we can run with the strictest `mypy` settings. It's going to take a long time to get there, but I think getting as much type information into new code as possible will help.

RishiDiwanTT:
> I've no particular criticisms of this code, but I think we should at least have type annotations on all methods. Could you add them?
> 
> Personally, I'd like to get us to the point where we can run with the strictest `mypy` settings. It's going to take a long time to get there, but I think getting as much type information into new code as possible will help.

@io7m Type hints added

PR #231, opened 2022-03-22
OPDS2 Accept header has been overridden from the OPDS value

RishiDiwanTT:
## Description

Make a minor change in how the JSON value for settings was managed. Only the last value in the SETTINGS array will be considered.

## Motivation and Context
The OPDS2 importer needed a different Accept header than the OPDS importer. However, due to the SETTINGS being an array it is not possible to overwrite a single configuration.
Hence a duplicate setting needed to be added, and thereafter deduped in the code.
The dedupe code relies on the fact that the child class always adds its settings to the end of the SETTINGS array

## How Has This Been Tested?

New test cases were written for different scenarios
Also tested this manually with the admin panel for the OPDS2 collections page

## Checklist

### DIFF
--- a/api/admin/controller/collection_settings.py
+++ b/api/admin/controller/collection_settings.py
@@ -28,6 +28,27 @@ class CollectionSettingsController(SettingsController):
         protocols = super(CollectionSettingsController, self)._get_collection_protocols(
             self.PROVIDER_APIS
         )
+
+        # dedupe and only keep the latest SETTINGS
+        # this will allow child objects to overwrite
+        # parent settings with the same key
+        # This relies on the fact that child settings
+        # are added after parent settings as such
+        # `SETTINGS + <configuration>.to_settings()`
+        for protocol in protocols:
+            if "settings" not in protocol:
+                continue
+            _found_settings = dict()
+            for ix, setting in enumerate(protocol["settings"]):
+                _key = setting["key"]
+                _found_settings[_key] = ix
+            _settings = []
+            # Go through the dict items and only use the latest found settings
+            # for any given key
+            for _, v in _found_settings.items():
+                _settings.append(protocol["settings"][v])
+            protocol["settings"] = _settings
+
         # If there are storage integrations, add a mirror integration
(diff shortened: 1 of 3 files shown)


api/admin/controller/collection_settings.py
```diff
@@ -28,6 +28,27 @@ def _get_collection_protocols(self):
         protocols = super(CollectionSettingsController, self)._get_collection_protocols(
             self.PROVIDER_APIS
         )
+
+        # dedupe and only keep the latest SETTINGS
+        # this will allow child objects to overwrite
+        # parent settings with the same key
+        # This relies on the fact that child settings
+        # are added after parent settings as such
+        # `SETTINGS + <configuration>.to_settings()`
```
tdilauro, line 37: Is this indeed how things work, in general (i.e., elsewhere in the codebase)?

api/admin/controller/collection_settings.py
```diff
@@ -28,6 +28,27 @@ def _get_collection_protocols(self):
...
+            for ix, setting in enumerate(protocol["settings"]):
+                _key = setting["key"]
+                _found_settings[_key] = ix
+            _settings = []
+            # Go through the dict items and only use the latest found settings
+            # for any given key
+            for _, v in _found_settings.items():
+                _settings.append(protocol["settings"][v])
+            protocol["settings"] = _settings
+
```
tdilauro, line 51: I think we should have some tests for this behavior.

tdilauro (commented):
The code looks good here, but I think we need some tests.

api/admin/controller/collection_settings.py
RishiDiwanTT: Yep, https://github.com/ThePalaceProject/circulation/blob/main/core/opds2_import.py#L126

api/admin/controller/collection_settings.py
RishiDiwanTT: Done

tdilauro (approved):
Looks good to me! 🥇

PR #234, opened 2022-03-23
Feature/bibliotheca cloudlink reserve

RishiDiwanTT:
## Description
Collection/Library associations now have a dropdown to select if unavailable works should be displayed
image

This is accomplished by adding this filter to the query all the way down at the DatabaseBackedWorkList level which affects all feeds that originate from lanes

## Motivation and Context
https://www.notion.so/lyrasis/Add-configuration-setting-to-filter-out-titles-when-all-available-copies-have-been-loaned-f488c3fa43704ed0a59323a80ee893fb

## Checklist

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -947,6 +947,8 @@ class ErrorParser(BibliothecaParser):
     error_mapping = {
         "The patron does not have the book on hold": NotOnHold,
         "The patron has no eBooks checked out": NotCheckedOut,
+        "the patron document status was CAN_HOLD and not one of CAN_LOAN,RESERVATION": NotOnHold,
+        "the patron document status was HOLD and not one of CAN_LOAN,RESERVATION": NotOnHold,
     }
 
     def process_all(self, string):
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -49,6 +49,7 @@ from core.model import (
     get_one,
     get_one_or_create,
 )
+from core.model.configuration import ConfigurationAttributeValue
 from core.monitor import CollectionMonitor, IdentifierSweepMonitor, TimelineMonitor
 from core.scripts import RunCollectionMonitorScript
 from core.testing import DatabaseTest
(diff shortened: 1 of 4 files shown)


core/lane.py
```diff
@@ -2240,6 +2246,50 @@ def works_from_database(self, _db, facets=None, pagination=None, **kwargs):
...
 
+    def _restrict_query_for_no_hold_collections(
+        self, _db: Session, qu: query.Query
+    ) -> query.Query:
+        """Restrict the query to not select works that are part of a collection
+        that dissallow holds through DONT_DISPLAY_RESERVES if they have no available copies
+        This is only meant for Bibliotheca collections, but since they're the only ones with
+        the DONT_DISPLAY_RESERVES setting it will implicitly only be active for them
+        """
+
```
RishiDiwanTT, line 2257: @tdilauro is there any time when we don't want this functionality to be active? Should this be configurable via its __init__ args?

api/bibliotheca.py
```diff
@@ -98,7 +99,20 @@ class BibliothecaAPI(BaseCirculationAPI, HasSelfTests):
     ] + BaseCirculationAPI.SETTINGS
 
     LIBRARY_SETTINGS = BaseCirculationAPI.LIBRARY_SETTINGS + [
-        BaseCirculationAPI.DEFAULT_LOAN_DURATION_SETTING
+        BaseCirculationAPI.DEFAULT_LOAN_DURATION_SETTING,
+        {
+            "key": ExternalIntegration.DISPLAY_RESERVES,
```
RishiDiwanTT, line 104: The semantics of this are completely upside down, DISPLAY_RESERVES is going to be NULL for most collections and the default behaviour is to always display reserves. We only act if it is set as "no".
So a DISPLAY_RESERVES=NULL value means we should be displaying reserves, this is because of the requirements per the ticket

core/lane.py
tdilauro: We would always want it available in all library associations to Bibliotheca collections.

api/bibliotheca.py
tdilauro: You could use the distinction between None (db NULL) and False to handle this properly.
```
some_var: Optional[bool] = ...

if some_var is None:
    """do default behavior"""
elif some_var is False:
    """do false behavior"""
else:
   """do true behavior"""
```

RishiDiwanTT: Yes the query only looks for the "no" value at the moment so the semantics work fine, it's just a weird semantic that None and True are equal as opposed to None and False

core/lane.py
RishiDiwanTT: Ok, this should take care of that then

tests/core/test_lane.py
```diff
@@ -2469,6 +2471,84 @@ def query_works(self, *args, **kwargs):
 
         assert [] == wl.search(self._db, query, RaisesException())
 
+    def test_worklist_for_resultset_no_holds_allowed(self):
```
tdilauro, line 2474: It would be nice to have analogous set of tests for when `DISPLAY_RESERVES` is unset or set to yes/true, but I don't want to hold up this functionality. If we don't get those test cases in this time, we can create a new PR.

tdilauro (approved):
This looks good. I think we need a couple more test cases, but we can handle in a future PR, if we do.

PR #240, opened 2022-03-28
An attributeerror was being raised whenever some data is missing from…

RishiDiwanTT:
… a work in to_search_document_in_app

This leads to batches of works not getting indexed

## Description
Defensive code has been added to ensure we always check for availability of the object
before accessing any values of it

## Motivation and Context

This issue will cause degradation in search result accuracy, so must be fixed asap

## How Has This Been Tested?
search_index_clear and search_index_refresh was run on a NJ db dump, where the issue manifested

## Checklist

### DIFF
--- a/core/model/work.py
+++ b/core/model/work.py
@@ -1696,14 +1696,14 @@ class Work(Base):
                 result["genres"].append(genre)
 
         result["identifiers"] = []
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
LGTM! 🎸

PR #244, opened 2022-03-29
Feature/use recursive equivalents

RishiDiwanTT:
## Description
Work.to_search_documents now uses the RecursiveEquivalentsCache table
The old to_search_documents has been appended with a __DONOTUSE statement, but not deleted.
Lest we forget the lessons of the past    

There was also a bug discovered where licensepools were not filtered out correctly in the method.
This has been fixed.

## Motivation and Context

This further reduces the CPU load on the DB cluster by avoiding the use of the recursive SQL procedure.
Only the to_search_documents function uses this table as of now since the other methods depend on a Policy that may change.
To_search_documents also took an optional Policy, but since the method is only used in one place it could be ignored

## How Has This Been Tested?
Unit test cases already cover the entirety of this functionality
They only had to be modified to prime the recursive equivalents cache before running

This was also manually tested with a sanitized NJ database

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -564,7 +564,7 @@ class ExternalSearchIndex(HasSelfTests):
             needs_add.append(work)
 
         # Add/update any works that need adding/updating.
-        docs = Work.to_search_documents_in_app(needs_add)
+        docs = Work.to_search_documents(needs_add)
 
         for doc in docs:
             doc["_index"] = self.works_index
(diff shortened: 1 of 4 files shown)


io7m (approved):
I can't claim to understand what the code is doing as a whole (very few people on the planet can :smile:) but I see no issues with this set of changes.

PR #247, opened 2022-03-30
Refactor Overdrive code to use new configuration API

io7m:
## Description

This refactors the Overdrive code to use the new configuration
API and improves type safety and encapsulation everywhere.
    
* The core OverdriveAPI class was renamed to OverdriveCoreAPI in order to stop having two classes in the core and API with the same name.
* Configuration values were renamed so that all Overdrive-specific values are now called `overdrive_*` in the database. Code was added to migrate from the old configuration structure to the new structure (and the old values are set to `None` in the database).
* Many attributes were renamed to turn them into "protected" attributes, and type annotations were added. External code needing access to the attributes now has to retrieve the values through methods.

## Motivation and Context

https://www.notion.so/lyrasis/Migrate-Overdrive-integration-to-new-configuration-API-1f8d105bf6cd4c99a23bd6baead15d9e

## How Has This Been Tested?

I tried creating a new Overdrive collection and checked that the configuration settings held.
I also opened an existing Overdrive collection that was created using the old API, and checked that the settings were preserved.

## Checklist

### DIFF
--- a/core/model/collection.py
+++ b/core/model/collection.py
@@ -82,6 +82,7 @@ class Collection(Base, HasSessionCache):
     # secret as the Overdrive collection, but it has a distinct
     # external_account_id.
     parent_id = Column(Integer, ForeignKey("collections.id"), index=True)
+    # SQLAlchemy will create a Collection-typed field called "parent".
     parent: "Collection"
 
     # When deleting a collection, this flag is set to True so that the deletion
--- a/core/model/collection.py
+++ b/core/model/collection.py
@@ -74,6 +74,7 @@ class Collection(Base, HasSessionCache):
     external_integration_id = Column(
         Integer, ForeignKey("externalintegrations.id"), unique=True, index=True
     )
+    _external_integration: ExternalIntegration
 
     # A Collection may specialize some other Collection. For instance,
     # an Overdrive Advantage collection is a specialization of an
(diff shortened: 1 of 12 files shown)


core/model/collection.py
```diff
@@ -74,13 +74,15 @@ class Collection(Base, HasSessionCache):
...
     )
+    _external_integration: ExternalIntegration
 
     # A Collection may specialize some other Collection. For instance,
     # an Overdrive Advantage collection is a specialization of an
     # ordinary Overdrive collection. It uses the same access key and
     # secret as the Overdrive collection, but it has a distinct
     # external_account_id.
     parent_id = Column(Integer, ForeignKey("collections.id"), index=True)
+    parent: "Collection"
```
RishiDiwanTT, line 85: Took me a while to figure out this is a backref placeholder, maybe add a comment to that effect

core/overdrive.py
```diff
@@ -199,7 +259,54 @@ def __init__(self, _db, collection):
...
         self._collection_token = None
 
-    def endpoint(self, url, **kwargs):
+    def _determine_hosts(self, configuration: OverdriveConfiguration) -> dict[str, str]:
+        # Figure out which hostnames we'll be using when constructing
+        # endpoint URLs.
+        server_nickname = (
+            configuration.server_nickname or OverdriveConfiguration.PRODUCTION_SERVERS
+        )
+        if server_nickname not in self.HOSTS:
```
RishiDiwanTT, line 268: In case of a misconfigured environment, why not fail rather than default to the Production API?

io7m: Good question... I didn't actually write that code, I just adapted it to the new system. Apparently falling back to the production API was the original intent.

io7m:
Hm, got some failing tests. Fixing!

io7m:
OK, I think we're good.

RishiDiwanTT (approved):
Looks all good

PR #257, opened 2022-04-05
Overdrive retry login: Use correct "allowed response codes"

io7m:
## Description

This adds the "allowed response codes" and the connection retry count
to the internal HTTP requests to allow for retrying on Overdrive
API failures.

## Motivation and Context

Affects: https://www.notion.so/lyrasis/overdrive_new_title-import-fails-due-to-timeouts-and-500-errors-87ec0491e0f94a889a3ef37cc2f440a5

## How Has This Been Tested?

I've run three complete Overdrive imports in total, which took about an hour each. Each one presented various errors, and the import continued after retrying. In particular, the imports would have been failing with 401 errors when the authentication token expired because the retry logic that was recently added to the low-level HTTP imports meant that the Overdrive code would have seen an exception instead of a 401 error code.

## Checklist

### DIFF
--- a/core/overdrive.py
+++ b/core/overdrive.py
@@ -431,7 +431,9 @@ class OverdriveCoreAPI(HasExternalIntegration):
         request_headers = dict(Authorization="Bearer %s" % self.token)
         request_headers.update(extra_headers)
 
-        response: Response = self._do_get(url, request_headers)
+        response: Response = self._do_get(
+            url, request_headers, allowed_response_codes=["2xx", "3xx", "401", "404"]
+        )
         status_code: int = response.status_code
         headers: CaseInsensitiveDict = response.headers
         content: bytes = response.content
@@ -650,11 +652,15 @@ class OverdriveCoreAPI(HasExternalIntegration):
     def _do_get(self, url: str, headers, **kwargs) -> Response:
         """This method is overridden in MockOverdriveAPI."""
         url = self.endpoint(url)
+        kwargs["max_retry_count"] = self._configuration.max_retry_count
+        kwargs["timeout"] = 120
         return HTTP.get_with_timeout(url, headers=headers, **kwargs)
 
     def _do_post(self, url: str, payload, headers, **kwargs) -> Response:
         """This method is overridden in MockOverdriveAPI."""
         url = self.endpoint(url)
+        kwargs["max_retry_count"] = self._configuration.max_retry_count
+        kwargs["timeout"] = 120
         return HTTP.post_with_timeout(url, payload, headers=headers, **kwargs)
 
     def website_id(self) -> bytes:
--- a/core/overdrive.py
+++ b/core/overdrive.py
@@ -425,7 +425,7 @@ class OverdriveCoreAPI(HasExternalIntegration):
         self._token = credential.credential
 
     def get(
-        self, url: str, extra_headers, exception_on_401=False
+        self, url: str, extra_headers={}, exception_on_401=False
     ) -> Tuple[int, CaseInsensitiveDict, bytes]:
         """Make an HTTP GET request using the active Bearer Token."""
         request_headers = dict(Authorization="Bearer %s" % self.token)
--- a/tests/core/util/test_mock_web_server.py
+++ b/tests/core/util/test_mock_web_server.py
@@ -1,0 +1,241 @@
+import logging
+import threading
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from typing import Dict, List, Optional, Tuple
+
+import pytest
+
+from core.util.http import HTTP, RequestNetworkException
+
+
+class MockAPIServerRequest:
+    """A request made to a server."""
+
+    headers: Dict[str, str]
+    payload: bytes
+    method: str
+    path: str
+
+    def __init__(self):
+        self.headers = {}
+        self.payload = b""
+        self.method = "GET"
+        self.path = "/"
+
+
+class MockAPIServerResponse:
+    """A response returned from a server."""
+
+    status_code: int
+    content: bytes
+    headers: Dict[str, str]
+    close_obnoxiously: bool
+
+    def __init__(self):
+        self.status_code = 200
+        self.content = "".encode("utf-8")
+        self.headers = {}
+        self.close_obnoxiously = False
+
+    def set_content(self, data: bytes):
+        """A convenience method that automatically sets the correct content length for data."""
+        self.content = data
+        self.headers["content-length"] = str(len(data))
+
+
+class MockAPIServerRequestHandler(BaseHTTPRequestHandler):
+    """Basic request handler."""
+
+    def _send_everything(self, _response: MockAPIServerResponse):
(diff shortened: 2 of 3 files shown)


core/overdrive.py
```diff
@@ -431,7 +431,9 @@ def get(
         request_headers = dict(Authorization="Bearer %s" % self.token)
         request_headers.update(extra_headers)
 
-        response: Response = self._do_get(url, request_headers)
+        response: Response = self._do_get(
+            url, request_headers, allowed_response_codes=["2xx", "3xx", "401", "404"]
```
RishiDiwanTT, line 435: This is an interesting change, how did this work before this change? Did it always have many failures?

io7m: What seemed to be happening was that we'd see a 401 and the retry logic added in the OPDS PR (https://github.com/ThePalaceProject/circulation/pull/123) would cause us to avoid going through the "refresh the OAuth token" logic needed to actually properly handle the 401.

RishiDiwanTT: Ah ok

RishiDiwanTT:
Maybe we should add a unit test to this end?
Otherwise all seems good

io7m:
Yes, I've been thinking about how to do that. Oddly enough, we do have existing tests that cover the things described above, through an unlikely series of events. :laughing: 

Essentially, we had tests that covered the credential refreshing on 401 errors. I then added part of the code above to enable retrying for other temporary errors, but this _disabled_ the 401 credential refreshing. I then added the rest of the code above, which re-enabled credential refreshing. We already have other tests that indicate that temporary errors cause retries thanks to the OPDS PR, and we already have tests that cover the 401 refreshing. It's an unusual case of an existing test suite already covering new functionality!

io7m:
I'm not sure what that lint error is trying to tell me. It looks like a bug in `black`.

RishiDiwanTT:
Looks like some python issue, maybe try your luck re-running the workflow

RishiDiwanTT:
Re: Testing, if the tests were passing already, clearly something in them was lacking as this was an existing issue

io7m:
> Re: Testing, if the tests were passing already, clearly something in them was lacking as this was an existing issue

Yeah, I think an extra slightly redundant test would be helpful. We essentially have this:

* We have a test covering "401 errors cause credentials to be refreshed"
* We have a test covering "HTTP temporary errors cause HTTP requests to be retried"

We can infer that temporary errors during Overdrive runs will be handled properly based on the above, but we don't have a test case that actually says this explicitly. I'll look into adding one.

io7m:
This is turning out to be annoyingly difficult. The test suite mocks enough of the HTTP library that it eliminates the retry logic that the real HTTP library does. The way to test this would be to have the code make real requests to an embedded web server in the test suite, but I think that's way more work than I want to do for the sake of a single unit test.

tdilauro:
> I'm not sure what that lint error is trying to tell me. It looks like a bug in `black` (doesn't happen locally, here).

There was, indeed a bug in `black`, which was caused by a breaking change in the `click` package. This is fixed in #263.

RishiDiwanTT:
RE: annoying tests:
You can just use magicmock on the get/post functions, or I would suggest using mock.path on the session.request function, so you can fake an exception (don't use the mocked api classes)
I agree, I did this for testing out some overdrive changes
https://github.com/ThePalaceProject/circulation/pull/260/files#diff-67d4b2612f315e1ef966127058266fd74d237c95851edc4d13f2608a15fa337eR1499

io7m:
> You can just use magicmock on the get/post functions, or I would suggest using mock.patch on the session.request function, so you can fake an exception (don't use the mocked api classes)

This won't really work. The actual property that I'd want to test would be: "If the Overdrive code makes a request, and that request results in one of the allowed error codes, then that request is transparently retried `n` times without the Overdrive code seeing that this has happened". In other words, I need to spy on the CM's internal HTTP library from the Overdrive code to check that one request from the Overdrive code results in three requests inside the HTTP code. We do already have tests on the HTTP code itself, and we have tests that show that the Overdrive code will correctly handle the error codes it gets back... We just don't have a single end-to-end test that shows those two things happening in combination. I would normally use [MockServer](https://www.mock-server.com/) to write an end-to-end test like this, but like I said, that's quite a bit of work to integrate for the sake of a single test.

RishiDiwanTT:
> I need to spy on the CM's internal HTTP library
This should work if you Mock.patch `Session.request` so that `HTTP::_request_with_timeout` uses tha patched version which can have your intended side_effects

io7m:
Hm, I see. I'll give that a shot now.

io7m:
It goes without saying that the underlying HTTP library is written to act differently based on whether you're using a `Session` or some other type. :-1:

Working on it...

io7m:
It's this specific line that's the problem:

https://github.com/ThePalaceProject/circulation/blob/feature/overdrive-retry/core/util/http.py#L280

I don't know how to provide a mock such that that code will be run _and_ that won't make a real request to an external server.

tests/core/util/test_mock_web_server.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def dequeue_response(
+        self, request: MockAPIServerRequest
+    ) -> Optional[MockAPIServerResponse]:
+        self._requests.append(request)
+        _by_method = self._responses.get(request.method) or {}
+        _by_path = _by_method.get(request.path) or []
+        if len(_by_path) > 0:
+            return _by_path.pop(0)
+        return None
```
RishiDiwanTT, line 159: Best to raise an error so the test fails predictably if there was no response enqueued for this request

io7m: That actually happens here:

https://github.com/ThePalaceProject/circulation/pull/257/files/aad54d9e8367b0ee2c2d3306206eb418c118ac79#diff-9712f5dfb46342a5dd008eed3f73df687da72fce71a2f897e3d814cb77000e13R77

io7m: It'll cause the connection to be terminated, so unless a test is specifically checking for connection failures, it won't accidentally succeed because the server happened to yield an HTTP status code that the test was looking for.

RishiDiwanTT (approved):
Although I philosophically prefer patching sessions as opposed to MockServers, this case may have been required.
Approved ! (caveat: linter fixes)

io7m:
> Although I philosophically prefer patching sessions as opposed to MockServers.

I used to agree. In the Android app, we had an HTTP abstraction similar to the one in the CM to allow for mocking. We ended up switching to using a mock web server over the years in order to better ensure that we were really testing the code we thought we were testing, and also to try to help readability in the tests. I'm not opposed to mocking in general, but I very much dislike the kind of deep internal patching and mocking of Python library classes that we're seeing in the CM test suite on the grounds of it being tough to understand and also prone to breakage as library internals change.

io7m:
> caveat: linter fixes

Those should go away automatically when I click the merge button thanks to @tdilauro's `black` upgrade.

PR #260, opened 2022-04-06
OverdriveAPI fulfillment credentials through patronauth

RishiDiwanTT:
using overdrive fulfillment key-secret on the backend we now
create an access token for the clients to use rather than
have them store the key-secret locally

Backend storage of key-secret is done via environment variables
for maximum security

## Description
The Configuration object now looks for SIMPLIFIED_OVERDRIVE_FULFILLMENT_CLIENT_{KEY,SECRET} environment variables
While returning an `OverdriveManifestFulfillmentInfo` we add into it the `X-Overdrive-Patron-Authentication` header with the fulfillment bearer token
The `OverdriveManifestFulfillmentInfo` object is the only object that returns the `X-Overdrive-Scope` header, so it is logically the only place we need to add this additional header

## Motivation and Context

The Overdrive security review stated the fulfilment key-secrets being stored on the mobile clients needed to be moved to the backend.
To this end, we now create fulfilment access tokens for the mobile clients during the fulfilment API requests and return them as an `X-Overdrive-Patron-Authentication` header

## Checklist

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -642,7 +642,7 @@ class OverdriveAPI(
                     patron,
                     pin,
                     is_fulfillment=True,
-                )
+                ).credential
                 return OverdriveManifestFulfillmentInfo(
                     self.collection,
                     download_link,
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -2,6 +2,7 @@ import datetime
 import json
 import re
 import urllib.parse
+from typing import Tuple, Union
 
 import dateutil
 import flask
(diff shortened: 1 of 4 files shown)


tdilauro:
@RishiDiwanTT If you can pick this up, we still need tests for both sets of credentials and for verifying the handling of both the `CannotLoadConfiguration` and `CannotFulfill` exceptions (and anything else you notice needs tests). If not, I'll come back to it tomorrow.

tdilauro:
@RishiDiwanTT Thank you! I really appreciate you keeping this moving forward. I have completed a couple pre-merge testing deployments and this implementation seems to be working well.

tdilauro (approved):
Looks good! 🚀

PR #263, opened 2022-04-06
Bugfix/upgrade black to stop exception

tdilauro:
## Description

Upgrade the version of `black` used by `pre-commit` and, by extension, the `git` pre-commit hook and the `Lint` GitHub workflow.

The first commit contains only the change to the version of `black`. The second commit contains changes to files that were reformatted by the new version of `black`.

## Motivation and Context

This prevents an exception caused by a breaking change to one of `blacks` underlying dependencies, `click`. 

See https://github.com/psf/black/issues/2964 for more details.

## How Has This Been Tested?

- Ran `pre-commit run --all-files` locally.
- CI workflows ran on `push`.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       - id: mixed-line-ending
 
   - repo: https://github.com/psf/black
-    rev: 21.12b0
+    rev: 22.3.0
     hooks:
       - id: black
         name: Run black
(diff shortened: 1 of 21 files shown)


io7m:
Looks good! Slightly disturbing that the dependencies of `black` are apparently configured in a way where something can break without us changing anything.

io7m (approved):

PR #275, opened 2022-04-12
Remove the complaint system

io7m:
## Description

This removes the entirety of the complaints system in anticipation
of the design of a new system.

## Motivation and Context

Affects: https://www.notion.so/lyrasis/Remove-Complaints-feature-from-Palace-Manager-7f7eea9e263643e099eec5b08adb3c4c

## How Has This Been Tested?

I ran the existing test suite.

## Checklist


io7m:
I'm not sure how or even if we handle table deletions. After this change, the `complaints` table is dead weight.

RishiDiwanTT:
The ticket says this may be a temporary removal, maybe we don't want to remove all the adjacent code, just the entry points?
@tdilauro is that right?

io7m:
In a conversation I had with @dwilcox, the intended route was to mercilessly remove the existing system as we'd be completely redesigning it. It's doubtful the new system would share any code or any endpoints.

dwilcox:
@io7m Yes, that's correct. I updated the ticket to remove the bit about a temporary removal.

RishiDiwanTT (approved):
Looks pretty merciless to me

PR #281, opened 2022-04-14
Feature/opds2 schema

RishiDiwanTT:
## Description

In an ideal world we would simply reference "https://raw.githubusercontent.com/opds-community/drafts/master/schema/feed.schema.json"
and be done with it, no changes required
However, the above in turn refers to "https://readium.org/webpub-manifest/schema/link.schema.json"
and "https://readium.org/webpub-manifest/schema/metadata.schema.json", both of which have
regexs that are not recognized under python regex library
They are valid regexs under PCRE and similar libraries but the jsonschema libraries
will be using the standard library

These regexs have been removed in the local version of the schema (only 3 places existed)
and a custom refresolver has been used to either pick up local or remote references

## Motivation and Context

An OPDS 2 schema validation tool is required to check new and existing OPDS 2 feeds for validity.
An executable tool has been added for this purpose `bin/opds2_schema_validate`

## How Has This Been Tested?

Tested manually against http://bookshelf-feed-demo.us-east-1.elasticbeanstalk.com/v1/publications

## Checklist

### DIFF
--- a/bin/odl2_schema_validate
+++ b/bin/odl2_schema_validate
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+"""Update the circulation manager server with new books from
+OPDS import collections."""
+import os
+import sys
+
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..")
+sys.path.append(os.path.abspath(package_dir))
+
+from webpub_manifest_parser.odl import ODLFeedParserFactory
+
+from api.odl2 import ODL2Importer
+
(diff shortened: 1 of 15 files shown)


RishiDiwanTT:
Moved to draft as I need to finalize the ODL implementation

io7m (approved):
Looks good to me!

This is just plain OPDS 2.0 with fixes, right? Presumably we need to do further work on top of this to encode the things that the CM specifically requires?

RishiDiwanTT:
@io7m What custom attributes does the CM look for specifically?
These changes include both OPDS2 and ODL schemas.

PR #292, opened 2022-04-20
Fix cache_marc_files script

dbernstein:
## Description
This PR fixes the broken cache_marc_files script.  The script does have any direct references to the core/s3.py library and as a result the MirrorUploader.IMPLEMENTATION_REGISTRY is not being loaded with the S3 implementations.  This PR ensures that registration occurs when loading the MARC Exporter.  

I added a test_marc2.py test to ensure that nothing from core/s3.py was being called directly.  The problem is that tests in both tests/api and tests/core rely on test classes (MockS3Client and MockS3Uploader) and thus if I put my unit test in core/test_marc.py, it passes with or without my fix.  

I contemplated doing a larger refactor to get the test classes out of core/s3.py so that the test could live in tests/core/test_marc.py but that would involve a larger refactor.   I'm happy to do that but wanted to check in before proceeding.

## Motivation and Context

https://www.notion.so/lyrasis/MARC-record-generator-not-running-8677e5dd7502412fa50900b404853eb2#7b1afef923f0419dab2b4e2abb9d1f19

## How Has This Been Tested?

I added a unit test to reproduce the problem and confirmed that the code change fixes it.
I also manually tested the script to verify that the issue has been resolved.

## Checklist

- [ ] I have updated the documentation accordingly. (N/A)


tdilauro (commented):
Good catch here. Just a couple of minor issues to address.

dbernstein:
Okay @tdilauro : thanks for the review. It should be good to go now.

tdilauro:
@dbernstein Hmm. Still some linter errors. Have you installed the `pre-commit` hook as discussed at the top of the [Code Style section of the README](https://github.com/ThePalaceProject/circulation#code-style)? It should run the linters and prevent commits that don't satisfy them.

tests/core/test_marc2.py
```diff
@@ -0,0 +1,15 @@
+from core.mirror import MirrorUploader
+
+
+class TestMiscellaneous:
+    def test_mirror_uploader_implementations_are_being_loaded(self):
+        """
+        This test verifies that the two S3 mirror implementations are being loaded when the MARCExporter is
+        imported.  It was not added to tests/core/test_marc.py because that test causes the implementations to be
+        loaded since it references the core.s3 package directly.
+        :return:
```
tdilauro, line 10: The `:return:` here seems extraneous.

Also, not a big deal, but to make your comments a little more readable, it would be helpful to wrap them at no more than 80 columns, where possible. I usually aim for around 72. Is that because I'm an old FORTRAN guy? Maybe?

tdilauro (approved):
Looks good! 🎸 

A comment with a couple of suggestions, but feel free to merge once you've reviewed and addressed as you deem appropriate.

tests/core/test_marc2.py
dbernstein: No problem.

dbernstein:
Will do.

PR #295, opened 2022-04-22
Case insensitive admin email auth

RishiDiwanTT:
## Description
Email Password auth for admins now searches for admin matches
with a case insensitive email

## Motivation and Context
While logging into the admin panel, email is considered to be case sensitive which can lead to confusion if the sign up was done with some capitalisation. 
More details here: https://www.notion.so/lyrasis/Change-Palace-Collection-Manager-login-screen-email-from-case-sensitive-to-not-case-sensitive-b5bb506842084cc98874c6de554a817e

## How Has This Been Tested?

A new unit test has been added and the login page was tested manually

## Checklist


tests/api/admin/test_password_admin_authentication_provider.py
```diff
@@ -34,6 +34,14 @@ def test_sign_in(self):
         assert PasswordAdminAuthenticationProvider.NAME == admin_details.get("type")
         assert "foo" == redirect
 
+        # Case insensitive test
+        admin_details, redirect = password_auth.sign_in(
+            self._db, dict(email="ADMin2@nyPL.Org", password="pass2", redirect="foo")
+        )
+        assert "admin2@nypl.org" == admin_details.get("email")
+        assert PasswordAdminAuthenticationProvider.NAME == admin_details.get("type")
+        assert "foo" == redirect
```
tdilauro, line 43: It would probably be useful to add an analogous test in [TestSignInController.test_password_sign_in](https://github.com/ThePalaceProject/circulation/blob/7c9ae5b9f66b3f3ad26950412ae6da05aade7d16/tests/api/admin/controller/test_controller.py#L702-L715).

tdilauro (commented):
Looking good! Suggested one additional test.

core/model/admin.py
```diff
@@ -53,13 +61,18 @@ def has_password(self, password):
...
+    def authenticate(cls, _db, email: str, password: str):
         """Finds an authenticated Admin by email and password
         :return: Admin or None
         """
 
         def lookup_hook():
-            return get_one(_db, Admin, email=str(email)), False
+            return (
+                _db.query(Admin)
+                .filter(func.lower(Admin.email) == email.lower())
```
io7m, line 72: It _might_ be worth uppercasing for comparisons, if a case folding function isn't available. There are characters out there that will map from uppercase to lowercase in surprising ways:

https://stackoverflow.com/questions/67106784/why-is-upper-casing-not-enough-for-case-insensitive-comparison

It's slightly safer to uppercase and compare than to lowercase and compare.

RishiDiwanTT: It's a marginal (and unknown) amount of safer, but still a valid thought. Updated.

tests/api/admin/test_password_admin_authentication_provider.py
tdilauro: @RishiDiwanTT 👆🏽

RishiDiwanTT: Something more than this commit is required?
https://github.com/ThePalaceProject/circulation/pull/295/commits/50307f92bf1cd30a1e39e7698802ea2c8d58af2b

tdilauro: I just went and looked and both the `/admin/sign_in` and `/admin/sign_in_with_password` routes end up in `PasswordAdminAuthenticationProvider.sig_in`, so yeah, this is fine.

tdilauro (commented):
A couple more comments.

Also, do we need to introduce a constraint (and associated test(s)) to ensure that "duplicate" admins (e.g., "user1@example.com" and "USER1@example.com") cannot be created?

RishiDiwanTT:
> A couple more comments.
> 
> Also, do we need to introduce a constraint (and associated test(s)) to ensure that "duplicate" admins (e.g., "[user1@example.com](mailto:user1@example.com)" and "[USER1@example.com](mailto:USER1@example.com)") cannot be created?

In this case the constraint should be on the database, by either doing an in-code lower() or using the `citext` extension

core/model/admin.py
```diff
@@ -53,13 +61,18 @@ def has_password(self, password):
         return self.password_hashed == bcrypt.hashpw(password, self.password_hashed)
 
     @classmethod
-    def authenticate(cls, _db, email, password):
+    def authenticate(cls, _db, email: str, password: str):
```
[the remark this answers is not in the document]
RishiDiwanTT: Do we need futures? Are we supporting python2.x?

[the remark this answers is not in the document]
tdilauro: They're not exclusive to Python 2. In this case, we need it for some, if not all, of our supported Python 3.x versions. See [this section of PEP 563](https://peps.python.org/pep-0563/#enabling-the-future-behavior-in-python-3-7).

core/model/admin.py
```diff
@@ -53,13 +61,18 @@ def has_password(self, password):
...
         """Finds an authenticated Admin by email and password
         :return: Admin or None
         """
 
         def lookup_hook():
-            return get_one(_db, Admin, email=str(email)), False
+            return (
+                _db.query(Admin)
+                .filter(func.upper(Admin.email) == email.upper())
+                .first(),
```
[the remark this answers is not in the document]
RishiDiwanTT: Updated to mimic the get_one logic

tdilauro:
> > Also, do we need to introduce a constraint (and associated test(s)) to ensure that "duplicate" admins (e.g., "[user1@example.com](mailto:user1@example.com)" and "[USER1@example.com](mailto:USER1@example.com)") cannot be created?
> 
> In this case the constraint should be on the database, by either doing an in-code lower() or using the `citext` extension

I think you could also define a unique index on `UPPER(email)`.

RishiDiwanTT:
> > > Also, do we need to introduce a constraint (and associated test(s)) to ensure that "duplicate" admins (e.g., "[user1@example.com](mailto:user1@example.com)" and "[USER1@example.com](mailto:USER1@example.com)") cannot be created?
> > 
> > 
> > In this case the constraint should be on the database, by either doing an in-code lower() or using the `citext` extension
> 
> I think you could also define a unique index on `UPPER(email)`.

Ok, I looked this up again, a Unique index can be created for the upper() email. (unique constraint cannot use functions, but an index can)
However, since there is already data in the production DBs this might be an un-enforcable index (in case conflicting data already exists). we can test this out by running the following query on the production DBs
`select * from (select UPPER(email), count(id) as c from admins group by UPPER(email)) as t where t.c > 1`

tdilauro:
I ran that query and found no duplicates in the prod CMs. 

That said, the migration should be robust to both (1) an already created index, in which case it should not create it, but note that this condition occurred; and (2) index creation failure caused by, e.g., duplicate entries, in which case an exception should be raised (with a helpful message) to abort the migration.

RishiDiwanTT:
> I think you could also define a unique index on UPPER(email).

Updated

tdilauro (approved):
This looks good to go!

PR #314, opened 2022-04-28
Make Overdrive migration explicit

io7m:
## Description

The previous Overdrive PR added some code to rename old configuration
settings to their new names for the current code. This had a bad
interaction with the nullable value column in the configuration
settings table. This change adds an explicit migration script, and
updates all of the unit tests that were now failing due to the old
configuration keys no longer being present.

## Motivation and Context

Affects: https://www.notion.so/lyrasis/Move-OverDrive-configuration-migration-code-into-the-current-standard-migration-mechanism-af3adcc60ba54958ad1c050279e34594

## How Has This Been Tested?

The existing test suite was updated to ensure the small configuration changes needed in the CM were tested.
I tested against my own local CM in various ways by manually inserting values into the configuration settings table in order to force the migration script to do work.

## Checklist

### DIFF
--- a/core/overdrive.py
+++ b/core/overdrive.py
@@ -231,17 +231,6 @@ class OverdriveCoreAPI(HasExternalIntegration):
         self._external_integration = collection.external_integration
         self._collection_id = collection.id
 
-        if collection.parent:
-            # This is an Overdrive Advantage account.
-            self.parent_library_id = collection.parent.external_account_id
-
-            # We're going to inherit all of the Overdrive credentials
-            # from the parent (the main Overdrive account), except for the
-            # library ID, which we already set.
-            collection = collection.parent
-        else:
-            self.parent_library_id = None
-
         # Initialize configuration information.
         self._configuration_storage = ConfigurationStorage(self)
         self._configuration_factory = ConfigurationFactory()
@@ -249,13 +238,34 @@ class OverdriveCoreAPI(HasExternalIntegration):
             configuration_storage=self._configuration_storage, db=_db
         )
 
-        self._migrate_configuration(
-            collection=collection, configuration=self._configuration
-        )
-        self._server_nickname = (
-            self._configuration.server_nickname
-            or OverdriveConfiguration.PRODUCTION_SERVERS
-        )
+        if collection.parent:
+            # This is an Overdrive Advantage account.
+            self.parent_library_id = collection.parent.external_account_id
+
+            # We're going to inherit all of the Overdrive credentials
+            # from the parent (the main Overdrive account), except for the
+            # library ID, which we already set.
+            parent_integration = collection.parent.external_integration
+
+            self._configuration.client_key = parent_integration.setting(
+                "overdrive_client_key"
+            )
+            self._configuration.client_password = parent_integration.setting(
+                "overdrive_client_secret"
+            )
+            self._configuration.website_id = parent_integration.setting(
+                "overdrive_website_id"
+            )
+        else:
+            self.parent_library_id = None
+
+        if not self._configuration.client_key:
+            raise ValueError("Overdrive client key is not configured")
+        if not self._configuration.client_password:
+            raise ValueError("Overdrive client password/secret is not configured")
+        if not self._configuration.website_id:
+            raise ValueError("Overdrive website ID is not configured")
+
         self._hosts = self._determine_hosts(configuration=self._configuration)
 
         # This is set by an access to .token, or by a call to
@@ -275,39 +285,6 @@ class OverdriveCoreAPI(HasExternalIntegration):
             server_nickname = OverdriveConfiguration.PRODUCTION_SERVERS
 
         return dict(self.HOSTS[server_nickname])
-
-    def _migrate_configuration(
-        self, collection: Collection, configuration: OverdriveConfiguration
-    ) -> None:
-        """Attempt to load configuration data that was saved using the old
-        external integration methods. Delete the old data when we're done."""
-        integration: ExternalIntegration = collection.external_integration
-        old_client_key: Optional[str] = integration.setting(
-            ExternalIntegration.USERNAME
-        ).value
-        if old_client_key:
-            configuration.client_key = old_client_key
-
-        old_client_secret: Optional[str] = integration.setting(
-            ExternalIntegration.PASSWORD
-        ).value
-        if old_client_secret:
-            configuration.client_password = old_client_secret
-
-        old_website_id: Optional[str] = integration.setting(self.WEBSITE_ID).value
-        if old_website_id:
-            configuration.website_id = old_website_id
-
-        old_server_nickname: Optional[str] = integration.setting(
-            self.SERVER_NICKNAME
-        ).value
-        if old_server_nickname:
-            configuration.server_nickname = old_server_nickname
-
-        integration.set_setting(ExternalIntegration.USERNAME, None)
-        integration.set_setting(ExternalIntegration.PASSWORD, None)
-        integration.set_setting(self.WEBSITE_ID, None)
-        integration.set_setting(self.SERVER_NICKNAME, None)
 
     def external_integration(self, db: Session) -> ExternalIntegration:
         return self._external_integration
@@ -739,9 +716,9 @@ class MockOverdriveCoreAPI(OverdriveCoreAPI):
         integration = collection.create_external_integration(
             protocol=ExternalIntegration.OVERDRIVE
         )
-        integration.username = client_key
-        integration.password = client_secret
-        integration.set_setting("website_id", website_id)
+        integration.set_setting("overdrive_client_key", client_key)
+        integration.set_setting("overdrive_client_secret", client_secret)
+        integration.set_setting("overdrive_website_id", website_id)
         library.collections.append(collection)
         OverdriveCoreAPI.ils_name_setting(_db, collection, library).value = ils_name
         return collection
--- a/core/overdrive.py
+++ b/core/overdrive.py
@@ -99,7 +99,7 @@ class OverdriveConfiguration(ConfigurationGrouping, BaseImporterConfiguration):
     PRODUCTION_SERVERS = "production"
     TESTING_SERVERS = "testing"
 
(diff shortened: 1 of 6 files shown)


migration/20220428-namespace-overdrive-configuration.py
```diff
@@ -0,0 +1,182 @@
...
+    if result_old_key.rowcount > 0:
+        existing_old_key_present = True
+        existing_old_key = result_old_key.fetchone()[0]
+
+    select_new_key = text(
+        """
+SELECT c.value FROM configurationsettings AS c
+  WHERE c.key = :key
+    AND c.external_integration_id = :id
+    """
```
tdilauro, line 72: Do you need to separate values for `select_old_key` and `select_new_key`? They look the same to me, though you would still need to use different params for `result_old_key` and `result_new_key`.

tdilauro, line 72: For these queries, it is important to distinguish between collection-level (external integration-level) settings (`external_integration_id` IS NOT NULL AND `library_id` IS NULL) and collection/library-level settings (`external_integration_id` IS NOT NULL AND `library_id` IS NOT NULL). If you're dealing with keys that appear only at the collection level, it would be good be clear about that by adding "AND c.library_id IS NULL" to the query.

tdilauro (commented):
This looks like it will work, but I've suggested a couple of changes that will improve error reporting and query safety.

migration/20220428-namespace-overdrive-configuration.py
io7m: I _think_ these settings are collection-level, but I'm not certain. Is there some way to find out?

io7m: What I mean is... Is there some way to tell from the code? Is a class extending `HasInternalIntegration` enough to know if settings are collection-level?

tests/api/test_overdrive.py
```diff
@@ -1521,6 +1521,9 @@ def test_refresh_patron_access_token_is_fulfillment(self):
         self._default_collection.external_integration.setting(
             "overdrive_website_id"
         ).value = "100"
+        self._default_collection.external_integration.setting(
+            "overdrive_server_nickname"
+        ).value = "testing"
```
tdilauro, line 1526: @io7m It's not obvious why this needs to be set to `testing` in order for this test to pass. If it's not set to `testing`, it should default to the production ones.

core/overdrive.py
```diff
@@ -231,31 +231,43 @@ def __init__(self, _db: Session, collection: Collection):
...
+        if not self._configuration.website_id:
+            raise CannotLoadConfiguration("Overdrive website ID is not configured")
 
-        self._migrate_configuration(
-            collection=collection, configuration=self._configuration
-        )
-        self._server_nickname = (
-            self._configuration.server_nickname
-            or OverdriveConfiguration.PRODUCTION_SERVERS
-        )
```
tdilauro, line 258: Ah, I think this is why the test broke. Lines 255-258 here are lines I added to include the server family (currently testing or production) in the object. I think you probably wanted to just remove lines 252-254, which invoked the instance's now-removed `_migrate_configuration` method.

core/overdrive.py
```diff
@@ -468,7 +447,8 @@ def token_authorization_header(self) -> str:
     def fulfillment_authorization_header(self) -> str:
         is_test_mode = (
             True
-            if self._server_nickname == OverdriveConfiguration.TESTING_SERVERS
```
tdilauro, line 471: This change won't work as expected, since `self._server_nickname` accounts for production servers being the default, whereas `self._configuration.server_nickname` does not.

tests/api/test_overdrive.py
tdilauro: It looks like this is happening because of change in `OverdriveCoreAPI.__init__` (removal of ll. 255-258).

migration/20220428-namespace-overdrive-configuration.py
tdilauro: I'm not sure that it's super obvious in the code. One way to tell is by looking at already set keys in `configurationsettings`.
- Site-wide settings have NULL `external_integration_id` and `library_id`.
- External Integration (including Collection) settings have a non-NULL `external_integration_id` and a NULL `library_id`.
- Library settings have a NULL `external_integration_id` and a non-NULL `library_id`.
- External Integration library-specific settings have non-NULL `external_integration_id` and `library_id`.

tdilauro (changes requested):
Currently, there are some changes here that would break the new OD audiobook fulfillment mechanism introduced in #260.

core/overdrive.py
io7m: Oops, yes. Over-zealous removal there.

core/overdrive.py
```diff
@@ -231,31 +231,48 @@ def __init__(self, _db: Session, collection: Collection):
...
+
+            self._configuration.client_key = parent_integration.setting(
+                "overdrive_client_key"
+            )
+            self._configuration.client_password = parent_integration.setting(
+                "overdrive_client_secret"
+            )
+            self._configuration.website_id = parent_integration.setting(
+                "overdrive_website_id"
+            )
```
tdilauro, line 258: The `_server_nickname`, which is one of the settings from the parent collection is not captured in the configuration here.

With the new configuration approach, is there a way to get *all* of the parent collection's settings without enumerating them individually?

tdilauro: We seem to have switched from using the class variables (e.g., "SERVER_NICKNAME", "WEBSITE_ID")  for the settings keys here. Because we also reference these keys multiple times in the tests, I think it is worthwhile to update them to their new values (or add them, in the case of the CLIENT_KEY and CLIENT_SECRET) and use them here and in the tests.

migration/20220428-namespace-overdrive-configuration.py
```diff
@@ -0,0 +1,174 @@
...
+    else:
+        logger.info("No updates were necessary")
+
+
+def execute_migration(db: Session) -> None:
+    select = text(
+        """
+SELECT e.id FROM externalintegrations AS e
+  WHERE e.protocol = 'Overdrive'
+    """
```
tdilauro, line 159: The properties enumerated in `update_configuration` are not associated with child OverDrive accounts (i.e., those that have a parent collection, like OverDrive Advantage collections). So we should not try to update the keys for OD collections that have a parent collection. Changing this query to the following should get the correct set of the `externalintegrations`:
```
    select = text(
        """
        SELECT e.id
        FROM collections c
        JOIN externalintegrations e on c.external_integration_id = e.id
        WHERE e.protocol = 'Overdrive' AND c.parent_id IS NULL
        """
    )
```

tdilauro (changes requested):
A few more comments for review.

core/overdrive.py
io7m: > With the new configuration approach, is there a way to get all of the parent collection's settings without enumerating them individually?

I think we can instantiate an `OverdriveConfiguration` value from the parent collection and copy settings that way. I'll try it.

io7m: > We seem to have switched from using the class variables (e.g., "SERVER_NICKNAME", "WEBSITE_ID") for the settings keys here. Because we also reference these keys multiple times in the tests, I think it is worthwhile to update them to their new values (or add them, in the case of the CLIENT_KEY and CLIENT_SECRET) and use them here and in the tests.

Yeah, I agree. The old configuration system uses these generic names in a vain attempt to provide a sort of common API over the different external integration types. The end result of which is that you end up with a dozen rows in the configuration settings table called `username`, and you have to work backwards to determine "Oh, _this_ `username` is an Overdrive client key" and so on. I'd much rather remove the generic names entirely and introduce a set of new `OVERDRIVE_*` class variables.

migration/20220428-namespace-overdrive-configuration.py
io7m: You're not wrong, but I'm not sure this is necessary. The purpose of this migration is solely to stop using the old names (Use `overdrive_client_key` instead of `username` and so on). Unless I'm mistaken, it shouldn't matter whether a collection has a parent or not when it comes to which names it's using.

core/overdrive.py
tdilauro: We could tell which `username` is a an OD client key because the `externalintegrations.protocol` is "Overdrive" and the `configurationsettings.key` was "username". These changes certainly give you a hint if you're looking at the database directly. But one value of the class constants is to avoid typos and the like and to allow us to change a value in a single place for, say, a database migration.

migration/20220428-namespace-overdrive-configuration.py
tdilauro: We fail the migration with a `RuntimeError` (ll. 81-85 above), if both old and new keys are missing. Because neither of those keys will be present for an OD Advantage integration, the migration would fail for any CM with any collection of that type.

core/overdrive.py
io7m: > We could tell which username is a an OD client key because the externalintegrations.protocol is "Overdrive" and the configurationsettings.key was "username".

It's that extra required step of backwards reasoning that I dislike, though. I'd rather be explicit than general, for the sake of my own and future developer's understanding. Now, if a developer wants to find an Overdrive client key, they look directly at the entries in the table that say `overdrive_client_key`.

core/overdrive.py
```diff
@@ -269,46 +289,14 @@ def _determine_hosts(self, configuration: OverdriveConfiguration) -> Dict[str, s
         # Figure out which hostnames we'll be using when constructing
         # endpoint URLs.
         server_nickname = (
-            configuration.server_nickname or OverdriveConfiguration.PRODUCTION_SERVERS
+            configuration.overdrive_server_nickname
+            or OverdriveConfiguration.PRODUCTION_SERVERS
         )
```
tdilauro, line 294: We should be able to use `self._server_nickname` here, rather than duplicating the logic used calculate it.

tdilauro (approved):
This looks good to go! 🥇

PR #322, opened 2022-05-03
Fix for empty string defaults for Configuration Settings

RishiDiwanTT:
These were considered as missing values and treated the same as a None type

## Description
Rather than using a `not` check, simply use an `is None` check
Its explicit and doesn't cause this confusion

## Motivation and Context
Saving a "Patron Authentication" with "Patron identifiers are not rendered as barcodes" was throwing an Error (400) 

## Checklist


tdilauro (approved):
LGTM! 🚀 

Bonus for catching the typo! 🥇

PR #325, opened 2022-05-04
Make audio/mpeg an "audiobook" format

io7m:
## Description

This makes an adjustment to the set of known content types so that
books using an audio/mpeg content type are classified as audio books
instead of eBooks.

## Motivation and Context

It seems that the audio/mpeg content type was introduced upstream one day without any explanation, and seems to have been (accidentally?) placed into the set of content types that denote eBooks instead of audio books. This means that OPDS feeds that are using the audio/mpeg content type have their books erroneously classified as eBooks.

Affects: https://www.notion.so/lyrasis/CM-considers-audio-mpeg-to-be-eBook-74b9dc74f8b74bb08007f1d2c18eb0fb

## How Has This Been Tested?

This has been tested against an old copy of the Biblioboard feeds (prior to the content type changes). I also wrote a one-line test that verifies that Editions are now assigned the expected type.

## Checklist

### DIFF
--- a/core/model/constants.py
+++ b/core/model/constants.py
@@ -302,13 +302,13 @@ class MediaTypes(object):
         AUDIOBOOK_MANIFEST_MEDIA_TYPE,
         AUDIOBOOK_PACKAGE_MEDIA_TYPE,
         AUDIOBOOK_PACKAGE_LCP_MEDIA_TYPE,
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
This makes sense to me. 🧠

PR #345, opened 2022-05-10
Fixed issue in SAMLWayfless postprocessing with nonetype patron

RishiDiwanTT:
## Description

Ensured none type patrons are not accessed while trying to post process a fulfilment
## Motivation and Context

Trying to fulfil a license without a patron (for open access licenses) would throw an error. 
No open access works were fulfil-able due to this
## How Has This Been Tested?

A new test case has been written targeting fulfilment post processing without any valid data
for collections not having an wayfless template url
## Checklist


io7m:
Nice! I'm glad this was so straightforward.

io7m (approved):

PR #348, opened 2022-05-11
Push notifications API implementation

RishiDiwanTT:
## Description
- model for DeviceTokens
  which stores the FCM token for patrons mobile devices
  It ensure uniqueness on the token string

- API to GET and PUT device tokens for patrons
- GET
  responds with 200 and token details
  404 if token is not found
- PUT
  responds with 201 if created
  400 if the token type is invalid
  409 if the token was not unique
- There is no DELETE method yet
  No use case is outlined for it

## Motivation and Context
The Palace Manager needs to support the following push notifications in the iOS and Android apps
This is the first step, allowing the apps to store the FCM tokens for the mobile devices

## How Has This Been Tested?
Unit tests were written and manual testing was done via Postman

## Checklist

### DIFF
--- a/core/model/devicetokens.py
+++ b/core/model/devicetokens.py
@@ -1,4 +1,4 @@
-from typing import TypeVar, Union
+from typing import Type, TypeVar, Union
 
 from sqlalchemy import Column, Enum, ForeignKey, Integer, Unicode
 from sqlalchemy.exc import IntegrityError
@@ -36,7 +36,7 @@ class DeviceToken(Base):
 
     @classmethod
     def create(
-        cls: type[T],
+        cls: Type[T],
         db,
         token_type: str,
         device_token: str,
--- a/core/model/devicetokens.py
+++ b/core/model/devicetokens.py
@@ -40,7 +40,7 @@ class DeviceToken(Base):
         db,
         token_type: str,
         device_token: str,
-        patron: "Union[Patron,int]",
+        patron: Union[Patron, int],
     ) -> T:
         """Create a DeviceToken while ensuring sql issues are managed.
         Raises InvalidTokenTypeError, DuplicateDeviceTokenError"""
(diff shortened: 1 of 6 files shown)


core/model/devicetokens.py
```diff
@@ -0,0 +1,75 @@
...
+
+from . import Base
+
+
+class DeviceTokenTypes:
+    FCM_ANDROID = "FCMAndroid"
+    FCM_IOS = "FCMiOS"
+
+
+T = TypeVar("T", bound="DeviceToken")
```
io7m, line 17: :+1: :+1: :+1: :+1:

io7m (approved):
This looks good.

io7m:
I like the use of generics. Most people don't think of them!

core/model/devicetokens.py
```diff
@@ -0,0 +1,75 @@
...
+
+    device_token = Column(Unicode, nullable=False, unique=True, index=True)
+
+    @classmethod
+    def create(
+        cls: Type[T],
+        db,
+        token_type: str,
+        device_token: str,
+        patron: "Union[Patron,int]",
```
tdilauro, line 43: Do we need to quote the type here?

RishiDiwanTT: Patron was earlier imported under `if Typing` I can change it to unquoted if needed

core/model/devicetokens.py
```diff
@@ -0,0 +1,75 @@
...
+        elif type(patron) is Patron:
+            kwargs["patron_id"] = patron.id
+
+        device = cls(**kwargs)
+        try:
+            db.add(device)
+            db.commit()
+        except IntegrityError as e:
+            if "device_token" in e.args[0]:
+                raise DuplicateDeviceTokenError()
```
tdilauro, line 63: Use raise from here: `raise DuplicateDeviceTokenError() from e`?

RishiDiwanTT: Done

PR #358, opened 2022-05-16
Always use fulfillment context when fetching OverDrive loans

tdilauro:
## Description

This branch changes the way we fetch OverDrive loans to always authorize the request with a patron bearer token that was obtained in the context of the fulfillment (rather than collection) client credentials. The method signature for `OverdriveAPI.get_loan` is changed because we no longer need to specify the context.

It also removes the unused `OverdriveAPI.get_loans` method, which exactly duplicated `OverdriveAPI.get_patron_checkouts`.

## Motivation and Context

Until now, when not actively fulfilling a loan, we would fetch loans using a patron bearer token that was obtained in the context of the collection (rather than fulfillment) client credentials. At least some of the time, the information returned did not contain the information we needed to adequately understand the formats available or the already selected.

## How Has This Been Tested?

Adjusted a test for method signature change. All tests pass locally.

## Checklist

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -588,10 +588,15 @@ class OverdriveAPI(
         )
 
     def get_fulfillment_link(
-        self, patron, pin, overdrive_id, format_type
+        self, patron: Patron, pin: Optional[str], overdrive_id: str, format_type: str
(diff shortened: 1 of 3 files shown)


RishiDiwanTT (approved):
LGTM!

PR #368, opened 2022-05-20
Proof of concept for exact terms json based search

RishiDiwanTT:
## Description
The following have been proved through testing (unit and manual)
- Basic exact matches
- And'd queries
- Or'd queries
- Mixed And'd and Or'd queries
- <> type queries
- Nested And'd and Or'd queries
- "Nested Type" ES object queries with And/Or operators

To move from POC to Production ready code the following is required
- ~~Test/implement all supported keys~~
- ~~Test/implement exception cases~~
- ~~Test/implement API entrypoint for json based search~~
- ~~Inline documentation for all new/modified code~~

The above are completed.
This is no longer a Proof of concept. This should be production ready now.
Simply adding a `&search_type=json` as a query parameter to the search endpoint will utilise the json query workflow.

## Motivation and Context
An advanced and exact terms based search was required to allow more control to the administrators
More details: https://www.notion.so/lyrasis/Proof-of-concept-Create-new-search-API-with-support-for-additional-metadata-elements-ef25fd3ec79241b9a003683403be752c

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -25,7 +25,7 @@ from elasticsearch_dsl.query import (
     Nested,
 )
 from elasticsearch_dsl.query import Query as BaseQuery
-from elasticsearch_dsl.query import Range, Term, Terms
+from elasticsearch_dsl.query import Range, Regexp, Term, Terms
 from flask_babel import lazy_gettext as _
 from spellchecker import SpellChecker
 
@@ -1941,6 +1941,8 @@ class JSONQuery(Query):
         LTE = "lte"
         LT = "lt"
         GT = "gt"
+        REGEX = "regex"
+        CONTAINS = "contains"
 
     _KEYWORD_ONLY = {"keyword": True}
 
@@ -1993,7 +1995,7 @@ class JSONQuery(Query):
         "classification": "classifications.term",
     }
 
-    def __init__(self, query: Dict, filter=None):
+    def __init__(self, query: Union[str, Dict], filter=None):
         if type(query) is str:
             try:
                 query = json.loads(query)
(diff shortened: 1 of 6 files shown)


core/external_search.py
```diff
@@ -1913,6 +1918,138 @@ def _hypothesize(cls, hypotheses, query, boost, filters=None, **kwargs):
...
+            es_query = Term(**{key: value})
+        elif op is self.Operators.NEQ:
+            es_query = Bool(must_not=[Term(**{key: value})])
+        elif op in {
+            self.Operators.GT,
+            self.Operators.GTE,
+            self.Operators.LT,
+            self.Operators.LTE,
+        }:
+            es_query = Range(**{key: {op: value}})
```
io7m, line 2030: I'm inclined to add a `raise` here if we encounter another `Operator` here. We want the code to loudly fail when a new operator is added, rather than being silently incomplete.

core/external_search.py
```diff
@@ -1913,6 +1918,138 @@ def _hypothesize(cls, hypotheses, query, boost, filters=None, **kwargs):
...
+        join = list(query.keys())[0]
+        to_join = []
+        for query_part in query[join]:
+            q = self._parse_json_query(query_part)
+            to_join.append(q)
+
+        if join == self.Conjunctives.AND:
+            joined_query = Bool(must=to_join)
+        elif join == self.Conjunctives.OR:
+            joined_query = Bool(should=to_join)
```
io7m, line 2048: Same goes here. We may never add a new `Conjunctive`, but the code should fail here if we ever do rather than being silently wrong.

io7m (changes requested):
Seems good, except for a couple of statements that could use some paranoid safety measures. :slightly_smiling_face: 

There's also a merge conflict now.

core/external_search.py
RishiDiwanTT: Yes, this is in draft for this reason, a lot of fine tuning is required, this is just to know whether the POC is in the right direction so far

RishiDiwanTT:
@io7m @tdilauro @dwilcox I will continue down this path, and move this from a POC state to a Production state by adding more tests, defensive checks and an API endpoint

core/external_search.py
RishiDiwanTT: We now have exception cases

PR #394, opened 2022-05-31
Attempt to work around a failure to update coverage records

io7m:
## Description

This catches the StaleDataError exception that might be raised
by a long-running Overdrive import when something else happens to
modify overlapping work coverage records at the same time the import
runs. This is, at best, a workaround, and may not actually fix the
problem.

This change essentially catches the exception, rolls back the transaction (which should only affect a single book), and then continues the import.

The real fix is to get rid of the work coverage records table.

## Motivation and Context

Long-running Overdrive imports fail due to this exception, and the exception _might_ indicate a problem that's actually harmless.

Affects: https://www.notion.so/lyrasis/Brooklyn-PL-overdrive_new_title-fails-updating-work-coverage-records-7c42a0c4278d4c5c8ad91c986a4a90b2

## How Has This Been Tested?

Unfortunately, I tried several imports locally and couldn't reproduce the issue. After a rather painful session of inspecting the code, and working from the stack traces given in the original ticket, I think this change might be the best I can do right now.

## Checklist

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -1351,7 +1351,9 @@ class OverdriveCirculationMonitor(CollectionMonitor, TimelineMonitor):
                         progress.exception = e
                     else:
                         time.sleep(1)
-                        self.log.warning(f"retrying book {book}")
+                        self.log.warning(
+                            f"retrying book {book} (attempt {attempt} of {OverdriveCirculationMonitor.MAXIMUM_BOOK_RETRIES})"
+                        )
 
             if self.should_stop(start, book, book_changed):
                 break
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 import re
+import time
 import urllib.parse
 from typing import Any, Dict, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


RishiDiwanTT:
I'm not sure I understand the problem. Do we know what has changed the WorkCoverageRecord, which I assume is only for Overdrive imports?
I can see the stack trace failed at get_one(Collection), did the collection get deleted??

RishiDiwanTT:
I'm not sure about the speculative approach, do we have cloudwatch logs to the same effect?
@tdilauro opinions?

io7m:
To be clear, I'm not happy about it either. The problem is neither the stack traces or the code give any indication as to why this might happen, and I think the issue is less about why this happens and more about the fact that it breaks the entire import when it happens. The above workaround will at least allow the import to continue.

io7m:
Amusingly, I just checked back here to see if there were comments are having implemented pretty much everything you said (except for 10 retries instead of 3-5).

io7m:
I'll make the retry log message a warning. :wrench:

tdilauro (approved):
LGTM. The retries code looks a little more complicated than needed and has the potential to generate more logging than I'd prefer, but looks like it should all work fine.

PR #395, opened 2022-06-01
Fix the 'bytes' object has no attribute 'encode' and support str and bytes in SAML validator and parser

pcustic:
## Description

Fixed the `AttributeError:'bytes' object has no attribute 'encode'` error in SAML validator/parser when running `saml_monitor` script. 
Added support for both `str` and `bytes` types in SAML metadata validator/parser.

## Motivation and Context

This is a result of this [ticket](https://www.notion.so/lyrasis/SAML-Investigate-the-saml_monitor_script-to-see-how-it-is-pulling-InCommon-federation-metadata-and--745955a430344b5aae3133f49b4f8ad4). 

Upon investigating the `saml_monitor` script I found out it doesn't work. Error was in metadata validator and parser due to the fact that `onelogin`'s function `fromstring` accepts `bytes` type. In the previous version of the code the `xml_metadata` and `metadata` vars were first encoded to bytes and then passed to `fromstring`. However, when `saml_monitor` script passes that `{xml_}metadata` to validator/parser it is already a `bytes` type and encoding of `bytes` to `bytes` fails. 

Not to break other parts of the system and other tests I added a check to see if the validator/parser gets `str` and encodes the variable to `bytes` if it does. 
 
## How Has This Been Tested?

Manually ran the `saml_monitor` script. 
Added tests with `bytes` metadata. All new and existing tests passed.

## Checklist

- [X] I have updated the documentation accordingly - [Docs](https://www.notion.so/lyrasis/saml_monitor-script-for-pulling-SAML-federation-metadata-921725d4deaa4e2098c02c939d5832f8)

### DIFF
--- a/api/saml/metadata/federations/validator.py
+++ b/api/saml/metadata/federations/validator.py
@@ -108,7 +108,7 @@ class SAMLFederatedMetadataExpirationValidator(SAMLFederatedMetadataValidator):
         )
 
         try:
-            root = fromstring(metadata.encode("utf-8"))
+            root = fromstring(metadata)
         except Exception as exception:
             raise SAMLFederatedMetadataValidationError(
                 "Metadata's XML is not valid", str(exception)
--- a/api/saml/metadata/federations/validator.py
+++ b/api/saml/metadata/federations/validator.py
@@ -1,10 +1,12 @@
 import datetime
 import logging
 from abc import ABCMeta
+from typing import Union
 
 from onelogin.saml2.utils import OneLogin_Saml2_Utils
 from onelogin.saml2.xmlparser import fromstring
 
+from api.saml.metadata.federations.model import SAMLFederation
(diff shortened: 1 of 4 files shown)


pcustic:
Added type hints in https://github.com/ThePalaceProject/circulation/pull/395/commits/f28c0a77ff32cb130e145bc22745ae659f04f544.

Comment about the tests was good - we had only tests with `str` so I duplicated those tests and encoded input metadata into `bytes` - https://github.com/ThePalaceProject/circulation/pull/395/commits/4ed0757353ed6c1b46201643219bee044c0ee849

tdilauro (approved):
Looks good to me! 🚀 

It's probably not strictly necessary to do `str` and `bytes` tests for every scenario, so long as you exercise each one of them at least once, but what you have is totally fine.

PR #410, opened 2022-06-07
Make the “do not block” SIP setting actually avoid blocking

io7m:
## Description

This adjusts the SIP code so that the checks that would cause a patron to be blocked from borrowing further books are fully disabled if the "do not block" setting is enabled. Previously, the "do not block" setting mistakenly only applied to flag-based blocks instead of fine-based blocks.

## Motivation and Context

The San Jose Public Library does not block users with fines from borrowing digital materials. The issue is that Palace is blocking users with more than $20 in fines from checking out in the app.

https://www.notion.so/lyrasis/Investigate-how-to-not-block-checkout-access-for-barcode-with-fines-f512ca133fae4a1796931f4a14331fe5

## How Has This Been Tested?

The test suite already included some SIP status samples that included excessive fines. I added a new test case to ensure that the user _would_ be blocked for excessive fines if the configuration setting is unset, and that the user _would not_ be blocked if the configuration setting _is_ set. :dizzy_face: 

## Checklist

### DIFF
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -192,10 +192,9 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
             _block = True
 
         if _block:
+            _deny_fields = SIPClient.PATRON_STATUS_FIELDS_THAT_DENY_BORROWING_PRIVILEGES
             self.patron_status_should_block = True
-            self.fields_that_deny_borrowing = (
-                SIPClient.PATRON_STATUS_FIELDS_THAT_DENY_BORROWING_PRIVILEGES
-            )
+            self.fields_that_deny_borrowing = _deny_fields
         else:
             self.patron_status_should_block = False
             self.fields_that_deny_borrowing = []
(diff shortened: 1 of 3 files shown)


api/sip/__init__.py
```diff
@@ -183,12 +184,20 @@ def __init__(
...
-        if patron_status_block is None or patron_status_block:
+
+        # Check if patrons should be blocked based on SIP status
+        # If nothing is specified, the default is True (block patrons!)
+        _block = integration.setting(self.PATRON_STATUS_BLOCK).json_value
+        if _block is None:
+            _block = True
+
+        if _block:
+            self.patron_status_should_block = True
```
RishiDiwanTT, line 195: I don't see this variable being used anywhere else

io7m: OK, this is disturbing. I thought I'd made a mistake and just left in an unused field, but in reality I've actually left out the `if self.patron_status_should_block ...` check later. 

What's disturbing about this? The new unit test shows that with `_block` set to `True`, the excessive fines block happens as it should, and with `_block` set to `False`, the excessive fines block doesn't occur... But we can clearly see that, because I managed to leave out the conditional, the test should fail and the excessive fines should always cause a block. This is into "this can't happen" territory.

Investigating...

RishiDiwanTT: The `info to block` function checks the block fields, I think thats the reason

io7m: I believe everything is consistent now.

RishiDiwanTT (approved):
Looks good!

PR #412, opened 2022-06-08
Skip saml_monitor script run if SAML is not used as patron auth protocol

pcustic:
## Description

Added check if SAML is being used as patron auth protocol. If it is not being used the `saml_monitor` script skips the fetching, parsing, validating and saving of the new SAML federated metadata.

## Motivation and Context

`saml_monitor` script is being run every day and it is updating SAML federated metadata in the database even for sites that are not using SAML as a patron authentication mechanism. Idea is to only update SAML federated metadata if SAML is used as a authentication mechanism to save some resources.

[Notion ticket](https://www.notion.so/lyrasis/Don-t-spend-time-updating-a-SAML-federation-s-metadata-if-the-server-doesn-t-use-it-86029a743bda4e2f852e2336a50dcb85)

## How Has This Been Tested?

Manually run `saml_monitor` script with and without SAML patron auth service.

## Checklist

### DIFF
--- a/api/saml/metadata/monitor.py
+++ b/api/saml/metadata/monitor.py
@@ -2,6 +2,8 @@ import datetime
 import logging
 
 from api.saml.metadata.federations.model import SAMLFederation
+from api.saml.provider import SAMLWebSSOAuthenticationProvider
+from core.model import ExternalIntegration
 from core.monitor import Monitor
 from core.util.datetime_helpers import utc_now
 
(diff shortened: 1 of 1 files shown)


pcustic:
Switching to draft since I realised ticket wanted to skip updating metadata if the federation is not used, not just if SAML is not used.

pcustic:
Blocked, see [ticket](https://www.notion.so/lyrasis/Don-t-spend-time-updating-a-SAML-federation-s-metadata-if-the-server-doesn-t-use-it-86029a743bda4e2f852e2336a50dcb85) comment section.

PR #414, opened 2022-06-08
Fix broken LCP prioritization

io7m:
## Description

This adjusts the way the LCP prioritization configuration settings are exposed in order to prevent the admin UI from placing string labels into the database.

## Motivation and Context

The previous "de-prioritize LCP content" feature was broken because the
configuration settings were exposed in a manner that allowed the Admin
UI to place string labels into the database instead of the enumeration
values that were intended to go in instead. This switches to a model
where we store boolean values, and includes a database migration to
fix up any databases that were affected.

Affects: https://www.notion.so/lyrasis/De-prioritize-LCP-switch-is-broken-c9b295a78c1142a1922ed2d7e3143836

## How Has This Been Tested?

I examined my local database, and also ran the migration to check that it works.

## Checklist

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -32,7 +32,7 @@ from core.model import (
     Patron,
     Session,
 )
-from core.model.formats import DePrioritizeLCPNonEPUBs, FormatPriorities
+from core.model.formats import FormatPriorities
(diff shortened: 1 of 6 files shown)


io7m:
Ugh, one sec. Created this branch from the wrong starting branch.

PR #416, opened 2022-06-08
Fix LCP prioritization

io7m:
## Description

This adjusts the way the LCP prioritization configuration settings are exposed in order to prevent the admin UI from placing string labels into the database.

## Motivation and Context

The previous "de-prioritize LCP content" feature was broken because the
configuration settings were exposed in a manner that allowed the Admin
UI to place string labels into the database instead of the enumeration
values that were intended to go in instead. This switches to a model
where we store boolean values, and includes a database migration to
fix up any databases that were affected.

Affects: https://www.notion.so/lyrasis/De-prioritize-LCP-switch-is-broken-c9b295a78c1142a1922ed2d7e3143836

## How Has This Been Tested?

I examined my local database, and also ran the migration to check that it works.

## Checklist


migration/20220608-fix-broken-deprioritize-lcp-config.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+        )
+    )
+    db.execute(
+        text(
+            """
+UPDATE configurationsettings
+  SET value = 'true'
+  WHERE value = 'De-prioritize'
+    """
```
tdilauro, line 36: We don't need to do all of our migrations using Python. This migration would be shorter and easier to understand if done in SQL:

```
-- The purpose of this migration is to fix any weird values that were inserted
-- into the configurationsettings table by the Admin API. Previous versions of
-- the code represented the LCP priority setting as an enum, and the Admin API
-- inserted the configuration labels rather than the enum values into the database.

UPDATE configurationsettings SET value = 'false' WHERE value = 'Do not de-prioritize';
UPDATE configurationsettings SET value = 'true' WHERE value = 'De-prioritize';
```

tdilauro (changes requested):
This looks fine. I'd like to change the migration from Python to strictly SQL. I'd like us to get back to using SQL where reasonable, as it is usually a lot more concise and easy to understand. The existing migration machinery handles a lot of the work for us.

tdilauro:
@dbernstein Would you give this a review, too?

dbernstein (approved):
Looks good to me as well.

migration/20220608-fix-broken-deprioritize-lcp-config.py
io7m: Noted! I actually didn't realize we could include SQL files directly and they'd be executed. I saw other files, but I thought they were being opened from Python scripts.

Had I known, this would have been pure SQL...

tdilauro: No worries. I figured this might be the case. As you suspected, the pure SQL migrations are indeed being opened by a python script. It's just that the script is part of the migration machinery, rather than something custom to a specific migration.

tdilauro (approved):
LGTM! 🎸

PR #419, opened 2022-06-09
Assert preview link imported

RishiDiwanTT:
## Description
Simple test addition to ensure SAMPLE links are tested for in ODL imports
The functionality always worked, this is to make sure it keeps working

## Motivation and Context
The Palace Marketplace feed introduced sample links, we have added an `assert` to mark the addition of this feature

## Checklist


pcustic (approved):
Seems good to me 👍

PR #439, opened 2022-06-23
Require library IDs in customlists import/export scripts

io7m:
## Description

This changes the customlist import and export scripts to explicitly take library short identifiers as parameters instead of extracting them from OPDS feeds. This allows for extracting only those lists that belong to a given library, and importing them into a specific library on a target CM.

## Motivation and Context

@tdilauro Noted that the existing scripts didn't quite fit the migration use case.

https://www.notion.so/lyrasis/Migrate-discrete-entries-in-client-library-lists-from-old-CM-to-new-CM-1b78b13b10234ee59830c343ef4862dc

The existing test suite was updated to include the new endpoint calls, and I tested this on a local CM across a couple of libraries.

## Checklist

### DIFF
--- a/customlists/README.md
+++ b/customlists/README.md
@@ -37,9 +37,9 @@ optional arguments:
 
 ```bash
 $ ./bin/customlist_import --help
-usage: customlist_import [-h] --server SERVER --username USERNAME
---password PASSWORD [--schema-file SCHEMA_FILE] [--schema-report-file SCHEMA_REPORT_FILE]
---file FILE --output OUTPUT [--dry-run] [--verbose]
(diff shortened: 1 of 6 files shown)


RishiDiwanTT (approved):
Seems straightforward enough. LGTM!

tdilauro:
I haven't had a chance to re-test this, but I guess we can open another PR if it's still not working.

io7m:
I accidentally pushed to a branch that had already been merged and deleted here. The new PR is here: https://github.com/ThePalaceProject/circulation/pull/442

PR #442, opened 2022-06-24
Attempt to use ID values from URIs instead of id elements

io7m:
## Description

This adjusts the code to take book ID and ID types from the links
inside OPDS feeds, instead of depending on the value of the "id"
element (which may not be usable as a work ID on the target CM). This
also adjusts the code to remove any percent encoding that may be in
use on the identifiers.

## Motivation and Context

The scripts failed in testing on real data due to differences between the IDs that appear in OPDS links,
and the IDs that appear in OPDS `id` elements. This should fix those failures.

Affects: https://www.notion.so/lyrasis/Migrate-discrete-entries-in-client-library-lists-from-old-CM-to-new-CM-1b78b13b10234ee59830c343ef4862dc

## How Has This Been Tested?

I updated the test suite and added some specific identifiers that were problematic in testing. I also tested on a local CM.

## Checklist

### DIFF
--- a/customlists/customlist_export.py
+++ b/customlists/customlist_export.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 from typing import IO, Iterable, List, Union
+from urllib.parse import unquote
 
 import feedparser
 import jsonschema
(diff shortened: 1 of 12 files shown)


RishiDiwanTT (approved):
LGTM!

io7m:
Sorry, but this is _still_ not correct. See the ticket above for details; I'm attempting to work out which minor thing was different about _these_ identifiers that's causing the CM to throw exceptions.

io7m:
I ... feel like I dispute that failed `mypy` check, especially given that it passed for all other checks.

PR #451, opened 2022-07-02
alpha3 only codes

barmintor:
## Description

Changes two of the mapping methods in core/utils/languages to support all the ISO-639-2 alpha-3 codes, including some edge cases that raise ValueErrors and interrupt OPDS processing:
- when there is no alpha-2
- when it's a terminologic code
- when it's a code reserved for local use

## Motivation and Context

Our lane processing fails with:
```bash
  File "/var/www/circulation/core/scripts.py", line 135, in run
    timestamp_data = self.do_run()
  File "/var/www/circulation/core/scripts.py", line 617, in do_run
    self.process_libraries(parsed.libraries)
  File "/var/www/circulation/core/scripts.py", line 621, in process_libraries
    self.process_library(library)
  File "/var/www/circulation/scripts.py", line 404, in process_library
    super().process_library(library)
  File "/var/www/circulation/core/scripts.py", line 734, in process_library
    self.process_lane(l)
  File "/var/www/circulation/scripts.py", line 428, in process_lane
    feed = self.do_generate(lane, facets, pagination)
  File "/var/www/circulation/scripts.py", line 654, in do_generate
    return feed_class.groups(
  File "/var/www/circulation/core/opds.py", line 626, in groups
    return CachedFeed.fetch(
  File "/var/www/circulation/core/model/cachedfeed.py", line 149, in fetch
    feed_data = str(refresher_method())
  File "/var/www/circulation/core/opds.py", line 614, in refresh
    return cls._generate_groups(
  File "/var/www/circulation/core/opds.py", line 719, in _generate_groups
    annotator.annotate_feed(feed, worklist)
  File "/var/www/circulation/api/opds.py", line 969, in annotate_feed
    if lane and lane.search_target:
  File "/var/www/circulation/core/lane.py", line 3072, in search_target
    display_name_parts.append(LanguageCodes.name_for_languageset(languages))
  File "/var/www/circulation/core/util/languages.py", line 615, in name_for_languageset
    raise ValueError("No native or English name for %s" % l)
ValueError: No native or English name for ang
```

This is apparently because of publications - Beowulf translations - that list their languages as `eng` and `ang`. The latter is appropriate and valid (English, Old (ca.450-1100)) but because it has no alpha-2 code, it is not present in the lookup table currently used to validate and normalize alpha-3 values.

## How Has This Been Tested?

Added relevant unit test cases to tests/core/test_languages.py for `LanguageCodes.name_for_languageset` and `LanguageCodes.string_to_alpha_3`

## Checklist

### DIFF
--- a/core/util/languages.py
+++ b/core/util/languages.py
@@ -30,6 +30,7 @@ class LanguageCodes:
 
     two_to_three = LookupTable()
     three_to_two = LookupTable()
+    terminologic_to_three = LookupTable()
     english_names: Dict[str, List[str]] = defaultdict(list)
     english_names_to_three = LookupTable()
     native_names: Dict[str, List[str]] = defaultdict(list)
@@ -540,6 +541,9 @@ zza|||Zaza; Dimili; Dimli; Kirdki; Kirmanjki; Zazaki|zaza; dimili; dimli; kirdki
         {"code": "sv", "name": "Swedish", "nativeName": "svenska"},
     ]
 
+    RESERVED_CODES = re.compile("^q[a-t][a-z]$")
+    RESERVED_CODE_LABEL = "Reserved for local use"
+
     for raw_name_data in RAW_DATA.split("\n"):
         (
             alpha_3,
@@ -553,6 +557,8 @@ zza|||Zaza; Dimili; Dimli; Kirdki; Kirmanjki; Zazaki|zaza; dimili; dimli; kirdki
             three_to_two[alpha_3] = alpha_2
             english_names[alpha_2] = names
             two_to_three[alpha_2] = alpha_3
+        if terminologic_code:
+            terminologic_to_three[terminologic_code] = alpha_3
         for name in names:
             english_names_to_three[name.lower()] = alpha_3
         english_names[alpha_3] = names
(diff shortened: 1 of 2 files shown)


barmintor:
There are 2 API failures:
### TestLaneCreation.test_create_lane_for_small_collection 
Test is failing because it is successfully creating a lane for the languages `["eng", "mul", "chi"]`, but expected not to create the lane. `mul` is the alpha-3 code for "Multiple Languages; it does not have an alpha-2 code but otherwise has the relevant data.

### TestLaneCreation.test_lane_for_tiny_collection
Test is failing because it is successfully creating a lane for the languages `["spa", "gaa", "eng"]`, but expected not to create the lane. `gaa` is the alpha-3 code for "Ga"; it does not have an alpha-2 code but otherwise has the relevant data.

These are both the same failure - is it the case that an alpha-2 code must be available? The codebase generally seems to use alpha-3; there are some legitimate languages for books that lack the alpha-2 code and failing lane processing when they are encountered is a problem that might not be addressable by the library.

io7m (changes requested):
I don't see anything wrong with the implementation, but there are some test failures. Could you take a look at those and update them?

io7m:
At first inspection, I don't see anything in the code that explicitly requires that there be an alpha-2 code, but it is the CM, where complicated relationships are usually implicit and undocumented. I'd personally be inclined to moving towards requiring an alpha-3 code, and actually start enforcing it.

barmintor:
@io7m I'm looking at the rest of that class - the dictionary `three_to_two` that maps *to* alpha2 is not really used in a mapping method (there's one more place where it is used to validate an alpha3 but should not).

So the places where an alpha2 is expected in the CM should be indicated by direct references to `LanguageCodes.three_to_two` or `two_to_three` in the codebase.

barmintor:
Ok, interestingly:

`LanguageCodes.three_to_two` is used with an alpha3 (presumably) default to effect the "shortest code available" requirement from BCP-47, without really documenting that. I have replaced that with an explicitly-named mapping method which should clarify intent.

`LanguageCodes.two_to_three` is used in the two places in the overall application configuration to normalize localization language codes - in those situations the more limited set of codes seems more justifiable, though maybe not ideal.

The failing tests are documented to be verifying that lane creation fails when a language code has no _name_. This was only incidentally true of the test data, which was for codes that _did_ have a name but did not have the alpha-2 that was used to look names up. I have replaced that with unambiguously bad test data, which should now successfully fail to create lanes.

I recommend taking up the question of whether some of the alpha3-only codes should be *specifically forbidden* separately, and maybe not until the lane creation logic is more generous (maybe by just requiring a non-zero number of mapped names rather than zero unmapped codes). Conversation in the slack pointed to several potentially special cases:

1. `mul` "Multiple languages", which would ideally be represented by enumerating the languages in question
2. `unk` for "Unknown"
3. the reserved codes (qaa-qtz)

io7m:
I've just kicked off the tests, so we'll see what happens. :slightly_smiling_face: 

Overall, this looks good.

PR #453, opened 2022-07-05
Feature/index edition published

RishiDiwanTT:
## Description
Added edition.published to the search index per work
Made the ES work index mapping updatable for new properties

## Motivation and Context
A new property `edition.published` was required in the search index in order to facilitate search on that property

## How Has This Been Tested?
Manual and Unit testing

## Checklist

### DIFF
--- a/tests/core/test_external_search.py
+++ b/tests/core/test_external_search.py
@@ -4503,9 +4503,6 @@ class TestSearchIndexCoverageProvider(DatabaseTest):
         # Top level keys should be the same
         assert len(result) == len(inapp)
 
-        # This test is no longer valid with new mapping properties
-        # assert result[0].keys() == inapp[0].keys()
-
         inapp_work1 = list(filter(lambda x: x["work_id"] == work1.id, inapp))[0]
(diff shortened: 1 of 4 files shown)


tests/core/test_external_search.py
```diff
@@ -4479,7 +4498,9 @@ def test_to_search_document(self):
 
         # Top level keys should be the same
         assert len(result) == len(inapp)
-        assert result[0].keys() == inapp[0].keys()
+
+        # This test is no longer valid with new mapping properties
```
dbernstein, line 4502: @RishiDiwanTT :  if this test is legitimately no longer valid,  it should be removed no?

dbernstein (changes requested):

tests/core/test_external_search.py
RishiDiwanTT: Yes, I seem to have this habit of not deleting old code. Will remove.

dbernstein (approved):
Nice work @RishiDiwanTT .  Feel free to merge it.

PR #454, opened 2022-07-06
Correct a possible mistake in import id checking

io7m:
## Description

This adjusts the comparison for import IDs so that the _unquoted_ OPDS entry ID is compared with the full book ID. My kingdom for a stronger type system.

## Motivation and Context

It appears that the code was comparing the wrong id elements, leading
to an import failure and a misleading error message. This should fix
that. This also introduces more debug logging, so we can see what's
gone wrong on the next inevitable failure.

Affects: https://www.notion.so/lyrasis/Migrate-discrete-entries-in-client-library-lists-from-old-CM-to-new-CM-1b78b13b10234ee59830c343ef4862dc

## How Has This Been Tested?

I ran through the usual "try it on a local CM" process, and updated the test suite again.

## Checklist

### DIFF
--- a/customlists/customlist_export.py
+++ b/customlists/customlist_export.py
@@ -408,6 +408,14 @@ class CustomListExporter:
                         link_id: str = unquote(link_id_quoted)
                         link_id_type: str = unquote(link_id_type_quoted)
 
+                        self._logger.debug(f"storing link id (id_value) {link_id}")
+                        self._logger.debug(
+                            f"storing link id type (id_type) {link_id_type}"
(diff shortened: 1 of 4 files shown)


RishiDiwanTT (approved):
I honestly have no idea what has changed here, since it's a trivial change. Code-wise it all seems fine

io7m:
> I honestly have no idea what has changed here, since it's a trivial change. Code-wise it all seems fine

Silly mistake in the comparison if ID values.

PR #458, opened 2022-07-12
Added the media_type of a sample link to the OPDS feed

RishiDiwanTT:
## Description
The sample links in an OPDS1.2 feed will now be of the format
`<link rel="http://opds-spec.org/acquisition/sample" href="https://market.feedbooks.com/item/4152047/preview" type="application/epub+zip"/>`

## Motivation and Context
Without the media-type in the sample link an App would only find out the kind of viewer required for the SAMPLE after downloading the content.
This way the right viewer can be opened before-hand.

## Checklist


io7m (approved):
Looks good to me.

PR #460, opened 2022-07-12
Run `update_lane_size` overnight to avoid daytime performance issues.

tdilauro:
## Description

Modify the script schedule to run `update_lane_size` overnight.

## Motivation and Context

When `update_lane_size` runs during the morning hours, it interacts with (an)other script(s) and causes response time issue. Killing the process associated with that script leads to an almost immediate return to responsiveness. It is possible that `update_lane_size` will continue to cause similar performance problems, just in the overnight hours, instead of the morning. But at least in that case, it will affect fewer patrons.

## How Has This Been Tested?

Carefully reviewed the change to the `crontab`.

## Checklist

  - We will need to update scripts page, once this change is deployed. 


io7m (approved):
Let's hope this doesn't mean someone has to be up at 2am to kill it. :laughing:

PR #462, opened 2022-07-14
Feature/loan notifications

RishiDiwanTT:
## Description
Loan notifications script for sending remote notifications to client apps
    
Uses FCM for push notifications
Pushes out notifications on day 5 and day 1 before loan expiry
poetry.lock has changed a lot by just adding fcm_admin, but all tests pass
Loans Notification is scheduled for 12:10 PM Server Time (EST) Daily

## Motivation and Context
In order to alert users to their expiring loans, we can send notifications to the Patrons apps
FCM is a widely used push notification service that works across both iOS and Android

## How Has This Been Tested?
Unit tests have been written, manual testing is impossible without an App that is built with the FCM service account keys

## Checklist


README.md
```diff
@@ -148,6 +148,7 @@ grant all privileges on database circ to palace;
 #### Environment variables
 
 To let the application know which database to use set the `SIMPLIFIED_PRODUCTION_DATABASE` env variable.
+The `SIMPLIFIED_FCM_CREDENTIALS_FILE` env variable should hold the path to the JSON format GCP service account key
```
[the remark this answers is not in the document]
RishiDiwanTT: If you don't run the `loan_notifications` job locally nothing goes wrong. If you do you will get an authentication or badconfiguration error depending on whether the variable is present or not

core/util/notifications.py
```diff
@@ -0,0 +1,49 @@
...
+    ):
+        responses = []
+        edition: Edition = loan.license_pool.presentation_edition
+        identifier: Identifier = loan.license_pool.identifier
+        library_short_name = loan.library and loan.library.short_name
+        for token in tokens:
+            msg = messaging.Message(
+                token=token.device_token,
+                data=dict(
+                    title=f"Only {days_to_expiry} {'days' if days_to_expiry != 1 else 'day'} left on your loan!",
```
io7m, line 39: Nice. I like it when people take extra care to get this sort of singular/plural thing right in messages.

io7m (approved):
This seems fine to me!

io7m (dismissed):
I've just realized there are test failures. Marking this as needing changes so that it doesn't accidentally get merged.

RishiDiwanTT:
Converted this to draft until we know what information the Mobile Apps require in the notifications

RishiDiwanTT:
@tdilauro @io7m @dbernstein This PR was in draft due to the loan/hold refresh requirements being in flux. This has been sorted out, so this (now monstrously large) PR is ready for review.
Note, the original PR was approved apart from some test failures, and the adjoining PR's (hold and patron sync) which were merged into this were approved themselves.

README.md
dbernstein, line 151: Can you add a sample fcm credentials file and any other information related to how to set one up so that users don't need to hunt it down?

core/jobs/holds_notification.py
```diff
@@ -0,0 +1,25 @@
...
+from core.util.notifications import PushNotifications
+
+
+class HoldsNotificationMonitor(SweepMonitor):
+    """Sweep across all holds that are ready to be checked out by the user (position=0)"""
+
+    MODEL_CLASS: Optional[Type[Base]] = Hold
+    SERVICE_NAME: Optional[str] = "Holds Notification"
+
+    def scope_to_collection(self, qu, collection):
```
dbernstein, line 15: param and return types please (for readability)

core/jobs/holds_notification.py
```diff
@@ -0,0 +1,25 @@
...
+        """Do not scope to collection"""
+        return qu
+
+    def item_query(self):
```
dbernstein, line 19: type please (for readability)

core/jobs/holds_notification.py
```diff
@@ -0,0 +1,25 @@
...
+        query = super().item_query()
+        query = query.filter(Hold.position == 0)
+        return query
+
+    def process_items(self, items):
```
dbernstein, line 24: param type please (for readability)

core/jobs/patron_activity_sync.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class PatronActivitySyncNotificationScript(PatronSweepMonitor):
+    """Find patrons with stale last_activity_sync timestamps, and also who have loans/holds
+    and notify said patron devices to re-sync their data"""
+
+    STALE_ACTIVITY_SYNC_DAYS = 2
+    SERVICE_NAME: Optional[str] = "Patron Activity Sync Notification"
+
+    def item_query(self):
```
dbernstein, line 21: return type in method signature

core/monitor.py
```diff
@@ -664,6 +664,16 @@ def item_query(self):
         return super().item_query().filter(not_presentation_ready)
 
 
+class PatronSweepMonitor(SweepMonitor):
+    """Sweep through all Patrons"""
+
+    MODEL_CLASS: Optional[Type[Base]] = Patron
+
+    def scope_to_collection(self, qu, collection):
```
dbernstein, line 672: param types and return type in method signature

core/util/notifications.py
```diff
@@ -0,0 +1,146 @@
...
+    @property
+    def fcm_app(cls):
+        if not cls._fcm_app:
+            cls._fcm_app = firebase_admin.initialize_app(
+                credentials.Certificate(Configuration.fcm_credentials_file())
+            )
+        return cls._fcm_app
+
+    @classmethod
+    def base_url(cls, _db):
```
dbernstein, line 43: param and return type in method signature

core/util/notifications.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+            resp = messaging.send(msg, dry_run=cls.TESTING_MODE, app=cls.fcm_app)
+            responses.append(resp)
+        return responses
+
+    @classmethod
+    def send_activity_sync_message(cls, patrons: List[Patron]) -> List[str]:
+        """Send notifications to the given patrons to sync their bookshelves
+        Enough information needs to be sent to identify a patron on the mobile Apps
+        and make the loans api request with the right authentication"""
```
dbernstein, line 88: Add punctuation (periods) to the docs.

dbernstein (changes requested):
Excellent - just some method signatures and supremely minor doc fixes.

dbernstein (approved):

RishiDiwanTT:
@jonathangreen @tdilauro I will hold off merging until told otherwise

tests/core/test_config.py
```diff
@@ -98,3 +100,27 @@ def load_cdns(cls, _db, config_instance=None):
         assert new_db != self._db
         assert isinstance(new_db, Session)
         assert None == none
+
+    def test_fcm_credentials_file(self):
+        env_var = os.environ.pop(
```
tdilauro, line 105: Should we use `monkeypatch`'s `setenv` and `delenv` here to isolate changes to the environment to make them safer?

tdilauro (commented):
This looks pretty good. A few questions and suggestions.

Also, I think it should be possible to configure the CM  with notifications turned off, in which case processing device tokens/notifications should basically be no-ops and not result in errors. This point might need a little more discussion on the ticket.

core/model/patron.py
```diff
@@ -60,6 +66,7 @@ class Patron(Base):
     # individual human being may patronize multiple libraries, but
     # they will have a different patron account at each one.
     library_id = Column(Integer, ForeignKey("libraries.id"), index=True, nullable=False)
+    library: "Library"
```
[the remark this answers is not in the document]
RishiDiwanTT: updated

tests/core/test_holds_notifications.py
```diff
@@ -0,0 +1,37 @@
...
+        work1 = self._work(with_license_pool=True)
+        work2 = self._work(with_license_pool=True)
+        work3 = self._work(with_license_pool=True)
+        work4 = self._work(with_license_pool=True)
+        hold1, _ = work1.active_license_pool().on_hold_to(patron1, position=1)
+        hold2, _ = work2.active_license_pool().on_hold_to(patron1, position=0)
+        hold3, _ = work3.active_license_pool().on_hold_to(patron1, position=0)
+        hold4, _ = work4.active_license_pool().on_hold_to(patron1, position=None)
+
+        # Only position 0 holds should be pushed queried for
```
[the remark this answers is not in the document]
RishiDiwanTT: Updated

tests/core/test_config.py
RishiDiwanTT: This is like the most basic method of testing and I don't know why I did what I did. Updated.

core/scripts.py
```diff
@@ -3581,6 +3584,61 @@ def _update_list_with_new_entries(self, custom_list: CustomList):
...
+    # Days before on which to send out a notification
+    LOAN_EXPIRATION_DAYS = [5, 1]
+    BATCH_SIZE = 100
+
+    def do_run(self):
+        self.log.info("Loan Notifications Job started")
+        _query = self._db.query(Loan).order_by(Loan.id)
+        last_loan_id = None
+        processed_loans = 0
+        for _ in range(100000):
```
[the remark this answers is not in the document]
RishiDiwanTT: Yyou can call him "Courage, the Cowardly while true".
I've added a comment to make it clear that this is a fail-safe max number of loops.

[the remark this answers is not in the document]
tdilauro: Can you say a little more in the comment about why you think we need a failsafe here (e.g., result set shifting with each batch) and possibly add a TODO to bring us back here to fix in the future.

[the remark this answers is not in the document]
RishiDiwanTT: There is absolutely no reason I have to believe this will actually become an infinite loop. It's just some defensive code, it can become a `while true` now if required. Should I change it?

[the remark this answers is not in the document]
tdilauro: My concern about the defensive coding here is that if it actually came into play, it would silently mask the underlying problem. So, if we keep it, it would be good to add an `else` clause to the `for` loop to emit an error about what just happened. We might also want to pick a number for the `range()` call this is tied more closely to our state (e.g., total-loan-count/batch-size). Or we could just go with `while True` and cross our fingers. 😄 🤞🏽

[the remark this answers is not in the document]
jonathangreen: From my perspective I wouldn't call this a blocker, but I thought I'd throw my 2 cents 🪙 in here, on my preferences for loops. 

Personally I prefer `while True` over `for _ in range(100000)`. But what I'd most prefer is something with a more explicit conditional like `while len(loans) != 0`, but that would take some restructuring of the logic here.

jonathangreen:
> Also, I think it should be possible to configure the CM with notifications turned off, in which case processing device tokens/notifications should basically be no-ops and not result in errors.

I agree here. For development and testing, it would be really nice if this was optional, and just threw a warning or info statement in the log that let you know this feature isn't configured.

I think we also need some more details in the README about what this configuration is used for and some outline about how these notifications work, how to test etc.

core/scripts.py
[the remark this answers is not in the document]
RishiDiwanTT: Updated this to be a `while true` since we already do a `len == 0: break` in the loop

RishiDiwanTT:
> I think we also need some more details in the README about what this configuration is used for and some outline about how these notifications work, how to test etc.

I added some information about the settings in the README, currently we don;t have a way to test push notifications either, unless we wait for a loan or hold reminder.

tests/core/test_config.py
```diff
@@ -98,3 +101,19 @@ def load_cdns(cls, _db, config_instance=None):
         assert new_db != self._db
         assert isinstance(new_db, Session)
         assert None == none
+
+    @patch.object(os, "environ", new=dict())
```
tdilauro, line 105: Another option that `pytest` gives us `monkeypatch.setenv` (and `monkeypatch.delenv`)... https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.setenv, but this should work, too.

core/jobs/holds_notification.py
```diff
@@ -0,0 +1,45 @@
...
+    def run_once(self, *ignore):
+        setting = ConfigurationSetting.sitewide(
+            self._db, Configuration.PUSH_NOTIFICATIONS_STATUS
+        )
+        if setting.value == ConfigurationConstants.FALSE:
+            self.log.info(
+                "Push notifications have been turned off in the sitewide settings, skipping this job"
+            )
+            return
+        return super().run_once(*ignore)
```
tdilauro, line 33: What happens if this is turned on, but the configuration file is not set or is not present?

Ideally, we'd want to log an error indicating the condition, but not have the jobs limp along throwing an error for each hold (in this case), loan, or patron sync notification.

tdilauro (commented):
A few more comments, which I think address my remaining concerns at this point. @jonathangreen may have additional thoughts.

tests/core/test_config.py
RishiDiwanTT: I generally like to stick to builtin unit-testing where available

core/jobs/holds_notification.py
RishiDiwanTT: The README states the jobs will run with an 'unset' setting, which basically means True is the default. I can switch it around to be False as default if this means the implementation teams will have their hands full during the deployment

README.md
```diff
@@ -360,6 +370,24 @@ the `xcode-select --install` command. If it does not work, you can try adding th
...
+#### hold_notifications
+
+Requires the `SIMPLIFIED_FCM_CREDENTIALS_FILE` to be present
+and the sitewide `PUSH_NOTIFICATIONS_STATUS` to be either `unset` or `true`
+
+#### loan_notifications
+
+Requires the `SIMPLIFIED_FCM_CREDENTIALS_FILE` to be present
+and the sitewide `PUSH_NOTIFICATIONS_STATUS` to be either `unset` or `true`
+
```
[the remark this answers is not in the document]
RishiDiwanTT: Done

README.md
```diff
@@ -337,6 +337,16 @@ Examining collection "Palace Bookshelf"
 We can see from the above output that the vast majority of the books in the _Open Bookshelf_ collection
 were indexed correctly.
 
+### Sitewide Settings
+
+Some settings have been provided in the admin UI that configure or toggle various functions of the Circulation Manager.
+These can be found at `/admin/web/config/SitewideSettings`
```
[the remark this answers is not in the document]
RishiDiwanTT: Done

README.md
```diff
@@ -337,6 +337,16 @@ Examining collection "Palace Bookshelf"
...
+
+#### Push Notification Status
+
+This setting is a toggle that may be used to turn on or off the ability for the the system
+to send the Loan and Hold reminder mobile notifications.
```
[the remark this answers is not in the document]
RishiDiwanTT: Done

RishiDiwanTT:
@tdilauro @jonathangreen Any additional feedback?

RishiDiwanTT:
@ThePalaceProject/backend Ping!

RishiDiwanTT:
@jonathangreen @io7m @tdilauro 
https://media.giphy.com/media/59fYAHs0Y4llMQ7Lnh/giphy.gif

RishiDiwanTT:
The rebase has caused an issue, [Firebase Admin SDK](https://github.com/firebase/firebase-admin-python) does not support python 3.11 yet.

RishiDiwanTT:
@jonathangreen @tdilauro python 3.11 works.
Upgraded firebase-admin to 6.0.1 (this was out last year, I didn't realize)
Had to upgrade PyJWT to 2.6.0 as a dependancy

jonathangreen:
@io7m I dismissed your review on this one, since it was from some time ago.

PR #465, opened 2022-07-19
Feature/patron pwd reset link

RishiDiwanTT:
## Description
Added Patron password reset link for Library admin configuration
Authentication document now has patron password reset link
Under rel="http://librarysimplified.org/terms/rel/patron-password-reset"

The authentication document will have the additional data as follows
image

## Motivation and Context
Patrons would need a way to reset forgotten passwords directly from the Palace App
This information needed to be configured per library
The authentication document links can be extended with URI type rels as per [the spec](https://drafts.opds.io/authentication-for-opds-1.0.html#232-links)

## How Has This Been Tested?
Unit and manual testing done

## Checklist

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -151,6 +151,7 @@ class Configuration(CoreConfiguration):
     ABOUT = "about"
     LICENSE = "license"
     REGISTER = "register"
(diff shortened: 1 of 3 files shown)


io7m (approved):
Looks good!

PR #468, opened 2022-07-22
Removed the uniqueness on device_tokens

RishiDiwanTT:
## Description
Introduced device_token, patron_id uniqueness

## Motivation and Context
Every User with multiple Libraries on their App will have multiple Patrons attached to the same device.
These Patrons may be on the same CM backend, which means they would share device tokens in the same DB.
The device_token uniqueness constraint would prevent this, so we change the device_token uniqueness to include the patron_id as well

## Checklist
Manual and Unit testing done


dbernstein (approved):
Looks good. Ready to merge.

PR #469, opened 2022-07-25
Feature/samlsettings refresh

RishiDiwanTT:
## Description
Self refreshing SAMLSettings for federated IdPs

2 changes had to be made for this
- The PatronAuthServicesController needed to have its self.protocols refresh           
- The SAMLSettings needed to be able to detect a change within the DB

PatronAuthServicesController.protocols was made a cacheable callable
with 30 minute caching, so the Federated IdPs will refresh only
30 minutes after the last access

SAMLSettings queries for the last updated time for incommon SAMLFederations
and refreshes the values as needed

A memoize function was written to facilitate the time-bound caching
for PatronAuthServicesController.protocols
It currently works in-memory only, but can be extended to fit
Redis cache when required, or wholly replaced by a python library

## Motivation and Context
The Federated IdPs list would not update unless the CM server was restarted, this needed to be rectified to be a more dynamic system.
[Ticket](https://www.notion.so/lyrasis/Make-current-list-of-SAML-federated-IdPs-appear-in-Admin-UI-without-restart-b30aa81ae65e4edeaf8ebf14aa0085de)

## How Has This Been Tested?
Unit and Manual testing, after running saml_import

## Checklist

### DIFF
--- a/api/admin/controller/patron_auth_services.py
+++ b/api/admin/controller/patron_auth_services.py
@@ -17,6 +17,7 @@ from api.saml.provider import SAMLWebSSOAuthenticationProvider
 from api.simple_authentication import SimpleAuthenticationProvider
 from api.sip import SIP2AuthenticationProvider
 from core.model import ConfigurationSetting, ExternalIntegration, get_one
+from core.util.cache import memoize
 from core.util.problem_detail import ProblemDetail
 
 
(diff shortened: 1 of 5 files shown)


dbernstein (approved):
Very slick, @RishiDiwanTT .

PR #472, opened 2022-07-29
Blank values in the Library settings are now considered as valid values

RishiDiwanTT:
## Description
Only NoneType values are ignored now

## Motivation and Context
Blank values are still valid values when optional fields are deleted, however, these values were being ignored as "falsey" values.

[Notion ticket](https://www.notion.so/lyrasis/Removing-a-value-from-an-Admin-UI-field-does-not-save-the-blank-value-55dacb3de5e94e34a1fdad6e450b64c5).

## How Has This Been Tested?

- New tests added.
- All tests ran successfully.

## Checklist
Unit and Manual testing was done

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -356,7 +356,9 @@ class LibrarySettingsController(SettingsController):
             if type == "image":
                 value = self.image_setting(setting) or default_value
             else:
-                value = self.scalar_setting(setting) or default_value
+                value = self.scalar_setting(setting)
+                # An empty "" value or 0 value is valid, hence check for None
+                value = default_value if value is None else value
         return value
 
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
This looks like it works as described, but I'm not sure which ticket it's associated with.

RishiDiwanTT:
> This looks like it works as described, but I'm not sure which ticket it's associated with.

Missed that, here's the [Notion ticket](https://www.notion.so/lyrasis/Removing-a-value-from-an-Admin-UI-field-does-not-save-the-blank-value-55dacb3de5e94e34a1fdad6e450b64c5)

tdilauro:
Thanks! I updated the main PR comment with the ticket and some info in the testing section. This looks good to go from my perspective.

PR #475, opened 2022-08-02
Patron activity sync scheduled job

RishiDiwanTT:
## Description
Should send out notifications to patron mobile apps that
    - Have a device token
    - Have either a loan or hold active
    - Have potentially stale loan/hold activity data

The job is scheduled for 1AM daily
A loans endpoint has also been added to this and the loans notifications

There is an issue with the current implementation of last_activity_sync_time
wherein the getter finds the sync time to be stale in 15 minutes and clears out the
last sync time on access, we may want to change this implementation to not clear out
the information and add a has_activity_sync_expired() method, allowing each type of
workflow define when a sync is stale

## Motivation and Context
The loans and holds information is often out of date due to out of system changes that occur. Due to this the patron must re-sync their bookshelf in a periodic fashion.
[Notion](https://www.notion.so/lyrasis/Push-Notifications-Patron-activity-sync-32fbabd6c78b4de2bf9264416ba3ee86)

## How Has This Been Tested?
Unit testing has been done, manual testing is not possible without device tokens

## Checklist

### DIFF
--- a/bin/patron_activity_sync_notifications
+++ b/bin/patron_activity_sync_notifications
@@ -1,0 +1,12 @@
+#!/usr/bin/env python
+"""Send out notifications to Patrons that need their loan/hold activity synced"""
+import os
+import sys
+
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..")
+sys.path.append(os.path.abspath(package_dir))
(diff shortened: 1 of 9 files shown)


dbernstein (approved):
Noice work. Feel free to merge it.

PR #479, opened 2022-08-09
Feature/holds notifications

RishiDiwanTT:
## Description
Hold notifications for patrons
    
The script will only pick holds that have position=0
There is currently no way of telling if a notification has been sent to a user or not
So if a hold remains in reserve for a long duration the patron will receive multiple notifications
as and when the job runs

## Motivation and Context
Patrons need to be informed when a hold is available for loan. This can be achieved through a push notification to the patron device as and when a hold is ready
[Notion](https://www.notion.so/lyrasis/Push-Notifications-Hold-available-f727cc8473bb4007b4cd8f3756b28789)

## How Has This Been Tested?
Unit testing is done. Manual testing is not possible without apps with FCM installed,

## Checklist

### DIFF
--- a/core/util/notifications.py
+++ b/core/util/notifications.py
@@ -23,7 +23,7 @@ class PushNotifications:
     VALID_TOKEN_TYPES = [DeviceTokenTypes.FCM_ANDROID, DeviceTokenTypes.FCM_IOS]
 
     @classmethod
-    def notifiable_tokens(cls, patron: Patron):
+    def notifiable_tokens(cls, patron: Patron) -> List[DeviceToken]:
         return [
             token
             for token in patron.device_tokens
--- a/core/util/notifications.py
+++ b/core/util/notifications.py
@@ -113,6 +113,7 @@ class PushNotifications:
 
     @classmethod
     def send_holds_notifications(cls, holds: List[Hold]) -> List[str]:
+        """Send out notifcations to all patron devices that their hold is ready for checkout"""
         if not holds:
             return []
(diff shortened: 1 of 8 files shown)


core/util/notifications.py
```diff
@@ -21,6 +22,14 @@ class PushNotifications:
 
     VALID_TOKEN_TYPES = [DeviceTokenTypes.FCM_ANDROID, DeviceTokenTypes.FCM_IOS]
 
+    @classmethod
+    def notifiable_tokens(cls, patron: Patron):
+        return [
+            token
```
dbernstein, line 28: This syntax is slick - I'm not familiar with it.

core/util/notifications.py
```diff
@@ -21,6 +22,14 @@ class PushNotifications:
 
     VALID_TOKEN_TYPES = [DeviceTokenTypes.FCM_ANDROID, DeviceTokenTypes.FCM_IOS]
 
+    @classmethod
+    def notifiable_tokens(cls, patron: Patron):
```
dbernstein, line 26: Add return type in the method signature.

dbernstein (changes requested):
Nice work on this ticket @RishiDiwanTT :  just a minor cosmetic suggestion.

core/util/notifications.py
RishiDiwanTT: list comprehensions FTW!

dbernstein (approved):

PR #480, opened 2022-08-10
Custom list API fix

RishiDiwanTT:
## Description
Fixed erroneous boolean type value conversion for custom_list POST request
Form requests are string values

## Motivation and Context
The API requests would not work from actual clients because form encoded boolean values are strings

## How Has This Been Tested?
Unit tests have been updated, also manually tested the API

## Checklist

### DIFF
--- a/core/util/flask_util.py
+++ b/core/util/flask_util.py
@@ -175,7 +175,7 @@ class OPDSEntryResponse(Response):
         super().__init__(response, **kwargs)
 
 
-def boolean_value(value):
+def boolean_value(value: str) -> bool:
     """Convert a string request value into a boolean, used for form encoded requests
     JSON encoded requests will get automatically converted"""
-    return True if value in ["true", "True", True] else False
+    return True if value in ["true", "True", True, "1"] else False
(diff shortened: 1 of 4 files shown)


core/util/flask_util.py
```diff
@@ -173,3 +173,9 @@ class OPDSEntryResponse(Response):
     def __init__(self, response=None, **kwargs):
         kwargs.setdefault("mimetype", OPDSFeed.ENTRY_TYPE)
         super().__init__(response, **kwargs)
+
+
+def boolean_value(value):
```
dbernstein, line 178: How about specifying the return type and parameter types? 
ie
```
def boolean_value(value:str) -> bool: 
```

dbernstein (approved):
Looks good.  One minor suggestion.

core/util/flask_util.py
```diff
@@ -173,3 +173,9 @@ class OPDSEntryResponse(Response):
...
+    """Convert a string request value into a boolean, used for form encoded requests
+    JSON encoded requests will get automatically converted"""
+    return True if value in ["true", "True", True] else False
```
tdilauro, line 181: It would be good to have some additional tests for "truthy" values for this condition: especially non-empty strings like "false", but also non-zero number, etc.

tdilauro (changes requested):
This looks good, but I think it needs some additional tests.

tdilauro (approved):
Looks good! 🎸

PR #481, opened 2022-08-10
Feature/maintain patron last activity

RishiDiwanTT:
## Description
last_loan_activity_sync accessor no longer clears out the underlying DB property
    
It continues to return None on an expired value, but the underlying property is still
queryable in the database

## Motivation and Context
The patron activity sync job needs the underlying _last_loan_activity_sync property to remain intact within the DB
in order to query for it in the future, beyond the hard coded staleness check
So we maintain the contract of the accessor, but do not clear out the property

## How Has This Been Tested?
Unit testing has been done

## Checklist

### DIFF
--- a/core/model/patron.py
+++ b/core/model/patron.py
@@ -249,7 +249,8 @@ class Patron(Base):
         """
         return 15 * 60
 
-    def has_last_loan_activity_expired(self):
+    def is_last_loan_activity_stale(self) -> bool:
+        """Has the last_loan_activity_sync timestamp outlived the loan_activity_max_age seconds"""
         if not self._last_loan_activity_sync:
             return True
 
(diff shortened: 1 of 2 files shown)


core/model/patron.py
```diff
@@ -249,6 +249,14 @@ def loan_activity_max_age(self):
         """
         return 15 * 60
 
+    def has_last_loan_activity_expired(self):
```
tdilauro, line 252: I think this might be more appropriately named `is_last_loan_activity_stale`, given that it might never have been set. Feel free to come up with a better name here.

tdilauro, line 252: Would be good to provide the return type for this method.

tdilauro (commented):
Generally looks good. Just a couple of minor comments.

tdilauro (approved):
Good to go! 👟

PR #484, opened 2022-08-12
Added auto_update_facets form parameter for /custom_list edit API

RishiDiwanTT:
## Description
Accepts a JSON of attributes that would generally go into a Facet object
Eg.
- order
- language
- media
- author

Added open API documentation for the modified endpoint which can be rendered using the index.html
image

## Motivation and Context
The auto update feature for custom lists was missing the facets which are a part of the search criteria. 
This is an important set of attributes which affect the results of the search and hence what may or may not be part of a list.
[Notion](https://www.notion.so/lyrasis/Enable-list-sharing-for-advanced-search-lists-35694eda31bd43f8a4b0b498dd8c2094)

## How Has This Been Tested?
API and facets have been unit tested

## Checklist

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -6,6 +6,7 @@ import os
 import sys
 import urllib.parse
 from datetime import date, datetime, timedelta
+from http.client import BAD_REQUEST
 from typing import Optional, Union
 
 import flask
@@ -796,6 +797,7 @@ class CustomListsController(AdminCirculationManagerController):
                         entry_count=list.size,
                         auto_update=list.auto_update_enabled,
                         auto_update_query=list.auto_update_query,
+                        auto_update_facets=list.auto_update_facets,
                     )
                 )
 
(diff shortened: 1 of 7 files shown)


api/admin/controller/__init__.py
```diff
@@ -796,6 +797,7 @@ def custom_lists(self):
                         entry_count=list.size,
                         auto_update=list.auto_update_enabled,
                         auto_update_query=list.auto_update_query,
+                        auto_update_facets=list.auto_update_facets,
```
dbernstein, line 800: Lines 782 to 800  and  804 to 824 are virtually identical.  It seems that these could be refactored in a single method to reduce code duplication.

dbernstein (changes requested):
Looks good.  One refactoring suggestion.

api/admin/controller/__init__.py
RishiDiwanTT: I was having similar anxiety while making this change. Will fix.

dbernstein:
Ready to merge.

dbernstein (approved):

PR #488, opened 2022-08-17
Feature/additional search keys

RishiDiwanTT:
## Description
Added published and data_source as searchable keys in the json format
    
Both keys now transform their values before exceuting the search
A data cache has been introduced that is persistent across the application process
It has the ability to have its _db session refreshed as required and
is written to be thread-safe for _db refreshes while data is being accessed through locks

Thread safety and session safety was tested by running the app on a `uwsgi` server with 10 threads, and then hitting the search endpoint with `ab -n100 -c20`. No broken sessions or lock conflicts were detected on multiple runs with cache ttl set to 1 second.

## Motivation and Context
The publish date and distributor are not part of the JSON search API keys, they needed to be added to the search criteria.
[Notion](https://www.notion.so/lyrasis/Add-published-date-and-datasource-name-as-search-keys-e7c408d429d04f37a41e083cb25e9c8c)

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -2034,7 +2034,8 @@ class JSONQuery(Query):
 
     class ValueTransforms:
         @staticmethod
-        def data_source(value):
+        def data_source(value: str) -> int:
+            """Transform a datasource name into a datasource id"""
             sources = CachedData.cache.data_sources()
             for source in sources:
                 if source.name == value:
@@ -2044,8 +2045,8 @@ class JSONQuery(Query):
             return 0
 
         @staticmethod
-        def published(value):
-            """Expects a YYYY-MM-DD format string"""
+        def published(value: str) -> int:
+            """Expects a YYYY-MM-DD format string and returns a timestamp from epoch"""
             try:
                 values = value.split("-")
                 return datetime.datetime(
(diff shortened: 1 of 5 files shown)


core/external_search.py
```diff
@@ -2020,6 +2029,36 @@ class Operators(Values):
...
+            sources = CachedData.cache.data_sources()
+            for source in sources:
+                if source.name == value:
+                    return source.id
+
+            # No such value was found, so return a non-id
+            return 0
+
+        @staticmethod
+        def published(value):
```
dbernstein, line 2047: please specify the return type in the method signature.

core/external_search.py
```diff
@@ -2020,6 +2029,36 @@ class Operators(Values):
         "open_access": "licensepools.open_access",
         "available": "licensepools.available",
         "classification": "classifications.term",
+        "data_source": "licensepools.data_source_id",
+    }
+
+    class ValueTransforms:
+        @staticmethod
```
dbernstein, line 2036: Please document the expected data type and format if applicable for value parameter and specify the return types in the signature.

dbernstein (changes requested):
Looks good overall - just a couple of minor suggestions.

dbernstein (approved):

PR #491, opened 2022-08-18
Share lists to all libraries

RishiDiwanTT:
## Description
Updated the list sharing functionality to share with all libraries on the CM
Rather than a point 2 point sharing method
Also removed the unrequired "share to libraries with collection" functionality
Added an is_owner boolean to the customlists GET endpoint so that the UI may segregate the lists

## Motivation and Context
Most users on a CM can’t see other libraries on the same CM. Sharing a list should share it with all libraries on the same CM
[Notion](https://www.notion.so/lyrasis/Sharing-a-list-automatically-shares-it-with-all-libraries-on-the-same-CM-0a293d3d2aac480d9b24b31e85603819)

## How Has This Been Tested?
All unit tests have been run and updated

## Checklist

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -7,7 +7,7 @@ import sys
 import urllib.parse
 from datetime import date, datetime, timedelta
 from http.client import BAD_REQUEST
-from typing import Dict, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Union
 
 import flask
 import jwt
@@ -773,7 +773,7 @@ class FeedController(AdminCirculationManagerController):
 
 
 class CustomListsController(AdminCirculationManagerController):
-    def _list_as_json(self, list: CustomList, is_owner=True):
+    def _list_as_json(self, list: CustomList, is_owner=True) -> Dict:
         """Transform a CustomList object into a response ready dict"""
         collections = []
         for collection in list.collections:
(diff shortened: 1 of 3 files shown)


api/admin/controller/__init__.py
```diff
@@ -773,7 +773,7 @@ def genres(self):
 
 
 class CustomListsController(AdminCirculationManagerController):
-    def _list_as_json(self, list: CustomList):
+    def _list_as_json(self, list: CustomList, is_owner=True):
```
dbernstein, line 776: Please add return type in method signature.

api/admin/controller/__init__.py
```diff
@@ -1061,48 +1062,35 @@ def custom_list(self, list_id):
...
-            return Response(200)
-        else:
-            self._db.rollback()
-            return response
-
-    def share_locally_with_library_collection(self, customlist_id):
-        """Share with all libraries having a collection"""
-        collection_id = flask.request.form.get("collection")
-        if not collection_id or not customlist_id:
+        """Share this customlist with all libraries on this local CM"""
```
dbernstein, line 1065: Please, add the param type for customlist_id in signature.

dbernstein (changes requested):
As usual, excellent work.  Just a few cosmetic touch-ups.

dbernstein (approved):

PR #493, opened 2022-08-19
Shared custom lists cannot be deleted

RishiDiwanTT:
## Description
The API protects against deleting shared lists with a `409` response.

## Motivation and Context
For the list sharing MVP, we do not want shared lists to be able to be deleted. This will prevent lists from being accidentally removed. In the future we will pursue a more robust solution.
[Notion](https://www.notion.so/lyrasis/Prevent-shared-lists-from-being-deleted-6b25a4f1e1254b57937f7cd7b116c2cd)

## How Has This Been Tested?
Unit tests have been run and updated.

## Checklist


tdilauro (approved):
Looks good! 🎺

PR #494, opened 2022-08-19
Axis360FulfillmentInfo fix

RishiDiwanTT:
## Description
Axis360's variant of FulfillmentInfo did not have the content_link_redirect attribute

Added with tests for the same workflow

## Motivation and Context
QA systems were down because of this

## How Has This Been Tested?
A Unit tests was written to cover this flow

## Checklist

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -1725,8 +1725,6 @@ class Axis360FulfillmentInfo(APIAwareFulfillmentInfo):
         self._content = str(manifest)
         self._content_type = manifest.MEDIA_TYPE
         self._content_expires = expires
-        # Never used in this object, required for consistency
-        self.content_link_redirect = False
 
 
 class Axis360AcsFulfillmentInfo(FulfillmentInfo):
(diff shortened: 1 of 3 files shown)


api/axis.py
```diff
@@ -1725,6 +1725,8 @@ def do_fetch(self):
         self._content = str(manifest)
         self._content_type = manifest.MEDIA_TYPE
         self._content_expires = expires
+        # Never used in this object, required for consistency
+        self.content_link_redirect = False
```
tdilauro, line 1729: It seems like this should be handled in `APIAwareFulfillmentInfo.__init__`, so this feature has consistent behavior for new integrations. I noticed that `APIAwareFulfillmentInfo.__init__` doesn't make a call to `super().__init__`, which seems problematic and might be why this property was not present as expected.

RishiDiwanTT: This is a fair observation, although Axis360FulfillmentInfo is the only one using this right now, this will make it more future proof

api/circulation.py
```diff
@@ -288,6 +288,8 @@ def __init__(self, api, data_source_name, identifier_type, identifier, key):
         self._content_type = None
         self._content = None
         self._content_expires = None
+        # Never used in this object, required for consistency
```
tdilauro, line 291: I don't think this comment is needed any longer.

tdilauro (approved):
Looks good!

PR #496, opened 2022-08-22
Makes code compatible with elastic 7.x

dbernstein:
## Description

This update adds a flag to two key elastic search calls that make it compatible with ElasticSearch 7.x.  
## Motivation and Context

There are no official docker images for users of OSX on M1 chips (ARM).  The ARM builds started after 7.x  Instead of adding a Dockerfile for ARM that we have to maintain, I've changed two lines of code which enables the circulation manager and scripts to work with ElasticSearch 7.x.  So we're effectively killing two birds with one stone:  smoothing the path for a future elastic search upgrade and allowing M1 users  to use docker to bring up ES. 

## How Has This Been Tested?
All tests pass, so we know it this change didn't break anything. 
I manually tested using ES 7.17.5

```
docker run -p 127.0.0.1:9200:9200 -d --name es-7.17.5  -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.17.5
```

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -295,9 +295,9 @@ class ExternalSearchIndex(HasSelfTests):
     def _update_index_mapping(self, dry_run=False) -> Dict:
         """Updates the index mapping with any NEW properties added"""
 
-        current_mapping: Dict = self.indices.get_mapping(self.works_index)[
-            self.works_index
-        ]["mappings"][self.work_document_type]["properties"]
+        current_mapping: Dict = self.indices.get_mapping(
+            self.works_index, include_type_name=True
+        )[self.works_index]["mappings"][self.work_document_type]["properties"]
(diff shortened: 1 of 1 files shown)


tdilauro (changes requested):
Marking this with `Request changes` to prevent merging, so that I can manually test this branch on AWS to ensure that there are no issues.

jonathangreen:
Haven't gotten a chance to test this in AWS, but taking a quick look at the PR, it would be nice to be testing against ES7 (as well as ES6) in CI since we're adding ES7 support here.

jonathangreen:
I'm going to close this one, as we're close to merging https://github.com/ThePalaceProject/circulation/pull/411 and that will allow running tests and development on a M1 machine.

PR #498, opened 2022-08-24
Feature/auto update with new lists

RishiDiwanTT:
## Description
Custom list create API also accepts auto_update parameters

## Motivation and Context
Initially the auto_update parameters were only accepted through the 'List edit' APIs
[Notion](https://www.notion.so/lyrasis/Can-t-create-a-new-list-with-auto_update-true-0ccfcbc0c35b4356962fc2268194945d)

## How Has This Been Tested?
Unit tests have been updated and manual testing was also done

## Checklist


tdilauro (approved):
Looks good! 🥇

One suggestion for a return type hint, but otherwise all good.

PR #502, opened 2022-08-26
Defensive code added for dashboard stats

RishiDiwanTT:
## Description
In the case of a parent collection that is not part of a library that a child is also part of
The dashboard stats API would error out otherwise

## Motivation and Context
Rochester Library was having this issue in their Dashboard page
[Notion](https://www.notion.so/lyrasis/Rochester-Public-Library-continually-getting-error-in-Collection-Manager-a196724ca9d04423bccb402f867c4a3d)

## How Has This Been Tested?
This has been manually tested with the VT collections
A unit test has been added for this scenario

## Checklist

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1465,6 +1465,11 @@ class DashboardController(AdminCirculationManagerController):
 
             library_collection_counts = dict()
             for collection in library.all_collections:
+                # sometimes a parent collection may be dissociated from a library
+                # in this case we may not have access to the collection as a library staff member
+                if collection.name not in collection_counts:
+                    continue
+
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
I can't see anything wrong with this PR.  However I am curious though about parent and child collections: How do those relationships get established through the UI?  I don't see anything about it in the CM System Configuration.

RishiDiwanTT:
> I can't see anything wrong with this PR. However I am curious though about parent and child collections: How do those relationships get established through the UI? I don't see anything about it in the CM System Configuration.

@dbernstein If you select "Overdrive" while creating a collection, a "parent" option appears. It's only valid for Overdrive sources.

PR #503, opened 2022-08-29
All outgoing requests will now have a default User Agent

RishiDiwanTT:
## Description
Currently set to the minimalistic Palace Manager/1.x.x

## Motivation and Context
Overdrive have requested the CM have a custom User Agent while making requests to their endpoints.
[Notion](https://www.notion.so/lyrasis/Make-user-agent-less-generic-4820ea007b3c4e1988c569c3988f08c5)

## How Has This Been Tested?
Unit tests have been updated.
Used a mock server to inspect the user-agent manually.

## Checklist

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -185,6 +185,8 @@ class Configuration(ConfigurationConstants):
 
     EXCLUDED_AUDIO_DATA_SOURCES = "excluded_audio_data_sources"
 
+    DEFAULT_USER_AGENT_VERSION = "1.x.x"
+
     SITEWIDE_SETTINGS = [
         {
             "key": BASE_URL_KEY,
--- a/core/config.py
+++ b/core/config.py
@@ -9,14 +9,14 @@ from flask_babel import lazy_gettext as _
 from sqlalchemy.engine.url import make_url
 from sqlalchemy.exc import ArgumentError
 
+# It's convenient for other modules import IntegrationException
+# from this module, alongside CannotLoadConfiguration.
+from core.exceptions import IntegrationException
+
 from .entrypoint import EntryPoint
 from .facets import FacetConstants
 from .util import LanguageCodes
 from .util.datetime_helpers import to_utc, utc_now
-
-# It's convenient for other modules import IntegrationException
-# from this module, alongside CannotLoadConfiguration.
-from .util.http import IntegrationException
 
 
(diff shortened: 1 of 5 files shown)


core/util/http.py
```diff
@@ -252,16 +252,19 @@ def _request_with_timeout(
...
-                    k = k.encode("utf8")
-                if isinstance(v, str):
-                    v = v.encode("utf8")
-                new_headers[k] = v
-            kwargs["headers"] = new_headers
+
+        headers = kwargs.get("headers", {})
+        # Set a user-agent if not already present
+        if "User-Agent" not in headers:
+            headers["User-Agent"] = "Palace Manager/1.x.x"
```
dbernstein, line 259: How much work would it be to pull in the version (and potentially revision)?  Not sure if this is out of scope.

RishiDiwanTT: We currently don't have the concept of numbered versions (like semver), we only use a commit hash. Hence we are using this stopgap example for now.

jonathangreen: This should perhaps reference the version number that gets set in the configuration so it returns the same info as the heartbeat endpoint. 

https://github.com/ThePalaceProject/circulation/blob/dd910452b4e7883f580df1854a38977a9fc767ef/core/config.py#L522

tdilauro: @jonathangreen It was my suggestion to include a generic version string until we could discuss any security implications. But, if you're good with it, I have no objection.

Also, good to see you here!

RishiDiwanTT: Updated to use app_version.

tdilauro (commented):
Requesting a small change, though it might require a change/addition to tests.

core/config.py
```diff
@@ -185,6 +185,8 @@ class Configuration(ConfigurationConstants):
 
     EXCLUDED_AUDIO_DATA_SOURCES = "excluded_audio_data_sources"
 
+    # In case an app version is not present, we can use this version as a fallback
+    # for all outgoing http requests with a custom user-agent
```
tdilauro, line 189: Did you mean for all requests withOUT a custom ua here?

RishiDiwanTT: Meant "default custom" UA, but it didn't come out right. "Without" is clearer.

tdilauro (approved):
LGTM! 🎺

PR #504, opened 2022-08-29
Fixed CollectionCrawlableWorklists without a library

RishiDiwanTT:
## Description
The collection_ids were missing due to no library being initialized
now we set the collection_ids directly when creating the SpecificWorklist if the library is missing

## Motivation and Context
`/collections/<name>/crawlable` endpoint was broken due to the above issue. This functionality is required if we want to have OPDS2 feeds per collection
[Notion](https://www.notion.so/lyrasis/Fix-collections-name-crawlable-endpoint-3abd8d94d2fc44f1b21efb0167a7ba77)

## How Has This Been Tested?
Unit tests have been updated
Manually queried the `/collections/<name>/crawlable` enpoint

## Checklist

### DIFF
--- a/core/lane.py
+++ b/core/lane.py
@@ -1892,6 +1892,10 @@ class WorkList:
         # performance isn't a big concern -- it's just ugly.
         wl = SpecificWorkList(work_ids)
         wl.initialize(self.get_library(_db))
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
Very nice work.

PR #506, opened 2022-08-31
JSON type search requests now ignore the scoring parameters

RishiDiwanTT:
## Description
When ordering was required, the search filter would add a min_score=500
Since we do not use any dis_max queries or boost parameters we do not have
any scoring for the JSON type queries

## Motivation and Context
A bug was discovered regarding json search and ordering
[Notion](https://www.notion.so/lyrasis/Title-and-author-sorts-for-list-search-don-t-return-results-6bfad0ab4e5c4705874f312ba13d2897)

## How Has This Been Tested?
A unit test to cover this situation has been introduced
Manual testing on the `/search` endpoint was done

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -2619,6 +2619,10 @@ class Filter(SearchBase):
             self.scoring_functions = []
             self.search_type = "default"
 
+        # JSON type searches are exact matches and do not have scoring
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good.

PR #507, opened 2022-09-01
Global announcement admin should get all announcements

tdilauro:
## Description

Changes global announcements admin `get` API to return all global announcements, rather than just the active ones.

## Motivation and Context

The sysadmin needs to be able to manage all announces, including those that have expired and those that are upcoming.

## How Has This Been Tested?

- Updated existing test to ensure that all global announcements are returned to the admin. Rant test. Test failed.
- Fixed response. Ran test. Test passed.

## Checklist

### DIFF
--- a/api/admin/controller/announcement_service.py
+++ b/api/admin/controller/announcement_service.py
@@ -36,7 +36,7 @@ class AnnouncementSettings(SettingsController):
         settings = Configuration.ANNOUNCEMENT_SETTINGS
         return dict(
             settings=settings,
-            announcements=[ann.json_ready for ann in announcements.active],
+            announcements=[ann.json_ready for ann in announcements.announcements],
         )
 
     def post(self) -> dict | ProblemDetail:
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
Looks good, @tdilauro .

dbernstein:
I'm rerunning the checks.  There may be a problem with the tests, FYI.

dbernstein:
Actually on closer examination it looks like there's a problem with one of your tests: 

  ______________________ TestAnnouncementService.test_post _______________________
  
  self = <tests.api.admin.controller.test_announcement_service.TestAnnouncementService object at 0x7f853d8b0610>
  
      def test_post(self):
          with self.request_context_with_admin("/", method="POST") as ctx:
              ctx.request.form = MultiDict([("announcements", json.dumps([self.active]))])
              response = AnnouncementSettings(self.manager).process_many()
      
          announcements = Announcements.for_all(self._db)
  >       assert len(announcements.announcements) == 2
  E       assert 1 == 2
  E        +  where 1 = len([<api.announcements.Announcement object at 0x7f8523ab3bb0>])
  E        +    where [<api.announcements.Announcement object at 0x7f8523ab3bb0>] = <api.announcements.Announcements object at 0x7f8523a8b2b0>.announcements
  
  tests/api/admin/controller/test_announcement_service.py:38: AssertionError

tdilauro:
> Actually on closer examination it looks like there's a problem with one of your tests:

Thanks! It was an inadvertent change. I meant to fix checks only in the `test_get` test, since the `get` method was the only place I saw the problem.

PR #509, opened 2022-09-02
Fixes problem of duplicate events appearing in the local analytics ex…

dbernstein:
## Description

Fixes problem of duplicate events appearing in the local analytics export.

This PR fixes a problem introduced when we added the library and collection to the local analytics export.  The problem is that I didn't specify explicitly how to join to the Collection table. As a result,  the join was occurring by default on the library  resulting in a row for each collection in the library, rather than the collection associated with the license pool (which has a 1 to 1 relationship with a collection.

## Motivation and Context
 https://www.notion.so/lyrasis/CSV-export-includes-duplicate-events-for-every-collection-on-the-CM-513bb0974cce46e399cd610c0cd7e9d9

## How Has This Been Tested?
I updated the test to use two collections rather than a single collection to ensure that the resulting row counts are correct.

I also manually checked it.
## Checklist


tdilauro (approved):
LGTM! 🎸

PR #512, opened 2022-09-09
Added an is_shared to the /custom_lists admin route

RishiDiwanTT:
## Description
Will be True in case a list is shared with any other library

## Motivation and Context
The UI needs to be aware if a list is shared to any other library to make appropriate changes in the view
[Notion](https://www.notion.so/lyrasis/Need-a-way-to-tell-if-a-custom-list-is-shared-when-you-are-the-owner-a8101f9aa97e47c381a99346a0318842)

## How Has This Been Tested?
Unit test has been updated
Manually tested with the admin panel "list editing" page

## Checklist


dbernstein (approved):
👍 💯

PR #516, opened 2022-09-15
Add mypy to more files in core

jonathangreen:
## Motivation and Context

Expand the mypy coverage in core.

## Checklist

### DIFF
--- a/core/opds2_import.py
+++ b/core/opds2_import.py
@@ -6,12 +6,17 @@ from typing import Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Un
 from urllib.parse import urljoin, urlparse
 
 import sqlalchemy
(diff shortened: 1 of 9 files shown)


core/opds2_import.py
```diff
@@ -883,7 +893,7 @@ def _record_publication_unrecognizable_identifier(
                 f"Publication # {original_identifier} ('{title}') has an unrecognizable identifier."
             )
 
-    def extract_next_links(self, feed: Union[str, opds2_ast.OPDS2Feed]) -> List[str]:
+    def extract_next_links(self, feed) -> List[str]:
```
dbernstein, line 896: Shouldn't the type of `feed` be added to the method signature?

dbernstein (changes requested):
Just one minor question.

PR #518, opened 2022-09-15
Fail-safe datetime conversions for Work.to_search_document()

RishiDiwanTT:
## Description
Timezone naive datetimes and dates are now set to UTC before converting to a timestamp

This was causing a problem in the work.to_search_document method when some works would appear with dates like 0001-01-01 Setting the timezone before converting the date would keep the timestamp method from trying to rewind time into year 0 (which is invalid in python)

Although setting the timezone averts this problem, exception handling has been added in as a fail-safe

## Motivation and Context
Some books have `published` values such as `0001-01-01` which would fail in case of non-UTC or times with a "fold" value.
[Notion](https://www.notion.so/lyrasis/Search-index-coverage-fails-on-invalid-date-785021498d2247de8c1e8ad6d2a5e61b)

## How Has This Been Tested?
Manually changed times in the database and ran the indexing scripts
Unit tests have been updated to avoid this error

## Checklist

### DIFF
--- a/core/model/work.py
+++ b/core/model/work.py
@@ -8,6 +8,7 @@ from datetime import date, datetime
 from decimal import Decimal
 from typing import TYPE_CHECKING, Any, TypeVar, cast
 
+import pytz
 from sqlalchemy import (
     Boolean,
     Column,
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
💯

PR #520, opened 2022-09-16
Support Non-nested Indirect Acquision Links in OPDS2 importer

jonathangreen:
## Description

The OPDS2 importer used to only support Indirect Acquisition links in the form:
```json
{
  "links": [
    {
      "type": "application/opds-publication+json",
      "rel": "http://opds-spec.org/acquisition/borrow",
      "href": "http://example.org/huckleberry-finn",
      "properties": {
        "availability": {
          "state": "available"
        },
        "indirectAcquisition": [
          {
            "type": "application/vnd.adobe.adept+xml",
            "child": [
              {
                "type": "application/epub+zip"
              }
            ]
          },
          {
            "type": "application/vnd.readium.lcp.license.v1.0+json",
            "child": [
              {
                "type": "application/epub+zip"
              }
            ]
          }
        ]
      }
    }
  ]
}
```
Where the type of both the content and the DRM are nested into `indirectAcquisition` properties. This seems like the least common case we will encounter, a more common case is the one we are trying to support here.

```json
{
  "links": [
    {
      "type": "application/vnd.adobe.adept+xml",
      "rel": "http://opds-spec.org/acquisition",
      "href": "http://foo",
      "templated": true,
      "properties": {
        "indirectAcquisition": [
          {
            "type": "application/epub+zip"
          }
        ]
      }
    },
    {
      "type": "application/vnd.adobe.adept+xml",
      "rel": "http://opds-spec.org/acquisition",
      "href": "http://bar",
      "templated": true,
      "properties": {
        "indirectAcquisition": [
          {
            "type": "application/pdf"
          }
        ]
      }
    }
  ]
}
```

## Motivation and Context

This change was prompted by testing with the Proquest OPDS2 feed, that uses `indirectAcquisition` properties. This feed was failing to properly import with the desired `DeliveryMechanism` because the importer was looking for DRM within nested `indirectAcquisition` properties.

## How Has This Been Tested?

The only testing here has been running unit tests on this code. I added a sample of the Proquest feed to the importer test, and added tests to make sure the delivery mechanisms are set correctly. The existing tests cover the old case with nested acquisition methods.

## Checklist


core/configuration/ignored_identifier.py
```diff
@@ -66,7 +66,7 @@ def set_ignored_identifier_types(
             if isinstance(item, str):
                 ignored_identifier_types.append(item)
             elif isinstance(item, IdentifierType):
-                ignored_identifier_types.append(item.name)
+                ignored_identifier_types.append(item.value)
```
jonathangreen, line 69: This change seems unrelated, but it was necessary when adding Proquest identifiers to the `test_opds2_importer_skips_publications_with_unsupported_identifier_types` test. Without this change, Proquest identifiers are not ignored correctly.

This function `set_ignored_identifier_types` is only used in tests in `test_odl2.py` and `test_opds2_import.py`, so the tests passing should suffice to validate the change. 

Because `item.name` and `item.value` are the same for `IdentifierType.ISBN` and `IdentifierType.URI` this wasn't noticed before, but for `IdentifierType.PROQUEST_ID` they are different, triggering this issue.

dbernstein (approved):
Looks good to me.

core/opds2_import.py
```diff
@@ -488,29 +488,40 @@ def _extract_media_types_and_drm_scheme_from_link(
...
+        ):
+            self._logger.info(f"Link unavailable. Skipping. {encode(link)}")
+            return []
 
-                    media_types_and_drm_scheme.append(
-                        (nested_acquisition_object.type, drm_scheme)
-                    )
+        # We need to take into account indirect acquisition links
+        if link.properties and link.properties.indirect_acquisition:
+            # We make the assumption that when we have nested indirect acquisition links
```
tdilauro, line 502: I think it would be useful to mention here that we don't *just* assume this. We verify that we indeed have that pairing for indirect links and, if not, we ignore the link entirely.

tests/core/test_opds2_import.py
```diff
@@ -283,18 +287,53 @@ def test_opds2_importer_correctly_imports_valid_opds2_feed(
             == huckleberry_finn_delivery_mechanisms[1].delivery_mechanism.content_type
         )
 
+        # 2.3 Edition with non open-access acquisition links (The Politics of Postmodernism)
+        postmodernism_license_pool = self._get_license_pool_by_identifier(
+            pools, self.PROQUEST_IDENTIFIER
+        )
+        assert True == isinstance(postmodernism_license_pool, LicensePool)
+        assert False == huckleberry_finn_license_pool.open_access
```
tdilauro, line 295: Did you mean `postmodernism_license_pool.open_access` here?

tests/core/test_opds2_import.py
```diff
@@ -80,6 +81,9 @@ def _get_work_by_identifier(works, identifier):
 class TestOPDS2Importer(OPDS2Test):
     MOBY_DICK_IDENTIFIER = "urn:isbn:978-3-16-148410-0"
     HUCKLEBERRY_FINN_IDENTIFIER = "http://example.org/huckleberry-finn"
+    PROQUEST_IDENTIFIER = (
```
tdilauro, line 84: I know this is not a product of the current work, but there is a weird asymmetry here: `<title>_IDENTIFIER`, `<title(ish)>_IDENTIFIER`, `<identifier-type>_IDENTIFIER`. Because they're used both in the context of the titles and as examples of identifier types, it might be helpful to make that more clear through their names. Maybe: `MOBY_DICK_ISBN_IDENTIFIER`, `HUCK_FINN_URI_IDENTIFIER`, and `POSTMODERNISM_PROQUEST_IDENTIFIER`.

tests/core/test_opds2_import.py
```diff
@@ -310,14 +349,27 @@ def test_opds2_importer_correctly_imports_valid_opds2_feed(
...
+            (
+                "URI",
+                [IdentifierType.ISBN, IdentifierType.PROQUEST_ID],
+                HUCKLEBERRY_FINN_IDENTIFIER,
+            ),
+            (
+                "ProQuest Doc ID",
+                [IdentifierType.ISBN, IdentifierType.URI],
+                PROQUEST_IDENTIFIER,
+            ),
```
tdilauro, line 366: See comment at ll. 81-84 re: identifier variable names.

tdilauro (commented):
Mostly looks good.

A possible paste-o and a couple of suggestions for variable name and comment changes.

core/opds2_import.py
jonathangreen: Added some more verbiage to this comment. Let me know if this addresses your comment, or if you have other suggestions for a comment here.

tests/core/test_opds2_import.py
jonathangreen: 👍🏻 fixed

tests/core/test_opds2_import.py
jonathangreen: Good call. Updated the names for these identifiers.

core/opds2_import.py
```diff
@@ -514,14 +515,17 @@ def _extract_media_types_and_drm_scheme_from_link(
                     nested_acquisition = first_or_default(nested_acquisition.child)
                 [drm_type, media_type] = nested_types[-2:]
 
+                # We then check this returned pair of content types to make sure they match known
+                # book or audiobook and DRM types. If they do not match known types, then we skip
+                # the item.
```
tdilauro, line 520: Are we skipping the `item` here or just this link? The link might be what you mean here, but `item` might be misconstrued to mean book, rather than link.

core/opds2_import.py
```diff
@@ -514,14 +515,17 @@ def _extract_media_types_and_drm_scheme_from_link(
...
                 if (
                     media_type in MediaTypes.BOOK_MEDIA_TYPES
                     or media_type in MediaTypes.AUDIOBOOK_MEDIA_TYPES
                 ) and drm_type in DeliveryMechanism.KNOWN_DRM_TYPES:
                     media_types_and_drm_scheme.append((media_type, drm_type))
 
         # There are no indirect links, then the link type points to the media, and
-        # there is no DRM on the item
+        # there is no DRM on the item.
```
tdilauro, line 528: Same concern about the word `item` here.

tdilauro (commented):
Looks good! My only remaining concern is that the word `item` in a couple of comments is referring to a specific acquisition link for a book, but might be misconstrued as referring to a book.

core/opds2_import.py
jonathangreen: Updated both these comments to refer to link instead of item.

tdilauro (approved):
Looks good to me! ⚡

PR #539, opened 2022-09-20
Escape special characters in Docker environment variable values.

tdilauro:
## Description

Changes Docker setup to escape special characters in the environment variables copied into app setup script in the scripts container.

## Motivation and Context

The variables are exported from this script with Bash, so are interpreted by the shell. Failure to escape certain special characters in these values caused errors when attempting to export the settings while activating the virtual environment. 

In addition, the way the variables were copied into the setup file could have allowed variable interpolation and created security vulnerabilities.

## How Has This Been Tested?

The changes were applied to fix a running environment that was affected by the error. More testing will be conducted after a container is built.

## Checklist


jonathangreen (approved):
This looks good to me. Did a bit of manual testing, and it looks like its correctly escaping in my env.

PR #540, opened 2022-09-20
Update poetry to 1.2.1

jonathangreen:
## Description

Update the version of Poetry we are using in Docker and CI to 1.2.1. 

Homebrew on my system updated Poetry to 1.2.1 today, then when I committed a update to a poetry lock file all of a sudden setuptools was being uninstalled by poetry. Apparently lock files generated by poetry 1.2.1 are not compatible with 1.1.x and will cause setuptools to be uninstalled 

https://github.com/python-poetry/poetry/issues/4242

## Motivation and Context

It seems reasonable to update the version of poetry we use, since 1.2 can read lock files from 1.1, but 1.1 can't correctly read lock files generated by 1.2.

## How Has This Been Tested?

Compared the packages we installed with Poetry 1.1 vs Poetry 1.2 by hand. CI running and building a docker container should show that this is working.

## Checklist

### DIFF
--- a/.github/actions/poetry/action.yml
+++ b/.github/actions/poetry/action.yml
@@ -5,7 +5,7 @@ inputs:
   version:
     description: "Version of poetry to install"
     required: false
-    default: "1.1.12"
+    default: "1.2.1"
 
 outputs:
   version:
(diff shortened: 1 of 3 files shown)


jonathangreen:
Convert this to a draft since tests are failing. Will have to debug whats going on here.

PR #541, opened 2022-09-21
Aligns the item count with the number of items in the list when works…

dbernstein:
… in list are no longer available.

## Description

This update ensures that the count of all items in the list uses the same query that is used to generate the list. 
## Motivation and Context
https://www.notion.so/lyrasis/Titles-missing-from-lists-lanes-in-CM-and-lanes-in-app-088f95698cdd4935898c01f0fc01f0b5

## How Has This Been Tested?
Existing tests were updated to account for changes.  
It should be manually tested to ensure that works that have been disassociated from the list are no longer being counted.

## Checklist

### DIFF
--- a/core/model/customlist.py
+++ b/core/model/customlist.py
@@ -113,7 +113,7 @@ class CustomList(Base):
         )
 
     @classmethod
-    def list_entries(cls, _db: Session, list_id: int):
+    def entries_having_works(cls, _db: Session, list_id: int):
         return (
             _db.query(Work)
             .join(Work.custom_list_entries)
@@ -327,7 +327,7 @@ class CustomList(Base):
         return qu
 
     def update_size(self, db: Session):
-        self.size = CustomList.list_entries(db, self.id).count()
+        self.size = CustomList.entries_having_works(db, self.id).count()
 
 
 customlist_sharedlibrary = Table(
--- a/core/model/customlist.py
+++ b/core/model/customlist.py
@@ -113,6 +113,15 @@ class CustomList(Base):
         )
 
     @classmethod
+    def list_entries(cls, _db: Session, list_id: int):
+        return (
+            _db.query(Work)
+            .join(Work.custom_list_entries)
+            .filter(CustomListEntry.list_id == list_id)
+            .order_by(Work.id)
+        )
+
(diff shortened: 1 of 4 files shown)


pcustic (commented):
This looks good to me. Tim made some comments in the ticket that we should think how to communicate that some books are no longer available - I'm not sure was that supposed to be in this PR or is that something for the future?

dbernstein:
Thanks @pcustic :  @tdilauro  can correct me if I'm wrong, but my sense was that we wanted a quick fix right now and then continue to think about how to communicate that some books are no longer available after the immediate issue is addressed.

core/model/customlist.py
```diff
@@ -317,8 +326,8 @@ def entries_for_work(self, work_or_edition):
         )
         return qu
 
-    def update_size(self):
-        self.size = len(self.entries)
+    def update_size(self, db: Session):
+        self.size = CustomList.list_entries(db, self.id).count()
```
tdilauro, line 330: Will this cause an already run query to be re-run unnecessarily?

tdilauro (commented):
Generally, I like the re-org here. Just one question...

core/model/customlist.py
dbernstein: I don't think so, but I'm not 100% sure.   I'm not sure if the previous approach was causing a query to run behind the scenes ie 
```
self.size = len(self.entries)
```
(I'm not familiar with the internal workings of sqlalchemy). 

That said, this method code is only called once in the entire codebase (tests excluded) and it is called in a cron job script it should have no impact whatsoever on the user experience.

core/model/customlist.py
```diff
@@ -112,6 +112,15 @@ def __lt__(self, other):
             other.name,
         )
 
+    @classmethod
+    def list_entries(cls, _db: Session, list_id: int):
```
tdilauro, line 116: Requesting one minor change to avoid confusion: can we rename this to `entries_having_works` or something along those lines to hint that this is possibly different from `self.entries`, which contains all entries in the list, regardless of whether or not they have works.

tdilauro (approved):
Looks good! 🎸

PR #542, opened 2022-09-22
Configurable setting for external search versions

RishiDiwanTT:
## Description
The versions supported are
- Elasticsearch 6.8.6
- Opensearch 1.2

The unit tests for ExternalSearchIndex now run in 2 different modes
depending on the environment available
Both search indexes can also be tested if both environments are available
The environment availability depends on the availability of the env variables
- SIMPLIFIED_TEST_ELASTICSEARCH
- SIMPLIFIED_TEST_OPENSEARCH

A pytest hook has been written that parameterizes any test class that has
been marked with a DUAL_SEARCH_TESTS variable
Each of the tests in those classes will be run with both search indexes
if both are available
By default the tests assume only elasticsearch is available

The main difference between both the versions is the removal of the document-type
from the index APIs, this is because OS1.2 (and ES7+) have removed multiple documents per index
as an option

Pending
- README updates
- Tox updates

## Motivation and Context
While Opensearch is the direction the CM is moving towards for the future, the current deployments
are all based on Elasticsearch, to move from one to the other we will first need to support both versions
and slowly change over on a per deployment basis

## How Has This Been Tested?
Unit tests have been updated
Manual testing has been done of listings

## Checklist

### DIFF
--- a/api/admin/controller/search_service_self_tests.py
+++ b/api/admin/controller/search_service_self_tests.py
@@ -24,6 +24,6 @@ class SearchServiceSelfTestsController(SelfTestsController, ExternalSearchTest):
     def look_up_by_id(self, identifier):
         return self.look_up_service_by_id(
             identifier,
-            ExternalIntegration.OPENSEARCH,
+            ExternalIntegration.ELASTICSEARCH,
             ExternalIntegration.SEARCH_GOAL,
         )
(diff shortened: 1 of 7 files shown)


jonathangreen:
I think this makes sense is is more aligned with the approach I was thinking. If you are okay with it @RishiDiwanTT, I'll merge this into the [feature/migrate-opensearch-1.2.0](https://github.com/ThePalaceProject/circulation/tree/feature/migrate-opensearch-1.2.0) branch, and we can continue the code review there.

PR #547, opened 2022-09-23
Proquest with authentication tokens as an OPDS2 feed

RishiDiwanTT:
## Description
The following changes are part of this PR
- Auth token endpoint parsing for OPDS2 feeds
- OPDS2 feed links that are marked as templated are stored as such
- ~~OPDS2 API class for generic OPDS2 circulation management with the authentication token endpoint~~
- OPDS2 authentication happens in a fulfillment post processor

~~The OPDS2 datasource now has an API class of its own, so all OPDS2 protocol feeds should work with it~~

~~This is not production ready, and is a draft for initial review of the architecture.~~

## Motivation and Context
The Proquest feed has it's own importer and API protocol as of now. 
To move this into a pure OPDS2 implementation some change have been made to the Proquest feed to allow authentication.
And some changes are being made to the OPDS2 circulation workflow to allow borrow/fulfill with the given authentication

[Notion](https://www.notion.so/lyrasis/Implement-token-service-extension-for-OPDS2-feed-c1b30a678a2444c4ba946dea09042ea3)

## How Has This Been Tested?
Manual testing has been done.
~~Unit tests are yet to be written or run in any format, this change probably breaks OPDS2 tests~~
Unit tests have been written for the changes

## Checklist

### DIFF
--- a/api/opds2.py
+++ b/api/opds2.py
@@ -6,11 +6,7 @@ from typing import TYPE_CHECKING
 from flask import url_for
 from uritemplate import URITemplate
 
-from api.circulation import (
-    BaseCirculationAPI,
-    CirculationFulfillmentPostProcessor,
-    FulfillmentInfo,
-)
+from api.circulation import CirculationFulfillmentPostProcessor, FulfillmentInfo
 from api.circulation_exceptions import CannotFulfill
 from core.model import ConfigurationSetting, ExternalIntegration
 from core.model.edition import Edition
@@ -78,38 +74,7 @@ class OPDS2NavigationsAnnotator(OPDS2Annotator):
         ]
 
 
-class OPDS2API(BaseCirculationAPI):
-    """
-    Generic OPDS2 Circulation APIs
-    This should be a pure implementation with no references to daatasources
-    """
-
-    @classmethod
-    def get_authentication_token(
-        cls, patron: Patron, token_auth_url: str
-    ) -> ProblemDetail | str:
-        """Get the authentication token for a patron"""
-        url = URITemplate(token_auth_url).expand(patron_id=patron.username)
-        log = logging.getLogger("OPDS2API")
-        response = HTTP.get_with_timeout(url)
-        if response.status_code != 200:
-            log.error(
-                f"Could not authenticate the patron({patron.username}): {response.content}"
-            )
-            return INVALID_CREDENTIALS
-
-        # The response should be the JWT token, not wrapped in any format like JSON
-        token = response.content
-        if not token:
-            log.error(
-                f"Could not authenticate the patron({patron.username}): {response.content}"
-            )
-            return INVALID_CREDENTIALS
-
-        return token.decode()
-
-
-class OPDS2TokenAuthenticationFulfillmentProcessor(CirculationFulfillmentPostProcessor):
+class TokenAuthenticationFulfillmentProcessor(CirculationFulfillmentPostProcessor):
     """In case a feed has a token auth endpoint and the content_link requires an authentication token
     Then we must fetch the required authentication token from the token_auth endpoint and
     expand the templated url with the received token.
@@ -129,7 +94,8 @@ class OPDS2TokenAuthenticationFulfillmentProcessor(CirculationFulfillmentPostPro
         if not fulfillment.content_link:
             return fulfillment
 
-        if "{authentication_token}" not in fulfillment.content_link:
+        templated = URITemplate(fulfillment.content_link)
+        if "authentication_token" not in templated.variable_names:
             return fulfillment
 
         token_auth = ConfigurationSetting.for_externalintegration(
@@ -138,12 +104,34 @@ class OPDS2TokenAuthenticationFulfillmentProcessor(CirculationFulfillmentPostPro
         if not token_auth or token_auth.value is None:
             return fulfillment
 
-        token = OPDS2API.get_authentication_token(patron, token_auth.value)
+        token = self.get_authentication_token(patron, token_auth.value)
         if type(token) == ProblemDetail:
             raise CannotFulfill()
 
-        fulfillment.content_link = URITemplate(fulfillment.content_link).expand(
-            authentication_token=token
-        )
+        fulfillment.content_link = templated.expand(authentication_token=token)
         fulfillment.content_link_redirect = True
         return fulfillment
+
+    @classmethod
+    def get_authentication_token(
+        cls, patron: Patron, token_auth_url: str
+    ) -> ProblemDetail | str:
+        """Get the authentication token for a patron"""
+        url = URITemplate(token_auth_url).expand(patron_id=patron.username)
+        log = logging.getLogger("OPDS2API")
+        response = HTTP.get_with_timeout(url)
+        if response.status_code != 200:
+            log.error(
+                f"Could not authenticate the patron({patron.username}): {response.content}"
+            )
+            return INVALID_CREDENTIALS
+
+        # The response should be the JWT token, not wrapped in any format like JSON
+        token = response.text
+        if not token:
+            log.error(
+                f"Could not authenticate the patron({patron.username}): {response.content}"
+            )
+            return INVALID_CREDENTIALS
+
+        return token
--- a/api/opds2.py
+++ b/api/opds2.py
@@ -1,17 +1,38 @@
-from typing import Dict
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
 
 from flask import url_for
+from uritemplate import URITemplate
 
+from api.circulation import BaseCirculationAPI, FulfillmentInfo, LoanInfo
+from api.circulation_exceptions import NoLicenses, NotCheckedOut
+from api.problem_details import CANNOT_FULFILL, INVALID_CREDENTIALS
+from core.model import (
+    Collection,
+    ConfigurationSetting,
+    ExternalIntegration,
+    LicensePoolDeliveryMechanism,
+)
 from core.model.edition import Edition
 from core.model.identifier import Identifier
 from core.model.resource import Hyperlink
 from core.opds2 import OPDS2Annotator
+from core.util.datetime_helpers import utc_now
+from core.util.http import HTTP
+from core.util.problem_detail import ProblemDetail
+
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
+    from core.model import LicensePool, Loan, Patron
 
 
 class OPDS2PublicationsAnnotator(OPDS2Annotator):
     """API level implementation for the publications feed OPDS2 annotator"""
 
-    def loan_link(self, edition: Edition) -> Dict:
+    def loan_link(self, edition: Edition) -> dict:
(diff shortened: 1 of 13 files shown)


api/opds2.py
```diff
@@ -58,3 +79,129 @@ def feed_links(self):
...
+        url = URITemplate(self.token_auth_url).expand(patron_id=patron.username)
+        response = HTTP.get_with_timeout(url)
+        if response.status_code != 200:
+            self.log.error(
+                f"Could not authenticate the patron({patron.username}): {response.content}"
+            )
+            return INVALID_CREDENTIALS
+
+        # The response should be the JWT token, not wrapped in any format like JSON
+        token = response.json().get("token")
```
jonathangreen, line 112: The comment and the logic here do not match. The comment is correct though, the token should be just plain text that we can directly template into the link.

jonathangreen: The plain text token matches what we saw in the actual feed.

api/opds2.py
```diff
@@ -58,3 +79,129 @@ def feed_links(self):
         return [
             {"href": self.url, "rel": "self", "type": self.OPDS2_TYPE},
         ]
+
+
+class OPDS2API(BaseCirculationAPI):
```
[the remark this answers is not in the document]
RishiDiwanTT: Currently the WAYFLess post processor is only setup for OPDS and not OPDS2.0, so it does not actually come in the way right now, also since it is the post processor it should work as it does not after the `.fulfill()` method is called.
I agree with the repetition of code between CirculationAPI and OPDS2API, I was however under the impression that all cicrulation calls with an OPDS2 collection that has a token endpoint would require authentication with that endpoint, should I be making those changes directly in the CirculationAPI? Or should we not authenticate those calls at all?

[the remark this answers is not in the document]
jonathangreen: Thats good, I thought that the WAYFLess URL processor applied to OPDS 2.0 as well. 

> agree with the repetition of code between CirculationAPI and OPDS2API, I was however under the impression that all cicrulation calls with an OPDS2 collection that has a token endpoint would require authentication with that endpoint

There are three endpoints we call, I think: 
- The feed URL
- The authentication URL
- The fulfillment URL

The feed URL and the authentication URL don't require the token, only the fulfillment do. Are there any other endpoint calls I'm forgetting here?

core/model/resource.py
```diff
@@ -69,6 +70,9 @@ class Resource(Base):
     # but it may also be a made-up URI.
     url = Column(Unicode, index=True)
 
+    # Is the href related to this resource templated?
+    templated = Column(Boolean, default=False)
```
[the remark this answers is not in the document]
jonathangreen: I was hoping that the Resource got passed along with the fulfillment, so we would be able to check if this is templated within the processor. I didn't realize we were just passing the URL there.

No reason to add this if we aren't using it.

api/opds2.py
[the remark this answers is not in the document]
RishiDiwanTT: Yes, they do not require the token, I was thinking of it as an extra layer of proquest authorization ([example here](https://github.com/ThePalaceProject/circulation/pull/547/commits/35e8affce7591d78f3cf64fc3129097a62e629d0#diff-b8e9c8ebb72f45f352cf91d3f85f476e946a8bde5d29606216462092f47fcd76R125)). If we don't want to do that, its fine.

api/opds2.py
```diff
@@ -1,17 +1,37 @@
...
 from core.model.identifier import Identifier
+from core.model.licensing import DeliveryMechanism
 from core.model.resource import Hyperlink
 from core.opds2 import OPDS2Annotator
+from core.problem_details import INVALID_CREDENTIALS
+from core.util.http import HTTP
+from core.util.problem_detail import ProblemDetail
+
+if TYPE_CHECKING:
+    pass
```
jonathangreen, line 26: Leftover from testing?

api/opds2.py
```diff
@@ -58,3 +78,74 @@ def feed_links(self):
...
+
+
+class OPDS2API(BaseCirculationAPI):
+    """
+    Generic OPDS2 Circulation APIs
+    This should be a pure implementation with no references to daatasources
+    """
+
+    @classmethod
+    def get_authentication_token(
```
jonathangreen, line 90: Since this isn't used anywhere else, should this function just get moved into `OPDS2TokenAuthenticationFulfillmentProcessor`?

core/opds2_import.py
```diff
@@ -937,6 +938,16 @@ def extract_last_update_dates(
 
         return dates
 
+    def _parse_feed_links(self, links: List["ast.core.Link"]) -> None:
```
jonathangreen, line 941: Should be able to not quote the type here with a future import.

jonathangreen (commented):
This is looking good, just a couple more minor comments.

I did some testing with this locally, and it seems like its working against the new proquest feed. Next step is to test this branch on the columbia staging server. I'd like do that before we do the final merge of this branch.

api/opds2.py
RishiDiwanTT: autoflake probably removed the imports

api/opds2.py
RishiDiwanTT: Since the {authentication_token} is a more generic OPDS2 workflow I thought having it accessible elsewhere would be better

core/opds2_import.py
```diff
@@ -898,7 +903,7 @@ def _record_publication_unrecognizable_identifier(
                 f"Publication # {original_identifier} ('{title}') has an unrecognizable identifier."
             )
 
-    def extract_next_links(self, feed: Union[str, opds2_ast.OPDS2Feed]) -> List[str]:
+    def extract_next_links(self, feed: str | opds2_ast.OPDS2Feed) -> list[str]:
```
jonathangreen, line 906: See union comment above

core/opds2_import.py
```diff
@@ -916,8 +921,8 @@ def extract_next_links(self, feed: Union[str, opds2_ast.OPDS2Feed]) -> List[str]
         return next_links
 
     def extract_last_update_dates(
-        self, feed: Union[str, opds2_ast.OPDS2Feed]
-    ) -> List[Tuple[str, datetime]]:
+        self, feed: str | opds2_ast.OPDS2Feed
```
jonathangreen, line 924: See union comment above

core/opds2_import.py
```diff
@@ -937,9 +942,19 @@ def extract_last_update_dates(
...
+                # Save the collection-wide token authentication endpoint
+                auth_setting = ConfigurationSetting.for_externalintegration(
+                    ExternalIntegration.TOKEN_AUTH, self.collection.external_integration
+                )
+                auth_setting.value = link.href
+
     def extract_feed_data(
-        self, feed: Union[str, opds2_ast.OPDS2Feed], feed_url: Optional[str] = None
-    ) -> Tuple[Dict, Dict]:
+        self, feed: str | opds2_ast.OPDS2Feed, feed_url: str | None = None
```
jonathangreen, line 956: See union comment above

tests/api/test_opds2.py
```diff
@@ -86,3 +101,110 @@ def test_navigation(self):
...
+        ff_info.content_link = (
+            "http://example.org/11234/fulfill?authToken=authentication_token"
+        )
+        ff_info.content_link_redirect = False
+        ff_info = processor.fulfill(patron, None, work.license_pools[0], None, ff_info)
+        assert ff_info.content_link_redirect == False
+
+        # No token endpoint config
+        ff_info.content_link = (
+            "http://example.org/11234/fulfill?authToken={authentication_token}"
```
jonathangreen, line 206: Might be good to add a test case for a complicated URI template, something like: 
```
http://example.org/11234/fulfill{?authentication_token}
```

expands to
```
http://example.org/11234/fulfill?authentication_token=foobarbaz
```

tests/core/files/opds2/pq_auth_token_feed.json
```diff
@@ -0,0 +1,158 @@
+{
+    "metadata": {
+        "title": "Ebook Central OPDS Feed for columbia",
+        "itemsPerPage": 2,
+        "currentPage": 1,
+        "numberOfItems": 2
+    },
+    "links": [
+        {
+            "href": "https://ebookcentral.proquest.com/auth/ws/auth/PlainPartnerAuthToken/columbia?userName={patron_id}",
```
jonathangreen, line 10: Might be good to do a find / replace on this file, and make the links a bit more generic. Maybe just replace `ebookcentral.proquest.com` with `example.com`, so we don't end up accidentally hitting the PQ servers in our tests. Maybe replace `lib/columbia` with `test` as well. And drop `pq_` from the filename.

tests/core/test_opds2_import.py
```diff
@@ -407,3 +409,18 @@ def test_opds2_importer_skips_publications_with_unsupported_identifier_types(
         # Ensure that it was parsed correctly and available by its identifier.
         edition = self._get_edition_by_identifier(imported_editions, identifier)
         assert edition is not None
+
+    def test_auth_token_feed(self):
+        content = self.sample_opds("pq_auth_token_feed.json")
```
jonathangreen, line 414: Since we have this feed as a fixture. It would be nice to have a test that tests one of the items in the feed all the way to fulfillment. We have unit tests for each piece, but since we're importing the feed, we should be able to do a more holistic test as well.

jonathangreen (commented):
Testing this on the Columbia staging server, and it seems to be working well so far. I went through the code a bit more thoroughly and added a few more minor comments. Once those are addressed and testing is complete in staging, I think this is ready to merge.

core/opds2_import.py
```diff
@@ -58,6 +61,10 @@
 from .util.http import BadResponseException
 from .util.opds_writer import OPDSFeed
 
+if TYPE_CHECKING:
+    from webpub_manifest_parser.core import ast as core_ast
+    from webpub_manifest_parser.opds2 import ast as opds2_ast
```
jonathangreen, line 66: Looks like this is already imported above (line 11)

core/opds2_import.py
```diff
@@ -76,7 +83,7 @@ def __init__(self, manifest_parser_factory: ManifestParserFactory):
         self._manifest_parser_factory = manifest_parser_factory
 
     def parse_manifest(
-        self, manifest: Union[str, Dict, Manifestlike]
+        self, manifest: str | dict | Manifestlike
```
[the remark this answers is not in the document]
tdilauro: The pipe Union syntax is supported if you import the `annotations` future.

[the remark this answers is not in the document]
jonathangreen: Thanks @tdilauro. I didn't know that. 

For my own future reference its documented by mypy here: 
https://mypy.readthedocs.io/en/stable/runtime_troubles.html?highlight=__future__#using-x-y-syntax-for-unions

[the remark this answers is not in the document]
jonathangreen: I'll resolve the comments about type Unions.

jonathangreen (approved):
🚀  looks good to me

PR #572, opened 2022-10-05
Crontab entry to run the auto-updating lists script

RishiDiwanTT:
## Description
Once an hour between 7AM and 1AM, 5 minutes past the hour (after update list size runs)

## Motivation and Context
Now that the auto update UI is implemented and ready to be deployed, the cron job for updating the lists should be activated.

## How Has This Been Tested?
Locally by installing and checking the validity of the cron expression.

## Checklist


tdilauro (approved):
Looks good! ⏲️

PR #576, opened 2022-10-06
Initial alembic integration

RishiDiwanTT:
## Description
The alembic/versions folder will hold the migration scripts 
alembic.ini has UTC timezone set for the migration times

The alembic/env.py MUST import any models not already imported in models/__init__.py 
This is to allow autogenerate to work correctly

Added the fkey name to the devicetokens model to avoid unnecessary migration

The migrate_database script has been co-opted to be the script that runs the alembic migration
This allows the ansible deployment script to remain as-is, it will only need changes to downgrade behaviour

## Motivation and Context
Currently the CM uses a home made solution for DB upgrades. SqlAlchemy has a recommended solution called [Alembic](https://alembic.sqlalchemy.org/en/latest/). It has some nice features, like keeping track of DB versions, and allowing both upgrades and downgrades from the current version. It can also generate migrations automatically for many DB changes.

[Notion](https://www.notion.so/lyrasis/Use-Alembic-for-Database-Upgrades-cc165881bbce4daaa87b56fda49eeccf)

## How Has This Been Tested?
Auto-generate the initial migration, reviewed the changes and ran the upgrade (via migrate_database).

## Checklist


alembic.ini
```diff
@@ -0,0 +1,106 @@
...
+# defaults to the current working directory.
+prepend_sys_path = .
+
+# timezone to use when rendering the date within the migration file
+# as well as the filename.
+# If specified, requires the python-dateutil library that can be
+# installed by adding `alembic[tz]` to the pip requirements
+# string value is passed to dateutil.tz.gettz()
+# leave blank for localtime
+timezone = UTC
```
RishiDiwanTT, line 23: This can also be US/Eastern

alembic/env.py
```diff
@@ -0,0 +1,91 @@
...
+target_metadata = Base.metadata
+
+# other values from the config, defined by the needs of env.py,
+# can be acquired:
+# my_important_option = config.get_main_option("my_important_option")
+# ... etc.
+
+
+# Import the models not included in models/__init__.py
+# This is required for autogenerate to work correctly
```
RishiDiwanTT, line 32: Some special import rules here

alembic.ini
```diff
@@ -0,0 +1,106 @@
...
+# version_path_separator = ;
+# version_path_separator = space
+version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+# This must be gotten from the env var in env.py
+# sqlalchemy.url = driver://user:pass@localhost/dbname
```
RishiDiwanTT, line 59: We pull this in from the env var in env.py

alembic/versions/6f96516c7a7b_initial.py
```diff
@@ -0,0 +1,90 @@
...
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "6f96516c7a7b"
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
```
RishiDiwanTT, line 19: Do we want to drop these old tables from the database entirely?

RishiDiwanTT:
@jonathangreen @tdilauro @dbernstein @io7m Requesting review

RishiDiwanTT:
@jonathangreen there are a few scripts that exist in the repository that I cannot find any usage of from the `hosting-playbook` repository
These are 
- 02_manage_simplified_database.sh
- initialize_instance (and respective class)

Should we delete these scripts?
Additionally the README mentions a SIMPLIFIED_DB_TASK env var, which is used in the `02_manage_simplified_database.sh`. But the script itself isn't invoked in either the ansible playbook or the circulation code itself.

RishiDiwanTT:
> @jonathangreen there are a few scripts that exist in the repository that I cannot find any usage of from the `hosting-playbook` repository These are
> 
> * 02_manage_simplified_database.sh
> * initialize_instance (and respective class)
> 
> Should we delete these scripts? Additionally the README mentions a SIMPLIFIED_DB_TASK env var, which is used in the `02_manage_simplified_database.sh`. But the script itself isn't invoked in either the ansible playbook or the circulation code itself.

@jonathangreen In the same line of thought, `SIMPLIFIED_DB_TASK` as an env var is pervasive throughout the docker builds in `circulation` and `hosting-playbook`, but is only ever use in `02_manage_simplified_database.sh`

RishiDiwanTT:
@jonathangreen I have updated the README with an explanation of how the migration will work in production. The method is divergent from the in-app paradign the library-registry follows, but it keeps with the current method of migration the circulation follows.

RishiDiwanTT:
@jonathangreen @tdilauro bump

RishiDiwanTT:
@jonathangreen @tdilauro 
https://media.giphy.com/media/59fYAHs0Y4llMQ7Lnh/giphy.gif

RishiDiwanTT:
@jonathangreen is there something that needs to be done in this PR to improve the confidence in these changes to be merged?
Like a test run on Dev?

io7m:
I'm sorry, I didn't realize this was going on (happened during the time off I took).

I'd like to have a read through this, because I'd like to understand Alembic. I'm very interested in how people handle relational databases in the context of doing infrastructure upgrades. It's a hard problem!

alembic.ini
io7m: I think UTC is the right choice.

core/bin/migrate_database
```diff
@@ -1,7 +1,51 @@
...
+for i in 1 2 3; do
+    if [ -f "$INI_FILE" ]; then
+        FOUND_INI=1
+        break
+    else
+        # Go up one directory to try and find the ini
+        INI_FILE="../${INI_FILE}"
+        SRC_DIR="../${SRC_DIR}"
+    fi
+done
```
io7m, line 21: I'm not keen on searching like this. Is there a way we can explicitly provide a config file path?

io7m (commented):
I think these changes look good (with one minor thing I didn't like, above).

I'm not going to click "approve" because I don't know if there are reasons we're holding off on merging this right now.

@jonathangreen ?

io7m:
Also: Of all of the migrations I've seen so far, it seems like we do data migrations _vastly_ more often than we do schema changes. I think in order to get deployments that we can revert, we'd need to start consistently providing code to downgrade in each migration script... How feasible is this to accomplish? If an upgrade involved dropping a table, we're not going to be able to get the data in that table back for a downgrade. We're almost certainly going to need a set of rules that we can follow for migration scripts (maybe tied to semantic versioning of the CM; "don't drop a table you might need to get back without a major version number increment").

core/bin/migrate_database
RishiDiwanTT: I could change this to just be `../alembic.ini`

jonathangreen:
Note for myself, as I'm reviewing this. There is one unreleased migration file [20230126-clean-up-proquest.sql](https://github.com/ThePalaceProject/circulation/blob/0df4ab68effd9d63d9dddfea15859304e979841a/migration/20230126-clean-up-proquest.sql ) from https://github.com/ThePalaceProject/circulation/pull/767. If this PR goes in before we make another release, we will have to recreate that migration with Alembic.

alembic/versions/6f96516c7a7b_initial.py
jonathangreen: I don't think this migration is needed anymore. When create a new CM instance, I don't see a complaints table.

scripts.py
```diff
@@ -1011,18 +1010,10 @@ def do_run(self, ignore_search=False):
                 # Elasticsearch isn't configured, so do nothing.
                 pass
 
-        # Set a timestamp that represents the new database's version.
```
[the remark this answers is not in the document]
jonathangreen: For an existing database, that has the `libraries` table, but doesn't have an alembic stamp, won't we miss setting the version here?

core/bin/migrate_database
jonathangreen: I wonder if we would be better off keeping the `migrate_database` script in python, rather then bringing a shell script into it. Like we are doing with library registry. But that is just a minor nit on my part, I prefer debugging and troubleshooting python scripts over shell scripts any day.

core/model/devicetokens.py
```diff
@@ -26,7 +26,7 @@ class DeviceToken(Base):
     id = Column("id", Integer, primary_key=True)
     patron_id = Column(
         Integer,
-        ForeignKey("patrons.id", ondelete="CASCADE"),
+        ForeignKey("patrons.id", ondelete="CASCADE", name="devicetokens_patron_fkey"),
```
jonathangreen, line 29: This change seems unrelated, why the change here?

alembic/versions/6f96516c7a7b_initial.py
jonathangreen: And actually in my testing these seem to be causing the migrations to fail.

jonathangreen (changes requested):
In order to merge this PR, I need to have some reasonable confidence it won't break our deployments when the code is running in docker. From my testing today, it doesn't seem like this was run under docker yet. I made a PR to add a `docker-compose` file that should help some with that. 

Here are the cases I need to see working before merging this: 
- ✅ New instance - Uninitialized DB
  - Creates all the tables, and the alembic version table
  - This wasn't working for me, when starting the docker container, because it wasn't resolving the path to `alembic.ini`. This is fixed in https://github.com/ThePalaceProject/circulation/pull/576/commits/d8e9f4bea8a8247f7ee84d99bf2ddb92a31401aa from my testing.
- Upgraded instance - Initialized DB, Uninitialized Alembic
  - I tested this by taking the most recent code, and initializing a database, then upgrading to this branch.
  - I see this when `initialize_instance` is called: `ERROR:Instance initialization:I think this site has already been initialized; doing nothing.` That is expected, although I believe because I get that error Alembic isn't creating its stamp table.
  - I see this when `migrate_database` is called: 
    ```
    Running 'alembic upgrade head'
    INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
    INFO  [alembic.runtime.migration] Will assume transactional DDL.
    INFO  [alembic.runtime.migration] Running upgrade  -> 6f96516c7a7b, initial
      File "/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
        self.dialect.do_execute(
      File "/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute
    psycopg2.errors.UndefinedObject: index "ix_libraryalias_language" does not exist
    ```
 - Upgraded Instance - Initialized DB, Initialized Alembic - Migrations needed
   - Because of the previous errors I didn't get to test this case
 - Downgraded Instance - Initialized DB, Initialized Alembic - Alembic database version pointing to a version that doesn't exist (a future migration). 
   - As with the library registry, I don't want to automatically downgrade, but this needs to log a reasonable message about what is going on, and not block startup.

core/model/devicetokens.py
jonathangreen: ahh I see in the PR description: 
> Added the fkey name to the devicetokens model to avoid unnecessary migration

I'm not sure that is true, or at least true anymore. You have to be somewhat careful with Alembic because it generates migrations based on the current state of your database. Is it possible your database was somewhat out of sync when you needed to add this foreigin key name?

jonathangreen: I stand corrected, I did some more testing, and you are right @RishiDiwanTT this is required. Will resolve this comment.

alembic/versions/6f96516c7a7b_initial.py
jonathangreen: I pushed a commit that dropped this migration, so I could get a bit further with testing.

scripts.py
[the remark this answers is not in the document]
RishiDiwanTT: Yes that's right, from the original intent of the script, this is meant only to "initialise a brand new database".
The total life cycle depends on the `02_manage_simplified_database.sh` script running in 'auto' mode.
First the `initialize_database` (this script) will run, then the `core/bin/migrate_database` will run.
`migrate_database` does an `upgrade head` which is idempotent and will create an `alembic_version` table if it is missing.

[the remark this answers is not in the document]
RishiDiwanTT: ONLY if we are an empty, brand new DB should we `stamp head` since SQLAlchemy will autocreate all tables and constraints.
If we are a non-empty DB without an alembic_version, then `upgrade head` will create the table and run all migrations in the repository. This should only happen during the FIRST run after deploying the alembic changes, thereafter the DB with a head stamp should exist.
Is there a case where we have a non-empty database and we DO NOT want to run the alembic revisions?

RishiDiwanTT:
> Also: Of all of the migrations I've seen so far, it seems like we do data migrations _vastly_ more often than we do schema changes. I think in order to get deployments that we can revert, we'd need to start consistently providing code to downgrade in each migration script... How feasible is this to accomplish? If an upgrade involved dropping a table, we're not going to be able to get the data in that table back for a downgrade. We're almost certainly going to need a set of rules that we can follow for migration scripts (maybe tied to semantic versioning of the CM; "don't drop a table you might need to get back without a major version number increment").

Alembic is just a framework for migrations, data migrations can be easily written using python and sqlalchemy within the alembic migration files (like in Django).
I also agree with the "do not delete tables/columns immediately" philosophy, they should be kept around until we know rollbacks aren't going to happen, like a couple of months down the line.

RishiDiwanTT:
> psycopg2.errors.UndefinedObject: index "ix_libraryalias_language" does not exist

Hmm, for my local DBs, I have a dump of both NJ and VT (albeit at least 10 months old), they both have the libraryalias and complaints tables, without any data. I can create exceptions in alembic to skip those databases out.

RishiDiwanTT:
> Downgraded Instance - Initialized DB, Initialized Alembic - Alembic database version pointing to a version that doesn't exist (a future migration).

I'm not sure this will work. I think the steps for a downgrade will have to be as follows
- Downgrade DB via alembic
- Deploy rolled back CM version

If we deploy a rolled back CM before downgrading the DB, we are deploying a version of code that is incompatible with the version of the database.
Alembic itself does not allow it, it will throw an error of the form `  FAILED: Can't locate revision identified by 'fa1abac4aef9'`

scripts.py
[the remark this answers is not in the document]
jonathangreen: No, this makes sense to me. Thank you for the clarification.

jonathangreen:
I made a couple small changes here while I was testing:
- I reverted a bunch of my commits from yesterday.
- I made the initial migration drop those old indexes and tables, but only if they exist.
- Changed `alembic.ini` so that the migration filenames are prefixed with a date. To make it easier to see the order they were added, when we want to cull old migrations.
- Added a few more tests.

jonathangreen (approved):
This looks really great @RishiDiwanTT 🚀. Thank you for working through this, and keeping this PR alive for so long. I think moving to Alembic for migrations will be a big win.

I just did another round of testing, going to put the results here:
- ✅ New instance - Uninitialized DB
  - Creates all the tables, and the alembic version table
- ✅ Upgraded instance - Initialized DB, Uninitialized Alembic
  - I tested this by taking the most recent code, and initializing a database, then upgrading to this branch.
  - `initialize_instance` is called and does nothing as expected.
  - `migrate_database` is called, runs the initial migration, and creates stamp table.
 - ✅ Upgraded Instance - Initialized DB, Initialized Alembic - Migrations needed
  - `initialize_instance` is called and does nothing as expected.
  - `migrate_database` is called, runs the all required migrations and updates stamp in DB.
 - ✅ Downgraded Instance - Initialized DB, Initialized Alembic - Alembic database version pointing to a version that doesn't exist (a future migration). 
   - This reacts as I would expect. It throws an error, but does not halt the startup of either the scripts or web container. So if the older code can run with the newer database, it will try.
   - I added a little bit of logging, so the error that happens in this case is easier to see. Something like: `Error: Can't locate revision identified by '057e7ec791a1'. No migrations performed.`

jonathangreen:
@RishiDiwanTT can you review the couple commits I pushed today and make sure they look good to you? If not feel free to change anything there before merging.

pcustic:
Too late to the party: it would be great if there was a bit of docs or short 'how to use alembic in the CM'. I was updating google oauth PR to use alembic and wasn't really sure if I am doing everything correctly.

PR #590, opened 2022-10-12
Enable the auto-updating list feature in the UI.

ray-lee:
## Description

This enables the auto-updating list feature in the UI by setting the feature flag.

## Motivation and Context

Since the script for populating auto-updating lists is now functional, we can allow creating auto-updating lists in the UI.

## How Has This Been Tested?

Create a new custom list. At the bottom of the "Search for Titles" section, there should be radio buttons for "Automatically update this list" and "Manually select titles".

## Checklist


tdilauro (approved):
🚀

PR #591, opened 2022-10-12
Cache OPDS for Distributor bearer token per collection

tdilauro:
## Description

Cache OPDS for Distributor bearer token per collection, rather than only per data source.

- New tests to ensure that tokens are stored per-collection.
- Migration to  remove any extant invalid credentials.

## Motivation and Context

Before this change, every `OPDS for Distributors` collection within a given CM shared a single cached account bearer token credentials per data source. So, when more than one collection shared a data source, collection import monitors and circulation APIs would use the datasource's most recently cached token -- which may have been obtained for a different collection -- until that credential expired. This resulted in the wrong books being imported for some collections and possibly in fulfillment from the wrong account. It may also have resulted in some patron-facing fulfillment errors. 

[Notion issue](https://www.notion.so/lyrasis/Full-OPDS-for-Distributors-import-produced-statistics-inconsistent-with-CM-dashboard-c20eaedcd93d4ad38a3fa1190e50842d)

## How Has This Been Tested?

- Created new tests. Manually ran tests that are most related.

## Checklist

- N/A ~~I have updated the documentation accordingly.~~


dbernstein (approved):
Looks good to me.

PR #598, opened 2022-10-18
Sirsidynix Authentication Provider

RishiDiwanTT:
## Description
Initial implementation of the sirsidynix authentication provider
Only implements the /user/patron/login API for remote authentication
This works as a Basic authentication provider

## Motivation and Context
CA state libraries require this proprietary authentication system implemented.
[Notion](https://www.notion.so/lyrasis/Implement-SirsiDynix-SSO-patron-authentication-491c16eef87440278425017f3b93178b)

## How Has This Been Tested?
Unit tests have been updated.
Manually tested the borrow workflow using this authentication provider.

## Checklist


tests/api/test_authenticator.py
```diff
@@ -3026,3 +3028,64 @@ def test_oauth_authentication_invalid_token(self):
...
+
+    @patch("api.sirsidynix_authentication_provider.HTTP.request_with_timeout")
+    def test_api_patron_login(self, mock_request):
+        response_dict = {"sessionToken": "xxxx"}
+        mock_request.return_value = MockRequestsResponse(200, content=response_dict)
+
+        response = self.api.api_patron_login("username", "pwd")
+
+        assert mock_request.call_count == 1
+        assert mock_request.call_args_list[0] == call(
```
tdilauro, line 3069: Could we use `call_args` -- rather than `call_args_list[0]` -- here, since we're only expecting one call?

```assert mock_request.call_args == call(...)```

**Review by tdilauro — COMMENTED** [review]
This is looking pretty good. We do appear to have a return type/annotation mismatch that we should sort out, though.

File: api/sirsidynix_authentication_provider.py [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+        if self.server_url[-1] != "/":
+            self.server_url = self.server_url + "/"
+        self.sirsi_client_id = integration.setting(self.Keys.CLIENT_ID).value
+        self.sirsi_app_id = integration.setting(self.Keys.APP_ID).value
+
+    def remote_authenticate(self, username: str, password: str) -> PatronData | None:
+        """Authenticate this user with the remote server"""
+        data = self.api_patron_login(username, password)
+        if not data:
+            return False
```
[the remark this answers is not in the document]
RishiDiwanTT: Will update the type annotation, `return False` is the way the other auth implementations are done

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,98 @@
...
+        return PatronData(username=username, authorization_identifier=username)
+
+    ###
+    # API requests
+    ###
+
+    def _request(self, method: str, path: str, json=None) -> Response:
+        """Request wrapper that adds the relevant request headers
+        :param method: The HTTP method for the request
+        :param path: The url path that will get joined to the server_url, should not have a leading '/'
```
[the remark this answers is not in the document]
RishiDiwanTT: Makes sense for now since we have no other type of url

tests/api/test_authenticator.py
```diff
@@ -3026,3 +3028,64 @@ def test_oauth_authentication_invalid_token(self):
...
+        )
+
+    def _headers(self):
+        return {
+            "SD-Originating-App-Id": self.api.sirsi_app_id,
+            "x-sirs-clientID": self.api.sirsi_client_id,
+        }
+
+    def test_settings(self):
+        # with leading slash
```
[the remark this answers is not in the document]
RishiDiwanTT: Updated with a more clear comment

tests/api/test_authenticator.py
RishiDiwanTT: That works too, updated.

RishiDiwanTT:
@tdilauro I'm moving this to draft since Carissa mentioned we may need to implement the Horizon ILS and not Symphony.

jonathangreen:
One additional thing I'd like to see here @RishiDiwanTT. One of our partner libraries mentioned a concern that sometimes barcodes can change for the same patron, for example when they renew a library card. Some ILS provide a more permanent identifier for patrons then barcodes. 

The `Patron` model supports this through the [`external_identifier` field](https://github.com/ThePalaceProject/circulation/blob/main/core/model/patron.py#L70). If we make a request for Patron information after we log in, is there a field that would work to more permanently identify a Patron?

It might also be good to fill in any more information we get back in this request that is supported in the `Patron` record like fines, blocks, etc.

RishiDiwanTT:
> One additional thing I'd like to see here @RishiDiwanTT. One of our partner libraries mentioned a concern that sometimes barcodes can change for the same patron, for example when they renew a library card. Some ILS provide a more permanent identifier for patrons then barcodes.
> 
> The `Patron` model supports this through the [`external_identifier` field](https://github.com/ThePalaceProject/circulation/blob/main/core/model/patron.py#L70). If we make an additional request for Patron information after we log in, is there a field that would work to more permanently identify a Patron?
> 
> It might also be good to fill in any more information we get back in this request that is supported in the `Patron` record like fines, blocks, etc.

As per the email from Brent Thompson, no changes are required for the authentication API between Symphony and Horizon.
Name changes have been made to clarify that we expect Horizon.
Self tests have been made available, without remote patron lookup since we have no way of storing the session token between requests, nor do we have a LibraryID.

jonathangreen:
So here is how the additional patron data is used by SIP2 and Millennium providers:

During a borrow request we call [`PatronUtility.assert_borrowing_privileges(patron)`](https://github.com/ThePalaceProject/circulation/blob/main/api/circulation.py#L984). The `patron` that is passed into this method, comes from the [`requires_auth` decorator](https://github.com/ThePalaceProject/circulation/blob/main/api/routes.py#L67-L78) on [the borrow route](https://github.com/ThePalaceProject/circulation/blob/main/api/routes.py#L573-L585). The decorator calls [`authenticated_patron_from_request`](https://github.com/ThePalaceProject/circulation/blob/main/api/base_controller.py#L57-L88) which eventually calls `AuthenticationProvider.authenticated_patron` which calls [`BasicAuthenticationProvider.authenticate`](https://github.com/ThePalaceProject/circulation/blob/main/api/authenticator.py#L2095-L2197). This calls the `remote_patron_lookup` and `remote_authenticate` functions in the `MilleniumPatronAPI` and `SIP2AuthenticationProvider` classes.

For this Sirsidynix provider, we probably need to implement the additional `_remote_patron_lookup` function that makes the additional request for more detailed patron data and fills in a complete `PatronData` record.

api/admin/controller/patron_auth_services.py
```diff
@@ -16,6 +16,9 @@
 from api.saml.provider import SAMLWebSSOAuthenticationProvider
 from api.simple_authentication import SimpleAuthenticationProvider
 from api.sip import SIP2AuthenticationProvider
+from api.sirsidynix_authentication_provider import (
+    SirsiDynixHorizonAuthenticationProvider,
```
jonathangreen, line 20: We will be using this integration with both Horizon and Symphony, so it probably makes sense to just call this `SirsiDynixAuthenticationProvider` unless we end up needing two different implementations, one for Horizon and one for Symphony.

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,120 @@
...
+            "label": _("Server URL"),
+            "description": _("The external server url."),
+            "required": True,
+            "default": "https://vendor1-sym.sirsidynix.net/ilsws_current/",
+        },
+        {
+            "key": Keys.APP_ID,
+            "label": _("Application ID"),
+            "description": _(
+                "This can be anything, it is used to group all requests to the API."
```
jonathangreen, line 36: I believe this should be a hard coded value, so all our CMs are sending the same thing.

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,120 @@
...
+class SirsiDynixHorizonAuthenticationProvider(BasicAuthenticationProvider):
+    NAME = "SirsiDynix Horizon Authentication"
+    DESCRIPTION = "SirsiDynix Horizon Webservice Authentication"
+    FLOW_TYPE = "http://librarysimplified.org/authtype/sirsidynix-horizon"
+
+    class Keys:
+        APP_ID = "APP_ID"
+        CLIENT_ID = "CLIENT_ID"
+
+    SETTINGS = [
```
jonathangreen, line 24: We should probably provide an appropriate `LIBRARY_SETTINGS` field, so that we can set a Library ID per library associate with this authentication mechanism. 

See the `LIBRARY_SETTINGS` in the base class https://github.com/ThePalaceProject/circulation/blob/main/api/authenticator.py#L1353-L1428 and what we are doing in the `MilleniumPatronAPI` class https://github.com/ThePalaceProject/circulation/blob/main/api/millenium_patron.py#L170-L187

api/admin/controller/patron_auth_services.py
jonathangreen, line 20: We will be using this integration with both Horizon and Symphony, so it probably makes sense to just call this `SirsiDynixAuthenticationProvider` unless we end up needing two different implementations, one for Horizon and one for Symphony.

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,120 @@
...
+        self.sirsi_client_id = integration.setting(self.Keys.CLIENT_ID).value
+        self.sirsi_app_id = integration.setting(self.Keys.APP_ID).value
+
+    def remote_authenticate(self, username: str, password: str) -> PatronData | bool:
+        """Authenticate this user with the remote server."""
+        data = self.api_patron_login(username, password)
+        if not data:
+            return False
+
+        return PatronData(
```
jonathangreen, line 75: We can probably sub-class `PatronData` so that we can use it to pass the our session information along to the new `_remote_patron_lookup` function without saving it out to the database.

api/sirsidynix_authentication_provider.py
RishiDiwanTT: Hard Coded, or environment variable?

api/admin/controller/patron_auth_services.py
RishiDiwanTT: Unless we want to be fancy and have some dynamism in the class based on what API we're hitting. Which seems unnecessarily complex.

jonathangreen: Two implementations is fine with me. They hopefully can share some code, since I think they will be fairly similar. Since we need support for both Horizon and Symphony we can either do this in this PR, or add an additional ticket for supporting the other.

api/sirsidynix_authentication_provider.py
jonathangreen: I was thinking hard coded, but you are right, an environment variable is better idea.

api/admin/controller/patron_auth_services.py
RishiDiwanTT: We can do it in this ticket in another ticket, we would need access to a symphony environment though. The current API endpoint is Horizon.

pcustic (commented):
This seems fine to me 👍 

Is there some Notion documentation about the Patron auth that needs to be updated with this change?

I won't approve this PR since Jonathan and Tim are still reviewing this.

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,279 @@
...
+    from requests import Response
+
+    from core.model.patron import Patron
+
+
+class SirsiDynixHorizonAuthenticationProvider(BasicAuthenticationProvider):
+    """SirsiDynix Authentication API implementation.
+
+    Currently is only used to authenticate patrons, there is no CRUD implemented for patron profiles.
+    It is recommended to have the environment variable `SIRSI_DYNIX_APP_ID` set.
```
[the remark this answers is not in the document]
RishiDiwanTT: Having an App ID is recommended by the SirsiDynix docs, but not mandatory. I'll clarify the point a bit.

RishiDiwanTT:
@jonathangreen bump. We can implement Symphony in another PR, when we get access and credentials.

RishiDiwanTT:
@jonathangreen @tdilauro The PR has been updated to work with the Horizon API and credentials provided

jonathangreen:
@RishiDiwanTT The code for this looks good. I have a couple questions from testing:

When I try to authenticate patrons I get
```json
{"messageList":[{"code":"invalidWorkingLibraryIDHeader","message":"SD-Working-LibraryID PALACE is invalid."}]}
```

What value were you using for testing?

RishiDiwanTT:
> @RishiDiwanTT The code for this looks good. I have a couple questions from testing:
> 
> When I try to authenticate patrons I get
> 
> ```json
> {"messageList":[{"code":"invalidWorkingLibraryIDHeader","message":"SD-Working-LibraryID PALACE is invalid."}]}
> ```
> 
> What value were you using for testing?

For the first patron example in the ticket we can use "coal" (Coalinga District) as the Library ID. Each library will have a unique Library ID which SirsiDynix maintains a table for, and we should add it to the Library-PatrinAuth integration.

jonathangreen:
@RishiDiwanTT Can you look at adding a couple more tests to get the coverage of this PR up while we wait to hear back about distinguishing between the patrons of different libraries.

RishiDiwanTT:
> @RishiDiwanTT Can you look at adding a couple more tests to get the coverage of this PR up while we wait to hear back about distinguishing between the patrons of different libraries.

`--cov-report=html` blew my mind, and should be the standard method of developers to ensure they have adequate code coverage

RishiDiwanTT:
@jonathangreen This should be ready for review and testing

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,332 @@
...
+            patrondata.block_reason = SirsiBlockReasons.NOT_APPROVED
+            return patrondata
+
+        # Remove the library prefix from the patron type
+        # we are left with the patron "suffix"
+        # This suffix can be part of the blocked patron types list
+        patron_suffix = patron_type[len(self.sirsi_library_prefix) :]
+        if patron_suffix in self.sirsi_disallowed_prefixes:
+            patrondata.block_reason = SirsiBlockReasons.PATRON_BLOCKED
+            return patrondata
```
jonathangreen, line 199: I'd rather see this enforced the `enforce_library_identifier_restriction` function https://github.com/ThePalaceProject/circulation/blob/03fe8e5d02bce41d3d68147d29ff0d8bd30b61af/api/authenticator.py#L1546-L1584

This is how we are enforcing patron restrictions in other integration.

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,332 @@
...
+
+        patrondata.external_type = patron_type
+
+        # Basic block reasons
+
+        # Does the patron type start with the library prefix
+        # This ensures the patron is a member of the library
+        # they are interacting with
+        if not patron_type.startswith(self.sirsi_library_prefix):
+            patrondata.block_reason = SirsiBlockReasons.INCORRECT_LOCATION
```
jonathangreen, line 186: I'm concerned with these being blocks, since blocks are enforced when a patron goes to actually checkout the content, not when they authenticate. Settings these as blocks will mean other library patrons will seem like they can log in, until they take an action in the library.

I'm also concerned because we have overrides for libraries that want to allow patrons to checkout content despite blocks, which would override these blocks getting set.

https://github.com/ThePalaceProject/circulation/blob/03fe8e5d02bce41d3d68147d29ff0d8bd30b61af/api/authenticator.py#L2095-L2197

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,332 @@
...
+            "SD-Working-LibraryID": self.sirsi_library_id,
+            "x-sirs-clientID": self.sirsi_client_id,
+        }
+        if session_token:
+            headers["x-sirs-sessionToken"] = session_token
+
+        url = urljoin(self.server_url, path)
+        # Adding a long timeout because /patronStatusInfo would fail often
+        return HTTP.request_with_timeout(
+            method, url, headers=headers, json=json, timeout=120
```
jonathangreen, line 264: 😟 that timeout. I think maybe we need to drop the timeout down... since a timeout this long will cause issues in production with other libraries.

jonathangreen (commented):
This is looking good @RishiDiwanTT

Enforcing the restrictions as blocks is the thing that concerns me the most.

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,332 @@
...
+            "required": True,
+        },
+        {
+            "key": Keys.LIBRARY_PREFIX,
+            "label": _("Library Prefix"),
+            "description": _(
+                "This is used to match a member of the Library to their member branch location."
+                " Should be a prefix letter, eg. 'p' or 'co' etc..."
+            ),
+            "required": True,
```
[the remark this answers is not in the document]
RishiDiwanTT: I have removed the required attribute, but this leaves no way to enforce patron-library restrictions for libraries that do not have a prefix defined.
This also stops the "disallowed suffixes" from being enforced.

api/sirsidynix_authentication_provider.py
RishiDiwanTT: The SD restrictions are a little different than what `_restriction_matches` allows (one positive, one negative), so I've implemented it differently for SD.

api/sirsidynix_authentication_provider.py
RishiDiwanTT: Moved the prefix check to the "enforce" method

RishiDiwanTT: We will always have the 12 hours gap in any restriction changes for the patron due to the needs_external_sync gap though

api/sirsidynix_authentication_provider.py
RishiDiwanTT: Removed the timeout. Today the API was snappy (kind of) but on other days I've had long waits for a response. Maybe production is better .

jonathangreen:
@RishiDiwanTT Re-reading Chris last email, and looking at his spreadsheet. I don't think this will meet the requirements he has there.

Here is how I understand the requirements:
- The prefix should determine if the patron can log in.
- The suffix should determine if the patron is blocked (if possible).

Does that match your understanding from his email as well? If so, I don't think you need to revert https://github.com/ThePalaceProject/circulation/commit/12b5f408ac007523fafdad12805613337885a71f, as that meets the first requirement. We could implement the suffix as a block, or just leave it not allowing the patron to log in.

jonathangreen:
Looking at his spreadsheet, the suffixes are per library as well, so they probably need to get moved into the library settings.

RishiDiwanTT:
> Looking at his spreadsheet, the suffixes are per library as well, so they probably need to get moved into the library settings.

You're right, I'm looking at all the possible patronTypes and there is no standardisation across the libraries... 
Like a Self-Registered Adult can be cadself (coalinga) or mzadself (Mariposa).. This is nightmare fuel...

RishiDiwanTT:
> @RishiDiwanTT Re-reading Chris last email, and looking at his spreadsheet. I don't think this will meet the requirements he has there.
> 
> Here is how I understand the requirements:
> 
> * The prefix should determine if the patron can log in.
> * The suffix should determine if the patron is blocked (if possible).
> 
> Does that match your understanding from his email as well? If so, I don't think you need to revert [12b5f40](https://github.com/ThePalaceProject/circulation/commit/12b5f408ac007523fafdad12805613337885a71f), as that meets the first requirement. We could implement the suffix as a block, or just leave it not allowing the patron to log in.

Made this change without `enforce_library_identifier`, during a remote patron lookup returning a problem detail serves the same purpose.

jonathangreen (commented):
@RishiDiwanTT these changes look good to me and I'd like to get this merged. I made a couple small comments that should be easy to resolve.

The only thing that is holding me back from approving this is that I'm a bit concerned about returning a `ProblemDetail` from `_remote_patron_lookup` instead of implementing `enforce_library_identifier`. 

You mentioned the two are equivalent, but I'm concerned that they are not. Can you check my reasoning here?

The main method that does authentication here is `BasicAuthenticationProvider.authenticate`
https://github.com/ThePalaceProject/circulation/blob/e84a8e7514c4683a18d16e2ecc78813384fe2e2d/api/authenticator.py#L2106-L2200

This method first calls `server_side_validation`, `remote_authenticate` and `enforce_library_identifier_restriction`. Then it checks if the patron `needs_external_sync`, if it doesn't then it returns the patron, otherwise it goes on to call `remote_patron_lookup`. 

What I'm worried about, by putting the identifier restriction in `remote_patron_lookup` is that we can successfully return a patron without ever getting to `remote_patron_lookup`, so we may not always enforce the library prefix.

RishiDiwanTT:
> What I'm worried about, by putting the identifier restriction in remote_patron_lookup is that we can successfully return a patron without ever getting to remote_patron_lookup, so we may not always enforce the library prefix.

This is right. However, we only get the "patronType" after we "read patron data" [L174](https://github.com/ThePalaceProject/circulation/blob/feature/sirsidynx/api/sirsidynix_authentication_provider.py#L174)
Semantically this is a patron lookup, hence I left it within the method. It also adds an API request to the authenticate workflow which will not be cached. The side effect, as you mentioned, will be that a blocked user will have a 12 hour margin for using the system after being blocked (or unblocked).

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,337 @@
...
+            username=username,
+            authorization_identifier=username,
+            permanent_id=data.get("patronKey"),
+            session_token=data.get("sessionToken"),
+            complete=False,
+        )
+
+    # Because of the odd way in which this method is written in AuthenticationProvider
+    # we have to exclude it from type checking, because mypy gets confused with the inheritance
+    def _remote_patron_lookup(  # type: ignore
```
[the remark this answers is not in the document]
RishiDiwanTT: Done :+1:

api/sirsidynix_authentication_provider.py
```diff
@@ -0,0 +1,337 @@
...
+        patrondata.external_type = patron_type
+
+        # Basic block reasons
+
+        # Does the patron type start with the library prefix
+        # This ensures the patron is a member of the library
+        # they are interacting with
+        if not patron_type.startswith(self.sirsi_library_prefix):
+            # We respond with a problem detail, stopping the upstream authentication immediately
+            return INVALID_CREDENTIALS
```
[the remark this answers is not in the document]
RishiDiwanTT: Done :+1:

jonathangreen (approved):

PR #606, opened 2022-10-21
Overdrive samples format fix

RishiDiwanTT:
## Description
Overdrive adds the formatType to each sample regardless of the parent format 
We must maintain the format provided by overdrive per sample

Additionally, when adding a link to an identifier we must update the representations media_type if required

## Motivation and Context
When attempting to view samples for some content in the Overdrive Integration Library a content protection error occurs, due to the incorrect media type.

[Notion](https://www.notion.so/lyrasis/Investigate-content-protection-issues-with-Overdrive-samples-9732715c05034bdaa5786a2d0342815d)

## How Has This Been Tested?
Manually ran the overdrive_format_sweep and saw different media type samples in the db now
Updated the unit tests accordingly

## Checklist


core/model/identifier.py
```diff
@@ -752,6 +752,24 @@ def add_link(
...
+                .filter(
+                    Representation.url == resource.url,
+                    Representation.media_type == media_type,
+                )
+                .count()
+            )
+            if not representation_exists:
+                # We have a representation that is not the same media type we previously knew of
+                resource.representation.media_type = media_type
+                resource.representation.url = resource.url
```
tdilauro, line 772: This should be fine for now, but could be a problem longer term. For example, what would happen if OD moved to a content negotiation approach, using the same URL but with differing media types to determine the HTTP response for differing sample formats.

If there's not a good way to handle this now, it would be helpful to put a TODO or other note here to acknowledge that possibility.

core/overdrive.py
```diff
@@ -1267,22 +1268,32 @@ def book_info_to_metadata(
...
-                            for sample_info in format["samples"]:
-                                href = sample_info["url"]
+                    for sample_info in format["samples"]:
+                        href = sample_info["url"]
+                        # Have we already parsed this sample? Overdrive repeats samples per format
+                        if href in sample_hrefs:
+                            continue
+
+                        # Every sample has its own format type
+                        overdrive_name = sample_info.get("formatType")
```
tdilauro, line 1278: `overdrive_format_name`? Not a big deal, since I think it's used only in this short block.

tests/core/files/overdrive/has_sample.json
```diff
@@ -1 +1,189 @@
...
+                },
+                {
+                    "source": "From the book",
+                    "formatType": "ebook-overdrive",
+                    "url": "https://samples.overdrive.com/?crid=9BD24F82-35C0-4E0A-B5E7-BCFED07835CF&.epub-sample.overdrive.com"
+                }
+            ]
+        }
+    ]
+}
```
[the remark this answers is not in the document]
RishiDiwanTT: I forced the newline, it seems the vscode json formatter doesn't agree

core/model/identifier.py
RishiDiwanTT: This depends on the replacement policy defined during metadata.apply()
https://github.com/ThePalaceProject/circulation/blob/main/core/metadata_layer.py#L1937
If it is set to replace, then its a complete replacement of all identifier links, so its a clean list every import.

core/overdrive.py
RishiDiwanTT: Better name though, updated

RishiDiwanTT:
@tdilauro bump

RishiDiwanTT:
@tdilauro bump

core/overdrive.py
```diff
@@ -1267,22 +1268,34 @@ def book_info_to_metadata(
...
+                            # Malformed sample
+                            continue
+                        internal_names = list(
+                            cls.internal_formats(overdrive_format_name)
+                        )
+                        if not internal_names:
+                            # Useless to us.
+                            continue
+
+                        for content_type, drm_scheme in internal_names:
```
jonathangreen, line 1289: It looks like we are throwing away the `drm_scheme` here. Is this is because all the samples don't have a `drm_scheme` set?

If `drm_scheme` isn't `None`, maybe we want to log a warning, and not create a representation so we don't end up with sample links that have DRM that we are not properly accounting for.

tests/core/files/overdrive/has_sample.json
```diff
@@ -1 +1,189 @@
...
+                {
+                    "type": "ExpirationRights",
+                    "value": 0
+                }
+            ],
+            "samples": [
+                {
+                    "source": "From the book",
+                    "formatType": "ebook-epub-adobe",
+                    "url": "http://excerpts.contentreserve.com/FormatType-410/1071-1/9BD/24F/82/BridesofConvenienceBundle9781426803697.epub"
```
jonathangreen, line 145: Hmmm... This sample, based on the `formatType` seems like it should be adobe protected... but the link to an epub doesn't look protected.

jonathangreen (commented):
This looks okay, I think. I'm a bit confused about the types of OD samples though. Do we know if they ever have DRM on them? We ignore the DRM part of the media type, which seems like it will fail if samples ever actually have DRM. 

Did we ever have discussion about DRM on these samples? If we know they won't have DRM can you add a comment that gives some detail why we ignore the `drm_type`?

jonathangreen (changes requested):
I had a convo in slack with Maurice about what we are expecting on mobile for previews. I don't think this PR matches that expectation yet. 

Slack thread here: 
https://lyrasis.slack.com/archives/CCY3PH8JD/p1671026666970849

Based on that convo the mobile apps will need some changes, after we have some changes to return the correct types from the CM, but we should get the CM returning correct types before we move forward on that mobile work.

From what I've seen digging through the overdrive feed, it looks like even though `formatType` is given for both samples, and the actual content, they have different meanings in the different contexts. We use the same lookup in these changes, just throwing away the DRM type. This gives us a somewhat reasonable value for epubs, but I don't think it gives us what we need for other types. 

So that the type tags we give for previews in our OPDS feed are semantically correct, we should be storing the actual content type of the preview. When I ran the code with this branch, I was getting types like `application/vnd.overdrive.circulation.api+json;profile=ebook` when we should have been returning `text/html` and types like `application/x-od-media` when we should be returning `audio/mpeg`.

jonathangreen:
Restarted the test that failed, it looks like `uwsgi` failed to run in the container. Hopefully will work this time, otherwise you may have to look into that failure @RishiDiwanTT.

core/overdrive.py
```diff
@@ -902,6 +902,18 @@ def link(self, page, rel):
         ),
     }
 
+    # A mapping of the overdrive format name to end sample content type
+    # Overdrive samples are not DRM protected so the links should be
+    # stored as the end sample content type
+    sample_format_to_content_type = {
+        "ebook-overdrive": "text/html",
+        "audiobook-wma": "application/octet-stream",
```
jonathangreen, line 910: It doesn't really matter, since the mobile apps can't play it anyway, but does wma here stand for [windows media audio](https://en.wikipedia.org/wiki/Windows_Media_Audio)? If it does, its probably better to use the media type listed on that wikipedia page `audio/x-ms-wma`.

RishiDiwanTT: Yes, that is what it represents. Will change the format.

PR #612, opened 2022-10-26
Update how we import DeMarque Audiobook Titles

jonathangreen:
## Description

Change how we are parsing the DRM formats for DeMarque Audiobook titles available through ODL. This was done after a discussion with DeMarque about why some titles were failing in the mobile apps. 

[Notion ticket](https://www.notion.so/lyrasis/Audible-content-not-working-in-app-e6a83717a0a84de19e8fc5666f08e963)

## Motivation and Context

Some titles (notably Audible, but there are others) are only available via LCP DRM, not through the streaming endpoint. This change makes sure that we correctly set the delivery mechanisms for these titles, so the mobile apps do not choose a delivery mechanism that is unavailable.

## How Has This Been Tested?

Running the included unit tests, which include feed samples for audiobooks that are and are not available for streaming.

## Checklist

### DIFF
--- a/api/odl2.py
+++ b/api/odl2.py
@@ -200,12 +200,7 @@ class ODL2Importer(OPDS2Importer, HasExternalIntegration):
                 if parsed_license is not None:
                     licenses.append(parsed_license)
 
-                # DPLA feed doesn't have information about a DRM protection used for audiobooks.
-                # We want to try to extract that information from the License Info Document it's present there.
                 license_formats = set(odl_license.metadata.formats)
-                if parsed_license and parsed_license.content_types:
-                    license_formats |= set(parsed_license.content_types)
-
                 for license_format in license_formats:
                     if (
                         skipped_license_formats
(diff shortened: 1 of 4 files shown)


io7m (approved):
Seems fine to me!

api/odl2.py
```diff
@@ -200,12 +200,7 @@ def _extract_publication_metadata(self, feed, publication, data_source_name):
                 if parsed_license is not None:
                     licenses.append(parsed_license)
 
-                # DPLA feed doesn't have information about a DRM protection used for audiobooks.
-                # We want to try to extract that information from the License Info Document it's present there.
                 license_formats = set(odl_license.metadata.formats)
-                if parsed_license and parsed_license.content_types:
-                    license_formats |= set(parsed_license.content_types)
```
tdilauro, line 207: Why can this go away? Was it vestigial/effectively a no-op? Or did I miss this moving somewhere else in another PR?

jonathangreen: Effectively we used to get `application/audiobook+json; protection=http://www.feedbooks.com/audiobooks/access-restriction` from the license status document, and the other types from the feed, but they are now all available in the feed. And hopefully only available if they are in fact available.

tdilauro: 🙏🏽

tdilauro (approved):
Progress! 🚀

jonathangreen:
The changes this PR takes advantage of are now live in the DeMarque system.

PR #623, opened 2022-11-03
Avoid exceptions for commas in currency values.

tdilauro:
## Description

- Fixes `MoneyUtility.parse` to avoid exceptions for commas in currency values.
- Refactors `MoneyUtility` tests and adds new test case with comma.

## Motivation and Context

`MoneyUtility.parse` raised an exception when parsing values with commas. This caused SIP2 patron authentication to fail under certain conditions. It could have an effect in other functionality, given the right conditions. [[Notion](https://www.notion.so/lyrasis/CA-Riverside-PL-patron-auth-self-test-fails-because-value-cannot-be-parsed-e6505183402d42ffb8483c62cdd93691)]

## How Has This Been Tested?

- Manual testing in development environment.
- New tests targeting the problem condition.
- CI tests. (TBD)

## Checklist

### DIFF
--- a/core/util/__init__.py
+++ b/core/util/__init__.py
@@ -1,5 +1,7 @@
 """Miscellaneous utilities"""
 
+from __future__ import annotations
+
 import re
 import string
 from collections import Counter
@@ -526,15 +528,21 @@ class MoneyUtility:
     DEFAULT_CURRENCY = "USD"
 
     @classmethod
-    def parse(cls, amount):
+    def parse(cls, amount: str | float | int | None) -> Money:
         """Attempt to turn a string into a Money object."""
         currency = cls.DEFAULT_CURRENCY
         if not amount:
             amount = "0"
         amount = str(amount)
+
         if amount[0] == "$":
             currency = "USD"
             amount = amount[1:]
+
+        # `Money` does not properly handle commas in the amount, so we strip
+        # them out of US dollar amounts, since it won't change the "value."
+        if currency == "USD":
+            amount = "".join(amount.split(","))
         return Money(amount, currency)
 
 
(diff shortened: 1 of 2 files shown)


core/util/__init__.py
```diff
@@ -526,15 +528,21 @@ class MoneyUtility:
...
         amount = str(amount)
+
         if amount[0] == "$":
             currency = "USD"
             amount = amount[1:]
+
+        # `Money` does not properly handle commas in the amount, so we strip
+        # them out of US dollar amounts, since it won't change the "value."
+        if currency == "USD":
+            amount = "".join(amount.split(","))
```
jonathangreen, line 545: This does the job but, I wonder though if we would be better off using something like the Babel number parser to create this value?

https://babel.pocoo.org/en/latest/api/numbers.html#babel.numbers.parse_decimal

Might handle some more of the edge cases here.

jonathangreen (approved):
One comment about an alternative approach, but can be merged in as is as well.

core/util/__init__.py
tdilauro: It  looked at it, but it seemed to be more trouble than it was worth. Its strict mode doesn't allow extra zeros after the decimal point (e.g., "4,444.4" is valid, but "4,444.40" is not) and we'd need to handle its exception or map it into a `ValueError`.

jonathangreen: Thats annoying! I thought Babel would have handled those cases better then that. I'll mark this as resolved then!

tdilauro: Yeah, it did seem kind of dumb, but that's why I gave it pass.

PR #629, opened 2022-11-08
Add new files to mypy type checking 

pcustic:
## Description
Adding new files to `mypy` type checking in the Circulation Manager. Ran `mypy` on the excluded files and resolved issues.

## Motivation and Context
This is part of the effort to have full type checking coverage in the CM:
https://www.notion.so/lyrasis/Get-Mypy-doing-type-checking-against-all-the-code-in-circulation-24f41995c1f74aec9963cb8b3d7b5238

## How Has This Been Tested?
Removed the files from the excluded list in the [pyproject.toml](https://github.com/ThePalaceProject/circulation/compare/feature/mypy-type-checking#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) and ran the `mypy`.

## Checklist


core/configuration/ignored_identifier.py
```diff
@@ -40,7 +40,7 @@ class IgnoredIdentifierConfiguration(ConfigurationTrait):
         format="narrow",
     )
 
-    def get_ignored_identifier_types(self) -> Set[str]:
+    def get_ignored_identifier_types(self) -> Any:
```
RishiDiwanTT, line 43: Why is this a more generic type than Set[int] now?

core/configuration/ignored_identifier.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, *args, **kargs) -> None:
 
     def _get_ignored_identifier_types(
         self, configuration: IgnoredIdentifierConfiguration
-    ) -> Set[int]:
+    ) -> Any:
```
RishiDiwanTT, line 90: Same question about less granular types

core/model/configuration.py
```diff
@@ -1154,7 +1159,11 @@ def __get__(
         return setting_value
 
     def __set__(
-        self, owner_instance: HasConfigurationSettings | None, value: Any
+        self,
+        owner_instance: HasConfigurationSettings
+        | IgnoredIdentifierConfiguration
```
RishiDiwanTT, line 1164: This is an interesting Union type, is IgnoreIdentifierConfiguration special, or should it also inherit from HasConfigurationSettings?

core/model/datasource.py
```diff
@@ -57,7 +57,7 @@ class DataSource(Base, HasSessionCache, DataSourceConstants):
     )
 
     # One DataSource can generate many Editions.
-    editions = relationship("Edition", backref="data_source")
+    editions = relationship("Edition", back_populates="data_source")
```
RishiDiwanTT, line 60: Should this also change to `uselist` too

RishiDiwanTT (commented):
Only some minor comments added.
Running mypy with 3.7 gives an additional typing issue in test_s3.py

core/model/datasource.py
pcustic: I suppose editions are not iterated over in the rest of the checked code so mypy is not complaining. I can add it if you think it makes sense for future use.

core/configuration/ignored_identifier.py
pcustic: This method returns `self.ignored_identifier_types or tuple()` so it definitely can't be `Set[str]` as it is now due to the empty tuple, whatever `self.ignored_identifier_types` is. 
If I understood it correctly, `self.ignored_identifier_types` is descriptor `ConfigurationMetadata` whose `__get__` has `Any` as a [return type](https://github.com/ThePalaceProject/circulation/pull/629/files#diff-5e9d4d0fb44a1a5f3015f053632b819c1202eac7bc445f2662135cb40f7f033dR1116) so I used Any from that return type.

core/configuration/ignored_identifier.py
pcustic: Pretty much the same as above since this is calling that method already mentioned.

core/model/configuration.py
pcustic: IgnoreIdentifierConfiguration is not inheriting from HasConfigurationSettings, instead it [inherits ConfigurationTrait](https://github.com/ThePalaceProject/circulation/blob/main/core/configuration/ignored_identifier.py#L15) (although that class doesn't do anything) so mypy complains here. 

```
core/configuration/ignored_identifier.py:75: error: Argument 1 to "__set__" of "ConfigurationMetadata" has incompatible type "IgnoredIdentifierConfiguration"; expected "Optional[HasConfigurationSettings]"
```

core/model/datasource.py
RishiDiwanTT: I think it does, as long as it doesn't break any tests.

core/model/configuration.py
RishiDiwanTT: Yea it looks like it should functionally be a `HasConfigurationSettings` but has the methods implemented with different names. This sounds like a needed change, but not in this ticket. We can keep it as is, for now.

core/configuration/ignored_identifier.py
RishiDiwanTT: Following the hierarchy of defined constants of the configuration it looks like each value in the `ALL_IDENTIFIER_TYPES` are strings.

pcustic: Yeah, this seems correct, thanks! Updated this and code from comment bellow in https://github.com/ThePalaceProject/circulation/pull/629/commits/0827f40d81efc2cbd485fed5a65ca09c36f116ec

RishiDiwanTT (approved):
Apart from the python 3.7 typing issue, this all seems good

pcustic:
> Apart from the python 3.7 typing issue, this all seems good

I'll check out that `test_s3.py` error that you mentioned before. I updated the `set[..` to `Set[..`,

pcustic:
@RishiDiwanTT Added `types-mock` package to clear that python 3.7 issue. Checked with `poetry run mypy --python-version=3.7` , seems fine for me.

PR #637, opened 2022-11-14
Feature/test fixtures

io7m:
## Description

This makes a number of changes to move to `pytest` test fixtures instead of using subclasses for unit tests. Currently, only the `core` test suite has been touched; the `api` tests still use the old test methods.

* All tests in the `core` test suite have been switched to using typed test fixtures. Tests must explicitly ask for fixtures; this aids maintenance because tests now explicitly state which fixtures they are affected by, and the use of type annotations makes the fixtures easy to locate in the IDE.
* Some tests that were using the `parameterized` library now used parameterized `pytest` marks. This appeared to be necessary to have parameterized tests work properly with injected fixtures.
* Many new fixtures have been developed, including but not limited to:
  *  `DatabaseFixture` encapsulates database startup/teardown
  *  `DatabaseTransactionFixture` encapsulates a database session similar to the old `DatabaseTest` class
  *  `TemporaryDirectoryConfigurationFixture` creates and configures a temporary directory for use in `Configuration.DATA_DIRECTORY`
  * `FilesFixture`, `CSVFilesFixture`, etc, provides access to sample files in tests. This makes tests largely position-independent rather than being dependent on test file locations.
  * `EndToEndSearchFixture` provides a strongly-typed repository for end-to-end search tests. This ensures that Elasticsearch is configured correctly, and provides a place for tests to store temporary search data without losing type safety.
  * `ExternalSearchPatchFixture` provides patching of the external search index for mocked search tests. The previous fixture was an autouse fixture that silently replaced class references without there being any indication to the developer that this was happening. Hilarity ensued.
  * `Time` provides a mock clock.
* Properties that have side effects now _look_ like they have side effects. For example, `transaction.fresh_url()` vs. `transaction._url`.

Current issues:

* The codebase has insufficient type annotations, currently. This causes a few hundred type errors that will likely already be fixed in a separate PR by @pcustic.
* One test had to be skipped:
```
    @pytest.mark.skip(
        reason="This test _may_ be incompatible with the use of a database fixture."
    )
    def test_cdns_loaded_dynamically(
        self, configuration_test_fixture: ConfigurationTestFixture
    ):
```
Thankfully, the test didn't really appear to be testing anything useful.
* There's quite a bit of repetitive code of the form:
`transaction, session = database_transaction, database_transaction.session()`
Unfortunately, I'm not sure how this can be reduced due to the way that `pytest` injects fixtures by name instead of by type (meaning long names are inevitable). Some use of subclassing in fixtures _might_ reduce some of this a bit.
_Edit: Quite a bit of this was eliminated through some renaming and pruning_
## Motivation and Context

The existing test suite has a number of problems that cause serious maintenance issues:

* Extensive use of subclassing everywhere means that developers have to read a lot of code in order to understand the behaviour of a single test.
* Use of `self` fields in test methods means that the behaviour of one test can affect the behaviour of another, because mutable state is carried between tests without being cleared/reset.
* Use of `autouse` fixtures means that, for any given test, developers must manually search for methods (that may or may not even exist) that _may_ be affecting the behaviour of a current test. Tests do not refer to these fixtures directly, and so changes in apparently unrelated parts of the test suite can break existing tests because they resulted in a change to a fixture that was being implicitly used in tests.
* Use of `autouse` fixtures means that tests become dependent on database initialization regardless of whether or not they use a database. This means that the database has to be configured in the IDE, even for running small unit tests, and it also means that the test suite spends a lot of time setting up and tearing down unused database instances.

## How Has This Been Tested?

The test suite was executed after each file of tests were converted to ensure that all tests pass and (importantly!) the same number of tests were executed.

## Checklist

- [x] All new and existing tests passed (except for one outdated test!)


tests/core/lcp/test_credential.py
```diff
@@ -34,27 +33,42 @@ def test_hashing(self):
...
+        data.patron = transaction.patron()
+        data.data_source = DataSource.lookup(
+            transaction.session(), DataSource.INTERNAL_PROCESSING, autocreate=True
         )
+        return data
+
+
+@pytest.fixture()
+def example_credential_fixture(
+    database_transaction: DatabaseTransactionFixture,
```
RishiDiwanTT, line 58: Purely from a legacy nomenclature perspective should we leave the session names as `db` or `_db` as it is in the application code

tests/core/lcp/test_credential.py
```diff
@@ -63,18 +77,23 @@ def test_getter(self, _, credential_type, method_name, expected_result):
...
-                self._db, self._data_source, credential_type, self._patron, None
+                session, data.data_source, credential_type, data.patron, None
             )
 
-    def test_get_patron_passphrase(self):
+    def test_get_patron_passphrase(
+        self, example_credential_fixture: ExampleCredentialFixture
+    ):
+        data = example_credential_fixture
+        session = data.transaction.session()
```
RishiDiwanTT, line 95: This is a little different than the previous method, can we add a little inline documentation in the Database[Transaction]Fixture about having to do `.session()` in the tests?

tests/core/lcp/test_credential.py
io7m: OK!

I hope the IDE can do this renaming. :crossed_fingers:

tests/core/lcp/test_credential.py
io7m: Yep, can do. The previous code worked in terms of `DatabaseTest`, which _kind of_ wrapped a database transaction, although did a lot of other things too. Some tests reached into the transaction and did things with the raw ORM session, and some did things with the transaction itself (although rarely).

RishiDiwanTT: Is there a case where a test using the DatabaseFixture will NOT use a session?

tests/core/models/test_collection.py
```diff
@@ -24,30 +24,50 @@
...
+from tests.fixtures.database import DatabaseTransactionFixture
 
 
-class TestCollection(DatabaseTest):
-    def setup_method(self):
-        super().setup_method()
-        self.collection = self._collection(
-            name="test collection", protocol=ExternalIntegration.OVERDRIVE
-        )
+class ExampleCollectionFixture:
```
RishiDiwanTT, line 32: A lot of these could have been dataclasses, no point changing this now since all the code is written now. So.. many.. lines

tests/fixtures/database.py
```diff
@@ -0,0 +1,1048 @@
...
+
+        # The CM uses SQLAlchemy's "declarative base" feature, which means that
+        # tables are only created if the classes containing references to them
+        # are actually loaded. The entire application database is therefore dependent on
+        # unspecified class initialization behaviour. The only safe way to guarantee
+        # that all tables are created is to refer to all of those classes before
+        # starting SQLAlchemy in order to ensure that all the classes are loaded
+        # in the interpreter.
+
+        tables = [
```
RishiDiwanTT, line 111: Would it be more future-proof to loop through the models directory and import all files? This way any new models are automatically picked up.
A special `import Lane` will have to be done too, unfortunately.

tests/fixtures/database.py
```diff
@@ -0,0 +1,1048 @@
...
+        # Also roll back any record of those changes in the
+        # Configuration instance.
+        for key in [
+            Configuration.SITE_CONFIGURATION_LAST_UPDATE,
+            Configuration.LAST_CHECKED_FOR_SITE_CONFIGURATION_UPDATE,
+        ]:
+            if key in Configuration.instance:
+                del Configuration.instance[key]
+
+    def database(self) -> DatabaseFixture:
```
RishiDiwanTT, line 258: database, transaction, session, default_collection and default_library are used as attributes, why not define them as `@property`'s

io7m: Don't `@property`s imply setters? These should be read-only.

tests/fixtures/database.py
io7m: I agree. In this case I did the dumbest and most explicit thing. I think the database initialization should be refactored so that neither the app or the test suite are dependent on class initialization. I made a quick attempt at doing this, but it quickly introduced circular dependencies, so I backed it out.

tests/fixtures/search.py
```diff
@@ -0,0 +1,293 @@
...
+        self.test_data = callback(self)
+
+        # Add all the works created in the setup to the search index.
+        SearchIndexCoverageProvider(
+            self.external_search.transaction.session(),
+            search_index_client=self.external_search.search,
+        ).run_once_and_update_timestamp()
+
+        # Sleep to give the index time to catch up.
+        time.sleep(2)
```
RishiDiwanTT, line 136: I always wondered why these tests took so long regardless of anything

tests/fixtures/search.py
```diff
@@ -0,0 +1,293 @@
...
+    """Note: You probably want EndToEndSearchFixture instead."""
+    data = ExternalSearchFixture.create(database_transaction)
+    yield data
+    data.close()
+
+
+T = TypeVar("T")
+
+
+class EndToEndSearchFixture(Generic[T]):
```
RishiDiwanTT, line 106: Why do we have this generic? I thought only `Work` was supported

tests/fixtures/search.py
RishiDiwanTT: Maybe we can add a speed up
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html

tests/fixtures/search.py
io7m: The way the previous code worked was that test classes extended an `EndToEndSearchTest` class, and defined a `populate_works()` method that assigned a completely test-specific set of attributes to `self`.

Replacing the `EndToEndSearchTest` class with a fixture meant that there could no longer be a `populate_works()` method that tests would implement, so we needed a different place for tests to store the attributes they needed. A first attempt at this produced a fixture that looked like this:

```
class EndToEndSearchFixture:
  def populate(f: Callable[EndToEndSearchFixture, ...]):
    f(self)
```

Tests would then call the `populate` method and assign fields to the fixture:

```
def _populate(data: EndToEndSearchFixture):
  data.something = "..."
  data.something_else = "..."
  data.something_else2 = "..."

def test_something(self, end_to_end_search_fixture: EndToEndSearchFixture):
  end_to_end_search_fixture.populate(self._populate_data);

  do_something_with_data(end_to_end_search_fixture.something, ...)
```

The problem with this is that it loses type-safety: The code in `test_something` would tell the fixture to call `self._populate_data`, and this would create new attributes on the fixture but the type checker would be completely unaware that the created attributes existed. The approach I then took was to have the `EndToEndSearchFixture` explicitly state the type of test-specific user data with type parameter `T`, and have the called `_populate_data` method be responsible for returning a value of type `T` that the tests can then use in a type-safe way.

It's not the way I would have written these tests from a blank canvas, but it was the least invasive way to convert the existing tests.

tests/fixtures/search.py
io7m: We should probably look into whether or not there's any way to have Elasticsearch actually tell us the index is ready.

io7m: I'm curious if the endpoint blocks until the indices are refreshed.

tests/core/lcp/test_credential.py
io7m: There are lots of cases where tests won't call `session()` directly, but they all use the session indirectly via the various methods on the fixture (like `lane()`, `work()`, etc).

tests/core/lcp/test_credential.py
io7m: Renamed the database transaction fixture to `db`, and added type annotations to test methods.

tests/fixtures/search.py
RishiDiwanTT: Yes it does, `Refresh requests are synchronous and do not return a response until the refresh operation completes.`

tests/fixtures/database.py
RishiDiwanTT: `@property` is a getter, whereas `@<property-name>.setter` is a setter
https://docs.python.org/3/library/functions.html#property

tests/fixtures/database.py
RishiDiwanTT: Yea, I tried something similar a while ago, self referential loops are all over the place, and an interpreted language is always going to rely on import order, don't think we can get around that.

tests/fixtures/search.py
io7m: I think I'm going to try implementing this, and removing that `sleep` call.

tests/fixtures/search.py
io7m: The `self.test_data` property is used, but it's only used inside the test methods. It's kind of an odd situation, because the tests are structured such that the fixture is initialized, and then has to have test-specific data assigned, and then data has to be indexed, and _then_ the test runs. In order to preserve type-safety between all of these steps, the fixture itself has to carry the type parameter.

RishiDiwanTT: Rather than accessing the works through the fixture, why not just have `self.data = TestExternalSearchWithWorksData()` in `TestExternalSearchWithWorks._populate_works`

tests/fixtures/search.py
io7m: This was straightforward to implement, but sadly it made no difference to execution times whatsoever. Presumably it takes exactly as long to `GET` as it does to sleep. Making the `GET` request is preferable though, as it presumably avoids a race condition.

tests/fixtures/search.py
io7m: This would yield the same kinds of mutable state problems that the fixtures are intended to prevent: Test methods have to be disciplined about not accidentally changing state in a way that could affect other test methods in the same class. Having the fixture carry the state guarantees this can't happen (because the fixture is always destroyed and recreated for each test method).

tests/fixtures/search.py
RishiDiwanTT: We can just use the `indices.refresh()` method.
https://elasticsearch-py.readthedocs.io/en/6.8.2/api.html#elasticsearch.client.IndicesClient.refresh

tests/fixtures/search.py
RishiDiwanTT: The test class is also destroyed and recreated between each test method. 
Now that I'm thinking about the lifecycle, does there need to be a callback function to create the works at all? 
The test method can just create the works up front within itself, and then ask the fixture to run the coverage provider indexing.
Since the coverage provider only cares about the works in the DB, the fixture doesn't need to know about what works have been created.

io7m: I was basing my approach on this comment made back in February:

https://github.com/ThePalaceProject/circulation/pull/178#discussion_r800983153

But I now realize that's referring only to class variables and not instance variables.

We could get rid of the callback if the tests only use instance variables.

io7m: @jonathangreen What do you think? ^^

tests/fixtures/database.py
io7m: I'm not sure I see the value of making them properties here. Could you elaborate a little?

I'm not opposed to doing it, it's just not clear to me what it gets us.

RishiDiwanTT: There is no functional change between the `@property` or current method implementation, it is only syntactic sugar where the test cases would be accessing `.session` as opposed to `.session()`

tests/fixtures/search.py
jonathangreen: I didn't do a very good job describing class variables vs instance variables in that comment did I 😅 

I think it makes sense for the tests to create the works as they see fit, then ask the fixture for indexing. Eliminating the callback seems like it would simplify things and provide better separation of concerns.

tests/fixtures/database.py
jonathangreen: I agree it might be nice to future proof this a bit with a loop to import the models, assuming it doesn't add too much complexity. 

@io7m I remember us discussing centralizing the database initialization, so we don't have to do it separately in tests vs prod code, and that causing a cascade of import loops. I forget though if we made a tech debt ticket to handle that refactoring or not. Do you remember? If not, can you put one into the queue, so that we come back around to it?

tests/fixtures/search.py
io7m: Thankfully, I know the difference, I just didn't realize that `pytest` would actually recreate class instances each time. I'll get rid of the callback.

tests/fixtures/search.py
jonathangreen: I didn't know about the `refresh()` method! That is really helpful. 

I've traced a flaky test failure in CI to this line before, but I wasn't sure a better way to make sure ES was caught up other then waiting longer 🕐. So I ended up just leaving it. It seems like it would be a win to just call refresh here if it works.

tests/fixtures/search.py
io7m: Turns out we got to go one better: None of the tests were dependent on using `self`, so the callback has been removed and everything has been reduced to local variables inside test methods.

tests/fixtures/database.py
io7m: There is a ticket somewhere... :hourglass:

tests/fixtures/search.py
io7m: I'll switch to `refresh()` instead of that `GET` call.

io7m: Actually, I'm going to hold off on using `refresh()` as we don't expose that as part of the CM code, and I'm trying to limit changes to the test suite only. It's an easy change to make later.

tests/fixtures/database.py
io7m: Replaced this with something that iterates over the modules with `importlib`.

tests/fixtures/database.py
```diff
@@ -97,69 +99,31 @@ def _get_database_connection() -> Tuple[Engine, Connection]:
...
+        path = Path(__file__)  # tests/fixtures/database.py
+        path = path.parent  # tests/fixtures
+        path = path.parent  # tests
+        path = path.parent  # .
+        path = os.path.join(path, "core", "model")
+
+        list_modules = os.listdir(path)
+        list_modules.remove("__init__.py")
+
+        for module_file in list_modules:
```
RishiDiwanTT, line 112: Do the modules need to be loaded on every request of the fixture or once during session? Since they're imported modules they probably stay imported after the first time

io7m: This particular fixture has session scope, so they should only be loaded once:

```
@pytest.fixture(scope="session")
def database(application: ApplicationFixture) -> Iterable[DatabaseFixture]:
```

tests/fixtures/database.py
io7m: Refactored most of these to properties, and that allowed for inlining some local variables.

io7m:
@jonathangreen @RishiDiwanTT OK to press the big green button on this one? Needs one approving review.

jonathangreen:
@io7m I took a look at this yesterday, and I have a concern with it. Looking at the test that is marked as skipped `TestConfiguration.test_cdns_loaded_dynamically`. When I take out the skip marker to see why it fails, it seems like it works when that test is run in isolation, but if its run with the other tests, it fails because its seeing the CDN from the previous test when doing the first assert `assert {} == cdns`. 

That concerned me a bit, since it seems to indicate some state is being transferred between tests and that wasn't happening before. I started debugging it, but I didn't get very far with it. Can you take a look at this and see whats going on there? The test itself doesn't seem that useful, but its not a good sign that state is being transferred between the tests.

jonathangreen:
Some more context about how I ran the tests:

- Remove the `@pytest.mark.skip` decorator on line 88
- Run the tests like this: `tox -e "py38-core-docker" -- -k "test_cdns_loaded_dynamically" tests/core/test_config.py`
  - Test passes 🎉 
- Run the tests like this: `tox -e "py38-core-docker" -- tests/core/test_config.py`
  - The `test_cdns_loaded_dynamically` fails, which is a bit suspicious 🤔

jonathangreen (approved):

io7m:
There has been a great disturbance in the force...

PR #645, opened 2022-11-17
Use zoned UTC when comparing search doc and edition publication dates.

tdilauro:
## Description

Fixes an [assertion in `tests/core/models/test_work.py::TestWork::test_to_search_document`](https://github.com/ThePalaceProject/circulation/blob/b30cf5a2803403095918302a685bd76fd505dc3a/tests/core/models/test_work.py#L1148), so that both sides of the comparison are based on UTC-zoned `datetime`s.

## Motivation and Context

Test was failing depending on the timezone and time configured for the test environment.

## How Has This Been Tested?

- Manually ran tests with different time zones configured in the test environment.
  - TZ = UTC: Test always passed -- both before and after the change.
  - EST: Before the fix, test always failed.
  - EST: After the fix, test always passed.
- CI tests.

## Checklist

### DIFF
--- a/tests/core/models/test_work.py
+++ b/tests/core/models/test_work.py
@@ -1145,7 +1145,12 @@ class TestWork(DatabaseTest):
         assert edition.publisher == search_doc["publisher"]
         assert edition.imprint == search_doc["imprint"]
         assert edition.permanent_work_id == search_doc["permanent_work_id"]
-        assert edition.published == datetime.date.fromtimestamp(search_doc["published"])
+        assert (
+            edition.published
+            == datetime.datetime.fromtimestamp(
+                search_doc["published"], tz=pytz.UTC
+            ).date()
+        )
         assert "Nonfiction" == search_doc["fiction"]
         assert "YoungAdult" == search_doc["audience"]
         assert work.summary_text == search_doc["summary"]


io7m:
This looks good. I'm curious why `assert_time_match` apparently isn't vulnerable to this problem. 

I might move these time/date comparisons into a module as part of the test suite refactoring, as there are a lot of comparisons all over the test suite and it'd be good to only have this kind of code once.

tdilauro:
`assert_time_match` is using a tz-aware, UTC `datetime` provided by our own `from_timestamp` helper method, so won't have the same problem.

tdilauro:
@RishiDiwanTT Requesting your review to make sure that I'm not breaking/failing to test any assumptions from this test's origin PR (#453).

RishiDiwanTT (approved):
Works for me too... I wonder why it started failing so much time after the merge though

tests/core/models/test_work.py
```diff
@@ -1145,7 +1145,12 @@ def assert_time_match(python, postgres):
         assert edition.publisher == search_doc["publisher"]
         assert edition.imprint == search_doc["imprint"]
         assert edition.permanent_work_id == search_doc["permanent_work_id"]
-        assert edition.published == datetime.date.fromtimestamp(search_doc["published"])
+        assert (
+            edition.published
+            == datetime.datetime.fromtimestamp(
+                search_doc["published"], tz=pytz.UTC
```
pcustic, line 1151: There already existsts `from_timestamp` helper function implemented [here](https://github.com/ThePalaceProject/circulation/blob/main/core/util/datetime_helpers.py#L20-L25) that does the same thing and it is [already imported](https://github.com/ThePalaceProject/circulation/blob/main/tests/core/models/test_work.py#L23) in this test file so maybe it is better to use that?

pcustic: @tdilauro Oh, I see you already mentioned `from_timestamp` to Mark so you are aware of `from_timestamp`, is there a reason you didn't use that?

PR #648, opened 2022-11-22
Fix s3 analytics none licensepool issue

dbernstein:
## Description
NEW_PATRON events occur when a user tries to check out a book for the first time.  Since these events are not associated with a collection,  this PR ensures, that the license pool that is passed down to the S3 analytics provider can be None type.
Additionally, the s3 path been changed to make purging older event data more straightforward.

## Motivation and Context

Events were not making it to S3 when the license pool for a particular event happens to be None.  The only case where this can happen is in the firing of NEW_PATRON events which are not associated with a work.   Now S3 will get NEW_PATRON events as well.  Additionally,  it is preferable that the json files containing event information are organized by year, month, and date to facilitate regular purging for already processed events.   We will likely want to keep them around in case we need to rerun the crawler/ETL process, but making it easy to script a periodic purging would be beneficial down the road.

## How Has This Been Tested?
Unit tests cover the code already.

## Checklist

### DIFF
--- a/api/s3_analytics_provider.py
+++ b/api/s3_analytics_provider.py
@@ -157,7 +157,7 @@ class S3AnalyticsProvider(LocalAnalyticsProvider):
             "series": work.series if work else None,
             "series_position": work.series_position if work else None,
             "language": work.language if work else None,
-            "open_access": license_pool.open_access,
+            "open_access": license_pool.open_access if license_pool else None,
         }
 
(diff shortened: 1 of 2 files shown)


io7m:
There seem to be some failing tests.

jonathangreen (approved):
Tests are passing now. Looks good.

PR #654, opened 2022-11-30
Open access fulfillment links

RishiDiwanTT:
## Description
Library based acquisition feeds will only send open-access fulfillment links if the library has no auth

Otherwise it will expect the patron to go through the regular borrow workflow

## Motivation and Context
The CM acts differently depending if it has patron authentication enabled or not. If a CM is configured without Patron authentication (Like the Palace Bookshelf CM) then users can borrow open access content without authentication. However if Patron authentication is enabled, then we require patrons to checkout the content, so we can keep the correct bookshelf entries for them. 

[Notion](https://www.notion.so/lyrasis/Investigate-why-the-CM-is-sending-an-unfulfillable-fulfillment-acquisition-link-in-the-results-of-an-5f391fa052a745cebd5ba9b81125fe20)

## How Has This Been Tested?
Manually tested the catalog and search feed, with and without patron authentication.
Updated the unit tests to cover the use case

## Checklist

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -253,7 +253,15 @@ class CirculationManagerAnnotator(Annotator):
         ).prioritize_for_pool(licensepool)
 
     def annotate_work_entry(
-        self, work, active_license_pool, edition, identifier, feed, entry, updated=None
+        self,
+        work,
+        active_license_pool,
+        edition,
+        identifier,
+        feed,
+        entry,
+        updated=None,
+        add_open_access_links=True,
     ):
         # If ElasticSearch included a more accurate last_update_time,
         # use it instead of Work.last_update_time
@@ -305,6 +313,7 @@ class CirculationManagerAnnotator(Annotator):
         can_revoke_hold=True,
         set_mechanism_at_borrow=False,
         direct_fulfillment_delivery_mechanisms=[],
+        add_open_access_links=True,
     ):
         """Generate a number of <link> tags that enumerate all acquisition
         methods.
@@ -452,7 +461,11 @@ class CirculationManagerAnnotator(Annotator):
 
         # If this is an open-access book, add an open-access link for
         # every delivery mechanism with an associated resource.
-        if active_license_pool and active_license_pool.open_access:
+        if (
+            add_open_access_links
+            and active_license_pool
+            and active_license_pool.open_access
+        ):
             for lpdm in active_license_pool.delivery_mechanisms:
                 if lpdm.resource:
                     open_access_links.append(
@@ -627,6 +640,15 @@ class LibraryAnnotator(CirculationManagerAnnotator):
         self._top_level_title = top_level_title
         self.identifies_patrons = library_identifies_patrons
         self.facets = facets or None
+
+        # Open access links are added only when the library needs no authentication,
+        # else we must not have direct fulfillment links in the feed.
+        # The Patron should go through the authenticated borrow workflow in that case.
+        _db = Session.object_session(self.library)
+        integrations = ExternalIntegration.for_library_and_goal(
+            _db, self.library, ExternalIntegration.PATRON_AUTH_GOAL
+        )
+        self.add_open_access_links = integrations.count() == 0
 
     @classmethod
     def _hidden_content_types(self, library):
@@ -1165,6 +1187,7 @@ class LibraryAnnotator(CirculationManagerAnnotator):
             ),
             set_mechanism_at_borrow=set_mechanism_at_borrow,
             direct_fulfillment_delivery_mechanisms=direct_fulfillment_delivery_mechanisms,
+            add_open_access_links=self.add_open_access_links,
         )
 
     def revoke_link(self, active_license_pool, active_loan, active_hold):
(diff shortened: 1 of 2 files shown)


api/opds.py
```diff
@@ -628,6 +641,15 @@ def __init__(
...
         self.facets = facets or None
 
+        # Open access links are added only when the library needs no authentication,
+        # else we must not have direct fulfillment links in the feed.
+        # The Patron should go through the authenticated borrow workflow in that case.
+        _db = Session.object_session(self.library)
+        integrations = ExternalIntegration.for_library_and_goal(
+            _db, self.library, ExternalIntegration.PATRON_AUTH_GOAL
+        )
+        self.add_open_access_links = integrations.count() == 0
```
RishiDiwanTT, line 651: Should this also depend on `api.can_fulfill_without_loan` ?

io7m:
Sorry, the comment I left two days ago was sitting there in _Pending_ mode without have been submitted. :roll_eyes:

api/opds.py
RishiDiwanTT: Hmm... Thanks for pointing out locations in code. After looking at them it seems only `OPDSForDistributorsAPI` implements this method, whereas other datasource APIs simply `return False` (from the Base API).
That would defeat the purpose here since we're looking at open-access links from all sources. I won't use it here.

io7m: Not sure. The `circulation` implementation in particular would seem to return `True` here:

```
def can_fulfill_without_loan(self, patron, pool, lpdm):
        """Can we deliver the given book in the given format to the given
        patron, even though the patron has no active loan for that
        book?

        In general this is not possible, but there are some
        exceptions, managed in subclasses of BaseCirculationAPI.

        :param patron: A Patron. This is probably None, indicating
            that someone is trying to fulfill a book without identifying
            themselves.

        :param delivery_mechanism: The LicensePoolDeliveryMechanism
            representing a format for a specific title.
        """
        if not lpdm or not pool:
            return False
        if pool.open_access:
            return True
        api = self.api_for_license_pool(pool)
        if not api:
            return False
        return api.can_fulfill_without_loan(patron, pool, lpdm)
```

Should there be an open access pool available?

RishiDiwanTT: Oh :sweat_smile: In which case it's redundant since we only use it for open access pools :laughing:

api/opds.py
```diff
@@ -253,7 +253,15 @@ def visible_delivery_mechanisms(
...
-        self, work, active_license_pool, edition, identifier, feed, entry, updated=None
+        self,
+        work,
+        active_license_pool,
+        edition,
+        identifier,
+        feed,
+        entry,
+        updated=None,
+        add_open_access_links=True,
```
jonathangreen, line 264: Is the new `add_open_access_links` parameter used anywhere in this function?

api/opds.py
jonathangreen: @RishiDiwanTT instead of querying for integration here and creating a new variable for it or using `can_fulfill_without_loan`, I think we can use `self.identifies_patrons`. If we identify patrons, we shouldn't include open access acquisition links in the feed.

api/opds.py
```diff
@@ -452,7 +461,11 @@ def acquisition_links(
 
         # If this is an open-access book, add an open-access link for
         # every delivery mechanism with an associated resource.
```
jonathangreen, line 463: Comment should get an update to reflect the additional condition.

api/opds.py
RishiDiwanTT: This variable depends on the `IDENTIFIES_INDIVIDUALS` class variable [[source](https://github.com/ThePalaceProject/circulation/blob/main/api/authenticator.py#L727)]. It looks like all AuthenticationProviders have `IDENTIFY_INDIVIDUALS=True` due to the [base class](https://github.com/ThePalaceProject/circulation/blob/main/api/authenticator.py#L1302).
Would it be equivalent to not having authentication (in this case only) if an authenticator is introduced that does not identify individuals? It seems this might make sense since there probably would be no "login" mechanism to the authenticator.

jonathangreen: @RishiDiwanTT The code is a bit of a maze, but I think (`LibraryAnnotator.library_identifies_patrons`)[https://github.com/ThePalaceProject/circulation/blob/3894035d6231685128cff97f9163d8fff2786cf9/api/opds.py#L608] is set when the class is instantiated. 

`LibraryAnnotator` gets instantiated by `CirculationManager` which sets it based on `library_identifies_patrons`:
https://github.com/ThePalaceProject/circulation/blob/1c565ad26555dec72cff819149c3004443e05a47/api/controller.py#L578-L581

So using `self.library_identifies_patrons` here, should cover two cases: 
- We are using an authentication method that doesn't identify individuals (which there are none)
- No authentication is configured

It seems like using this existing variable is a better choice then doing a new query here.

RishiDiwanTT: Sounds good, will make the change with the assumption IDENTIFY_INDIVIDUALS will not be False anytime soon.

jonathangreen: I think if `IDENTIFY_INDIVIDUALS` does become false in the future, this will give us the behaviour we want.

api/opds.py
RishiDiwanTT: Removed

jonathangreen (approved):
Looks good

PR #657, opened 2022-11-30
Ensures that the "Children and Middle Grade" lanes and sublanes expli…

dbernstein:
…citly have an explicit target age range set.

Resolves: https://www.notion.so/lyrasis/Adult-titles-showing-in-Children-and-Middle-Grades-lane-in-Palace-app-St-Mary-s-County-Library-f2914ecfd97c42a4a7554cf08f995d5e

## Motivation and Context
If a lane has specified a "Children" audience and has no specified age range, then a book with audience of "Children" will appear in the lane even if the specified age range on the book is 18+.

However, setting the lane's "target age" will automatically add the appropriate audiences and should thus provide a double check that both the audience and target age matches.  

In the case of the MD libraries, this fix should resolve the issue.

## How Has This Been Tested?

I added a test to ensure that the target_age is being set appropriately on the Children and Middle Grade lane.

## Checklist

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -589,7 +589,9 @@ def create_lanes_for_large_collection(_db, library, languages, priority=0):
     ya_nonfiction_priority += 1
 
     children_common_args = dict(common_args)
-    children_common_args["audiences"] = CHILDREN
+    children_common_args["target_age"] = Classifier.range_tuple(
+        0, Classifier.YOUNG_ADULT_AGE_CUTOFF - 1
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
Looks good. Will existing lanes need to be recreated for this change to be visible on existing CM?

dbernstein:
@jonathangreen : Yes I believe you're right.  We will have to regenerated the affected lanes.

PR #663, opened 2022-12-02
Adds a script that pulls Overdrive Advantage account info and formats…

dbernstein:
… it into a CSV

## Checklist


jonathangreen:
@dbernstein This looks good and will be helpful to have. 

So we have some test coverage here, can you add a test or two that covers the new `GenerateOverdriveAdvantageAccountList` class?

dbernstein:
Yes - will do.

dbernstein:
Although I'm seeing this as a stopgap until we automate the setup of overdrive advantage accounts.  So I don't expect this code to last more than a few months.

jonathangreen:
It would be nice to have a couple tests here anyway. Stopgap commits have a way of sticking around longer then expected sometimes 😄

dbernstein:
I added a field to the output file that specifies whether or not the account has already been configured in order to help the I-Team zero in on what needs to be done.

pcustic:
The PR template is not really filled out, it would be great if we would have some more info along with the code.

dbernstein:
@pcustic :  the PR template is filled out now.

jonathangreen (approved):

PR #676, opened 2022-12-09
Convert PNGs to a valid format before saving if required

RishiDiwanTT:
## Description
Eg. CMYK does not save as a PNG via PIL
Had to use an internal variable `_OUTMODES` from PIL since it's documentation and API aren't very expansive.

## Motivation and Context
Some JPEG images may be in the CMYK format, so converting to PNG would fail during the upload workflow.
[Notion](https://www.notion.so/lyrasis/Collection-Manager-Error-when-attempting-to-save-a-library-logo-77f0aa914a174ef0a64e8434bc98b335)
[Related Notion](https://www.notion.so/lyrasis/Investigate-why-logos-are-not-automatically-resizing-in-the-Collection-Manager-108b38e17f074f52851bb5dbcd67111a?d=91f02f8935094b01a825cb7589b21b8b)

## How Has This Been Tested?
Manually uploaded the example image provided.
An additional test case was written for CMYK type images

## Checklist

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -9,6 +9,7 @@ import wcag_contrast_ratio
 from flask import Response
 from flask_babel import lazy_gettext as _
 from PIL import Image
+from PIL.PngImagePlugin import _OUTMODES as PNG_MODES
 
 from api.admin.announcement_list_validator import AnnouncementListValidator
 from api.admin.geographic_validator import GeographicValidator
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
@RishiDiwanTT This looks good to me. 

I pushed a commit with a suggestion. It would be nice to avoid importing the PIL internal variable, perhaps we can just always convert if the image mode is not "RGB" or "RGBA".

PR #677, opened 2022-12-09
Case-insensitive search for all keyword terms

RishiDiwanTT:
## Description
All keyword type search properties are now case-insensitive, and have ascii folding
Added audience and language to the keyword type properties

Since this is a change in current mapping properties, the index version had to be moved from v4 to v5.
There is no way to move the `-current` alias though, so it won't be moved and there will be a failing test for this.
[This test will fail](https://github.com/ThePalaceProject/circulation/blob/d402b736b045ebbeb1b2a95b27fe28e500383e03/tests/core/test_external_search.py#L206). 
I don't quite understand the motivation behind the test, it mentions the alias will never switch once created, and then tests for the alias anyways, so an updated version will always fail, in the test and in a deployed environment,
The [transfer current alias](https://github.com/ThePalaceProject/circulation/blob/d402b736b045ebbeb1b2a95b27fe28e500383e03/tests/core/test_external_search.py#L206) method is never used, and does what is required.
EDIT: We now do a "delete alias" before running the problematic test, The comment and assert statements still don't match though

## Motivation and Context
Needed case-insensitivity for all keyword type properties.
Only some properties had the `filterable-strings` normalizer.
[Notion](https://www.notion.so/lyrasis/Remove-case-sensitivity-from-list-searches-b89aef467ed34296aaf0dd950a39d7ed)

## How Has This Been Tested?
Manually re-indexed and tested case-insensitive terms
Updated the test cases to include case-insensitivity

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -862,6 +862,10 @@ class MappingDocument:
             "normalizer": "filterable_string",
         }
 
+    def keyword_property_hook(self, description):
+        """Hook method to ensure the keyword type attributes are case-insensitive"""
+        description["normalizer"] = "filterable_string"
+
 
 class Mapping(MappingDocument):
     """A class that defines the mapping for a particular version of the search index.
@@ -956,7 +960,7 @@ class CurrentMapping(Mapping):
     * contributors -- these Contributors worked on the Work
     """
 
-    VERSION_NAME = "v4"
+    VERSION_NAME = "v5"
 
     # Use regular expressions to normalized values in sortable fields.
     # These regexes are applied in order; that way "H. G. Wells"
@@ -1135,6 +1139,7 @@ class CurrentMapping(Mapping):
             "sort_author_keyword": ["sort_author"],
             "integer": ["series_position", "work_id"],
             "long": ["last_update_time", "published"],
+            "keyword": ["audience", "language"],
         }
         self.add_properties(fields_by_type)
 
@@ -1982,7 +1987,7 @@ class JSONQuery(Query):
 
     # The fields mappings in the search DB
     FIELD_MAPPING: Dict[str, Dict] = {
-        "audience": _KEYWORD_ONLY,
+        "audience": dict(),
         "author": _KEYWORD_ONLY,
         "classifications.scheme": _KEYWORD_ONLY,
         "classifications.term": _KEYWORD_ONLY,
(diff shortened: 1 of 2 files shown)


RishiDiwanTT:
Regarding the failing test `tests/core/test_external_search.py::TestFeaturedFacets::test_run`.
It would seem the test is itself incorrectly written, the random score is not reproducible across different index names (in this case v4 -> v5) 
`a salt that is computed based on the index name and shard id ` [[source](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-random)]

I need to rewrite this test, but I'm not sure what the last `assert` of the function is meant to test. The list will be reproducably random for every different index name, should we be aiming to reset the "expected" list everytime we change the index name? One would also expect a shard id to change across tests, especially in the CI workflow where the ES cluster is spun anew everytime.

jonathangreen (approved):
I pushed a couple new commits to this, that you should review @RishiDiwanTT. 

One commit just reverts https://github.com/ThePalaceProject/circulation/pull/677/commits/5da6a356ed3101bf35c2f3f2bd9ea20ac96f21cc. I was wondering if maybe this was an artifact from a test failing in your environment? The tests seem to pass okay for me locally, and in CI without the changes introduced there. 

https://github.com/ThePalaceProject/circulation/pull/677/commits/eaa4e73e348c601cbe15e3aef28b7e8085dcc72f the other just clarifies some comments, and tries to make the test that includes the version number a little less brittle.

RishiDiwanTT:
> I pushed a couple new commits to this, that you should review @RishiDiwanTT.
> 
> One commit just reverts [5da6a35](https://github.com/ThePalaceProject/circulation/commit/5da6a356ed3101bf35c2f3f2bd9ea20ac96f21cc). I was wondering if maybe this was an artifact from a test failing in your environment? The tests seem to pass okay for me locally, and in CI without the changes introduced there.
> 
> [eaa4e73](https://github.com/ThePalaceProject/circulation/commit/eaa4e73e348c601cbe15e3aef28b7e8085dcc72f) the other just clarifies some comments, and tries to make the test that includes the version number a little less brittle.

The test should always pass in a tox env since the docker is spun down at the end of it.
Running a test locally via pytest, which I do when iteratively making changes will cause the test to fail if we don't delete the alias first.
Since that's probably only something I do we can skip it for now.

jonathangreen:
@RishiDiwanTT do you still see the test failing locally? I also run the test outside of docker iteratively when developing. But for me the text fixture cleans up the ES indexes. That's why I was wondering if this was a test failure that didn't clean up the index that caused your issue.

PR #685, opened 2022-12-14
API for unsharing a shared customlist

RishiDiwanTT:
## Description
Unsharing is only allowed if the shared list is not in use by any other library 
To unshare do an HTTP DELETE on the share endpoint

## Motivation and Context
Currently, there is no way to unshare a list once it has been shared. This can be problematic if a library administrator accidentally shares a list. 
[Notion](https://www.notion.so/lyrasis/Allow-users-to-unshare-a-list-in-the-Collection-Manager-ddda10ea6dcd4ea699dc90205686e0ca)

This PR also has changes for generating automatic documentation through in-code annotations and request-response models.
[Pydantic](https://docs.pydantic.dev/) and [Flask-Pydantic-Spec](https://github.com/turner-townsend/flask-pydantic-spec) were used to achieve the model setup and autogeneration respectively.
The API swagger can be seen on a local setup on the `http://localhost:6500/apidoc/swagger` endpoint

## How Has This Been Tested?
Test cases have been update to test unshares

## Checklist

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -12,6 +12,7 @@ import flask
 import jwt
 from flask import Response, redirect
 from flask_babel import lazy_gettext as _
+from pydantic import BaseModel
 from sqlalchemy.sql import func
 from sqlalchemy.sql.expression import and_, desc, distinct, join, nullslast, select
 
@@ -781,6 +782,10 @@ class FeedController(AdminCirculationManagerController):
 
 
 class CustomListsController(AdminCirculationManagerController):
+    class CustomListSharePostResponse(BaseModel):
+        successes: int = 0
+        failures: int = 0
+
     def _list_as_json(self, list: CustomList, is_owner=True) -> Dict:
         """Transform a CustomList object into a response ready dict"""
         collections = []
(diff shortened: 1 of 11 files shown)


RishiDiwanTT:
Moved this PR to `Draft` for the initial implementation of Pydantic based documentation

api/admin/controller/__init__.py
```diff
@@ -777,6 +778,21 @@ def genres(self):
 
 
 class CustomListsController(AdminCirculationManagerController):
+    class CustomListSharePostResponse(BaseModel):
+        successes: int = 0
+        failures: int = 0
+
+    class CustomListPostRequest(BaseModel):
+        name: str
+        id: int = None
```
io7m, line 787: I think the type should be `Optional[int]`.

RishiDiwanTT: Pydantic, oddly, does not seem to cover this combination of Typed variable with a None value. It just... works as None, no problem.
I'll add the `Optional` to remove ambiguity.

RishiDiwanTT:
@io7m bump

io7m (approved):
Looks good!

PR #695, opened 2022-12-19
Feature/remove python 3.7 support

tdilauro:
## Description

Remove support for Python 3.7. Specifically:
- Removes Python 3.7 from CI.
- Updates `pyupgrade` pre-commit configuration to support Python 3.8+.
- Removes dependency on standalone `mock` that was needed for 3.7.
- Updates README, including some old stuff we've missed on previous Python version deprecations.

## Motivation and Context

While Python 3.7 has not yet reached its end of life (soon, though; very soon) it is preventing us using some very useful Python features (e.g., PEP 544 Protocols).

## How Has This Been Tested?

- Locally ran `core` and `api` test suites for Python 3.8 and 3.11.
- GH Actions CI will run.

## Checklist

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
         module: [Api, Core]
 
     # We want to run on external PRs, but not on our own internal PRs as they'll be run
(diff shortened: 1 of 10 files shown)


jonathangreen:
One thing I was concerned about is what version of Python we have in our docker containers, looks like we've got Python 3.8.10 running there, so we should be good to drop Python 3.7. Will be nice to be able to get rid of the `mock` imports.

jonathangreen (approved):

tdilauro:
Thanks! Yeah, I should've noted that I checked on the version in our containers before moving forward with this work.

PR #702, opened 2022-12-20
Fuzzy search for json type queries, eq and neq operators

RishiDiwanTT:
## Description
Fuzzy search will only take place for text type properties.
The `contains` type operator cannot make use of this since regex's are not well suited for this kind of query, and using a `match` query for the `contains operator` turned out to be far too promiscuous with the results.
This will be achieved by replacing Term queries with Match queries for fields that have text values.

## Motivation and Context
The current list search functionality is overly restrictive, and does not allow for slight misspellings, missing spaces, etc. (see related ticket). We need to add some fuzziness to the search fields to allow for more flexibility.
[Notion](https://www.notion.so/lyrasis/Add-fuzziness-to-list-search-queries-674c63ef3a974f59a659484519772cee)

## Checklist

### DIFF
--- a/tests/core/test_external_search.py
+++ b/tests/core/test_external_search.py
@@ -5092,18 +5092,30 @@ class TestJSONQuery:
     def _jq(query):
         return JSONQuery(dict(query=query))
 
+    match_args = dict(
+        auto_generate_synonyms_phrase_query=False,
+        max_expansions=10,
+        fuzziness="AUTO",
+    )
+
     def test_elasticsearch_query(self, external_search_fixture: ExternalSearchFixture):
         q = {"key": "medium", "value": "Book"}
         q = self._jq(q)
-        q.elasticsearch_query.to_dict() == {"term": {"medium.keyword": "Book"}}
+        q.elasticsearch_query.to_dict() == {
+            "match": {"medium.keyword": {"query": "Book", **self.match_args}}
+        }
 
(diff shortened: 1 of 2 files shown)


tests/core/test_external_search.py
```diff
@@ -5092,18 +5092,30 @@ def _leaf(key, value, op="eq"):
     def _jq(query):
         return JSONQuery(dict(query=query))
 
+    match_args = dict(
+        auto_generate_synonyms_phrase_query=False,
+        max_expansions=10,
+        fuzziness="AUTO",
+    )
```
io7m, line 5099: Might be worth replacing the arguments to this with constants that are defined somewhere and also used at the same call in [core/external_search.py](https://github.com/ThePalaceProject/circulation/pull/702/files#diff-bd35fcc1f1282a6accc97a70b10cf52835dc43297097ea923569bfd4d0a1b13f).

You know someone will update the parameters in one place and forget the other, some day. :slightly_smiling_face:

RishiDiwanTT: Sounds about right, both places now use the same source.

RishiDiwanTT:
@io7m Bump

io7m (approved):
Looks good to me!

tdilauro:
One more administrative comment: 
- The PR description doesn't really say much about what this PR is doing.
- There's a small typo in the `Motivation` section: "...and does not all for..." -> "...and does not allow for..."

PR #703, opened 2022-12-20
Fix duplicate Bibliotheca test classes.

tdilauro:
## Description

Updates Bibliotheca `ErrorParser` tests in order to:
- Avoid a duplicate test class name.
- Make tests a little more uniform and manageable.

## Motivation and Context

The `TestErrorParser` class was duplicated, which meant that only the tests in the last defined of the duplicates were actually being run. [[Notion](https://www.notion.so/lyrasis/Duplicate-TestErrorParser-class-causing-results-in-some-tests-not-being-run-47810343361944deb293f3dfad530a9c)]

## How Has This Been Tested?

- Manually ran affected tests locally.
- CI tests will run.

## Checklist

### DIFF
--- a/tests/api/test_bibliotheca.py
+++ b/tests/api/test_bibliotheca.py
@@ -1,8 +1,11 @@
+from __future__ import annotations
+
 import json
 import os
 import random
 from datetime import datetime, timedelta
 from io import BytesIO, StringIO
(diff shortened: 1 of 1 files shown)


jonathangreen (approved):
🚀

PR #727, opened 2023-01-03
Fix search test causing issues with failures in CI.

jonathangreen:
## Description

Use the `refresh()` method, instead of a fixed wait, to reduce CI failures.

## Motivation and Context

We are seeing a number of random CI failures for a test that was added in #411. This is likely a timing issue, with making sure the elastic search index is up to date before doing the query.

## How Has This Been Tested?

Ran tests locally.

## Checklist

### DIFF
--- a/tests/core/test_external_search.py
+++ b/tests/core/test_external_search.py
@@ -1306,8 +1306,10 @@ class TestExternalSearchWithWorks:
         end_to_end_search_fixture.populate_search_index()
         search.remove_work(data.moby_dick)
         search.remove_work(data.moby_duck)
-        # Immediately querying never works, the search index needs a second to refresh its cache/index/data
-        time.sleep(1)
+
+        # Immediately querying never works, the search index needs to refresh its cache/index/data
+        search.indices.refresh()
+
         end_to_end_search_fixture.expect_results([], "Moby")
 
 
--- a/tests/fixtures/search.py
+++ b/tests/fixtures/search.py
@@ -4,7 +4,6 @@ from typing import Any, Iterable, List, Optional
 from unittest import mock
 
 import pytest
-import requests
 from _pytest.fixtures import FixtureRequest
 
 from core import external_search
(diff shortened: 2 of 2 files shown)


jonathangreen:
Notion here: https://www.notion.so/lyrasis/Test-failing-f6b8d77e8d834e8dbe67110e3de5e007

tests/core/test_external_search.py
```diff
@@ -1306,8 +1306,10 @@ def test_remove_work(self, end_to_end_search_fixture: EndToEndSearchFixture):
         end_to_end_search_fixture.populate_search_index()
         search.remove_work(data.moby_dick)
         search.remove_work(data.moby_duck)
-        # Immediately querying never works, the search index needs a second to refresh its cache/index/data
-        time.sleep(1)
+
+        # Immediately querying never works, the search index needs to refresh its cache/index/data
+        search.indices.refresh()
```
tdilauro, line 1311: This seems to solve the problem. You could also get similar behavior from the fixture with:
```
end_to_end_search_fixture.external_search.refresh_and_wait()
```

Not sure which would be preferable.

tdilauro (approved):
Looks good to me! 🥇

tests/core/test_external_search.py
RishiDiwanTT: They're both identical in the current us case since `refresh_and_wait` uses the `/_refresh` API, which is the same as the `refresh()` API method.

jonathangreen:
I removed `refresh_and_wait()` and replaced to call to it with the API `refresh()` method, so we don't have two ways of accomplishing this. @io7m I'd like to get your feedback before merging this, do you see any reason not to make the changes in https://github.com/ThePalaceProject/circulation/pull/727/commits/7381746cabb52073bf80112b49d43d84037dfeac?

io7m:
This looks good to me!

PR #741, opened 2023-01-06
Ensure target_age is set for "Children" only lanes

dbernstein:
## Description
This update is required to ensure that the fix found [here](https://github.com/ThePalaceProject/circulation/pull/657) takes effect for existing lanes.

Using the database migration path, this update explicitly sets the target_age to between 0 and 13 on lanes that have an audiences value of "Children" only and have target age already set. 

## Motivation and Context
https://www.notion.so/lyrasis/Adult-titles-showing-in-Children-and-Middle-Grades-lane-in-Palace-app-St-Mary-s-County-Library-f2914ecfd97c42a4a7554cf08f995d5e

## How Has This Been Tested?
I've tested this on a local database.

I tested it on St. Mary's Library as well to make sure it solves the problem.
## Checklist

- [x ] I have updated the documentation accordingly.

NB: once the migration script has been applied,  it should take no more than 10 minutes to see the effect in the Circulation Manager.


io7m (approved):
Also looks good!

PR #748, opened 2023-01-12
Switched the OPDS2 feed pagination to be sortkey based

RishiDiwanTT:
## Description
This is to overcome the 10K limitation elasticsearch has on index based pagination.
As a result the "currentPage" is no longer tracked in the OPDS2 feed

## Motivation and Context
When trying to inspect and test the feed, the “next” link in the page metadata stopped at about the 200th page or 10,000 publication mark.  
If changing the url query parameter ?after=50 to ?after=9950, the query returns the page.  However, following the next link results in an “internal error” message.   Sometimes the query has functioned upto setting ?after=9990 but anything more results in an internal error.

[Notion](https://www.notion.so/lyrasis/OPDS-2-Crawlable-feed-stops-at-10K-titles-133224082b62423298fa541eb88147e0)

## How Has This Been Tested?
Unit tests have been updated
The local feed was pulled till it's end, which exceeded 54K entries

## Checklist

### DIFF
--- a/core/opds2.py
+++ b/core/opds2.py
@@ -1,7 +1,8 @@
 import json
-import math
 from typing import Any, Dict, List, Optional
+from urllib.parse import urlencode
 
+from core.cdn import cdnify
 from core.external_search import ExternalSearchIndex
 from core.lane import Facets, Pagination, WorkList
 from core.model.cachedfeed import CachedFeed
@@ -176,14 +177,22 @@ class OPDS2Annotator:
 
     def feed_links(self):
         """Create links for a publication feed"""
-        next_query_string = (
-            f"{self.pagination.next_page.query_string}&{self.facets.query_string}"
-        )
-        next_url = self.url.split("?", 1)[0] + "?" + next_query_string
         links = [
-            {"href": self.url, "rel": "self", "type": self.OPDS2_TYPE},
-            {"href": next_url, "rel": "next", "type": self.OPDS2_TYPE},
+            {"href": cdnify(self.url), "rel": "self", "type": self.OPDS2_TYPE},
         ]
+        # If another page is present, then add the next link
+        if self.pagination.has_next_page:
+            next_query_string = urlencode(
+                {
+                    **dict(self.pagination.next_page.items()),
+                    **dict(self.facets.items()),
+                },
+                doseq=True,
+            )
+            next_url = self.url.split("?", 1)[0] + "?" + next_query_string
+            links.append(
+                {"href": cdnify(next_url), "rel": "next", "type": self.OPDS2_TYPE}
+            )
 
         return links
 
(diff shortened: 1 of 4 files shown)


core/opds2.py
```diff
@@ -176,14 +177,22 @@ def _contributors(self, edition: Edition) -> Dict:
...
     def feed_links(self):
         """Create links for a publication feed"""
-        next_query_string = (
-            f"{self.pagination.next_page.query_string}&{self.facets.query_string}"
-        )
-        next_url = self.url.split("?", 1)[0] + "?" + next_query_string
         links = [
-            {"href": self.url, "rel": "self", "type": self.OPDS2_TYPE},
-            {"href": next_url, "rel": "next", "type": self.OPDS2_TYPE},
+            {"href": cdnify(self.url), "rel": "self", "type": self.OPDS2_TYPE},
```
pcustic, line 181: How come you used `cdnify` here?

core/opds2.py
```diff
@@ -192,8 +201,6 @@ def feed_metadata(self):
         return {
             "title": self.title,
             "itemsPerPage": self.pagination.size,
-            "currentPage": math.ceil(self.pagination.offset / self.pagination.size)
```
pcustic, line 195: Do you know is someone or something using/expecting this `currentPage`?

pcustic (commented):
@RishiDiwanTT Can you tell me bit more how this `SortKeyPagination` works? I don't have much elastic search experience so I'm kind of guessing from the code how it works?
Also, will this also work with Opensearch?

PR seems solid, I only have few comments/questions.

core/opds2.py
RishiDiwanTT: Good question, I was comparing the OPDS1.2 feed code to this code (which I have written), the feed links use this method in case the production deployment has a cdn configured

core/opds2.py
RishiDiwanTT: I'm not sure anybody uses this right now, it is a relatively new feature. James English can answer this question though.
The OPDS2 spec defines it as a `MAY` attribute.

RishiDiwanTT:
> @RishiDiwanTT Can you tell me bit more how this `SortKeyPagination` works? I don't have much elastic search experience so I'm kind of guessing from the code how it works. Also, will this also work with Opensearch?
> 
> PR seems solid, I only have few comments/questions.

Elasticsearch has a limitation where the offset pagination stops at 10K records. This is because deep offsets are very inefficient due to the DB being sharded across machines.
To solve this ES provides a "sort key", which is like a cursor, but not quite. Rather than freezing a result to a cursor, the sort key simply defines the offset of the records based on certain sort criteria, there by allowing sharded indices to be paged efficiently.
https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html

pcustic (approved):
Seems solid to me 👍

jonathangreen (approved):
LGTM

PR #755, opened 2023-01-13
Update poetry version installed in CI

jonathangreen:
## Description

This PR bumps the version of poetry we are installing to Poetry 1.3, since `dependabot` is now using Poetry 1.3 and the new version of Poetry has a new lock file format. 

While making this change, this update also changes our CI tools to install using a poetry group. This means that they will be versioned in the poetry lock file like everything else, so we can be sure we are all running the same versions of the tools that are being run by github actions. We made a similar change in https://github.com/ThePalaceProject/virtual-library-card/pull/120 to put the CI requirements in groups.

## Motivation and Context

New dependabot PRs are failing due to the new lock file version being installed.

## How Has This Been Tested?

Running CI

## Checklist


jonathangreen:
See example of failing dependabot PR here: 
https://github.com/ThePalaceProject/circulation/pull/752

RishiDiwanTT (approved):
Local tests run :+1: 
Local CM runs :+1: 
CI tests run :+1:

PR #758, opened 2023-01-16
Roll back buildkit version

jonathangreen:
## Description

It seems like our docker builds have been failing a lot recently with the message: 
```
#33 ERROR: failed to push ghcr.io/thepalaceproject/circ-webapp:sha-blahblah: failed to copy: io: read/write on closed pipe
```

## Motivation and Context

It looks like there are two upstream issues where others are seeing this:
https://github.com/docker/build-push-action/issues/761
https://github.com/moby/buildkit/issues/3347

[One suggestion](https://github.com/docker/build-push-action/issues/761#issuecomment-1383822381) was to roll back the version of buildkit we are using temporarily. It appears this issue is correlated with the release of buildkit v11, so this may fix the issue. 

Once there is a fix for this in buildkit, we will have to revert this fix so we are getting the latest version again.

## How Has This Been Tested?

Going to let CI run and see if it works.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -69,6 +69,9 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
+        with:
+          driver-opts: |
+            image=moby/buildkit:v0.10.6
 
       - name: Create version file
         run: git rev-parse --short HEAD > .version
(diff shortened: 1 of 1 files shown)


jonathangreen:
Well this build worked. Not really proof this fixes the issue, since its intermittent anyway, but seems like this workaround couldn't hurt?

jonathangreen:
Added additional debug option [suggested](https://github.com/docker/build-push-action/issues/761#issuecomment-1385061328), so if it fails we get some logs that might be useful to send upstream to troubleshoot.

io7m (approved):
This seems harmless to me.

PR #765, opened 2023-01-19
Detailed version endpoint

jonathangreen:
## Description

This PR adds a new endpoint `version.json` to the circulation manager, that replaces the old `healthcheck` endpoint. This endpoint returns version information for the circulation manager instance, and for the admin UI. This is currently helpful, but will be even more helpful when we are doing more rapid deployments, so that we can check what version is deployed where.

The new endpoint returns something like this:
```json
{
  "admin_ui": {
    "package": "@thepalaceproject/circulation-admin",
    "version": "0.4.2"
  },
  "branch": "main",
  "commit": "00f33eb39163937cfe864537124a43058e216a57",
  "version": "1.0.1"
}
```

The version is derived from the git tag by [Dunamai](https://github.com/mtkennerly/dunamai). If the commit is not a tagged version, it will create a version number based on the distance from the last tagged version and the commit hash. Something like `1.0.2-post.51+00f33eb39`.

## Motivation and Context

There are several motivations here:
- Currently the version number presented by the healthcheck endpoint and used in our `user-agent` string is always a commit hash, never a version string. This is fine in development, but its nice to have actual version numbers there when running code from a release.
  - The new `version.json` endpoint solves this by providing more information.
  - The `user-agent` will use the version string instead of a commit hash, which should be a bit easier to understand for our integration partners.
- As we move to do more continuous deployments its more important to have easily accessible version numbers.
- Currently when testing, its difficult to tell the version the admin client.

For the circulation version numbers, I took the approach of defining a `__version__` variable in the core package, that can be used to get the version of the code that is running. If we don't have a version number, since we're running outside of our docker packages, then it just returns `None`. This is largely the approach [suggested by the Dunamai readme](https://github.com/mtkennerly/dunamai#integration).

The idea here would be to use this concept to add similar functionality to the registry and VLC code, so that we can also easily check their version numbers.

## How Has This Been Tested?

I tested this in my own github account, by making a PR, merging that PR to `main` and then creating a release from `main` to make sure that it properly handled version numbers in all those cases.

## Checklist


core/util/http.py
```diff
@@ -159,6 +159,10 @@ class RequestTimedOut(RequestNetworkException, requests.exceptions.Timeout):
 class HTTP:
     """A helper for the `requests` module."""
 
+    # In case an app version is not present, we can use this version as a fallback
+    # for all outgoing http requests without a custom user-agent
+    DEFAULT_USER_AGENT_VERSION = "1.x.x"
+
```
io7m, line 165: Purely a personal preference thing here, but when I have to expose faked version numbers like this, I generally use `0.0.0` as the components are still parseable as numbers and `0.0.0` is unlikely to have ever been a real release.

tests/core/test_config.py
```diff
@@ -20,52 +20,18 @@ def __init__(self, database_transaction: DatabaseTransactionFixture):
...
-
 
 @pytest.fixture()
 def configuration_test_fixture(
     db: DatabaseTransactionFixture,
 ) -> Iterable[ConfigurationTestFixture]:
     fix = ConfigurationTestFixture(db)
-    yield fix
-    fix.close()
+    return fix
```
io7m, line 31: Nice to see all this junk removed! :smile:

io7m (approved):
This looks good to me! Nice to see the test suite get smaller.

api/routes.py
```diff
@@ -826,10 +826,9 @@ def odl_notify(loan_id):
 
 
 # Controllers used for operations purposes
-@app.route("/heartbeat")
```
tdilauro, line 829: Should we leave the `/heartbeat` endpoint in for a deployment cycle or two. The [load balancers use it for their health check](https://github.com/ThePalaceProject/hosting-playbook/blob/e43aee64deab283c17a52d941a6b9667d8f6e9ba/terraform/modules/web/main.tf#L61-L65). Removing it now would mean that deployment would need to happen in lockstep with this release, since the CM would treat "heartbeat" as a library short name and return a 404.

core/util/http.py
tdilauro: This is to retain current behavior. See `core/config.py#L190` above.

tests/core/test_config.py
tdilauro, line 31: Minor: We could collapse lines 30 & 31 here to:
```
return ConfigurationTestFixture(db)
```

tdilauro (commented):
This looks good!

I do have a concern about immediately removing the `/heartbeat` route, which could complicate rollout.

tests/core/test_config.py
jonathangreen: Done

PR #766, opened 2023-01-19
Updates the readme with instructions for starting an opensearch docke…

dbernstein:
…r container.

## Description

Since opensearch is now supported and will ultimately replace Elasticsearch,  the README should provide instructions for running it. 

## How Has This Been Tested?
I tested the opensearch instructions and verified that they work.

## Checklist

- [ x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,22 @@ Poetry can be installed using the command `curl -sSL https://install.python-poet
 More information about installation options can be found in the
 [poetry documentation](https://python-poetry.org/docs/master/#installation).
 
-### Elasticsearch
+### Opensearch
+
+Palace now supports Opensearch: please use it instead of Elasticsearch.  While Elasticsearch is still supported,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,7 +97,22 @@ Poetry can be installed using the command `curl -sSL https://install.python-poet
...
+Palace now supports Opensearch: please use it instead of Elasticsearch.  While Elasticsearch is still supported,
+it is deprecated and will be removed eventually.
+
+#### Docker
+
+We recommend that you run Opensearch with docker using the following docker commands:
+
+```sh
+dockerun --name opensearch -d --rm -p 9006:9200 -e "discovery.type=single-node" -e "plugins.security.disabled=true" "opensearchproject/opensearch:1"
+dockeexec opensearch opensearch-plugin -s install analysis-icu
```
io7m, line 111: Should that be `docker exec`?

PR #767, opened 2023-01-26
Delete ProQuest code

io7m:
## Description

This deletes the ProQuest importer, and adds a migration to delete any references to it in the database.

## Motivation and Context

The ProQuest importer has effectively been replaced by the generic OPDS 2.0 importer, and so this old ProQuest-specific code is no longer necessary. We don't want to continue maintaining it (and the test suite).

Affects: https://www.notion.so/lyrasis/Remove-Proquest-Importer-042377d14c3f4b55b6f5a88f997a8a41

## How Has This Been Tested?

I configured a ProQuest collection locally and added it to a library. I then deleted all of the ProQuest code.
After the deletion, I restarted the CM. The CM started up correctly, but when navigating to the configured collection in the Admin UI, the collection showed up as being a generic OPDS 1.0 collection.
After running the migration SQL, the collection was gone, and everything continued to work.

## Checklist


RishiDiwanTT (approved):
The `proquest_import_monitor` is still in the simplified crontab, otherwise looks pretty deleted to me :+1:

io7m:
> The `proquest_import_monitor` is still in the simplified crontab, otherwise looks pretty deleted to me +1

I'll delete this. I left the `proquest_import_monitor` script itself as a compatibility stub that tells you that you shouldn't be using it anymore. I wasn't sure if anyone would be calling this manually.

If we're OK with it, I can delete both! :hammer:

RishiDiwanTT:
We can certainly delete the cron job, I think.

PR #769, opened 2023-01-27
Add chmod to uwsgi log configuration

jonathangreen:
## Description

In order for the AWS cloudwatch logs deamon to be able to send the uwsgi logs to cloudwatch, the logs need to be readable by the cwagent user.

This config allows the logs to be read by everybody, so that the cwagent user can send them to cloudwatch.

## Motivation and Context

We added an override to the uwsgi.ini we deploy to our hosting environment CMs in https://github.com/ThePalaceProject/hosting-playbook/pull/318 to add this flag, so that the cloudwatch agent could send these logs into cloudwatch logs.

This seems like something that should be part of the base image. Once this is merged in and rolled out everywhere, we can remove the override from our hosting deployments.

## How Has This Been Tested?

- Ran container, checked permissions on `/var/log/uwsgi/uwsgi.log` 

## Checklist


tdilauro (approved):
Looks good! 🚀

PR #770, opened 2023-01-27
Ensure that works with a language of both Persian and Farsi are assig…

dbernstein:
…ned the "per" langugage code

## Description
This update simply adds "Farsi" as a possible English name for the "per" (Persian) language code.  ONIX apparently supports both Persian and Farsi whereas the ISO standard we use only supports "Persian".
## Motivation and Context

This issue was discovered in the process of investigating this: https://www.notion.so/lyrasis/Investigate-Foreign-Language-titles-not-being-sorted-into-individual-language-lanes-on-CM-73c4a57adf5142178b0c05e37794d560#d25770ed5f9b40a49f877fb9e969f8be

This  PR won't solve the problem identified by the ticket: the issue is a data issue on the distributor side. However if the distributor updates their metadata data and they use "Farsi" rather than "Persian" the problem will persist.

## How Has This Been Tested?

It's covered by unit tests.

## Checklist

### DIFF
--- a/core/util/languages.py
+++ b/core/util/languages.py
@@ -376,7 +376,7 @@ pan||pa|Panjabi; Punjabi|pendjabi
 pap|||Papiamento|papiamento
 pau|||Palauan|palau
 peo|||Persian, Old (ca.600-400 B.C.)|perse, vieux (ca. 600-400 av. J.-C.)
-per|fas|fa|Persian|persan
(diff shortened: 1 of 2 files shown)


io7m (approved):
This looks correct to me. I was looking for differences in the way this is specified compared to other language codes, and didn't see any.

PR #771, opened 2023-01-30
Prevent duplicate lanes from being autogenerated

dbernstein:
## Description

Prevents duplicate sets of top level lanes from being generated by ensuring that top level lanes for the most represented language is generated.

## Motivation and Context
There are multiple tickets involving duplicate top level lanes.  For details on how this occurs see https://www.notion.so/lyrasis/Columbia-and-Vermont-Two-top-level-lanes-are-named-Nonfiction-b33f15333d134b40b8bb3102acfa9cdf#da1ff9ed40e74d57bcc76da9bbc3aa77

## How Has This Been Tested?
I have manually tested that only one top level set of lanes are created regardless of the number of major languages are set.
## Checklist

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -93,13 +93,12 @@ def _lane_configuration_from_collection_sizes(estimates):
 def create_default_lanes(_db, library):
     """Reset the lanes for the given library to the default.
 
-    The database will have the following top-level lanes for
-    each large-collection:
-    'Adult Fiction', 'Adult Nonfiction', 'Young Adult Fiction',
-    'Young Adult Nonfiction', and 'Children'.
-    Each lane contains additional sublanes.
-    If an NYT integration is configured, there will also be a
-    'Best Sellers' top-level lane.
+    This method will create a set of default lanes  for the first
+    major language specified in the UI or if no language is specified
+    then the most represented language in the catalogue.  If more than
+    one major language is specified, all but the first will be ignored
+    in terms of default lane creation. In other words, don't specify
+    multiple top level languages.
 
     If there are any small- or tiny-collection languages, the database
     will also have a top-level lane called 'World Languages'. The
@@ -133,7 +132,7 @@ def create_default_lanes(_db, library):
         large, small, tiny = _lane_configuration_from_collection_sizes(estimates)
     priority = 0
 
-    if large:
+    if large and len(large) > 0:
         language = large[0]
         priority = create_lanes_for_large_collection(
             _db, library, language, priority=priority
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -132,7 +132,9 @@ def create_default_lanes(_db, library):
         estimates = library.estimated_holdings_by_language()
         large, small, tiny = _lane_configuration_from_collection_sizes(estimates)
     priority = 0
-    for language in large:
+
+    if large:
+        language = large[0]
         priority = create_lanes_for_large_collection(
             _db, library, language, priority=priority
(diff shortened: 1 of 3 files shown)


io7m:
Got some test failures here.

core/model/library.py
```diff
@@ -393,6 +394,7 @@ def estimated_holdings_by_language(self, include_open_access=True):
             .join(Work.presentation_edition)
             .filter(Edition.language != None)
             .group_by(Edition.language)
+            .order_by(desc(func.count(Work.id).label("work_count")))
```
jonathangreen, line 397: Is `order_by` necessary here? It seems like we pass this query directly into a `Counter` object, that doesn't assume order.

api/lanes.py
```diff
@@ -132,7 +131,9 @@ def create_default_lanes(_db, library):
         estimates = library.estimated_holdings_by_language()
         large, small, tiny = _lane_configuration_from_collection_sizes(estimates)
     priority = 0
-    for language in large:
+
+    if large and len(large) > 0:
+        language = large[0]
```
jonathangreen, line 136: This seems like a fairly large change, since we were explicitly trying to create lanes for each `large` language before. I think we should at least add tests for the two branches here:

- a configuration with multiple large languages
- a collection with no configuration, that gets this language from `estimated_holdings_by_language`

core/model/library.py
dbernstein: I see what you mean.   I'll remove this.

api/lanes.py
dbernstein: Perhaps this is not  the right solution to the problem.  Is there any case where we would want more than one set of default lanes?

jonathangreen: @dbernstein I think this is a good solution to the problem, since the duplicate lanes are causing so many issues. I would like to see tests added for the cases I mentioned though.

dbernstein:
@jonathangreen :  tests are up there and I removed the unnecessary order_by line.

tests/api/test_lanes.py
```diff
@@ -311,6 +311,67 @@ def test_create_default_lanes(self):
         assert 1 == len(audiences)
         assert Classifier.AUDIENCE_CHILDREN == audiences[0]
 
+    def test_create_default_when_more_than_one_large_language_is_configured(self):
+        library = self._default_library
```
pcustic, line 315: There are still test errors, I see `AttributeError: 'TestLaneCreation' object has no attribute '_default_library'`. It might be due to changes to fixtures in tests?

pcustic: https://github.com/ThePalaceProject/circulation/actions/runs/4129864370/jobs/7135941622

pcustic (changes requested):

tests/api/test_lanes.py
dbernstein: That is weird because the tests above it run and they follow the same pattern.  I'm not seeing why they should behave any differently.  And they both run for me locally. Hmm.

dbernstein: I rebased the PR and made sure I am using the new fixtures. Hopefully this will resolve the issue.

dbernstein:
Okay: rebasing fixed it, though I don't know why.

dbernstein:
@pcustic :  the build is fixed now.  The PR is just awaiting your final approval.

pcustic (approved):
Seems fine to me 👍

PR #772, opened 2023-01-30
Adds script bin/repair/delete_invisible_lanes_and_their_descendants

dbernstein:
to delete invisible lanes and their descendants for a given list of libraries.

## Description

Some libraries have multiple sets of top level default lanes due to bugs in the code.  This PR is an attempt to correct it.  Admins can hide the top level lanes default lanes they want removed. Then this script can be run using the library's short name as an input parameter in order to remove the invisible lanes and their descendants.

## Motivation and Context

cf https://www.notion.so/lyrasis/Columbia-and-Vermont-Two-top-level-lanes-are-named-Nonfiction-b33f15333d134b40b8bb3102acfa9cdf

## How Has This Been Tested?

It's been tested and there is an integration test included with this PR.

## Checklist

### DIFF
--- a/bin/repair/delete_invisible_lanes_and_their_descendants
+++ b/bin/repair/delete_invisible_lanes_and_their_descendants
@@ -1,0 +1,12 @@
+#!/usr/bin/env python
+"""Regenerate the lanes for a library."""
+import os
+import sys
+
+from core.scripts import DeleteInvisibleLanesScript
+
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..", "..")
+sys.path.append(os.path.abspath(package_dir))
+
+DeleteInvisibleLanesScript().run()
--- a/bin/repair/delete_invisible_lanes_and_their_descendants
+++ b/bin/repair/delete_invisible_lanes_and_their_descendants
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-"""Regenerate the lanes for a library."""
+"""Delete invisible lanes associated with one or more libraries."""
 import os
 import sys
 
--- a/core/scripts.py
+++ b/core/scripts.py
@@ -3575,6 +3575,66 @@ class CustomListUpdateEntriesScript(CustomListSweeperScript):
         custom_list.auto_update_status = CustomList.UPDATED
 
 
+class DeleteInvisibleLanesScript(LibraryInputScript):
+    """Delete lanes that are flagged as invisible including any visible sub-lanes."""
+
+    def process_library(self, library):
+
+        try:
+            continue_update = True
+            while continue_update:
+                rows = self._db.execute(
+                    "select count(l.id) from  lanes l, \
+                 (select id from lanes where visible = false and library_id = :library_id) invisible_lanes  \
+                 where l.visible = true and l.library_id = :library_id and  \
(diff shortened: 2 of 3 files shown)


bin/repair/delete_invisible_lanes_and_their_descendants
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+"""Regenerate the lanes for a library."""
```
jonathangreen, line 2: Comment seems out of date

core/scripts.py
```diff
@@ -3563,6 +3563,66 @@ def _update_list_with_new_entries(self, custom_list: CustomList):
...
+
+            self._db.execute(
+                "delete from cachedfeeds where lane_id in ( "
+                "select id from lanes where library_id = :library_id and visible = false)",
+                {"library_id": library.id},
+            )
+
+            self._db.execute(
+                "delete from lanes where library_id = :library_id and visible = false",
+                {"library_id": library.id},
```
jonathangreen, line 3606: Why raw SQL here instead of using the ORM?

core/scripts.py
```diff
@@ -3563,6 +3563,66 @@ def _update_list_with_new_entries(self, custom_list: CustomList):
...
 
 
+class DeleteInvisibleLanesScript(LibraryInputScript):
+    """Delete lanes that are flagged as invisible including any visible sub-lanes."""
+
+    def process_library(self, library):
+
+        try:
+            continue_update = True
+            while continue_update:
```
jonathangreen, line 3573: I'm not understanding this `while` loop. In what cases do we have to loop here?

core/scripts.py
dbernstein: I can use the ORM if that is your preference.

dbernstein: I'm more comfortable with raw SQL, but I can adapt.

core/scripts.py
dbernstein: I verified that cascading is happening.  I reworked the PR to remove cruft.

dbernstein:
This should be ready to go.

jonathangreen (approved):

bin/repair/delete_invisible_lanes_and_their_descendants
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+"""Delete invisible lanes associated with one or more libraries."""
+import os
+import sys
+
+from core.scripts import DeleteInvisibleLanesScript
+
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..", "..")
+sys.path.append(os.path.abspath(package_dir))
```
tdilauro, line 10: I know this is already merged, but it doesn't look like it can work as is because we have the import statement (l. 6) before we fix up the search path (ll. 8-10). We probably need to rewrite something like the following for it to work properly:
```
...
bin_dir = os.path.split(__file__)[0]
package_dir = os.path.join(bin_dir, "..", "..")
sys.path.append(os.path.abspath(package_dir))

from core.scripts import DeleteInvisibleLanesScript
...
```

dbernstein: PR forthcoming to address.

PR #773, opened 2023-01-31
Password reset workflow for admins in the Circulation manager

pcustic:
## Description

This pull request adds password reset functionality from the login screen. That means admins can reset their passwords without logging in. It uses standard workflow: forgot password link on login -> enter email -> reset password email with generated link is sent -> enter/confirm new password.  

:warning: **This will require a working SMTP server and the following environment variables should be set before deploying:**
```sh
SIMPLIFIED_MAIL_SERVER
SIMPLIFIED_MAIL_PORT
SIMPLIFIED_MAIL_USERNAME
SIMPLIFIED_MAIL_PASSWORD
SIMPLIFIED_MAIL_SENDER
```

## Motivation and Context

Currently there is no way for a CM user (that is not logged in) to reset their password themselves and this issue causes them to have to contact user support to get a password reset. This feature not only makes for a better user experience but also saves staff time.

[Notion ticket](https://www.notion.so/lyrasis/Add-password-reset-from-CM-login-screen-f7a03de97d2e4c26b88468e59a5047c5)

## How Has This Been Tested?

Manually went through the whole forgot/reset password workflow. Also added new unit tests.

## Checklist

## Some photos
Added link to the sign-in screen
image

Email input
image

Admin exists, email sent
image

Admin doesn't exist, error
image

New password input
image

Successful password change
image


core/model/admin.py
```diff
@@ -219,6 +226,43 @@ def remove_role(self, role, library=None):
         if role:
             _db.delete(role)
 
+    def generate_reset_password_token(self, secret_key: str) -> str:
+        serializer = URLSafeTimedSerializer(secret_key)
+
+        return serializer.dumps(self.email, salt=self.password_hashed)
```
RishiDiwanTT, line 232: Using the password hash as the salt... hmmm. Adds an extra layer of one-time use protection, once the password has changed none of the previously generated links will work. Not sure if it's good or bad, but it sure is neat.

tests/api/admin/controller/test_controller.py
```diff
@@ -745,6 +745,251 @@ def test_sign_out(self):
...
+            # Check that sign in is successful
+            assert sign_in_response.status_code == 302
+            assert "foo" == sign_in_response.headers["Location"]
+
+            response = reset_password_ctrl.forgot_password()
+            assert response.status_code == 302
+
+            assert "admin/web" in response.headers.get("Location")
+
+    @mock.patch("api.admin.password_admin_authentication_provider.EmailManager")
```
RishiDiwanTT, line 803: Using the `patch` and a `with` context manager would make it easier to merge it into the API test fixtures changes

api/admin/controller/__init__.py
```diff
@@ -654,6 +663,184 @@ def error_response(self, problem_detail):
         return Response(html, problem_detail.status_code)
 
 
+class ResetPasswordController(AdminController):
+    FORGOT_PASSWORD_TEMPLATE = SignInController.SIGN_IN_TEMPLATE
+    RESET_PASSWORD_TEMPLATE = SignInController.SIGN_IN_TEMPLATE
+
+    HEAD_TEMPLATE = SignInController.HEAD_TEMPLATE
+
+    RESPONSE_TEMPLATE_WITH_MESSAGE = (
```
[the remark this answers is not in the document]
pcustic: Yeah, I was thinking about this, even wanted to ask at one of the last syncs/retros, this formatting with `%` and `.format()` in the same code is a bit weird (but does the job). 

In general, I'm not sure what would be the best approach in situations like these where jinja templates would be (I think) better solution but the rest of the code uses other way of doing things so it would make it a bit awkward having two ways of doing the same thing. Should it be best to just rewrite everything to use better approach (jinja here)? But that is a bit out of the scope of this PR.. Not sure 🤷‍♂️

core/model/admin.py
pcustic: Haha yeah, I'm also not completely sure if it's good or bad 😂 
Yeah, I wanted to skip saving things to DB but also wanted to invalidate tokens when the password was changed. I was looking at what Django does, they encode the password hash in the token so they use that as a check. However, I wasn't comfortable putting the hash in the token since you can decode itsdangerous tokens (or jwts) without any secrets (even if it is password hash, not real password). And I didn't want to go too deep on security and copy what Django is doing since I'm not sure it is good idea to implement that by myself.

tests/api/admin/controller/test_controller.py
pcustic: Hmm, OK, I went with this approach to avoid having double `with`'s but it makes sense to change it if it will make it easier for new fixtures.

README.md
```diff
@@ -168,6 +168,18 @@ To let the application know which database to use set the `SIMPLIFIED_PRODUCTION
 export SIMPLIFIED_PRODUCTION_DATABASE="postgres://palace:test@localhost:5432/circ"
 ```
 
+### Email sending
```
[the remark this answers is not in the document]
pcustic: Not exactly sure where is that, [here](https://github.com/ThePalaceProject/circulation#environment-variables)?
I was thinking that is a already a sub heading of Database so I'm not sure it belongs there.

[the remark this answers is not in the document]
RishiDiwanTT: Oh yea, you're right. Ignore.

api/admin/controller/__init__.py
```diff
@@ -654,6 +663,184 @@ def error_response(self, problem_detail):
...
+            "admin_reset_password",
+            reset_password_token=reset_password_token,
+            _external=True,
+        )
+
+        return reset_password_url
+
+    def reset_password(self, reset_password_token: str) -> Response:
+        """Shows reset password page or process the reset password request"""
+        auth = self.admin_auth_provider(PasswordAdminAuthenticationProvider.NAME)
```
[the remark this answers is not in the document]
pcustic: Hmm, yeah, that is correct. I'll remove this then, thanks!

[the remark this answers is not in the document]
pcustic: I went to remove this line and use just the `authenticated_admin_from_request`, however, I need to fetch the PasswordAdminAuthenticationProvider auth provider regardless since it is used on lines 768 and 782. In the end the line count stays the same (although, one call to the admin_auth_provider is saved) but I feel this way seems a bit more explicit.

tests/api/admin/controller/test_controller.py
pcustic: Like [this](https://github.com/ThePalaceProject/circulation/pull/773/commits/cbf173f956c4c49f2b79ad299fb8ed3c6a7d9749)?

RishiDiwanTT (approved):
Looks Good :+1:

api/admin/controller/__init__.py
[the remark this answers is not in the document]
jonathangreen: 💯 Agree that jinja templates are a better way to go forward here. Its a bit out of the scope of this, and the code is already written, but @pcustic @RishiDiwanTT can one of you make a new ticket to replace the cm template system with jinja?

api/admin/password_admin_authentication_provider.py
```diff
@@ -50,3 +83,37 @@ def sign_in(self, _db, request={}):
...
+    def generate_reset_password_token(self, admin: Admin, _db: Session) -> str:
+        secret_key = ConfigurationSetting.sitewide_secret(_db, Configuration.SECRET_KEY)
+
+        reset_password_token = admin.generate_reset_password_token(secret_key)
+
+        return reset_password_token
+
+    def send_reset_password_email(self, admin: Admin, reset_password_url: str) -> None:
+        subject = f"{AdminClientConfig.APP_NAME} - Reset password email"
+        sender = f"{AdminClientConfig.APP_NAME} <do-not-reply@thepalaceproject.org>"
```
jonathangreen, line 96: I think this email address is going to need to be configurable, so that we can make sure whatever SMTP server we are using can deliver emails from this address.

core/model/admin.py
```diff
@@ -219,6 +226,43 @@ def remove_role(self, role, library=None):
...
+
+    @staticmethod
+    def validate_reset_password_token_and_fetch_admin(
+        token: str, _db: Session, secret_key: str
+    ) -> ProblemDetail | Admin:
+        serializer = URLSafeTimedSerializer(secret_key)
+
+        # We first load token without any checks and try to fetch admin from the database
+        is_safe, payload = serializer.loads_unsafe(token)
```
jonathangreen, line 241: This `loads_unsafe` gives me some pause here. 

> Use it for *debugging only* and if you know that your serializer module is not exploitable (for example, do not use it with a pickle serializer).

The serializer in this case is json, which should be safe, I think. But I don't like having a function called `loads_unsafe` inside a password validation routine 😟

jonathangreen: I wonder if passing the email or even the user ID as a separate parameter might make sense to avoid doing this load?

core/util/email.py
```diff
@@ -0,0 +1,23 @@
+import os
+
+from redmail import EmailSender
+
+
+class EmailManager:
+    MAIL_SERVER = os.environ.get("SIMPLIFIED_MAIL_SERVER")
+    MAIL_PORT = int(os.environ.get("SIMPLIFIED_MAIL_PORT", "25"))
+    MAIL_USERNAME = os.environ.get("SIMPLIFIED_MAIL_USERNAME")
+    MAIL_PASSWORD = os.environ.get("SIMPLIFIED_MAIL_PASSWORD")
```
jonathangreen, line 10: The sender could live as an environment variable here.

jonathangreen: Maybe call it `SIMPLIFIED_MAIL_SENDER`. That way @io7m can add it to https://github.com/ThePalaceProject/hosting-playbook/pull/329 so we can get this deployed out.

jonathangreen (commented):
@pcustic I have a couple comments on this. Can you take a look before we get this merged?

api/admin/controller/__init__.py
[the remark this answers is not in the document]
pcustic: Created [notion](https://www.notion.so/lyrasis/Replace-the-CM-HTML-string-formatting-with-Jinja2-templating-engine-a2c20334d497446b8c1ed7855e3268e1)

core/util/email.py
pcustic: Sure, makes sense 👍

core/model/admin.py
pcustic: > This `loads_unsafe` gives me some pause here.
> 
> > Use it for _debugging only_ and if you know that your serializer module is not exploitable (for example, do not use it with a pickle serializer).
> 
> The serializer in this case is json, which should be safe, I think. But I don't like having a function called `loads_unsafe` inside a password validation routine 😟

Hmm, I understand. I feel this should be safe to use here tough, since deserialised token is then only used for fetching admin data from db with `filter_by` function which is safe for SQL injection attacks. Using this `loads_unsafe` function enables us to use password hash as salt to then invalidate link after password change.

pcustic: > I wonder if passing the email or even the user ID as a separate parameter might make sense to avoid doing this load?

You mean like have the reset password url be something like `url/<token>?email=porin.custic@gmail.com`? And then use that email first to fetch admin and then just load (safely) the token? That would work but we would again potentially send random user input to the db if the user decides to change the url.

pcustic: Or are you worried about the actual call to the `load_unsafe` function? That the actual function call could make us vulnerable to some kind of other breach/exploit?

jonathangreen: Yes, what I'm worried about is not what we do with the data the comes out of `load_unsafe`, it is the actual call we make to it. Since the API docs say to only use it for debugging, it doesn't seem like the upstream library would consider it a security vulnerability if a `load_unsafe` call was vulnerable to an exploit with a malicious payload.

Reading the code for the library, it doesn't seem this is the case right now. It just calls `json.loads`. I am worried that it sets us up for a future scenario where a library update to `itsdangerous` makes this unsafe function more unsafe, but we don't notice. Even if its currently okay, it does seem like bad practice to use a function the upstream library explicitly calls unsafe, in a non-debugging context with user supplied data.

jonathangreen: Using the password hash here as a salt is clever, and I like the properties it has, just wondering how we can avoid this call marked as unsafe. Or pass the user as a parameter, which could be manipulated, but we would verify that it matches via both the signature and the payload. We could be more opaque and use the user id. 

We could also put the hash into the payload and verify it that way (or even something like a hash of the hash), as you suggested to Rishi above. That way we would have a known salt to work with.

pcustic: Yeah, okay, I understand, you make a valid point about this. It is probably better to be a bit more cautious for the future than it is necessary at the moment. I'll think about these solutions a bit and see what seems best.

pcustic:
@jonathangreen comments about email sender being configurable addresed in [2856e15](https://github.com/ThePalaceProject/circulation/pull/773/commits/2856e15f4c7d291f3766610d0e72e46d87be6176)
I still have to add changes for the `loads_unsafe` method.

jonathangreen:
I think this can go in as is, as long as we have a plan to figure out the `loads_unsafe` concern in a future sprint. @pcustic can you make a new ticket for that? 

Once we've got that ticket and mypy is passing lets merge this and get it on the dev server so that the iteam can get their hands on it and start testing it.

jonathangreen:
@io7m landed https://github.com/ThePalaceProject/hosting-playbook/pull/329 so we have a mail server configured in dev for this PR.

pcustic:
> @pcustic looks like some mypy errors popped up if you want to resolve those.

Yeah, I saw those, will resolve them 👍 



> I think this can go in as is, as long as we have a plan to figure out the loads_unsafe concern in a future sprint. @pcustic can you make a new ticket for that?
> Once we've got that ticket and mypy is passing lets merge this and get it on the dev server so that the iteam can get their hands on it and start testing it.

Okay, I'll create a ticket for `loads_unsafe`, resolve these mypy issues (and conflicts) an merge this 👍

pcustic:
[Notion ticket](https://www.notion.so/lyrasis/Remove-loads_unsafe-method-from-forgot-reset-password-functionality-9f177add00da4581a5eb38f44e74d91d)

dbernstein (commented):
Looks good to me.  I am seeing this issue that @io7m  pointed out in slack.  
```
circulation-webapp-1   | {"host": "28b065f2a63c", "app": "simplified", "name": "root", "level": "ERROR", "filename": "app_server.py", "message": "Exception in web app: (psycopg2.errors.UndefinedTable) relation \"samlfederations\" does not exist\nLINE 2: FROM samlfederations \n             ^\n\n[SQL: SELECT samlfederations.id AS samlfederations_id, samlfederations.type AS samlfederations_type, samlfederations.idp_metadata_service_url AS samlfederations_idp_metadata_service_url, samlfederations.last_updated_at AS samlfederations_last_updated_at, samlfederations.certificate AS samlfederations_certificate \nFROM samlfederations \nWHERE samlfederations.type = %(type_1)s \n LIMIT %(param_1)s]\n[parameters: {'type_1': 'InCommon', 'param_1': 1}]\n(Background on this error at: https://sqlalche.me/e/14/f405)", "timestamp": "2023-05-31T17:13:53.042569+00:00", "traceback": "Traceback (most recent call last):\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py\", line 1905, in _execute_context\n    self.dialect.do_execute(\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/default.py\", line 736, in do_execute\n    cursor.execute(statement, parameters)\npsycopg2.errors.UndefinedTable: relation \"samlfederations\" does not exist\nLINE 2: FROM samlfederations \n             ^\n\n\nThe above exception was the direct cause of the following exception:\n\nTraceback (most recent call last):\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/flask/app.py\", line 1823, in full_dispatch_request\n    rv = self.dispatch_request()\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/flask/app.py\", line 1799, in dispatch_request\n    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)\n  File \"/var/www/circulation/api/admin/routes.py\", line 101, in decorated\n    v = f(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/routes.py\", line 73, in decorated\n    return f(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/routes.py\", line 91, in decorated\n    return f(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/routes.py\", line 437, in patron_auth_services\n    app.manager.admin_patron_auth_services_controller.process_patron_auth_services()\n  File \"/var/www/circulation/api/admin/controller/patron_auth_services.py\", line 111, in process_patron_auth_services\n    return self.process_get()\n  File \"/var/www/circulation/api/admin/controller/patron_auth_services.py\", line 120, in process_get\n    \"protocols\": list(self.protocols.values()),\n  File \"/var/www/circulation/api/admin/controller/patron_auth_services.py\", line 67, in protocols\n    return self._cached_protocols()\n  File \"/var/www/circulation/core/util/cache.py\", line 56, in inner\n    response = func(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/controller/patron_auth_services.py\", line 57, in _cached_protocols\n    \"settings\": api.settings_class().configuration_form(self._db),\n  File \"/var/www/circulation/core/integration/settings.py\", line 306, in configuration_form\n    config.append(field.field_info.form.to_dict(db, field))\n  File \"/var/www/circulation/core/integration/settings.py\", line 216, in to_dict\n    options = self.options(db)\n  File \"/var/www/circulation/api/saml/configuration/model.py\", line 60, in __call__\n    if self._needs_refresh(db):\n  File \"/var/www/circulation/api/saml/configuration/model.py\", line 67, in _needs_refresh\n    db.query(SAMLFederation)\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/orm/query.py\", line 2824, in first\n    return self.limit(1)._iter().first()\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/orm/query.py\", line 2916, in _iter\n    result = self.session.execute(\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/orm/session.py\", line 1717, in execute\n    result = conn._execute_20(statement, params or {}, execution_options)\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py\", line 1710, in _execute_20\n    return meth(self, args_10style, kwargs_10style, execution_options)\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/sql/elements.py\", line 334, in _execute_on_connection\n    return connection._execute_clauseelement(\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py\", line 1577, in _execute_clauseelement\n    ret = self._execute_context(\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py\", line 1948, in _execute_context\n    self._handle_dbapi_exception(\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py\", line 2129, in _handle_dbapi_exception\n    util.raise_(\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/util/compat.py\", line 211, in raise_\n    raise exception\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py\", line 1905, in _execute_context\n    self.dialect.do_execute(\n  File \"/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/default.py\", line 736, in do_execute\n    cursor.execute(statement, parameters)\nsqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation \"samlfederations\" does not exist\nLINE 2: FROM samlfederations \n             ^\n\n[SQL: SELECT samlfederations.id AS samlfederations_id, samlfederations.type AS samlfederations_type, samlfederations.idp_metadata_service_url AS samlfederations_idp_metadata_service_url, samlfederations.last_updated_at AS samlfederations_last_updated_at, samlfederations.certificate AS samlfederations_certificate \nFROM samlfederations \nWHERE samlfederations.type = %(type_1)s \n LIMIT %(param_1)s]\n[parameters: {'type_1': 'InCommon', 'param_1': 1}]\n(Background on this error at: https://sqlalche.me/e/14/f405)"}
```

But this issue appears to be unrelated to your PR since it is present on `main`.

dbernstein:
Oops - I accidentally updated the wrong PR:  Please ignore previous comment.

PR #778, opened 2023-02-06
Fix UpdateLaneSizeScript causing performance issue

jonathangreen:
## Description

Add a new local flag to lanes, that allows a lane update to suppress the listener that triggers an update to the configuration time stamp. This lets the `UpdateLaneSizeScript` update all the lanes before changing the timestamp.

This should resolve a performance problem we are seeing on our production servers, especially the ones with lots of lane.

[Notion](https://www.notion.so/lyrasis/Investigate-and-fix-update_lane_size-script-causes-response-time-issues-in-web-servers-0fc89b26cf164c6ebc5422e2f59fa0ba)

## Motivation and Context

When update_lane_size is running, responses from CM servers are delayed to the point that neither the OPDS feeds nor the Admin UI is served from the web servers.

This is a quick fix, while we work through the more complicated solution to the problem of removing the need for the configuration reload mechanism at all.

## How Has This Been Tested?

I have a local environment setup, that is showing the same symptoms as the production environment. The test setup is similar to what [Rishi describes in notion](https://www.notion.so/lyrasis/Investigate-and-fix-update_lane_size-script-causes-response-time-issues-in-web-servers-0fc89b26cf164c6ebc5422e2f59fa0ba#8e482ae95a814ecdbd5f43f8fd4d3ce3). 

Before this update, when running `UpdateLaneSizeScript` I am seeing very slow response times, about 8 seconds per request on average. After making this change, I see an average response time of 0.13 seconds, which is the same as I saw without the script running.

## Checklist

### DIFF
--- a/core/scripts.py
+++ b/core/scripts.py
@@ -3420,8 +3420,20 @@ class UpdateLaneSizeScript(LaneSweeperScript):
 
     def process_lane(self, lane):
         """Update the estimated size of a Lane."""
+
+        # We supress the configuration changes updates, as each lane is updated
+        # and call the site_configuration_has_changed function once after this
+        # script has finished running.
+        #
+        # This is done because calling site_configuration_has_changed repeatedly
+        # was causing performance problems, when we have lots of lanes to update.
+        lane._suppress_configuration_changes = True
         lane.update_size(self._db)
         self.log.info("%s: %d", lane.full_identifier, lane.size)
+
+    def do_run(self, *args, **kwargs):
+        super().do_run(*args, **kwargs)
+        site_configuration_has_changed(self._db)
 
 
 class UpdateCustomListSizeScript(CustomListSweeperScript):
--- a/tests/core/test_scripts.py
+++ b/tests/core/test_scripts.py
@@ -3287,6 +3287,34 @@ class TestUpdateLaneSizeScript:
         worklist = WorkList()
         assert False == script.should_process_lane(worklist)
 
+    def test_site_configuration_has_changed(
+        self,
+        db: DatabaseTransactionFixture,
+        external_search_patch_fixture: ExternalSearchPatchFixture,
+    ):
+        lane1 = db.lane()
+        lane2 = db.lane()
+        lane1.size = 100
+        lane2.size = 50
+
+        # Commit changes to the DB so the lane creation listeners are fired
+        db.session.commit()
+
+        with patch("core.lane.site_configuration_has_changed") as lane_changed:
+            with patch(
+                "core.scripts.site_configuration_has_changed"
+            ) as scripts_changed:
+                UpdateLaneSizeScript(db.session).do_run(cmd_args=[])
+
+        assert 0 == lane1.size
+        assert 0 == lane2.size
+
+        # The listeners in lane.py shouldn't call site_configuration_has_changed
+        lane_changed.assert_not_called()
+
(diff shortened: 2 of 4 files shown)


core/scripts.py
```diff
@@ -3420,9 +3420,21 @@ def should_process_lane(self, lane):
...
     def process_lane(self, lane):
         """Update the estimated size of a Lane."""
+
+        # We supress the configuration changes updates, as each lane is updated
+        # and call the site_configuration_has_changed function once after this
+        # script has finished running.
+        #
+        # This is done because calling site_configuration_has_changed repeatedly
+        # was causing performance problems, when we have lots of lanes to update.
+        lane._suppress_configuration_changes = True
```
io7m, line 3430: Is this ever set back to `False` afterwards? Maybe I'm missing it.

io7m: I'd be inclined to wrap `update_size()` in a `try ... finally` and set the flag back to `False` in the `finally`.

jonathangreen: My thought was, that since `_suppress_configuration_changes` is set at on the instance, and not persisted to the database, that there is no need to set back to false, since any new loads will give new lane objects with the default of `False`, and the lifetime of these `Lane` objects is just within the script, where we want the config changes suppressed. 

I pushed another commit to make this a bit more explicit, but defining `_suppress_configuration_changes` on the instance, inside the `__init__` instead of on the class. Previously it was defined on the class, but only set at the instance level.

io7m: Yeah, my comment only applies to the class variable version. :slightly_smiling_face:

io7m (approved):
I don't think this does any serious harm. :laughing:

tests/core/test_scripts.py
```diff
@@ -3287,6 +3287,34 @@ def test_should_process_lane(self, db: DatabaseTransactionFixture):
...
+        lane2.size = 50
+
+        # Commit changes to the DB so the lane creation listeners are fired
+        db.session.commit()
+
+        with patch("core.lane.site_configuration_has_changed") as lane_changed:
+            with patch(
+                "core.scripts.site_configuration_has_changed"
+            ) as scripts_changed:
+                UpdateLaneSizeScript(db.session).do_run(cmd_args=[])
```
tdilauro, line 3307: Fun fact: You can write that as a single `with` statement and, thus, add only a single level of indenting. Warning: it looks hideous. 😱

```
with patch(
    "core.lane.site_configuration_has_changed"
) as lane_changed, patch(
    "core.scripts.site_configuration_has_changed"
) as scripts_changed:
    UpdateLaneSizeScript(db.session).do_run(cmd_args=[])
```

You could also use backslashes for continuation to get each context manager on it's own line, but I'm not sure how our linter overlords will feel about that. 😅

Python 3.10 supports [parenthesized context managers](https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers), so we might just want to wait until that's our minimum support level before doing this kind of combining. But imagine...

```
with (
    patch("core.lane.site_configuration_has_changed") as lane_changed,
    patch("core.scripts.site_configuration_has_changed") as scripts_changed,
):
    UpdateLaneSizeScript(db.session).do_run(cmd_args=[])
```

jonathangreen: 🙏🏻 Thank you. I didn't know you could use multiple context managers like that.

PR #786, opened 2023-02-07
Remove 🔥 unused core config

jonathangreen:
## Description

This is a follow up PR to #785. As part of that work I noticed how much cursed 💀 💣 code is buried in the core config class. This code seems like its leftover from before config settings were added to the database, and config used to be loaded from a file. 

Most of the code in here is actually dead, and never called. And what is being called, was just returning defaults, since a config file was never loaded. 

This PR refactors the small bits of the class that are left, and removes as much of the old code as possible.

## Motivation and Context

The old dead code makes refactoring hard, and imposes an additional cognitive burden when trying to understand the CM code. Cleaning it up, as we see it, will help us reduce complexity.

This PR is in draft right now, because it is targeting the branch from #785. Once that PR goes in, then this can retarget to main. I broke this out to two PRs to make review easier.

## How Has This Been Tested?

Ran unit tests

## Checklist


core/external_list.py
```diff
@@ -4,134 +4,18 @@
...
     Edition,
     Identifier,
     Subject,
-    Work,
     get_one_or_create,
 )
 from .util.datetime_helpers import utc_now
 
 
-class CustomListFromCSV(CSVMetadataImporter):
```
RishiDiwanTT, line 22: Was this poor chap just collateral damage?

scripts.py
```diff
@@ -163,48 +156,6 @@ def q(self):
         )
 
 
-class UpdateStaffPicksScript(Script):
```
RishiDiwanTT, line 166: Flamethrowers cast a wide net I suppose :fire:

RishiDiwanTT (approved):
:fire: :fire:

scripts.py
jonathangreen: This one was removed because it used the old configuration settings to configure the script: 
`Configuration.integration_url(Configuration.STAFF_PICKS_INTEGRATION, True)`

Since we never setup this config file, it seemed like this script was dead as well.

core/external_list.py
jonathangreen: This got removed because it was only used by the `UpdateStaffPicksScript`. So it seemed good to remove it. I should have annotated that when I submitted the PR since its not at all clear why I removed these. Thanks for checking on them.

PR #791, opened 2023-02-07
Automate overdrive advantage account setup

dbernstein:
## Description
This PR contains a new script and a test for importing new advantage accounts associated with a main account. 
It should not yet be merged because it currently is importing all accounts associated with a main account.   Some Overdrive advantage libraries are not signed up with Palace and so should not be imported into the system.  Until we can figure out how to prevent this from happening,  we should hold off on merging this PR.
## Motivation and Context
https://www.notion.so/lyrasis/Automatically-import-OD-Advantage-Accounts-7f604219d69a421a881c3711c91cdfb7

## How Has This Been Tested?

It has been tested manually and the unit test passes.

## Checklist

### DIFF
--- a/bin/import_new_overdrive_advantage_accounts
+++ b/bin/import_new_overdrive_advantage_accounts
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+"""Import new overdrive advantage collections associated with a main overdrive account"""
+import os
+import sys
+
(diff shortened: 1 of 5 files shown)


jonathangreen:
I'm going to close this one, since we implemented it in the CSV export. @dbernstein feel free to reopen if I shouldn't have closed it.

PR #792, opened 2023-02-08
Remove 🔥 metadata wrangler

jonathangreen:
## Description

Since I've been working on 🔥 🧨  unused code, all the metadata wrangler stuff seemed like a good target as well.

## Motivation and Context

Since the typical bug rate per LOC is [[1]](https://stackoverflow.com/questions/2898571/basis-for-claim-that-the-number-of-bugs-per-line-of-code-is-constant-regardless):
> 1-25 errors per 1000 lines of code for delivered software

This PR is statistically fixing between 3 and 75 bugs 😄 

## How Has This Been Tested?

Ran tests

## Checklist


io7m (approved):
This looks good to me! I don't think there's anyone still running a MW anywhere on the planet, so I'm glad to see it gone.

PR #801, opened 2023-02-09
Make passwords required (in most cases)

io7m:
## Description

This updates the code in the Admin API controller that handles password changes. Now, passwords are always required for new admins.

There are numerous issues with the original password checking code that will require a lot of refactoring, but these are all outside of the scope of the ticket associated with this work.

## Motivation and Context

Passwords have been practically required for a while, but the code doesn't enforce the fact that they are.

Affects: https://www.notion.so/lyrasis/Make-password-a-required-field-when-creating-an-account-on-the-CM-ab4313f97fc4465687dfbafc9f64dc1e

## How Has This Been Tested?

I created a new admin locally, and ensured all of the existing tests pass.

## Checklist

### DIFF
--- a/tests/api/admin/controller/test_individual_admins.py
+++ b/tests/api/admin/controller/test_individual_admins.py
@@ -738,3 +738,46 @@ class TestIndividualAdmins(SettingsControllerTest):
 
         [role] = admin_match.roles
         assert AdminRole.SYSTEM_ADMIN == role.role
+
+    def test_individual_admins_post_create_second_admin(self):
+        """Creating a second admin with a password works."""
+        for admin in self._db.query(Admin):
+            self._db.delete(admin)
+
+        system_admin, ignore = create(self._db, Admin, email=self._str)
+        system_admin.add_role(AdminRole.SYSTEM_ADMIN)
+
+        with self.request_context_with_admin("/", method="POST", admin=system_admin):
+            flask.request.form = MultiDict(
+                [
+                    ("email", "second_admin@nypl.org"),
+                    ("password", "pass"),
+                    ("roles", []),
+                ]
+            )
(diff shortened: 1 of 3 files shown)


io7m:
I should point out that the Admin web interface won't show the password field as being required. I believe it's a one line change in that project to make it show up as required, but I don't have permission to push to the `circulation-admin` project:

```
diff --git a/src/components/IndividualAdminEditForm.tsx b/src/components/IndividualAdminEditForm.tsx
index 4e7652d5..cbf281fe 100644
--- a/src/components/IndividualAdminEditForm.tsx
+++ b/src/components/IndividualAdminEditForm.tsx
@@ -134,7 +134,7 @@ export default class IndividualAdminEditForm extends React.Component<
             name="password"
             label="Password"
             ref={this.passwordRef}
-            required={this.context.settingUp}
+            required={true}
             error={this.props.error}
           />
         )}
```

tests/api/admin/controller/test_individual_admins.py
```diff
@@ -676,9 +691,12 @@ def test_individual_admins_post_create_on_setup(self):
                 AdminNotAuthorized,
                 self.manager.admin_individual_admin_settings_controller.process_post,
             )
-            self._db.rollback()
 
-        # The password is required.
+    def test_individual_admins_post_create_requires_password(self):
```
RishiDiwanTT, line 695: This tests scenario 1, setting up the first admin, but not scenario 2, creating another admin with or without a password.

io7m: Oops, yes, I mangled one of the tests. Fixing... :hourglass:

RishiDiwanTT (approved):
Looks Good :+1:

PR #802, opened 2023-02-09
Remove Google OAuth admin authentication provider 

pcustic:
## Description

This PR removes Google OAuth admin authentication provider (`GoogleOAuthAdminAuthenticationProvider`) and all the code related to id. However, it leaves supporting code for having another external authentication provider even though Google OAuth was the only one currently implemented.

## Motivation and Context

The circulation manager has an admin authentication integration meant to allow admins to authenticate via Google to log into the circulation manager. We are not using this integration anywhere, and the library that it is built on oauth2client is deprecated and not meant to be used anymore.

[Notion](https://www.notion.so/lyrasis/Remove-google-oauth-admin-authentication-provider-567b21054fc844cd88415ca52ba66bb2)

## How Has This Been Tested?

Removed tests that used Google OAuth, all other tests passed. Also, manually tested sign-in functionality.

## Checklist


pcustic:
I see there is significant drop in the code coverage in `api/admin/controller/admin_auth_services.py`, I'll try to figure out what lines are currently missed and try to add some more tests.

pcustic:
@jonathangreen I left the option (existing code) for other external admin auth providers in the future, is that OK? Should I just completely remove option for other auth providers except classic username/password?

jonathangreen:
@pcustic I would completely remove the code for external admin auth providers. That should help the code coverage. We can revist it if we need to in the future, but having dead code around that is never called just slows down our development.

pcustic:
> @pcustic I would completely remove the code for external admin auth providers. That should help the code coverage. We can revist it if we need to in the future, but having dead code around that is never called just slows down our development.

Sure, makes sense 👍

api/admin/routes.py
```diff
@@ -393,22 +387,6 @@ def collection_library_registrations():
     )
 
 
-@app.route("/admin/admin_auth_services", methods=["GET", "POST"])
```
pcustic, line 396: This will require update to the admin UI, we will need to remove the Admin Auth from the settings so users on the CM cannot (try) to access these endpoints.

migration/20230214-remove-admin-credential.sql
```diff
@@ -0,0 +1,6 @@
+--
+-- Remove admin credential field since we are not using any external admin auth service.
+--
+
+-- Remove the admin.credential
+ALTER TABLE admins DROP COLUMN if exists credential;
```
pcustic, line 6: This migration would need to be run on deployment to remove the extra column.

pcustic:
@jonathangreen (I think) I removed all the external admin auth stuff. I'll create PR from removing admin auth views in admin UI. 

Btw, I see the codecov report wasn't created for these recent couple of commits, only for one of the first ones.

api/admin/routes.py
pcustic: https://github.com/ThePalaceProject/circulation-admin/pull/75

pcustic:
> @pcustic now that [ThePalaceProject/circulation-admin#75](https://github.com/ThePalaceProject/circulation-admin/pull/75) is merged, and I made a release that includes it https://github.com/ThePalaceProject/circulation-admin/releases/tag/v1.5.1 can you update the package version here: https://github.com/ThePalaceProject/circulation/blob/main/api/admin/config.py#L16
> 
> Once that is updated, this should be good to go.

Updated [0845185](https://github.com/ThePalaceProject/circulation/pull/802/commits/0845185c4fe2dee12a61ecd07d40036543a0867a)

jonathangreen (approved):

PR #804, opened 2023-02-09
Inject dashboard stats generator

tdilauro:
## Description

Factors `Flask` out of the dashboard statistics method and then injects that stats method into the controller.

There is no new functionality.

## Motivation and Context

Improve testability: Isolate the function from the web framework to make testing more straightforward.

It would be good to be able to use this method (and, potentially, others) outside of the `Flask` context, including potentially with other web application frameworks.

## How Has This Been Tested?

- Manual testing locally and adjusted tests work properly.
- CI tests for the branch all pass.

## Checklist


pcustic (approved):
If I understand correctly - this is just old code moved around a bit? Not much new code added, right? Seems fine to me.

PR #805, opened 2023-02-09
Remove 🔥 feedbooks importer

jonathangreen:
## Description

Since I've been working on 🪓  unused code, the feedbooks importer seems like a good target to go. As I understand it was written to do an initial import of the Palace bookshelf. DPLA has since moved on to pull in these books outside of the CM, so this importer isn't used anywhere and can go.

## Motivation and Context

Looking through our automated security scanning alerts, this class throws an automated warning for *Incomplete URL substring sanitization*, which looking at the code is true. Instead of bothering to fix the URL sanitation though its probably better to remove the code entirely.

## How Has This Been Tested?

Running unit tests.

## Checklist


RishiDiwanTT (approved):
Looks pretty :axe: to me

PR #806, opened 2023-02-10
Remove 🔥 the public_key_document endpoint

jonathangreen:
## Description

Building on the work in https://github.com/ThePalaceProject/circulation/pull/792, this removes the `public_key_document` endpoint. This endpoint was only being used by the metadata wrangler integration. 🗑️

## Motivation and Context

This could probably have happened as part of the wrangler removal PR, but I didn't notice it could go as well until after that PR was submitted. 

I queried through the logs on our circulation managers, and going back as far as we have access logs, no one has ever hit the `public_key_document` endpoint.

## How Has This Been Tested?

- Ran tests locally
- Queried logs in cloudwatch

## Checklist


api/controller.py
```diff
@@ -628,24 +625,6 @@ def authentication_for_opds_document(self):
             value = json.dumps(value)
         return value
 
-    @property
```
RishiDiwanTT, line 631: This is identical to the `key_pair` implementation for the `authentication_document` [here](https://github.com/ThePalaceProject/circulation/blob/feature/remove-sitewide-key/api/authenticator.py#L1200).
Does it server the same purpose? Should it also be removed?

jonathangreen: It would be nice to be able to remove that one as well, but the library specific keypair in the authentication_document is used by the library registry to verify the identity of the library, so it can't just be dropped like this one. 

This served the same purpose, but for the metadata wrangler integration. But since that integration is gone, this one can be removed.

RishiDiwanTT (approved):

PR #815, opened 2023-02-14
Palace Marketplace (ODL 2.0) collections with loan and hold limits

RishiDiwanTT:
## Description
Both are implemented as collection settings for ODL2 type collections
Initially, ODL 2 only differed from the ODL 1 implementation with respect to the import format.
This adds a deviation to the loan and hold functions, only ODL 2 supports the loan and hold limits.

## Motivation and Context
Suffolk County Library System requested the ability to configure in the CM the amount of loans and   holds that are allowed for Palace Marketplace content.

[Notion](https://www.notion.so/lyrasis/In-the-Collection-Manager-Add-ability-to-define-the-number-of-loans-and-holds-available-for-Palace--c4efa7e8ccb04e28aa677eee5055ec92)

## Checklist

### DIFF
--- a/api/odl2.py
+++ b/api/odl2.py
@@ -1,10 +1,14 @@
+from __future__ import annotations
+
 import logging
+from typing import TYPE_CHECKING
 
 from contextlib2 import contextmanager
 from flask_babel import lazy_gettext as _
 from webpub_manifest_parser.odl import ODLFeedParserFactory
 from webpub_manifest_parser.opds2.registry import OPDS2LinkRelationsRegistry
 
+from api.circulation_exceptions import PatronHoldLimitReached, PatronLoanLimitReached
 from api.odl import ODLAPI, ODLImporter
 from core.metadata_layer import FormatData
 from core.model import Edition, RightsStatus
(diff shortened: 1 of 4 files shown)


api/odl2.py
```diff
@@ -34,11 +41,68 @@ class ODL2APIConfiguration(OPDSImporterConfiguration):
         default=["text/html"],
     )
 
+    loan_limit = ConfigurationMetadata(
+        key="odl2_loan_limit",
+        label=_("Loan limit per patron"),
+        description=_(
+            "The number of assets a patron can have loaned out at any given time."
```
RishiDiwanTT, line 48: Maybe "assets" is not the right word

tests/fixtures/odl.py
```diff
@@ -260,3 +263,47 @@ def odl_api_test_fixture(odl_test_fixture: ODLTestFixture) -> ODLAPITestFixture:
     return ODLAPITestFixture(
         odl_test_fixture, library, collection, work, license, api, patron, client
     )
+
+
+class ODL2TestFixture(ODLTestFixture):
```
RishiDiwanTT, line 268: @io7m I have used inheritance here for the ODL2 fixtures because they ODL2 API has almost no deviations from the ODL1 API.
This PR introduces the only changes between the 2 versions with respect to API actions.

io7m: Normally I wouldn't like this, but there is an inheritance relationship between ODLAPI and ODL2API, so it makes sense that the fixtures might end up having one as well.

io7m (approved):
This looks good to me.

io7m:
That coverage reduction is due to code that isn't really "executed" as such.

PR #751, opened 2023-01-12
Refactor a lot of API tests to use fixtures

io7m:
## Description

This refactors any API tests that weren't in a submodule (LCP, SAML, etc) such that they now use fixtures.

This work had already been performed for the core tests, so all of the "how" and "why" is explained in that original PR:

https://github.com/ThePalaceProject/circulation/pull/637

## Motivation and Context

See above!

The ticket for the API test suite is here: https://www.notion.so/lyrasis/Refactor-API-test-suite-9852f7a8e4484cc6bed8235acb20a7fe

## How Has This Been Tested?

All tests were run locally, repeatedly.

## Checklist

- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

PR #825, opened 2023-02-16
Migrate tests/api/clever to fixtures

pcustic:
## Description

This is the continuation of the effort (started by @io7m) to move CM tests to use fixtures instead of class inheritance. 

## Motivation and Context

Motivation is to move all the tests to use fixtures.
[Notion](https://www.notion.so/lyrasis/Refactor-Clever-API-test-suite-83cbf3c6ec7849848d1a66f0063e7e21)

More can be read in previous PRs and tickets. 
[Here](https://github.com/ThePalaceProject/circulation/pull/637) and [here](https://github.com/ThePalaceProject/circulation/pull/751)

## How Has This Been Tested?

Ran all changed unit tests.

## Checklist


tests/api/clever/test_clever.py
```diff
@@ -89,190 +93,238 @@ def mock_integration(self):
         integration.setting(MockAPI.OAUTH_TOKEN_EXPIRATION_DAYS).value = 20
         return integration
```
io7m, line 94: I think it was wrong for the original code to make this a `@property`; accessing it has a side effect (creates a new external integration every time). I suspect someone wrote this not realizing the original method had a side effect. The new fixtures make it clearer when something has a side effect.

tests/api/clever/test_clever.py
```diff
@@ -67,10 +68,13 @@ def test_external_type_from_clever_grade(self):
...
-        self.api = MockAPI(self._default_library, self.mock_integration)
+class CleverAuthenticationFixture:
+    db: DatabaseTransactionFixture
+    api: MockAPI
+
+    def __init__(self, db: DatabaseTransactionFixture):
+        self.db = db
+        self.api = MockAPI(db.default_library(), self.mock_integration)
         os.environ["AUTOINITIALIZE"] = "False"
         from api.app import app
```
io7m, line 79: I think this carries the risk of polluting the environment for further test runs, unless I'm wrong.

The code should probably set `os.environmen["AUTOINITIALIZE"]` and then restore it in a `close` method. This will require making the fixture method into one that yields a value and then performs the close operation afterwards. Take a look at, for example, the end-to-end search fixture:

https://github.com/ThePalaceProject/circulation/blob/main/tests/fixtures/search.py#L278

io7m (changes requested):
Think this needs a couple of small changes.

tests/api/clever/test_clever.py
pcustic: Removed `@property` https://github.com/ThePalaceProject/circulation/pull/825/commits/a537f274feadb80e1fa9336a1f313e7f2a35b5a6

tests/api/clever/test_clever.py
pcustic: make sense 👍 https://github.com/ThePalaceProject/circulation/pull/825/commits/a537f274feadb80e1fa9336a1f313e7f2a35b5a6

tests/api/clever/test_clever.py
```diff
@@ -67,20 +74,30 @@ def test_external_type_from_clever_grade(self):
...
+
         os.environ["AUTOINITIALIZE"] = "False"
         from api.app import app
 
+        fixture.app = app
+
+        return fixture
+
+    def close(self):
         del os.environ["AUTOINITIALIZE"]
```
io7m, line 96: To be pedantic, it should probably restore the old value of the environment variable here. We don't actually know if it was unset before we started, so we should probably leave it how we found it.

io7m: Er, what I mean is:

1. Save the old value
2. Set the new value to `False`
3. Restore the old value in `close()`

jonathangreen: The `monkeypatch` fixture that pytest supplies, automates the process for patching env variables. Maybe we want to use that here instead of doing this ourselves? 

https://docs.pytest.org/en/6.2.x/monkeypatch.html#monkeypatching-environment-variables

something like: 

```python
def clever_fixture(
    db: DatabaseTransactionFixture, monkeypatch: MonkeyPatch
):
    fixture = CleverAuthenticationFixture.create(db)
    monkeypatch.setenv("AUTOINITIALIZE", "False")
    yield fixture
    fixture.close()
```

pcustic: Monkeypatching is neat 👍  I went with that in https://github.com/ThePalaceProject/circulation/pull/825/commits/b8bb780e9a1ee0129c6668f7e01154814f0d8e1c and removed the create/yield/close/del pattern and reverted back to something similar to the first version, just using monkeypatch with it.

io7m (approved):
This seems OK now. :+1:

PR #833, opened 2023-02-17
Fix links in OPDS feeds containing facets objects

jonathangreen:
## Description

Fix a bug introduced in https://github.com/ThePalaceProject/circulation/pull/785

## Motivation and Context

This relates to work @io7m is doing to re-enable some tests that stopped working in https://github.com/ThePalaceProject/circulation/pull/751

## How Has This Been Tested?

Running the tests on this branch.

## Checklist

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -964,7 +964,6 @@ class OPDSFeedController(CirculationManagerController):
             "acquisition_groups",
             lane_identifier=lane_identifier,
             library_short_name=library.short_name,
-            _facets=facets,
             _external=True,
         )
 
@@ -1005,7 +1004,6 @@ class OPDSFeedController(CirculationManagerController):
             "feed",
             lane_identifier=lane_identifier,
             library_short_name=library_short_name,
-            _facets=facets,
             _external=True,
         )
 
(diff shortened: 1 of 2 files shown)


api/controller.py
```diff
@@ -1005,7 +1004,6 @@ def feed(self, lane_identifier, feed_class=AcquisitionFeed):
             "feed",
             lane_identifier=lane_identifier,
             library_short_name=library_short_name,
-            _facets=facets,
```
jonathangreen, line 1008: Same reason as above, we can just drop `_facets` here.

tests/api/test_controller_opdsfeed.py
```diff
@@ -193,7 +193,7 @@ def page(cls, **kwargs):
                 "feed",
                 lane_identifier=lane_id,
                 library_short_name=circulation_fixture.db.default_library().short_name,
-                _facets=load_facets_from_request(),
+                _external=True,
```
jonathangreen, line 196: In these tests we can drop the `_facets` parameter for the same reason. But we have to add the `_external=True` parameter because the the old `url_for` function used to implicitly add it:
https://github.com/ThePalaceProject/circulation/blob/6ba5bf6a9d6bcd7b99dc3ff84c69067a17c52a9a/api/controller.py#L396-L399

io7m (approved):
Looks good!

io7m:
> Base: 55.03% // Head: 89.36% // Increases project coverage by +34.32%

Feel like these may not be entirely accurate...

jonathangreen:
I think its because we are lacking a token: https://github.com/ThePalaceProject/circulation/pull/832

Seems like some of the requests to record coverage are failing, and if it happens to miss the coverage for the api tests it thinks coverage has dropped significantly

PR #834, opened 2023-02-17
Fix pytest warning

jonathangreen:
## Description

Fix pytest warning that I introduced in https://github.com/ThePalaceProject/circulation/pull/765

```
tests/core/test_app_server.py::TestApplicationVersionController::test_version[None-None-None-None-None]
test_app_server.py:50: PytestWarning: Value of environment variable TPP_CIRCULATION_ADMIN_PACKAGE_NAME type should be str, but got None (type: NoneType); converted to str implicitly
    monkeypatch.setenv(AdminUiConfig.ENV_ADMIN_UI_PACKAGE_NAME, ui_package)

tests/core/test_app_server.py::TestApplicationVersionController::test_version[None-None-None-None-None]
test_app_server.py:51: PytestWarning: Value of environment variable TPP_CIRCULATION_ADMIN_PACKAGE_VERSION type should be str, but got None (type: NoneType); converted to str implicitly
    monkeypatch.setenv(AdminUiConfig.ENV_ADMIN_UI_PACKAGE_VERSION, ui_version)
```

## Motivation and Context

Warning is fairly benign, but its good to reduce the number of warnings that we have in our tests.

## How Has This Been Tested?

Ran tests on the branch.

## Checklist

### DIFF
--- a/tests/core/test_app_server.py
+++ b/tests/core/test_app_server.py
@@ -47,8 +47,12 @@ class TestApplicationVersionController:
         monkeypatch.setattr(core, "__branch__", branch)
 
         # Mock the admin ui version strings
-        monkeypatch.setenv(AdminUiConfig.ENV_ADMIN_UI_PACKAGE_NAME, ui_package)
-        monkeypatch.setenv(AdminUiConfig.ENV_ADMIN_UI_PACKAGE_VERSION, ui_version)
+        monkeypatch.setenv(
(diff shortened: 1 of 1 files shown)


jonathangreen:
Good catch @io7m, I may be been a bit trigger happy on this one 😅. I updated the PR to use `delenv` when its None, which should remove the warning, and preserve the meaning of the test.

io7m (approved):
This seems OK now.

_insert generic public service announcement about rampant monkeypatching_

PR #835, opened 2023-02-17
Fix pytest warnings and (masked) failing tests

io7m:
## Description

This fixes the `pytest` warnings that aren't related to deprecation. I can take a look at the deprecation warnings separately if you like.

## Motivation and Context

We're trying to get towards having no warnings at all in the test suite, with the hope that we can eventually turn warnings into errors.

https://www.notion.so/lyrasis/Fix-pytest-warnings-a4cf9fc8ec3c4b5ea8d62633f1d60ba6

## How Has This Been Tested?

I ran the test suite.

## Checklist


api/controller.py
```diff
@@ -1037,7 +1037,12 @@ def navigation(self, lane_identifier):
         )
         annotator = self.manager.annotator(lane, facets)
         return NavigationFeed.navigation(
-            self._db, title, url, lane, annotator, facets=facets
+            _db=self._db,
+            title=title,
+            url=url,
+            worklist=lane,
+            annotator=annotator,
+            facets=facets,
```
jonathangreen, line 1045: Why does this call need to be updated?

tests/api/test_controller_opdsfeed.py
```diff
@@ -745,7 +740,7 @@ def test_qa_feed(self, circulation_fixture: CirculationControllerFixture):
 
         # For the most part, we're verifying that the expected values
         # are passed in to _qa_feed.
-        assert AcquisitionFeed.groups == kwargs.pop("feed_method")  # type: ignore
+        assert AcquisitionFeed.groups == kwargs.pop("feed_factory")  # type: ignore
```
jonathangreen, line 743: What caused this kwarg to change?

api/controller.py
io7m: The error was that the test code was failing to pass in a required positional `annotator` argument. Since it's not actually needed in the test code, I made them all into named arguments.

tests/api/test_controller_opdsfeed.py
io7m: I actually don't know. Presumably some recent refactoring inside the CM that would have caused this test to fail had it actually been running. I inspected the `kwargs` array in the debugger, and the bound method was returned via `feed_factory` instead of `feed_method`.

jonathangreen (approved):

PR #838, opened 2023-02-20
Attempt to require passwords (again)

io7m:
## Description

The original code to handle the creation or editing of admins had a serious bug in that, even if the endpoint returned an error, the code would still commit the admin that it might have created in the process of trying to check if the admin existed. This refactors the method into something much more explicit, where the three possible code paths are kept distinct, and an explicit rollback is made if anything other than success occurs.

There are still problems with this from a UI perspective, and I think it's something that needs design work. For example, the admin UI always says "Updated password successfully" even if you didn't specify a password (and therefore didn't ask the admin password to be updated). It always marks the password field as being required, when in fact it's only required if you want to update the password. The user interface doesn't give any way to express intent.

These changes at least mean that the server side always does the right thing, but the UI needs serious work.

## Motivation and Context

We're trying to make sure that admins can never have missing (or empty) passwords.

Affects: https://www.notion.so/lyrasis/Make-password-required-when-changing-already-existing-admin-account-information-in-the-CM-36df0e10860547d4a6a693eea452f37d

## How Has This Been Tested?

I tried creating new admins, and editing existing admins. I also ran the test suite, and added a new test.

## Checklist

### DIFF
--- a/api/admin/controller/individual_admin_settings.py
+++ b/api/admin/controller/individual_admin_settings.py
@@ -98,6 +98,104 @@ class IndividualAdminSettingsController(SettingsController):
             individualAdmins=admins,
         )
 
+    def process_post_create_first_admin(self, email: str):
+        """Create the first admin in the system."""
+
+        # Passwords are always required, so check presence and validity up front.
+        password: Optional[str] = flask.request.form.get("password")
+        if not self.is_acceptable_password(password):
+            return self.unacceptable_password()
+
+        success = False
+        try:
+            admin, _ = get_one_or_create(self._db, Admin, email=email)
+            self.check_permissions(admin, settingUp=True)
+
+            # Update the roles, if requested.
+            roles = flask.request.form.get("roles")
+            if roles:
+                roles = json.loads(roles)
+            else:
+                roles = []
+
+            roles_error = self.handle_roles(admin, roles, settingUp=True)
+            if roles_error:
+                return roles_error
+
+            # Update the password, if requested.
+            self.handle_password(password, admin, is_new=True, settingUp=True)
+
+            success = True
+            return self.response(admin, is_new=True)
+        finally:
+            if not success:
+                self._db.rollback()
+
+    def process_post_create_new_admin(self, email: str):
+        """Create a new admin (not the first admin in the system)."""
+
+        # Passwords are always required, so check presence and validity up front.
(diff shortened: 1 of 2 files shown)


io7m:
Going to update a couple of the tests for better coverage. Those new code coverage annotations are pretty good!

api/admin/controller/individual_admin_settings.py
```diff
@@ -98,6 +98,104 @@ def append_role(roles, role):
...
+        if not self.is_acceptable_password(password):
+            return self.unacceptable_password()
+
+        success = False
+        try:
+            admin, _ = get_one_or_create(self._db, Admin, email=email)
+            self.check_permissions(admin, settingUp=True)
+
+            # Update the roles, if requested.
+            roles = flask.request.form.get("roles")
```
io7m, line 115: You'll see this code repeated a couple of times.

My feeling was that, due to how twisted the rest of the code is and given the critical nature of authenticating and creating admins, it was more important to have something straightforward and readable for each case rather than save a few lines and thread more "are we setting up", "are we creating a new admin" flags through the existing code.

api/admin/controller/individual_admin_settings.py
```diff
@@ -98,6 +98,104 @@ def append_role(roles, role):
...
+
+            # Update the password, if requested.
+            self.handle_password(password, admin, is_new=True, settingUp=True)
+
+            success = True
+            return self.response(admin, is_new=True)
+        finally:
+            if not success:
+                self._db.rollback()
+
```
io7m, line 133: This is the main issue that was being fixed; previously, the code could "fail" and yet _still_ create an admin with an empty password.

api/admin/controller/individual_admin_settings.py
```diff
@@ -98,6 +98,104 @@ def append_role(roles, role):
...
+            roles_error = self.handle_roles(admin, roles, settingUp=False)
+            if roles_error:
+                return roles_error
+
+            # Update the password, if requested.
+            self.handle_password(password, admin, is_new=True, settingUp=False)
+            success = True
+            return self.response(admin, is_new=True)
+        finally:
```
RishiDiwanTT, line 162: Smart use of finally :D

RishiDiwanTT (approved):
Looks Good :+1:

PR #845, opened 2023-02-22
Migrate tests/api/lcp to fixtures

pcustic:
## Description

This is the continuation of the effort (started by @io7m) to move CM tests to use fixtures instead of class inheritance. 

## Motivation and Context

Motivation is to move all the tests to use fixtures.
[Notion](https://www.notion.so/lyrasis/Refactor-LCP-API-test-suite-b14c81e3a77d45f587c55eb30ebe3499)

More can be read in previous PRs and tickets. 
[Here](https://github.com/ThePalaceProject/circulation/pull/637) and [here](https://github.com/ThePalaceProject/circulation/pull/751)

## How Has This Been Tested?

Ran all changed unit tests.

## Checklist


tests/api/lcp/test_controller.py
```diff
@@ -10,13 +10,13 @@
 from api.lcp.server import LCPServer
 from core.lcp.credential import LCPCredentialFactory, LCPUnhashedPassphrase
 from core.model import ExternalIntegration
-from tests.api.lcp import fixtures
-from tests.api.test_controller import ControllerTest
+from tests.api.lcp import lcp_strings
```
pcustic, line 13: I renamed this file since it is just a bunch of strings and also not to confuse with fixtures we are using now.

@io7m I decided to leave this just as strings, not to convert it to fixture that contain strings. I know there is this [block](https://www.notion.so/lyrasis/Test-Fixtures-in-the-CM-4848df744ef141d0ab48392c448f60dd?pvs=4#93e056c47a474122a279c435d3148e98) in the fixtures docs but I'm not sure there is any benefits to moving this to fixture since these are just strings, no other data manipulation or side effects. 
Let me know what you think, if you think this would be better to convert to some kind of `LCPStringsFixture` or something like that I'm open to that.

io7m: I think it's fine to just have a module full of constants. :+1:

io7m: The difference is that those LCP strings are constants. The example data in the documentation are actually created at runtime and so are "different" every time.

tests/api/lcp/test_controller.py
```diff
@@ -35,14 +36,14 @@ def test_get_lcp_passphrase_returns_the_same_passphrase_for_authenticated_patron
             patron = controller_fixture.default_patron
             manager = CirculationManager(controller_fixture.db.session, testing=True)
             controller = LCPController(manager)
-            controller.authenticated_patron_from_request = MagicMock(
+            controller.authenticated_patron_from_request = MagicMock(  # type: ignore
```
pcustic, line 39: error: Cannot assign to a method  [assignment]

tests/api/lcp/test_server.py
```diff
@@ -219,10 +219,10 @@ def test_generate_license(
             configuration.max_printable_pages = max_printable_pages
             configuration.max_copiable_pages = max_copiable_pages
 
-            lcp_server_fixture.credential_factory.get_patron_id = MagicMock(
+            lcp_server_fixture.credential_factory.get_patron_id = MagicMock(  # type: ignore
```
pcustic, line 222: error: Cannot assign to a method  [assignment]

tests/api/lcp/test_server.py
```diff
@@ -219,10 +219,10 @@ def test_generate_license(
...
                 return_value=expected_patron_id
             )
-            lcp_server_fixture.credential_factory.get_patron_passphrase = MagicMock(
+            lcp_server_fixture.credential_factory.get_patron_passphrase = MagicMock(  # type: ignore
```
pcustic, line 225: error: Cannot assign to a method  [assignment]

io7m (approved):
Looks good!

We must be getting pretty close to being able to delete a ton of "test support" code in the CM now...

PR #850, opened 2023-02-23
Language expansion and contraction for the JSON search endpoint

RishiDiwanTT:
## Description
In the search field value API expand the language codes into full english names.
In the json search API contract the english language names into language codes again.
This does not mean language codes cannot be used in the API at all, they will still work in the search.

Incidental changes: In #702 a change was made to search with some fuzziness on text fields. This also added fuzziness to the language field. This is not required for languages, so this PR also fixes that inadvertent change.

## Motivation and Context
When creating a list and searching using the language filter, the language field executes the search based on language code instead of the full name of a language. For example, the user must search for “Eng” instead of “English” to return the expected result of materials written in English. If the user searches for materials in English, no results are returned.

[Notion](https://www.notion.so/lyrasis/Add-support-to-list-creation-search-for-full-name-of-language-88cbaa729ee8429789ace553bc382f85)

## How Has This Been Tested?
Unit tests have run.
Manually tested the JSON search endpoint.
Example: http://localhost:6500/localtest/search?q={%22query%22:{%22value%22:%22english%22,%22key%22:%22language%22}}&search_type=json

## Checklist

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -6,7 +6,17 @@ import os
 import sys
 import urllib.parse
 from datetime import date, datetime, timedelta
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    TypeVar,
+    Union,
+)
 
 import flask
 from flask import Request, Response, redirect, url_for
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -2413,12 +2413,14 @@ class AdminSearchController(AdminController):
         collection_ids = [coll.id for coll in library.collections]
         return self._search_field_values_cached(collection_ids)
 
+    @classmethod
+    def _unzip(cls, values: List[Tuple[str, int]]) -> dict:
+        """Covert a list of tuples to a {value0: value1} dictionary"""
+        return {a[0]: a[1] for a in values if type(a[0]) is str}
+
     # 1 hour in-memory cache
     @memoize(ttls=3600)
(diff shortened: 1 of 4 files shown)


core/external_search.py
```diff
@@ -2052,7 +2053,9 @@ class Operators(Values):
         "identifiers.identifier": dict(path="identifiers"),
         "identifiers.type": dict(path="identifiers"),
         "imprint": _KEYWORD_ONLY,
-        "language": dict(),
+        "language": dict(
+            type="_text"
+        ),  # Made up keyword type, because we don't want text fuzzyness on this
```
RishiDiwanTT, line 2058: This change avoids fuzziness on the language field, since it is no longer of `type=text` (by default)

api/admin/controller/__init__.py
```diff
@@ -2266,7 +2277,7 @@ def search_field_values(self) -> dict:
     # 1 hour in-memory cache
     @memoize(ttls=3600)
     def _search_field_values_cached(self, collection_ids: List[int]) -> dict:
-        def _unzip(values: List[tuple]) -> dict:
+        def _unzip(values: List[Tuple[str, int]]) -> dict:
```
dbernstein, line 2280: Just curious:  is there any reason to have the  _unzip function defined within _search_field_values_cached?  I'm not sure how the python interpreter works on this level, but I'm wondering if it would be more clear from a readability as well as code execution point of view  to have this method moved up to the class level.

dbernstein (approved):
@RishiDiwanTT:  I defer to your python expertise regarding the comment below.  Change it if you think it is warranted.  Otherwise it looks good to me.

RishiDiwanTT:
> @RishiDiwanTT: I defer to your python expertise regarding the comment below. Change it if you think it is warranted. Otherwise it looks good to me.

This was good feedback, I made the change. From a performance perspective the inner code block is not copied in the interpretter every time the method is run, just  a new pointer to the code block is created for that method.

PR #855, opened 2023-02-25
Upgrade Flask and Werkzeug

jonathangreen:
## Description

Upgrade the version of Flask and Werkzeug we are using to the latest versions.

## Motivation and Context

There are some outstanding CVE against Werkzeug that make it desirable to upgrade it. However once we start upgrading we quickly end up in a dependency hell 🔥 that ends in `flask-sqlalchemy-session` not supporting the most recent flask version.

Since [flask-sqlalchemy-session](https://github.com/dtheodor/flask-sqlalchemy-session) hasn't seen an update since 2015, and is very small, only about 50 lines of code. We pull that code into `core` and apply some patches to get it working with the latest versions of flask and werkzeug.

## How Has This Been Tested?

Tested by running CI against the code.

## Checklist


api/admin/controller/__init__.py
```diff
@@ -386,7 +388,11 @@ def __call__(self, collection, book, path=None):
         if not setting_up:
             admin = self.authenticated_admin_from_request()
             if isinstance(admin, ProblemDetail):
-                redirect_url = flask.request.url
+                url = urlparse(flask.request.url)
```
jonathangreen, line 391: At some point, flask started returning the unescaped URL in `flask.request.url`, however we were expected the escaped URL here. So the new code does this escaping.

api/controller.py
```diff
@@ -2693,12 +2693,10 @@ def revoke_hold(self, collection_name, hold_id):
...
 class StaticFileController(CirculationManagerController):
     def static_file(self, directory, filename):
-        cache_timeout = ConfigurationSetting.sitewide(
+        max_age = ConfigurationSetting.sitewide(
             self._db, Configuration.STATIC_FILE_CACHE_TIME
         ).int_value
-        return flask.send_from_directory(
-            directory, filename, cache_timeout=cache_timeout
-        )
+        return flask.send_from_directory(directory, filename, max_age=max_age)
```
jonathangreen, line 2699: Name of the `cache_timeout` parameter changed to `max_age` in the flask API.

pyproject.toml
```diff
@@ -248,8 +247,7 @@ tox-gh-actions = "^2.10.0"
 boto3-stubs = "^1.26.71"
 botocore-stubs = "^1.29.71"
 freezegun = "~1.2"
-Jinja2 = "2.11.3"
-markupsafe = "2.0.1"
```
jonathangreen, line 252: We had `markupsafe` as a dependency here, just to pin the version, we don't actually rely on it directly. So we are able to drop it now that we can update flask & werkzeug.

tests/api/test_axis.py
```diff
@@ -1831,8 +1831,11 @@ class TestAxis360AcsFulfillmentInfo:
     @pytest.fixture
     def mock_request(self):
         # Create a mock request object that we can use in the tests
+        response = MagicMock(return_value="")
+        type(response).headers = PropertyMock(return_value=[])
+        type(response).status = PropertyMock(return_value=200)
```
jonathangreen, line 1836: This test needed to be updated, because werkzeug started doing some parameter validation, and the old mock returned a status code of "" (empty string), which caused these tests to fail. Updating the mock to return a valid status code (200) makes the tests pass again.

tests/api/test_opds.py
```diff
@@ -1243,11 +1243,9 @@ def test_active_loan_feed(self, annotator_fixture: LibraryAnnotatorFixture):
         # last_loan_activity_sync is tracked at the millisecond level
         # and Last-Modified is tracked at the second level.)
 
-        # Putting the last loan activity sync into an Flask Response
-        # strips timezone information from it,
-        # so to verify we have the right value we must do the same.
-        last_sync_naive = patron.last_loan_activity_sync.replace(tzinfo=None)
-        assert (last_sync_naive - response.last_modified).total_seconds() < 1
+        assert (
+            patron.last_loan_activity_sync - response.last_modified
```
jonathangreen, line 1247: `response.last_modified` now returns a timezone aware datetime, so we no longer have to strip off the timezone for this test.

tests/core/test_flask_sqlalchemy_session.py
```diff
@@ -0,0 +1,95 @@
...
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session, sessionmaker
+
+from core.flask_sqlalchemy_session import current_session, flask_scoped_session
+
+
+@pytest.fixture
+def sqlite_engine():
+    """Return an sqlite engine"""
+    return create_engine("sqlite://")
```
jonathangreen, line 14: These are the original tests from the upstream library we pulled in. The old tests used sqllite, so I left them like that, although we could change them to use postgres instead here.

api/admin/controller/__init__.py
RishiDiwanTT: Don't url.query and fragment need to be quoted too?

pyproject.toml
```diff
@@ -248,8 +247,7 @@ tox-gh-actions = "^2.10.0"
...
+Jinja2 = "^3.1.2"
```
RishiDiwanTT, line 250: Flask implicitly depends on Jinja2>3.0, do we need this dependency now?

RishiDiwanTT (commented):
Just some minor comments, overall looks good

api/admin/controller/__init__.py
jonathangreen: Good catch, yes. Those are quoted now.

RishiDiwanTT (approved):
Looks Good :+1:

PR #857, opened 2023-02-25
Remove loggly integration

jonathangreen:
## Description

The loggly integration was only used by NYPL. We are not using it on any of our servers, so we can remove it.

## Motivation and Context

Simplifying our code and requirements where possible.

## How Has This Been Tested?

Ran CI.

## Checklist


io7m (approved):
Nice.

:fire:

PR #863, opened 2023-02-28
Fix a regex typo

io7m:
## Description

This regex appears to have been introduced years ago, and matches a few more characters than it should due to a lowercase/uppercase typo. The code scanning tool flags this as a vulnerability, but it probably isn't. It's still a bug worth correcting though.

## Motivation and Context

We're trying to clean up security-related code scanning issues. This was one of them.

Affects: https://github.com/ThePalaceProject/circulation/security/code-scanning/33
Affects: https://www.notion.so/lyrasis/Address-codeql-code-scanning-security-alerts-94b682553e9f43fb98496582278a41b1

## How Has This Been Tested?

I ran the test suite.

## Checklist

### DIFF
--- a/core/model/contributor.py
+++ b/core/model/contributor.py
@@ -365,7 +365,7 @@ class Contributor(Base):
 
     # Regular expressions used by default_names().
     PARENTHETICAL = re.compile(r"\([^)]*\)")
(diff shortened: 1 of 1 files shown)


pcustic (approved):
👌

PR #867, opened 2023-03-01
Refactor strange authprovider inheritance pattern.

jonathangreen:
## Description

Refactor the strange inheritance pattern being used in `AuthenticationProvider`. Where we defined `remote_patron_lookup` as `_remote_patron_lookup` on `AuthenticationProvider`, and then called that method from `BasicAuthenticationProvider` and override it in subclasses of `BasicAuthenticationProvider`.

## Motivation and Context

This inheritance pattern causes mypy some issues when trying to understand the types of the inherited methods, and makes this code a bit confusing to read.

We should be able to refactor in this way, because the two classes that inherit from `AuthenticationProvider` other then `BasicAuthenticationProvider` both override `remote_patron_lookup` and don't call `_remote_patron_lookup` at all. So it makes sense to move its definition up the class hierarchy.

## How Has This Been Tested?

Running CI tests.

## Checklist

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -8,7 +8,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 from abc import ABCMeta
-from typing import Any, Dict, Iterable, List, Optional
+from typing import Any, Dict, Iterable, List, Optional, Union
 
 import flask
 import jwt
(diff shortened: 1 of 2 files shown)


jonathangreen:
Tagging you for a review @RishiDiwanTT since you also ran into this in https://github.com/ThePalaceProject/circulation/pull/598

RishiDiwanTT (approved):
This looks good :+1:

PR #868, opened 2023-03-01
Add mypy to additional files

jonathangreen:
## Description

This PR removes two of the imports that we have set to `skip`, so we are doing some more type checking. Tagging you for review @io7m since most of the changes come from the fixtures.

## Motivation and Context

This PR has to go in first: https://github.com/ThePalaceProject/circulation/pull/867 as this one is based on it.

## How Has This Been Tested?

Running mypy

## Checklist

### DIFF
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -1,13 +1,13 @@
 import json
 from datetime import datetime
-from typing import Union
+from typing import Optional, Union
 
 from flask_babel import lazy_gettext as _
 
 from api.authenticator import BasicAuthenticationProvider, PatronData
 from api.sip.client import SIPClient
 from api.sip.dialect import Dialect as Sip2Dialect
-from core.model import ExternalIntegration
+from core.model import ExternalIntegration, Patron
 from core.util import MoneyUtility
 from core.util.http import RemoteIntegrationException
 
@@ -238,7 +238,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         except OSError as e:
             raise RemoteIntegrationException(self.server or "unknown server", str(e))
 
-    def _remote_patron_lookup(self, patron_or_patrondata):
+    def _remote_patron_lookup(self, patron_or_patrondata) -> Union[None, PatronData]:
         info = self.patron_information(
             patron_or_patrondata.authorization_identifier, None
         )
@@ -305,7 +305,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
                     ("Raw test patron information"), raw_patron_information
                 )
 
-    def info_to_patrondata(self, info, validate_password=True):
+    def info_to_patrondata(self, info, validate_password=True) -> Optional[PatronData]:
 
         """Convert the SIP-specific dictionary obtained from
         SIPClient.patron_information() to an abstract,
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -238,7 +238,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         except OSError as e:
             raise RemoteIntegrationException(self.server or "unknown server", str(e))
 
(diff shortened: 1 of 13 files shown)


api/sip/__init__.py
```diff
@@ -238,7 +238,7 @@ def patron_information(self, username, password):
         except OSError as e:
             raise RemoteIntegrationException(self.server or "unknown server", str(e))
 
-    def _remote_patron_lookup(self, patron_or_patrondata):
+    def _remote_patron_lookup(self, patron_or_patrondata) -> Union[None, PatronData]:
```
io7m, line 241: `Union[None, PatronData]` is isomorphic to `Optional[PatronData]`. I think `Optional` is clearer.

core/model/configuration.py
```diff
@@ -329,6 +329,7 @@ class ExternalIntegration(Base):
         "Library",
         back_populates="integrations",
         secondary=lambda: externalintegrations_libraries,
+        uselist=True,
```
io7m, line 332: I don't get what happened here.

tests/fixtures/api_controller.py
```diff
@@ -102,7 +102,7 @@ def set_base_url(self):
         base_url.value = "http://test-circulation-manager/"
 
     def circulation_manager_setup_with_session(
-        self, session: Session, overrides: ControllerFixtureSetupOverrides = None
+        self, session: Session, overrides: ControllerFixtureSetupOverrides | None = None
```
io7m, line 105: `ControllerFixtureSetupOverrides | None` <-> `Optional[ControllerFixtureSetupOverrides]`?

tests/fixtures/api_controller.py
```diff
@@ -168,7 +168,7 @@ def circulation_manager_setup_with_session(
         return self.manager
 
     def circulation_manager_setup(
-        self, overrides: ControllerFixtureSetupOverrides = None
+        self, overrides: ControllerFixtureSetupOverrides | None = None
```
io7m, line 171: As above.

tests/fixtures/odl.py
```diff
@@ -65,7 +66,7 @@ class ODLTestFixture:
     def __init__(
         self,
         db: DatabaseTransactionFixture,
-        files: ODLAPIFilesFixture,
+        files: APIFilesFixture,
```
io7m, line 69: What happened here?

io7m (changes requested):
Looks mostly good, just a few incredibly minor tweaks and questions.

tests/fixtures/api_controller.py
jonathangreen: Yes, those two are equivalent. This new `|` syntax is enabled by importing `from __future__ import annotations` at the top of the file.

In any file that does this future import the pyupgrade linter prefers the `|` union notion over the old `Optional[]` form, see [docs here](https://github.com/asottile/pyupgrade#pep-604-typing-rewrites).

tests/fixtures/odl.py
jonathangreen: Without this change, mypy gives this error: 
```
tests/fixtures/odl.py:296: error: Argument 2 to "ODL2TestFixture" has incompatible type "ODL2APIFilesFixture"; expected "ODLAPIFilesFixture"  [arg-type]
```

`ODL2TestFixture`:
https://github.com/ThePalaceProject/circulation/blob/5ecedbc042927e219a4faab3a730732756fff19c/tests/fixtures/odl.py#L285-L286

I think this is the correct change, as both `ODLAPIFilesFixture` and `ODL2APIFilesFixture` use `APIFilesFixture` as their base class.

api/sip/__init__.py
jonathangreen: Updated in most recent commit

io7m (approved):
OK, seems good!

PR #871, opened 2023-03-01
Speed up controller tests

jonathangreen:
## Description

Doing a bit more profiling of our code, it seems that the single most expensive function we call when running the tests is now the call to set `password` in `AdminControllerTest`. This removes the need to hash this password on each test setup, by directly setting the the hashed password.

This doesn't speed up the tests dramatically, but does shave about a minute off the test run on my machine.

## Motivation and Context

This came up when doing other profiling, and it seemed worth it to address. Save a bit of time every time we run the tests.

## How Has This Been Tested?

Running CI against it.

## Checklist

### DIFF
--- a/tests/api/admin/controller/test_controller.py
+++ b/tests/api/admin/controller/test_controller.py
@@ -85,7 +85,11 @@ class AdminControllerTest(CirculationControllerTest):
             Admin,
             email="example@nypl.org",
         )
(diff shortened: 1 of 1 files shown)


io7m (approved):
Looks good.

I wonder if this will trigger a code scanning issue. :laughing:

PR #887, opened 2023-03-06
Fix failing tests due to werkzeug escaping changes

jonathangreen:
## Description

Currently our tests are failing on the `main` branch, due to a conflict between that tests that started running in #835 and #855. This is caused by Werkzeug not escaping as many characters in URLS see issue here https://github.com/pallets/werkzeug/issues/262. These URLs are still valid, but have less characters in them escaped.

## Motivation and Context

https://github.com/ThePalaceProject/circulation/commit/cf09bf08e22055b5ee1136aaeeafdb37326e0ea0 gets our tests running again. 

Working through this issue, and looking at the linked Werkzeug issue, made me reconsider the fix I made in #855 to fix the `admin/controller` URL escaping tests. 

https://github.com/ThePalaceProject/circulation/commit/d1d663a63f1f0310409301bd2df0cb8ed0c28936 mostly reverts the changes from https://github.com/ThePalaceProject/circulation/pull/855/commits/f0cca3232e51f4e379a6816e169189628e905763 what went in with #855 and takes a different approach to solving that test failure.

In these tests `(` was previously escaped, but it isn't escaped anymore. This is also due to Werkzeug changes from the linked issue. Based on that I'm imported the workzeug `url_quote_plus` function, and used that to do the escaping instead of the python stdlib function we were using before.

## How Has This Been Tested?

Ran tests locally

## Checklist

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -17,7 +17,7 @@ from pydantic import BaseModel
 from sqlalchemy.orm import Session
 from sqlalchemy.sql import func
 from sqlalchemy.sql.expression import and_, desc, distinct, join, nullslast, select
-from werkzeug.urls import url_fix, url_parse, url_quote, url_unparse
+from werkzeug.urls import url_fix, url_parse, url_quote, url_quote_plus, url_unparse
 from werkzeug.wrappers import Response as werkzeug_response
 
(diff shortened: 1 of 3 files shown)


dbernstein (approved):
LTGM

PR #892, opened 2023-03-07
This update fixes ./bin/informational/disappearing-books

dbernstein:
Resolves: https://www.notion.so/lyrasis/bin-informational-disappearing_books-script-is-broken-cf9e190339d94eefb62fd36db6ebf0dc

## Description

The above mentioned script was broken on VT scripts.

```
(env) root@f1d3e761b8a0:/var/www/circulation# ./bin/informational/disappearing_books
Identifier	Title	Author	First seen	Last seen (best guess)	Current licenses owned	Current licenses available	Changes in number of licenses	Changes in title availability
{"host": "f1d3e761b8a0", "app": "simplified", "name": "root", "level": "ERROR", "filename": "[scripts.py](http://scripts.py/)", "message": "Fatal exception while running script: sequence item 0: expected str instance, bytes found", "timestamp": "2023-03-07T19:25:54.532217+00:00", "traceback": "Traceback (most recent call last):\n  File \"/var/www/circulation/core/scripts.py\", line 133, in run\n    timestamp_data = self.do_run()\n  File \"/var/www/circulation/scripts.py\", line 980, in do_run\n    self.explain(pool)\n  File \"/var/www/circulation/scripts.py\", line 1091, in explain\n    print(\"\\t\".join([str(x).encode(\"utf-8\") for x in data]))\nTypeError: sequence item 0: expected str instance, bytes found"}
File "./bin/informational/disappearing_books", line 12, in <module>
DisappearingBookReportScript().run()
File "/var/www/circulation/core/scripts.py", line 133, in run
timestamp_data = self.do_run()
File "/var/www/circulation/scripts.py", line 980, in do_run
self.explain(pool)
File "/var/www/circulation/scripts.py", line 1091, in explain
print("\t".join([str(x).encode("utf-8") for x in data]))
TypeError: sequence item 0: expected str instance, bytes found
```
## Motivation and Context
I was trying to figure out why a certain book was flagged as not presentation ready.  Tim suggested I use this script, but when I did, I hit an error.  

## How Has This Been Tested?
I manually tested the fix against the VT CM.  
I must admit I don't fully understand why this fixes it because it's not clear to me why it is failing in the first place.

## Checklist

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1088,7 +1088,7 @@ class DisappearingBookReportScript(Script):
         ]
         data.append(", ".join(title_removals))
 
-        print("\t".join([str(x).encode("utf8") for x in data]))
+        print("\t".join([str(x) for x in data]))
 
 
 class NYTBestSellerListsScript(TimestampScript):


scripts.py
```diff
@@ -1088,7 +1088,7 @@ def explain(self, licensepool):
         ]
         data.append(", ".join(title_removals))
 
-        print("\t".join([str(x).encode("utf8") for x in data]))
+        print("\t".join([str(x) for x in data]))
```
dbernstein, line 1091: I'm not sure why this fixes the problem but it does.   I'm seeing titles and authors with UTF-8 characters.  And the stacktrace I was getting doesn't make sense it me.  I think there must be something in the way that pythons handles strings that I don't understand.

io7m: I have no idea how this code was ever valid.

```
>>> "a".join("b")
'b'
>>> "a".join("b".encode("utf-8"))
  File "<stdin>", line 1, in <module>
TypeError: sequence item 0: expected str instance, int found
```

Maybe old Python versions were lax about strings vs bytes?

jonathangreen: This must be leftover from the Python 2 -> Python 3 migration. The biggest difference between Python 2 and Python 3 is that strings were bytes in Python 2, and there was a separate unicode type. 

https://python-future.org/compatible_idioms.html#strings-and-bytes
http://python3porting.com/problems.html#bytes-strings-and-unicode

Most of these bytes vs string issues have been sorted out by now, although there is still some conversion code that could be removed now that we are Python 3 only.

PR #893, opened 2023-03-08
Add logging around patron authentication calls

io7m:
## Description

This adds elapsed time logging around the calls to external services. This is achieved by adding logging to the AuthenticationProvider interface, and requiring that subclasses provide an implementation name to appear in logs. In order to ensure that all of the relevant subclasses have been updated, AuthenticationProvider is now marked as an abstract base class (by extending ABC).

## Motivation and Context

We want to know how long these calls take so that we can set sensible timeout values, amongst other uses.

Affects: https://www.notion.so/lyrasis/Add-logging-instrumentation-to-our-patron-auth-requests-9e86b9c0e1294899afaf726b92a5eb9b

## How Has This Been Tested?

I made sure the test suite ran, but I haven't tested authentication yet.

## Checklist

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -7,7 +7,7 @@ import sys
 import urllib.error
 import urllib.parse
 import urllib.request
-from abc import ABC, ABCMeta, abstractmethod
+from abc import ABC, ABCMeta
 from typing import Any, Dict, Iterable, List, Optional, Union
 
 import flask
@@ -1280,9 +1280,6 @@ class LibraryAuthenticator:
 class AuthenticationProvider(OPDSAuthenticationFlow, ABC):
     """Handle a specific patron authentication scheme."""
 
-    # The logger used for logging timing information for authentication requests (amongst other things).
-    log = logging.getLogger("api.authenticator.AuthenticationProvider")
-
     # NOTE: Each subclass MUST define an attribute called NAME, which
     # is displayed in the admin interface when configuring patron auth,
     # used to create the name of the log channel used by this
@@ -1592,12 +1589,10 @@ class AuthenticationProvider(OPDSAuthenticationFlow, ABC):
     def external_integration(self, _db):
         return get_one(_db, ExternalIntegration, id=self.external_integration_id)
 
-    @abstractmethod
-    def logging_name(self) -> str:
-        """The name used for this authentication provider in log messages."""
-        raise NotImplementedError()
+    @classmethod
+    def _logger(cls) -> logging.Logger:
+        return logging.getLogger(f"{cls.__module__}.{cls.__name__}")
 
-    @log_elapsed_time(log_method=log.info, message_prefix="authenticated_patron")
     def authenticated_patron(self, _db, header):
         """Go from a WWW-Authenticate header (or equivalent) to a Patron object.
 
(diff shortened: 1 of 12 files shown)


api/authenticator.py
```diff
@@ -1277,9 +1277,12 @@ def create_authentication_headers(self):
         return headers
 
 
-class AuthenticationProvider(OPDSAuthenticationFlow):
+class AuthenticationProvider(OPDSAuthenticationFlow, ABC):
     """Handle a specific patron authentication scheme."""
```
io7m, line 1281: `AuthenticationProvider` was conceptually an abstract class without actually saying so. By extending `ABC`, we're making it a hard requirement to fill in the methods marked as abstract; it's no longer possible to instantiate the class directly. This helped massively in finding all of the places in the codebase that needed updating when a new method was added.

jonathangreen: IIRC `ABC` is new in Python 3.x, so this code was written before it was a thing. Makes sense to start using it here.

api/authenticator.py
```diff
@@ -1589,6 +1592,12 @@ def library(self, _db):
     def external_integration(self, _db):
         return get_one(_db, ExternalIntegration, id=self.external_integration_id)
 
+    @abstractmethod
+    def logging_name(self) -> str:
+        """The name used for this authentication provider in log messages."""
+        raise NotImplementedError()
+
+    @log_elapsed_time(log_method=log.info, message_prefix="authenticated_patron")
     def authenticated_patron(self, _db, header):
```
jonathangreen, line 1601: Its slightly different from what we have done in other places, but maybe we could implement this as a `with` instead of a decorator. It would make naming the loggers a lot easier.

```python
@classmethod
def logger(cls):
    return logging.getLogger(f"{cls.__module__}.{cls.__name__}")

def authenticated_patron(self, _db, header):
    with elapsed_time_logging(
        log_method=self.logger.info
    ):
        ...
```

Then since the logger has the class name, we probably don't need to implement `logging_name` in each subclass.

jonathangreen (commented):
Looks good, just a comment about how we might simplify the implementation as bit.

api/authenticator.py
io7m: That looks good. I wasn't aware it was an option!

jonathangreen: We use different logger names all over the place, but I'd like to try to standardize on something like this, since its easy to implement, and it will let adjust the log level for different modules / classes if we need.

io7m: Yep. The standard in Java land is to use the fully qualified class name. I don't think I've ever seen anything more than that be necessary, in practical terms.

api/clever/__init__.py
```diff
@@ -76,8 +77,7 @@ def external_type_from_clever_grade(grade):
     return CLEVER_GRADE_TO_EXTERNAL_TYPE_MAP.get(grade, None)
 
 
-class CleverAuthenticationAPI(OAuthAuthenticationProvider):
-
+class CleverAuthenticationAPI(OAuthAuthenticationProvider, ABC):
```
jonathangreen, line 80: I think this class is meant to be instantiated directly, since its included in the list here: 
https://github.com/ThePalaceProject/circulation/blob/e8d2e96c28f3cf7caad14f9aa73bfb42cacf2b1e/api/admin/controller/patron_auth_services.py#L30-L40

So I don't think we should be giving it the `ABC` type.

jonathangreen (commented):
Once comment about the `CleverAuthenticationAPI` otherwise this looks good to me.

api/clever/__init__.py
io7m: Erk, yes. Not sure why that one was marked ABC...

jonathangreen (approved):

PR #898, opened 2023-03-08
Initialize CM instance outside of first request

dbernstein:
Resolves: https://www.notion.so/lyrasis/Initialize-CM-instance-outside-of-first-request-d024b0b64174442d8bd2c2171a2d4aeb?pvs=4

## Description
Moves all the initialization logic from that was running in the `@before_first_request` decorator into the application initialization stage in order to ensure that the first user to hit the application doesn't experience a multi-second delay while the application is being initialized.

## Motivation and Context
 https://www.notion.so/lyrasis/Initialize-CM-instance-outside-of-first-request-d024b0b64174442d8bd2c2171a2d4aeb?pvs=4

## How Has This Been Tested?
All the previous unit tests still pass post refactor.   I also stood up an instance and tested it manually to ensure that the first request was snappy.

## Checklist


jonathangreen:
@dbernstein looks like `mypy` is failing here

api/app.py
```diff
@@ -61,6 +70,37 @@ def initialize_database(autoinitialize=True):
     logging.getLogger().info("Application debug mode==%r" % app.debug)
 
 
+def setup_admin(_db=None):
```
dbernstein, line 73: maybe I should rename to "intialize_admin" for consistency.

api/util/profilers.py
```diff
@@ -44,7 +44,6 @@ def configure(cls, app: Flask):
         # Don't import if we are not profiling
         from pyinstrument import Profiler
 
-        @app.before_first_request
```
dbernstein, line 47: I removed this tag because I couldn't see how it was doing anything more than @app.before_request (below: line 48) was doing.

api/app.py
```diff
@@ -87,6 +127,7 @@ def run(url=None):
 
         socket.setdefaulttimeout(None)
 
+    initialize()
```
jonathangreen, line 130: The `run` function here isn't ever called in the production environment, its only used when running locally. So in production this initialize function will never get called.

jonathangreen (changes requested):
This works when running the server locally, but it doesn't work in the docker image builds, since `run` never gets called. 

Try standing up a server with `docker-compose up -d` to test this out in an environment closer to what we are using in production. 

I pushed a PR https://github.com/ThePalaceProject/circulation/pull/906 to fix the docker-compose build for m1 macs, that might be helpful.

api/app.py
jonathangreen: When I try running this in the dockerized environment I see this exception in the logs and a 502 error come back: 
```
  File "/var/www/circulation/env/lib/python3.8/site-packages/flask/app.py", line 2551, in __call__
    return self.wsgi_app(environ, start_response)
  File "/var/www/circulation/env/lib/python3.8/site-packages/flask/app.py", line 2531, in wsgi_app
    response = self.handle_exception(e)
  File "/var/www/circulation/env/lib/python3.8/site-packages/flask/app.py", line 1727, in handle_exception
    server_error = self.ensure_sync(handler)(server_error)
  File "/var/www/circulation/api/routes.py", line 98, in wrapped_function
    patron_web_domains = app.manager.patron_web_domains
AttributeError: 'Flask' object has no attribute 'manager'
```

dbernstein: Okay - I'll fix that.

dbernstein (commented):
I don't love this solution but it does work.  The problem before this commit was that while the test environment sets and deletes the "TESTING" environment in core/testing.py and tests/fixtures/database.py, the initialize() method is called before the test fixtures before and after code is executed.   Thus the errors. 

Initially I tried to add the initialization routines to the end of the Flask.run() command. But apparently that method is not called in production.  I still don't fully understand what is the entrypoint for the flask application.

RishiDiwanTT:
> Initially I tried to add the initialization routines to the end of the Flask.run() command. But apparently that method is not called in production. I still don't fully understand what is the entrypoint for the flask application, ie what does uwsgi call to "run" flask.

UWSGI imports app.py as a python module. Flask supports the uwsgi protocol so once the "app" in found in app.py UWSGI simply runs it under the uwsgi protocol (no flask.run()).

jonathangreen:
Here is where the app gets setup in the [uwsgi config](https://uwsgi-docs.readthedocs.io/en/latest/Options.html?highlight=callable#callable):
https://github.com/ThePalaceProject/circulation/blob/0c0f21281857c5596330fe8eb6379c722290fad6/docker/services/uwsgi.d/50_palace.ini#L7-L9

I'd really like to move the CM code away from checking environment variables to determine if tests are running, and from using global variables as much as possible, so it would be nice to refactor this a bit more. 

We should be able to make the `initialize` function something like this: 
```python
def initialize_application() -> Flask:
    initialize_database(app)
    initialize_circulation_manager(app)
    initialize_admin(app)
    return app
```

We can probably even pull of the initialization from here into this function: 
https://github.com/ThePalaceProject/circulation/blob/b27413584dd3de4e160d0ab01d884b57261b704c/api/app.py#L23-L42

Then update the config for uwsgi to use this new callable:
```ini
 module = api.app 
 callable = initialize_application 
```

dbernstein:
Thanks @jonathangreen  and @RishiDiwanTT for the review and suggestions.  I'll make those changes.

dbernstein:
@jonathangreen : I tried to pull in the additional lines into the initialization routine but ran into issues with  circular dependencies.  I did fix the other issues.  In the process of testing, I kept running into race conditions with multiple flask processes initializing at the same time which was leading to database update conflicts.  So I brought in the uwsgi `cheaper` configuration to ensure that only one instance of flask  initialized at startup.  Over time two more will be added.

jonathangreen:
I'd really like to avoid having to to use cheaper to start the worker threads, since it creates some unpredictability around how many processes we are running at any time and creating new worker processes is still slow. It also ties us to using this method, since the app server fails otherwise.

I did some testing with this today, without cheaper and here are the details of what I was seeing: 

```
circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [296] ERROR:  duplicate key value violates unique constraint "ix_configurationsettings_external_integration_id_key"
circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [296] DETAIL:  Key (external_integration_id, key)=(154, ssl_verification) already exists.
circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [296] STATEMENT:  INSERT INTO configurationsettings (external_integration_id, library_id, key, value) VALUES (154, NULL, 'ssl_verification', NULL) RETURNING configurationsettings.id
circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [294] ERROR:  duplicate key value violates unique constraint "ix_configurationsettings_external_integration_id_key"
circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [294] DETAIL:  Key (external_integration_id, key)=(154, ssl_verification) already exists.
circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [294] STATEMENT:  INSERT INTO configurationsettings (external_integration_id, library_id, key, value) VALUES (154, NULL, 'ssl_verification', NULL) RETURNING configurationsettings.id
```

These errors were happening in `initialize_database`, `initialize_admin` and `initialize_circulation_manager` all due to conflicts writing `configurationsettings`. These conflicts were happening because all the workers would try to write out any non-existent configuration settings at the same time, resulting in some of the workers having their transactions rolled back. Since we never just query for configuration settings, we always try to create any non-existing ones, this conflict happens every time.

You didn't provide many details @dbernstein, is this the same behaviour you were seeing?

Another possible solution would be to lock the `configurationsettings` table when we are initializing, that way we know that only one worker is creating config settings. Since we have `_db.commit()` already in all these functions, we would just have to add something like `_db.execute('LOCK TABLE configurationsettings IN ACCESS EXCLUSIVE MODE;')`.

I did some testing with this, and it does work. It wouldn't tie us to starting the application with cheaper and it eliminates the possibility of this race happening when two containers come up at the same time. But I don't really like this solution either, since it means our config settings table is going to have a lock on it every time we are initializing a worker. 

@RishiDiwanTT @tdilauro do either of you have any ideas how we might best handle this?

api/app.py
```diff
@@ -42,25 +48,63 @@
...
 
 
-@app.before_first_request
-def initialize_database(autoinitialize=True):
-    testing = "TESTING" in os.environ
+def initialize_application() -> Flask:
+
+    initialize_database()
+    initialize_circulation_manager()
+    initialize_admin()
```
jonathangreen, line 55: I think in order to make sure that the DB session we are opening here is cleaned up correctly, we need to do something like this: 

```python
def initialize_application() -> Flask:
    with app.application_context():
        initialize_database()
        initialize_circulation_manager()
        initialize_admin()
```

api/app.py
```diff
@@ -42,25 +48,63 @@
...
+    return app
 
+
+def initialize_database(autoinitialize=True, is_testing=False):
```
jonathangreen, line 59: I'm torn about adding a `is_testing` parameter here. On one hand, It is *really* nice to get rid of:
```python
    testing = "TESTING" in os.environ
```
And I think we should be cleaning things like this up as we go.

My concern though is that we still use that environment variables in other places in the code base. Even [in other functions called from this one](https://github.com/ThePalaceProject/circulation/blob/main/core/config.py#L311). If you do a search for "TESTING", you'll find a number of places we use that check to decide if we are in test mode or not.

Adding this parameter here, means that anyone calling this function in test needs to remember both to set the "TESTING" env var (otherwise they will get a connection to the [wrong DB URL](https://github.com/ThePalaceProject/circulation/blob/main/core/config.py#L311)) and also set the `is_testing` parameter to get consistent behavior. Which feels like a bit of a trap we are laying for our future selves.

Maybe getting rid of the "TESTING" env var here needs to happen in sync with removing it in other places in the code in a future PR?

dbernstein: That's valid.  I'll revert that change and create a ticket to remove "TESTING" from the codebase.

api/app.py
dbernstein: Good catch.  I will fix.

dbernstein:
> I'd really like to avoid having to to use cheaper to start the worker threads, since it creates some unpredictability around how many processes we are running at any time and creating new worker processes is still slow. It also ties us to using this method, since the app server fails otherwise.
> 
> I did some testing with this today, without cheaper and here are the details of what I was seeing:
> 
> ```
> circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [296] ERROR:  duplicate key value violates unique constraint "ix_configurationsettings_external_integration_id_key"
> circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [296] DETAIL:  Key (external_integration_id, key)=(154, ssl_verification) already exists.
> circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [296] STATEMENT:  INSERT INTO configurationsettings (external_integration_id, library_id, key, value) VALUES (154, NULL, 'ssl_verification', NULL) RETURNING configurationsettings.id
> circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [294] ERROR:  duplicate key value violates unique constraint "ix_configurationsettings_external_integration_id_key"
> circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [294] DETAIL:  Key (external_integration_id, key)=(154, ssl_verification) already exists.
> circulation-pg-1  | 2023-03-15 16:36:21.149 UTC [294] STATEMENT:  INSERT INTO configurationsettings (external_integration_id, library_id, key, value) VALUES (154, NULL, 'ssl_verification', NULL) RETURNING configurationsettings.id
> ```
> 
> These errors were happening in `initialize_database`, `initialize_admin` and `initialize_circulation_manager` all due to conflicts writing `configurationsettings`. These conflicts were happening because all the workers would try to write out any non-existent configuration settings at the same time, resulting in some of the workers having their transactions rolled back. Since we never just query for configuration settings, we always try to create any non-existing ones, this conflict happens every time.
> 
> You didn't provide many details @dbernstein, is this the same behaviour you were seeing?
> 
> Another possible solution would be to lock the `configurationsettings` table when we are initializing, that way we know that only one worker is creating config settings. Since we have `_db.commit()` already in all these functions, we would just have to add something like `_db.execute('LOCK TABLE configurationsettings IN ACCESS EXCLUSIVE MODE;')`.
> 
> I did some testing with this, and it does work. It wouldn't tie us to starting the application with cheaper and it eliminates the possibility of this race happening when two containers come up at the same time. But I don't really like this solution either, since it means our config settings table is going to have a lock on it every time we are initializing a worker.
> 
> @RishiDiwanTT @tdilauro do either of you have any ideas how we might best handle this?
> 
> [Here is the branch I was testing the lock table approach on](https://github.com/dbernstein/circulation/compare/refactor_before_first...jonathangreen:circulation:refactor_before_first), if its helpful.

I'll try bringing in your suggestion.  I agree with you re `cheaper` - > it's definitely a more "expensive" option compared to your table lock approach.

RishiDiwanTT:
I'm not a fan of the locking mechanism because it will exasperate the issue of slow startups, since the threads and processes and load balanced containers will all be in queue for the same lock.
This solves the problem of only the first startup if I'm understanding this correctly, maybe we can mitigate by testing if some configuration value has been created before doing exclusive locks?
Maybe have a special `CONFIGURATION_EXISTS` configuration, that is created at startup, if it does not exist, then we do the locks, and not otherwise.
The `cheaper` solution will give the same problem when 2 load balanced containers come up together for the first time.

dbernstein:
@jonathangreen :  I tried running your proposed changes but ran into a couple of issues:  when trying to run the tests I ran into a timeout on the following test. 
```
tests/api/test_controller_scopedsession.py
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Captured stderr ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2023-03-16 13:45:27,980:root:INFO:config.py:Connecting to database: postgresql://simplified_test:***@localhost:9005/simplified_circulation_test

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Stack of MainThread (4365321600) ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/bin/pytest", line 8, in <module>
    sys.exit(console_main())
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/config/__init__.py", line 190, in console_main
    code = main()
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/config/__init__.py", line 167, in main
    ret: Union[ExitCode, int] = config.hook.pytest_cmdline_main(
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/main.py", line 317, in pytest_cmdline_main
    return wrap_session(config, _main)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/main.py", line 270, in wrap_session
    session.exitstatus = doit(config, session) or 0
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/main.py", line 324, in _main
    config.hook.pytest_runtestloop(session=session)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/main.py", line 349, in pytest_runtestloop
    item.config.hook.pytest_runtest_protocol(item=item, nextitem=nextitem)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/runner.py", line 112, in pytest_runtest_protocol
    runtestprotocol(item, nextitem=nextitem)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/runner.py", line 131, in runtestprotocol
    reports.append(call_and_report(item, "call", log))
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/runner.py", line 220, in call_and_report
    call = call_runtest_hook(item, when, **kwds)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/runner.py", line 259, in call_runtest_hook
    return CallInfo.from_call(
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/runner.py", line 339, in from_call
    result: Optional[TResult] = func()
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/runner.py", line 260, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/runner.py", line 167, in pytest_runtest_call
    item.runtest()
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/python.py", line 1789, in runtest
    self.ihook.pytest_pyfunc_call(pyfuncitem=self)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/_pytest/python.py", line 195, in pytest_pyfunc_call
    result = testfunction(**testargs)
  File "/Users/danielbernstein/code/circulation/tests/api/test_controller_scopedsession.py", line 92, in test_scoped_session
    api.app.initialize_database()
  File "/Users/danielbernstein/code/circulation/api/app.py", line 69, in initialize_database
    _db.execute("LOCK TABLE configurationsettings IN ACCESS EXCLUSIVE MODE;")
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/sqlalchemy/orm/scoping.py", line 163, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 1295, in execute
    return self._connection_for_bind(bind, close_with_result=True).execute(
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1011, in execute
    return meth(self, multiparams, params)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/sqlalchemy/sql/elements.py", line 298, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1124, in _execute_clauseelement
    ret = self._execute_context(
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
    self.dialect.do_execute(
  File "/Users/danielbernstein/code/circulation/.tox/py38-api-docker/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute

+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
ERROR: InvocationError for command /Users/danielbernstein/code/circulation/.tox/py38-api-docker/bin/pytest tests/api/test_controller_scopedsession.py (exited with code 1)
```

dbernstein:
When running  docker-compose I ran into errors likes this: 
```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateTable) relation "licensepools" already exists
...
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateTable) relation "circulationevents" already exists
...
sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "pg_type_typname_nsp_index"
DETAIL:  Key (typname, typnamespace)=(holds_id_seq, 2200) already exists.
```

dbernstein:
@jonathangreen : I've got an idea I'm trying out. Hold off until I get back to you.

dbernstein:
@jonathangreen :  check out https://github.com/dbernstein/circulation/pull/1 - it puts your updates on top of this PR and adds a couple of fixes that makes the tests run for me and as well as fix the occasional errors on start up that I was seeing.

dbernstein:
This PR has been rebased and I finally worked out the locks needed to get the tests,  the test docker builds, and `docker-compose up` working.  See my annotations for explanation.

api/app.py
```diff
@@ -42,25 +49,65 @@
 PalaceXrayProfiler.configure(app)
 
 
-@app.before_first_request
+def initialize_application() -> Flask:
```
dbernstein, line 52: This is the entry point for uwsgi

api/app.py
```diff
@@ -42,25 +49,65 @@
...
     _db.commit()
+
     logging.getLogger().info("Application debug mode==%r" % app.debug)
 
 
+def initialize_admin(_db=None):
+    if getattr(app, "manager", None) is not None:
+        setup_admin_controllers(app.manager)
+    _db = _db or app._db
+    _db.execute("LOCK TABLE configurationsettings IN ACCESS EXCLUSIVE MODE;")
```
dbernstein, line 81: This lock is required. Otherwise multiple processes initializing the app for the first time end up trying to write configuration to the database at the same time.

core/model/__init__.py
```diff
@@ -334,38 +334,40 @@ def initialize(cls, url, initialize_data=True, initialize_schema=True):
...
-        )
-        result = connection.execute(query)
-        result = list(result)
-
-        # If it doesn't, create it.
-        if not result and initialize_data:
-            resource_file = os.path.join(
-                cls.resource_directory(), cls.RECURSIVE_EQUIVALENTS_FUNCTION
+        with engine.connect() as connection:
+            tx = connection.begin()
```
dbernstein, line 338: This was an especially tricky bit.  We need a way to lock the database on startup when no schema exists.  Since there are no cm tables yet,  I'm using postges internal stored procedure metadata table. We need to open a transaction and close it only when we're done with our initialization.  The reason is that without a transaction,  I believe line 342 (`cls.initialize_schema(connection)`) will spawn a separate transaction causing a deadlock.  I'm not totally clear on the mechanics here - the sqlalchemy code is pretty dense.

core/model/__init__.py
```diff
@@ -334,38 +334,40 @@ def initialize(cls, url, initialize_data=True, initialize_schema=True):
...
-            .where(literal_column("proname") == "fn_recursive_equivalents")
-        )
-        result = connection.execute(query)
-        result = list(result)
-
-        # If it doesn't, create it.
-        if not result and initialize_data:
-            resource_file = os.path.join(
-                cls.resource_directory(), cls.RECURSIVE_EQUIVALENTS_FUNCTION
+        with engine.connect() as connection:
```
dbernstein, line 337: Using `with` construct to ensure connection is closed automatically.

core/model/__init__.py
```diff
@@ -334,38 +334,40 @@ def initialize(cls, url, initialize_data=True, initialize_schema=True):
...
-        if initialize_data:
-            session = Session(connection)
-            cls.initialize_data(session)
+            if initialize_data:
+                session = Session(connection)
+                cls.initialize_data(session)
 
-        if connection:
-            connection.close()
+            tx.commit()
```
dbernstein, line 370: I'm not 100% this line is necessary.  Does closing a connection automatically commit any associated transactions?

core/model/__init__.py
dbernstein: Most of the changes below are just side effects of the indent required by the transaction and with statement.

core/model/__init__.py
```diff
@@ -403,17 +406,16 @@ def session(cls, url, initialize_data=True, initialize_schema=True):
                 initialize_schema=initialize_schema,
             )
         session = Session(connection)
-        if initialize_data:
```
dbernstein, line 406: This block can be removed because `initialize_data` is called in `initialize`

api/app.py
```diff
@@ -42,25 +49,65 @@
...
+    # The secret key is used for signing cookies for admin login
+    app.secret_key = ConfigurationSetting.sitewide_secret(_db, Configuration.SECRET_KEY)
+    # Create a default Local Analytics service if one does not
+    # already exist.
+
+    local_analytics = LocalAnalyticsProvider.initialize(_db)
+    _db.commit()
```
tdilauro, line 88: I worry a bit about having these locks around in the code. Could we improve the situation a bit creating a context manager, pulling it back into the ./core/model module somewhere, and just invoking it from here?

So you might have some like this in `core/model/configuration.py`:

```python
from contextlib import contextmanager

if TYPE_CHECKING:
    from sqlalchemy.orm import Session

...

@contextmanager
def access_exclusive_lock_configurationsettings(session: Session):
    session.begin_nested()
    session.execute("LOCK TABLE configurationsettings IN ACCESS EXCLUSIVE MODE;")
    try:
        yield session
    finally:
        session.commit()
```


Then `initialize_admin` might look something like this (note that the `_db.commit()` at the end is gone):
```python
def initialize_admin(_db=None):
    if getattr(app, "manager", None) is not None:
        setup_admin_controllers(app.manager)

    _db = _db or app._db
    with access_exclusive_lock_configurationsettings(_db) as db_session:
        # The secret key is used for signing cookies for admin login
        app.secret_key = ConfigurationSetting.sitewide_secret(db_session, Configuration.SECRET_KEY)
        # Create a default Local Analytics service if one does not
        # already exist.
        local_analytics = LocalAnalyticsProvider.initialize(db_session)
```

core/model/__init__.py
tdilauro: If `autocommit` is on, it does. I think it's best to explicitly commit (or rollback), where appropriate.

core/model/__init__.py
```diff
@@ -334,38 +334,40 @@ def initialize(cls, url, initialize_data=True, initialize_schema=True):
...
-        result = connection.execute(query)
-        result = list(result)
-
-        # If it doesn't, create it.
-        if not result and initialize_data:
-            resource_file = os.path.join(
-                cls.resource_directory(), cls.RECURSIVE_EQUIVALENTS_FUNCTION
+        with engine.connect() as connection:
+            tx = connection.begin()
+            connection.execute("LOCK TABLE pg_proc IN ACCESS EXCLUSIVE MODE;")
```
tdilauro, line 339: Assuming that a CM doesn't have "root" access to the DB server, will it be able to successfully get a lock on this table?

api/app.py
dbernstein: @tdilauro : that's very slick.  I'm willing it give it a try.

core/model/__init__.py
dbernstein: I'm not sure.  If there were a problem, wouldn't docker-compose or the other docker tests have failed?

dbernstein: I'll try it on one of the dev rds databases.

dbernstein: Good catch @tdilauro :  this doesn't work on RDS - you need to be root.

api/app.py
dbernstein: https://github.com/ThePalaceProject/circulation/pull/898/commits/e2445dfc03075e19076ae597431850f7b2a0594d

dbernstein: Added to the latest.

tdilauro (approved):
🚀

PR #905, opened 2023-03-09
Fixes broken palace cm logo.

dbernstein:
## Description
I appears that the Palace CM Logo link broke when we upgraded Flask.
The max_age argument is no longer supported in the send_from_directory method.
Instead the default max age is set elsewhere:  see https://flask.palletsprojects.com/en/2.2.x/api/ for details ( search for SEND_FILE_MAX_AGE_DEFAULT ).
 
## Motivation and Context

https://www.notion.so/lyrasis/Palace-CM-Logo-is-broken-78d753d9a5db491fa7ce3463d37c314c?pvs=4

## How Has This Been Tested?
Manually verified that it is fix by going to the sign in page.

## Checklist

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -36,6 +36,14 @@ def setup_admin(_db=None):
     # already exist.
     local_analytics = LocalAnalyticsProvider.initialize(_db)
 
+    # Configure the default max age for the static file cache.
+    max_age = ConfigurationSetting.sitewide(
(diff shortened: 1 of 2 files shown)


jonathangreen:
I think based on the [thread in slack](https://lyrasis.slack.com/archives/C03E3JHEFK8/p1678388371544909) this one can be closed. Is that correct @dbernstein?

PR #919, opened 2023-03-14
Changed socket timeouts for SIP2 connections to 3 seconds

RishiDiwanTT:
## Description
To ensure speedy failures for faulty SIP2 providers
During a timeout the API will respond with a 
`{"type": "http://librarysimplified.org/terms/problem/remote-integration-failed", "title": "Failure contacting external service", "status": 502, "detail": "The server tried to access 127.0.0.1 but the third-party service experienced an error."}`

## Motivation and Context
We have seen an incident twice in the past two weeks, where a SIP integration has gone offline, and this has caused one of our CM to become unresponsive for all the libraries in the CM. In order to reduce this happening, we should reduce the timeout that we wait for a SIP connection to respond.

Rationale explained in [Notion](https://www.notion.so/lyrasis/Reduce-the-length-of-timeouts-for-SIP-connections-d841aa38edfe4d54a2bc3bc62129e866) 

3 seconds is based on some statistics collected from our production CM instances. See [comment](https://github.com/ThePalaceProject/circulation/pull/919#issuecomment-1496009603) below.

## Checklist

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -222,6 +222,8 @@ class SIPClient(Constants):
 
     # Maximum retries of a SIP message before failing.
     MAXIMUM_RETRIES = 5
+    # Timeout in seconds
+    TIMEOUT = 3
 
     # These are the subfield names associated with the 'patron status'
     # field as specified in the SIP2 spec.
@@ -386,7 +388,7 @@ class SIPClient(Constants):
             else:
                 self.connection = self.make_insecure_connection()
 
-            self.connection.settimeout(2)
+            self.connection.settimeout(self.TIMEOUT)
             self.connection.connect((self.target_server, self.target_port))
         except OSError as message:
             raise OSError(
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -386,7 +386,7 @@ class SIPClient(Constants):
             else:
                 self.connection = self.make_insecure_connection()
 
-            self.connection.settimeout(12)
+            self.connection.settimeout(2)
             self.connection.connect((self.target_server, self.target_port))
         except OSError as message:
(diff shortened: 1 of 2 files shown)


io7m (approved):
Looks good! :rocket:

jonathangreen (changes requested):
I'd like to hold off on merging this one, until we test two things:
- Write a small benchmark and see if 2s is a long enough timeout for our existing production servers. 
- Make sure that in the event we do timeout, we are getting a reasonable log message that we can query for. 
  - I see we are re-raising the exception, but I'm not sure what that will end up looking like in out logs.

jonathangreen:
Since the https://github.com/ThePalaceProject/circulation/pull/893 is now deployed to production, we can get some real statistics from Cloudwatch insights about how long our connections are taking.

Cloudwatch insights query I used:
```
fields @timestamp, @message, @logStream, @log
| filter @logStream like "uwsgi" and message like "authenticated_patron - authenticate: Completed."
| parse message 'authenticated_patron - authenticate: Completed. (elapsed time: * seconds)' as seconds
| stats pct(seconds,98), avg(seconds), count() as c by name, @logStream
| sort c desc
```

The SIP2 p98 connections time ends up at about 1.5 seconds overall. 

If we wanted to use p99 as our metric here, we would end up with a time of 2.5 seconds, which would imply we should increase this from 2 to 3 seconds for a timeout. So it seems this value is fairly reasonable. 

Maybe to be on the safe side, we could bump it up to 3 second @RishiDiwanTT, then we can get it out to production and monitor how many connections are failing.

jonathangreen (approved):
🚀

PR #924, opened 2023-03-15
Ensure uwsgi logs go to stdout

io7m:
## Description

This adjusts the `uwsgi` configuration so that access logs go to stdout.

## Motivation and Context

We want all logs to go to stdout so that we can collect logs in a central location when running on Fargate.

Affects: https://www.notion.so/lyrasis/Update-our-docker-webapp-build-so-uwsgi-access-logs-are-on-stdout-15dd47b802e145358ff20aff0301fded

## How Has This Been Tested?

I built Docker images locally, ran them, and made requests. I saw logs such as this on stdout:

```
- (172.17.0.1) - - [15/Mar/2023:12:24:07 +0000] "GET / HTTP/1.1" 302 399 "-" "curl/7.88.1" host_hdr=localhost:32770 req_time_elapsed=11745
```

## Checklist

### DIFF
--- a/docker/services/uwsgi.d/40_log.ini
+++ b/docker/services/uwsgi.d/40_log.ini
@@ -1,5 +1,2 @@
 [uwsgi]
-# location of log files
-logto = /var/log/uwsgi/uwsgi.log
 log-format = %(var.HTTP_X_FORWARDED_FOR) (%(addr)) - - [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)" host_hdr=%(host) req_time_elapsed=%(msecs)
-logfile-chmod = 644
--- a/docker/services/uwsgi.d/40_log.ini
+++ b/docker/services/uwsgi.d/40_log.ini
@@ -1,2 +1,5 @@
 [uwsgi]
 log-format = %(var.HTTP_X_FORWARDED_FOR) (%(addr)) - - [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)" host_hdr=%(host) req_time_elapsed=%(msecs)
+logfile-chmod = 644
+logger = stdio:
+logger = file:/var/log/uwsgi/uwsgi.log


io7m:
There is more we could do here.

Disable syslog entirely and make sure everything in the image logs to `stdout`. That would also allow us to eliminate `logrotate`.

io7m:
We could/should also find a way to prefix log messages for easier filtering.

dbernstein (commented):
I tried it out and it works as advertised.  I have a question:  if we're bringing the logs into cloud watch is there anything in the logs that would enable us to locate the specific container the log messages are originating from?  Will the ip address be sufficient to tie it back to a specific docker container?

io7m:
That is a good question. I don't know if Cloudwatch allows filtering based on source address... I assume it would have to, otherwise actually using it would be nightmarish.

RishiDiwanTT:
> That is a good question. I don't know if Cloudwatch allows filtering based on source address... I assume it would have to, otherwise actually using it would be nightmarish.

In the ECS implementation each shard will have its own log group eg `tpp-prod` and then each "service" and "task" will have it's own log stream. So `<shard>/ECS/<service-name>/<task-id>` Eg. `/tpp-prod/ECS/cm-web/8suduu88ed8`

jonathangreen:
@io7m is it possible to get these logs going out to both the log file, and to stdout? This would be useful while we are transitioning over to using fargate. Otherwise we'll lose the logs that are currently getting pushed into cloudwatch from the ec2 instances during the transition.

io7m:
> @io7m is it possible to get these logs going out to both the log file, and to stdout?

I'll check, but the documentation suggests that it's not possible. As soon as you specify a `logto` statement, that turns off logging to `stdout`.

We might be able to do something with `tee`, but that would probably run into problems with log rotation.

io7m:
This doesn't seem possible unless we're willing to run `uwsgi` behind `tee` (not a good idea), or write a new logging plugin in C (not a good idea).

jonathangreen:
@io7m I think something like this will get the logs going to both places:
```ini
[uwsgi]
log-format = %(var.HTTP_X_FORWARDED_FOR) (%(addr)) - - [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)" host_hdr=%(host) req_time_elapsed=%(msecs)
logfile-chmod = 644
logger = stdio:
logger = file:/var/log/uwsgi/uwsgi.log
```

jonathangreen:
The `uwsgi` docs are a nightmare. Honestly it may be worth switching away from it in the future just because the docs are so damn bad.

I took a look at this page on the pluggable logger options: https://uwsgi-docs.readthedocs.io/en/latest/Logging.html#pluggable-loggers

I was wondering if we could do something a bit more involved like log to syslog, and configure syslog to output to both stdout and the log file for this logstream. In those docs, I saw it mentioned `--logger-list` to see what loggers are compiled into uwsgi, so I gave it a go on my local system, mostly to see if syslog was enabled by default, and it gave me this list:

```
$ uwsgi --logger-list

*** uWSGI loaded loggers ***
python
syslog
rsyslog
socket
redislog
mongodblog
file
fd
stdio
--- end of loggers list ---
```

Then I went looking for `stdio` in the docs, to see if it did what we wanted, and also noticed its pretty much absent. Seems like it does what we need though.

io7m:
There's me thinking that the logger list wouldn't include items that aren't in the documentation.

I should remember what kind of libraries we're working with here.

io7m:
Sorry, accidentally clicked the "re-review" button. It's not ready yet.

io7m:
OK, this seems ready. That configuration worked.

jonathangreen (approved):
Looks good to me

PR #933, opened 2023-03-20
Remove default Sirsidynix URL

dbernstein:
Resolves: https://www.notion.so/lyrasis/Do-not-auto-populate-URL-when-SirsiDynix-Horizon-Authentication-is-selected-in-the-CM-b0f6cb86719f48d49929fb813badf820?pvs=4

## Description

This PR simply changes the default SirsiDynix url from a mock value to "". 
## Motivation and Context

https://www.notion.so/lyrasis/Do-not-auto-populate-URL-when-SirsiDynix-Horizon-Authentication-is-selected-in-the-CM-b0f6cb86719f48d49929fb813badf820?pvs=4
## How Has This Been Tested?

I tested it manually by firing up a fresh cm, created a sirsidynix patron auth and validated that the default value is no longer set.
## Checklist

### DIFF
--- a/api/sirsidynix_authentication_provider.py
+++ b/api/sirsidynix_authentication_provider.py
@@ -57,7 +57,6 @@ class SirsiDynixHorizonAuthenticationProvider(BasicAuthenticationProvider):
             "label": _("Server URL"),
             "description": _("The external server url."),
             "required": True,
-            "default": "",
         },
         {
             "key": Keys.CLIENT_ID,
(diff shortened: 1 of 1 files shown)


api/sirsidynix_authentication_provider.py
```diff
@@ -57,7 +57,7 @@ class Keys:
             "label": _("Server URL"),
             "description": _("The external server url."),
             "required": True,
-            "default": "https://vendor1-sym.sirsidynix.net/ilsws_current/",
+            "default": "",
```
jonathangreen, line 60: Can probably just completely drop `default`, instead of setting it to a blank string.

jonathangreen (approved):
Made a minor comment you might want to address, but looks good.

PR #936, opened 2023-03-21
Removed the Hold object creation from within the ODLAPI implementation

RishiDiwanTT:
## Description
All other integrations depend on the CirculationAPI to create the Hold object 
They create a HoldInfo object with the required information, which is what the ODLAPI also does now 
The CirculationAPI is also aware of integration_client based Holds now
There should be no behavioural changes while booking a Hold for ODL type collections, except for the bugfix which allows the analytics event to be captured. 

## Motivation and Context
During a recent test, we noticed that the circulation_manager_hold_place event was not being captured in either Redshift or the CM CSV export for Palace Marketplace content. The event is being captured for all other distributors.

[Notion](https://www.notion.so/lyrasis/Hold-placed-events-are-not-captured-for-Palace-Marketplace-content-03b00ca9543b4b76b0fc56474aa42373)

## How Has This Been Tested?
Manually Tested the borrow/hold workflow for the ODLAPI
All unit tests were run

## Checklist

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -396,12 +396,14 @@ class HoldInfo(CirculationInfo):
         end_date,
         hold_position,
         external_identifier=None,
+        integration_client=None,
     ):
         super().__init__(collection, data_source_name, identifier_type, identifier)
         self.start_date = start_date
         self.end_date = end_date
         self.hold_position = hold_position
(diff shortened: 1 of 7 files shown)


io7m (approved):
This seems fine to me! :+1:

PR #949, opened 2023-03-26
Upgrade to tox 4.x

jonathangreen:
## Description

Upgrades tox to the 4.x version. In order to upgrade tox, we also have to upgrade our tox plugins.

## Motivation and Context

Previously we were not able to upgrade tox, because we were not able to use the new version of tox-docker as outlined in https://github.com/tox-dev/tox-docker/issues/154. Now that https://github.com/tox-dev/tox-docker/pull/156 has been merged, we can create a new Dockerfile (called dockerfile.ci in this PR), and use that with tox-docker.

## How Has This Been Tested?

Tested locally. This should be good to go if CI passes.

## Checklist


tdilauro (approved):
LGTM! 🚀

PR #964, opened 2023-03-30
Closed the leaking connection pool sessions

RishiDiwanTT:
## Description
In the PatronActivityThreads the api objects would access the db using the flask_scoped_session which creates a new Session per python thread 
Since the PatronActivityThread is spawned for every bookshelf_sync and the newly created Session is never torn down, we end up with orphaned Sessions using up the connection pool and never being closed 
The leak occurs slowly because python reuses completed threads which are identified via get_ident()

The fix __isn't__ a catchall thread safety change to the entire CirculationAPI architecture, but a fix specifically for the PatronActivityThread. 
Any new threads introduced into the CM will have to be handled again.

## Motivation and Context
We are seeing a number of errors coming back from our production webservers:
> `Exception in web app: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30 (Background on this error at: http://sqlalche.me/e/13/3o7r)`
>

[Notion](https://www.notion.so/lyrasis/QueuePool-limit-reached-fd9d8a7f6f1e4e7785c2040cbc1d00ef)

## How Has This Been Tested?
- Ran the CM under uwsgi locally (3 processes, 2 threads)
- Changed the bookshelf_sync code to ignore caching
- Hit the `/patron/loans` endpoint continuously using `siege`
- Observed the lack of leaking connections with `engine.pool.status()` [[source](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_3/lib/sqlalchemy/pool/impl.py#L192)]

## Checklist

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -1634,6 +1634,14 @@ class CirculationAPI:
                     "Synced %s in %.2f sec", self.api.__class__.__name__, after - before
                 )
 
+                # While testing we are in a Session scope
+                # we need to only close this if api._db is a flask_scoped_session.
+                if type(api._db) != Session:
+                    # Since we are in a Thread using a flask_scoped_session
+                    # we can assume a new Session was opened due to the thread activity.
+                    # We must close this session to avoid connection pool leaks
+                    api._db.close()
+
         threads = []
         before = time.time()
         for api in list(self.api_for_collection.values()):
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -1636,11 +1636,11 @@ class CirculationAPI:
 
                 # While testing we are in a Session scope
                 # we need to only close this if api._db is a flask_scoped_session.
-                if type(api._db) != Session:
+                if getattr(self.api, "_db", None) and type(self.api._db) != Session:
                     # Since we are in a Thread using a flask_scoped_session
                     # we can assume a new Session was opened due to the thread activity.
                     # We must close this session to avoid connection pool leaks
-                    api._db.close()
+                    self.api._db.close()
 
         threads = []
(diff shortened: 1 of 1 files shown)


RishiDiwanTT:
The drop in code coverage is due to the added line specifically meant to be avoided during a test run.

api/circulation.py
```diff
@@ -1634,6 +1634,14 @@ def run(self):
                     "Synced %s in %.2f sec", self.api.__class__.__name__, after - before
                 )
 
+                # While testing we are in a Session scope
+                # we need to only close this if api._db is a flask_scoped_session.
+                if getattr(self.api, "_db", None) and type(self.api._db) != Session:
```
jonathangreen, line 1639: I wonder if it would be better to not do this check here, but create a `db_close()` function that we can mock out when we are calling tests, so we're sure that our sessions all get closed?

jonathangreen (commented):
This is really great work Rishi! Thank you for hunting this mystery down.

I have one question / comment about the test for `Session`, but otherwise I'd like to get this rolled out to production very quickly, since we're seeing lots of occurrences of this error.

api/circulation.py
RishiDiwanTT: I can mock the scoped session close() method itself, I think that would work.

RishiDiwanTT: While doing this I can see dozens of places this will need to be mocked, making this a little messy.
Additionally, it would be confusing while writing new tests and seeing them fail for no apparent reason.

jonathangreen (approved):
Approved this one, but not going to merge until I can make a hotfix branch for it

PR #974, opened 2023-04-03
Fixing exception in CustomListsController - missing url_for

pcustic:
## Description

Switched out the last old-style CDN aware url_for with flask's url_for. 

## Motivation and Context

This was a leftover from this [PR](https://github.com/ThePalaceProject/circulation/pull/785). It caused exception on `CustomListsController` since it couldn't create the next page custom list url because the `self.url_for` method was missing. It wasn't triggered in tests since this method is only used for lists that needed to paginate their entries.

[Notion](https://www.notion.so/lyrasis/Exception-in-web-app-CustomListsController-object-has-no-attribute-url_for-e41ce7f4b1284bf8b48e5a432d838370)

## How Has This Been Tested?

Launched CM locally and checked that this error is not popping up. Also, added one test with list that contains more than 50 (default pagination setting) entries to trigger creation of the next pagination link for custom lists.  

## Checklist


dbernstein (approved):
LGTM.  Nice work.

PR #980, opened 2023-04-04
Disable failing patron authentication (DB Migration Required)

RishiDiwanTT:
## Description
The LibraryAuthenticator skips authentication mechanisms based on their status
    
There is also a random sampling built into the disable logic
so the disabled integrations get sampled during the timeout window
Now instead of raising an error the authenticated_patron method will return a problem detail
This is so the error is written to the DB and not rolled back at the end of the request

## Motivation and Context
Over the last several weeks, we have seen a number of patron authentication mechanisms fail (the remote ILS goes offline). Because these connections take some time to fail, they cause a backup of requests on the webserver, causing other library on the same CM not to be able to authenticate.

[Notion](https://www.notion.so/lyrasis/Disable-patron-authentication-mechanism-after-repeated-failures-459919ba61034183be2221ee762d94a5)

## How Has This Been Tested?
Manually Set and unset the status while hitting an authenticated endpoint
New unit tests were written

## Checklist


api/authenticator.py
```diff
@@ -843,14 +852,44 @@ def authenticated_patron(self, _db, header):
...
+            error=str(error),
+        )
+        return record
+
+    def _is_provider_available(self, _db, provider):
+        if provider.external_integration(_db).status != ExternalIntegration.STATUS.RED:
+            return True
+        # Provider status is RED, should we still try a request?
+        # Let's sample 1 in 10 requests to keep the errors up to date
+        return random.random() >= 0.9
```
RishiDiwanTT, line 891: Randomly chosen sample rate.

api/sip/client.py
```diff
@@ -351,7 +352,9 @@ def login(self):
                 self.location_code,
             )
             if response["login_ok"] != "1":
-                raise OSError("Error logging in: %r" % response)
+                raise PatronAuthorizationFailedException(
+                    "Error logging in: %r" % response
+                )
```
RishiDiwanTT, line 357: In case the credentials provided are incorrect, it should not contribute to the failure counts of the auth mechanism, hence this exception type was changed to differentiate the login failure vs API failure

dbernstein (approved):
Looks good to me. Nice work, @RishiDiwanTT.

core/jobs/update_integration_status.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class UpdateIntegrationStatusScript(Script):
+    """Script to update the external integration status fields
+    based on any errors logged in the db
+    """
+
+    # 10 errors in the last 30 minutes
+    ERROR_WINDOW_DELTA = datetime.timedelta(minutes=30)
+    ERROR_WINDOW_COUNT = 10
```
RishiDiwanTT, line 17: I forgot to draw attention to these numbers, they were chosen because they are numbers, and for no other reason.
Should they be changed to something with more rationale behind them?
@dbernstein @jonathangreen

core/model/configuration.py
```diff
@@ -122,12 +123,32 @@ class ExternalIntegrationLink(Base, HasSessionCache):
...
+
 class ExternalIntegration(Base):
 
     """An external integration contains configuration for connecting
     to a third-party API.
     """
 
+    class STATUS:
+        RED = 0
+        GREEN = 1
```
tdilauro, line 150: Any reason not to just make this an enum?

tdilauro (changes requested):
This is looking pretty good!

I have a couple concerns, which are discussed in the comments below, and one minor general general suggestion, which is to end sentences in comments and docstrings with a period/full stop.

api/sip/client.py
RishiDiwanTT: A failed login would usually mean the API server is running just fine and the User has incorrectly input their credentials. This should not impact other Users on the system who would be using the functioning API for authentication.

The response being logged and returned in the Exception is a json value, we cannot reliably depend on it string format to differentiate the types. I also believe having a different Exception type altogether is more explicit and future-proof as opposed to differentiating based on the string messages within them.

core/model/configuration.py
RishiDiwanTT: Laziness. This has been fixed. Not in me, in the code.

jonathangreen: I was reading through my GitHub notifications this morning while I was eating breakfast, and this comment gave me a good laugh 🤣

core/jobs/update_integration_status.py
dbernstein: @RishiDiwanTT :  I like numbers.  Numbers are good.  ;) It seems like as good a guess as any.  I think 10 errors in 1/2  hour is a good place to start.

RishiDiwanTT:
@tdilauro I'll wait for your approval before merging.

api/sip/client.py
tdilauro: I was focused in on SIP2 service, where login is distinct from patron lookup. In that (SIP2) case, login failure would, in a practical sense, be tantamount to service unavailability. But that is not the case for all of our integrations.

api/authenticator.py
tdilauro: I think this one might've been accidentally resolved, so unresolving. ☝🏽

RishiDiwanTT: I removed this logic in this [commit](https://github.com/ThePalaceProject/circulation/pull/980/commits/d456f194e94fe86fc77828f81e0a883d50912ca7) hence marked it as resolved

tdilauro (approved):
I think this is ready to go! 🏁

PR #986, opened 2023-04-05
Fix attribute error in the single item feed

RishiDiwanTT:
## Description
Protect against the rare case where we may not have a licensepool or the licensepool does not have the right relationships

## Motivation and Context
We are seeing the following exception happening on our production CMs:
`AttributeError: 'NoneType' object has no attribute 'work'`
in the file `  File "/var/www/circulation/api/opds.py", line 1725, in single_item_feed`

[Notion](https://www.notion.so/lyrasis/Exception-in-web-app-NoneType-object-has-no-attribute-work-8a2e407b710d42e69c1f05638982dcd1)

## How Has This Been Tested?
A unit test has been written to mock the situation.

## Checklist

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -1843,6 +1843,10 @@ class LoanController(CirculationManagerController):
             feed = LibraryLoanAndHoldAnnotator.single_item_feed(
                 self.circulation, loan, fulfillment=fulfillment
             )
+            if isinstance(feed, ProblemDetail):
+                # This should typically never happen, since we've gone through the entire fulfill workflow
+                # But for the sake of return-type completeness we are adding this here
+                return feed
             if isinstance(feed, Response):
                 return feed
             if isinstance(feed, OPDSFeed):
--- a/api/controller.py
+++ b/api/controller.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import email
 import json
 import logging
@@ -5,7 +7,7 @@ import os
 import urllib.parse
 from collections import defaultdict
 from time import mktime
-from typing import List, Optional
+from typing import TYPE_CHECKING, Any
 from wsgiref.handlers import format_date_time
 
 import flask
@@ -125,6 +127,9 @@ from .opds import (
 from .problem_details import *
 from .shared_collection import SharedCollectionAPI
 from .testing import MockCirculationAPI, MockSharedCollectionAPI
+
+if TYPE_CHECKING:
+    from werkzeug import Response as wkResponse
 
 
 class CirculationManager:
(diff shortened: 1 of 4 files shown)


api/opds.py
```diff
@@ -1722,7 +1723,17 @@ def single_item_feed(
...
+
+        # Sometimes the pool or work may be None
+        # In those cases we have to protect against the exceptions
+        try:
+            work = license_pool.work or license_pool.presentation_edition.work
+        except AttributeError:
+            return NOT_FOUND_ON_REMOTE
+
+        if not work:
+            return NOT_FOUND_ON_REMOTE
```
tdilauro, line 1735: Minor, because this looks like it would work fine: Instead of using both a try/except and the conditional, could we just check all these conditions at once and then either return the problem detail or set `work`?

api/opds.py
```diff
@@ -1722,7 +1723,17 @@ def single_item_feed(
             )
 
         _db = Session.object_session(item)
-        work = license_pool.work or license_pool.presentation_edition.work
+
```
tdilauro, line 1726: Could we move the definition of `_db` below the definition of `work` here? If we fail to acquire a work, I don't think we'll need the db session.

api/opds.py
```diff
@@ -1722,7 +1723,17 @@ def single_item_feed(
...
+        # Sometimes the pool or work may be None
+        # In those cases we have to protect against the exceptions
+        try:
+            work = license_pool.work or license_pool.presentation_edition.work
+        except AttributeError:
+            return NOT_FOUND_ON_REMOTE
```
tdilauro, line 1732: Because we might return a problem detail, it would be good to update the docstring `:return:` of this method. While we're at it, it would be useful to add type hints, including the return types.

api/controller.py
```diff
@@ -1843,6 +1843,10 @@ def fulfill_part_url(part):
             feed = LibraryLoanAndHoldAnnotator.single_item_feed(
                 self.circulation, loan, fulfillment=fulfillment
             )
+            if isinstance(feed, ProblemDetail):
+                # This should typically never happen, since we've gone through the entire fulfill workflow
+                # But for the sake of return-type completeness we are adding this here
+                return feed
```
tdilauro, line 1849: We're adding a new return type to this method, so it would be good to add type hints to it.

tdilauro (commented):
Overall, this looks good.

I have a few specific comments below and one minor general documentation style observation: It would be good to put a period/full stop at the end of sentences in comments and docstrings.

api/opds.py
RishiDiwanTT: This was my original approach but it ends being a double nested and/or statement which is confusing to understand and hence would be confusing to debug in the future.

RishiDiwanTT:
@tdilauro I added a log statement to allow us to track the cause of the missing relationship better.

PR #988, opened 2023-04-05
Provides a configuration script to add SAML federations to the CM.

tdilauro:
## Description

Provides a configuration script to add SAML federations to the CM.

## Motivation and Context

The method for adding this configuration had been previously captured -- incorrectly -- in a database migration. That migration has since been removed from current versions of the project.

## How Has This Been Tested?

- Tested locally in my development environment.
- No new tests have been added for this script.


pcustic (approved):
👍

PR #989, opened 2023-04-06
Remove work_controller.report endpoint

pcustic:
## Description

This PR removes endpoint for the `WorkController` `report` method. The code for the method was removed in [this PR](https://github.com/ThePalaceProject/circulation/pull/275), however, the actual endpoint (route) was never removed.

## Motivation and Context

[Notion](https://www.notion.so/lyrasis/WorkController-object-has-no-attribute-report-4b2811328a304738b7f7858f55d6cd3a)

Since the code for the method was removed but the endpoint was still there we were getting errors because someone was trying to access that endpoint. Endpoint should have been removed in the original PR.

## How Has This Been Tested?

Updated tests and run them.

## Checklist

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -767,19 +767,6 @@ class LibraryAnnotator(CirculationManagerAnnotator):
     def annotate_work_entry(
         self, work, active_license_pool, edition, identifier, feed, entry
     ):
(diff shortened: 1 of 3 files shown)


io7m (approved):
Looks good to me.

PR #1006, opened 2023-04-14
Collection based coverage records

RishiDiwanTT:
## Description
Changed the OPDS and OPDS2 importers to maintain coverage records PER collection

Earlier they maintained one record per identifier, now it will be multiple records per identifier This will facilitate the re-import of a deleted collection

:warning: Merging this change causes all OPDS and OPDS2 feeds to re-import from scratch since there are currently no CoverageRecords with collections associated with them.

## Motivation and Context
A collection will fail to re-import after being deleted and re-added. This appears to be a problem with the collection deletion cleanup logic; some record of the collection remains after being deleted, so the importer skips over all the entries because they appear to already be present in the CM.

[Notion](https://www.notion.so/lyrasis/Improve-collection-deletion-clean-up-logic-35e81b1ec14a4f3f82670311ae6da8ca)

## How Has This Been Tested?
Deleted and recreated both OPDS and OPDS2 type collections, both times the titles were re-imported.
Unit tests have been updated for collection based CoverageRecords.

## Checklist

### DIFF
--- a/alembic/versions/20230501_f9985f6b7767_remove_import_coverage_records_without_.py
+++ b/alembic/versions/20230501_f9985f6b7767_remove_import_coverage_records_without_.py
@@ -1,0 +1,24 @@
+"""Remove import coverage records without collections
+
+Revision ID: f9985f6b7767
+Revises: 3ee5b99f2ae7
+Create Date: 2023-05-01 10:07:45.737475+00:00
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "f9985f6b7767"
+down_revision = "3ee5b99f2ae7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "DELETE FROM coveragerecords WHERE collection_id IS NULL AND operation='import'"
+    )
+
+
+def downgrade() -> None:
+    pass
--- a/alembic/versions/20230501_f9985f6b7767_remove_import_coverage_records_without_.py
+++ b/alembic/versions/20230501_f9985f6b7767_remove_import_coverage_records_without_.py
@@ -1,7 +1,7 @@
 """Remove import coverage records without collections
 
 Revision ID: f9985f6b7767
-Revises: 3ee5b99f2ae7
+Revises: 5dcbc92c20b2
 Create Date: 2023-05-01 10:07:45.737475+00:00
 
 """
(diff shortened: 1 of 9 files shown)


io7m (approved):
I don't see any problems with this.

jonathangreen:
Added the incompatible changes label here, so that the re-import mentioned gets noted in the release notes.

tdilauro:
@RishiDiwanTT This seems to work as expected. Do we need a DB migration to clean out the non-collection-based import coverage records for existing OPDS 1/2 collections?

RishiDiwanTT:
> @RishiDiwanTT This seems to work as expected. Do we need a DB migration to clean out the non-collection-based import coverage records for existing OPDS 1/2 collections?

This CoverageRecord type is used for OPDS, OPDS2 and ODL as far as I could tell but this issue is only for OPDS v1. So I'm not sure if removing CoverageRecords should be done unless we're sure which identifier it's for, and I'm not sure how to check that.

tdilauro:
@RishiDiwanTT: Updating with my comments from our slack thread...

Is this question about how to distinguish between those different import (license) types or something else? Generally speaking, it should be safe to remove import coverage records, in that this will result only in potential re-import.

It is possible that, because we have not included collection as a discriminator in those import coverage records, we might remove import coverage from an unexpected collection, but the worst case would be that we need to re-import. To me that seems preferable to carrying forward "zombie" records that are missing the collection.

We could select just those records that have at least one OPDS-based license integration, but not sure if it's worth the trouble, especially if we're using import coverage for only these kinds of imports anyway.

tdilauro:
@RishiDiwanTT Does this seem reasonable to you?

RishiDiwanTT:
> @RishiDiwanTT: Updating with my comments from our slack thread...
> 
> Is this question about how to distinguish between those different import (license) types or something else? Generally speaking, it should be safe to remove import coverage records, in that this will result only in potential re-import.
> 
> It is possible that, because we have not included collection as a discriminator in those import coverage records, we might remove import coverage from an unexpected collection, but the worst case would be that we need to re-import. To me that seems preferable to carrying forward "zombie" records that are missing the collection.
> 
> We could select just those records that have at least one OPDS-based license integration, but not sure if it's worth the trouble, especially if we're using import coverage for only these kinds of imports anyway.

Sounds good, I'll create a migration to remove "orphaned" coverage records.

tdilauro:
> Sounds good, I'll create a migration to remove "orphaned" coverage records.

I think this is what you mean, but just to be completely clear, we'd want to delete _only_ the IMPORT_OPERATION ("import") coverage records.

RishiDiwanTT:
> > Sounds good, I'll create a migration to remove "orphaned" coverage records.
> 
> I think this is what you mean, but just to be completely clear, we'd want to delete _only_ the IMPORT_OPERATION ("import") coverage records.

@tdilauro This is done. All OPDS, OPDS2, ODL and ODL2 collections will re-import after merge.

PR #1018, opened 2023-04-18
Moved from elasticsearch-py to opensearch-py

RishiDiwanTT:
## Description
Just replaced the elasticsearch imports and classes with opensearch.
This change does not change names in the code to favour opensearch.
However, it seemed only right to replace the comments to have opensearch in them.
Moved to opensearch 1.x as per the [compatibility table](https://github.com/opensearch-project/opensearch-py/blob/1.x/COMPATIBILITY.md).

## Motivation and Context
Since the use of elasticsearch with the CM is now deprecated 
Opensearch-py (atleast 1.x) works with elasticsearch as well

[Notion](https://www.notion.so/lyrasis/Update-our-dependencies-for-opensearch-3bbe0e86dab74aa1a23ce04794a79052)

## Checklist


core/external_search.py
```diff
@@ -1546,17 +1546,17 @@ def __init__(self, query_string, filter=None, use_query_parser=True):
             # run the 'fuzzy' hypotheses at all.
             self.fuzzy_coefficient = 0
 
-    def build(self, elasticsearch, pagination=None):
-        """Make an Elasticsearch-DSL Search object out of this query.
+    def build(self, opensearch, pagination=None):
```
jonathangreen, line 1549: Looks like the param is changing here. Happy to see it change, but just wanted to make sure we searched for usages of this function that might be using the old param name.

tests/core/test_external_search.py
```diff
@@ -158,11 +158,12 @@ def test_setup_index_creates_new_index(
         # The alias hasn't been passed over to the new index.
         alias = "test_index-" + external_search_fixture.search.CURRENT_ALIAS_SUFFIX
         assert alias == external_search_fixture.search.works_alias
+        print(current_index, alias)
```
jonathangreen, line 161: Print statement from debugging maybe?

tests/core/test_external_search.py
```diff
@@ -158,11 +158,12 @@ def test_setup_index_creates_new_index(
...
         assert True == external_search_fixture.search.indices.exists_alias(
-            current_index, alias
+            alias, index=current_index
```
jonathangreen, line 163: Did the param order here change between the ES and OS library? Were there any other changes we should be watching out for?

core/external_search.py
RishiDiwanTT: Within the function the variable has been refactored and the calling methods don't use an elasticsearch specific name

tests/core/test_external_search.py
RishiDiwanTT: How embarrassing

tests/core/test_external_search.py
RishiDiwanTT: Yes the signature changed between the libraries, this is the only code change that occured.

core/external_search.py
```diff
@@ -24,9 +22,11 @@
...
     Nested,
 )
-from elasticsearch_dsl.query import Query as BaseQuery
-from elasticsearch_dsl.query import Range, Regexp, Term, Terms
-from flask_babel import lazy_gettext as _
+from opensearch_dsl.query import Query as BaseQuery
+from opensearch_dsl.query import Range, Regexp, Term, Terms
+from opensearchpy import OpenSearch
+from opensearchpy.exceptions import OpenSearchException, RequestError
+from opensearchpy.helpers import bulk as opensearch__bulk
```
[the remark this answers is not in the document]
RishiDiwanTT: Nope, no reason, fixed.

jonathangreen (approved):
Looks good

pcustic:
Do we want to remove the mention of the Elasticsearch from the README since this drops support for it?

RishiDiwanTT:
There is a similar ticket to remove all elasticsearch items that I will be
doing next

PR #1023, opened 2023-04-19
Do not initialise SQLA schemas and data during an AlembicMigration

RishiDiwanTT:
## Description
Set the initialize_schema and initialize_data to false for the DB session when running the script.
Now onwards the alembic migrations MUST contain the new table schemas.

## Motivation and Context
~~Since SQLA creates new types and tables on startup, this becomes a conflict.~~
~~Since the migrate database script is a python script the type gets created and then the script tries to recreate it, failing the migration entirely~~
Running an upgrade with a new schema type was failing due to, both, the CM and alembic trying to create the types.
In response to this [Notion Update](https://www.notion.so/lyrasis/Alembic-Issue-where-migrations-appear-to-be-applied-without-the-migration-stamp-being-updated-e5b4bea669b149a794f358c468c8d1a8?pvs=4#1a7b4cac1fed4faca30107c7b7a3939f)

## How Has This Been Tested?
Ran the migrate_database script multiple times while upgrading and downgrading the database.
Ran the migrate_database in parallel with itself multiple times to ensure there was no race condition.

## Checklist


core/bin/migrate_database
```diff
@@ -3,6 +3,11 @@
 
 import startup  # noqa: autoflake
 
+from core.config import Configuration
+from core.model import SessionManager
 from core.scripts import AlembicMigrateVersion
 
-AlembicMigrateVersion().run()
+session = SessionManager.session(
+    Configuration.database_url(), initialize_data=False, initialize_schema=False
```
jonathangreen, line 11: Might be helpful to add a comment here with some context as to why we need to set `initialize_data` and `initialize_schema` to `False`.

jonathangreen (approved):
Looks good to me

PR #1026, opened 2023-04-19
Fall back to the patron external_identifier if no username has been s…

dbernstein:
…et for the patron.

## Description

This update very simply allows the patron external_identifier to be used in case there is no username as is the case when a patron record is created by the SAML authentication process.

## Motivation and Context
https://www.notion.so/lyrasis/For-PQ-content-a-500-error-is-generated-during-acquisition-process-resulting-into-the-app-restartin-531f60890b564e54bd0140ece63a96b6

## How Has This Been Tested?

Once I get a preliminary approval I will test this on the columbia test instance before merging it.

## Checklist


jonathangreen (approved):
Looks good to me @dbernstein

pcustic (approved):
Looks good to me also 👍  I added bug label so it's properly labeled in the release notes

PR #1032, opened 2023-04-21
Update configuration for Sirsi Auth

jonathangreen:
## Notes

This PR is based on the work in https://github.com/ThePalaceProject/circulation/pull/1027 and that PR will have to go in before this one does.

## Description

While working on updating how we are storing our configuration settings, I ran into a bit of an issue with the Sirsi auth provider, because it overrides `SETTINGS` and `LIBRARY_SETTINGS` without inheriting the values from the base class. So as we move to concrete settings objects, the base class constructor starts expecting settings that don't exist on the `SirsiDynixHorizonAuthenticationProvider` class. This is different from the other auth providers, which just add additional settings.

## Motivation and Context

This PR has two main changes, each is a separate commit, so it should be easier to review. 

- https://github.com/ThePalaceProject/circulation/commit/85fc3d724c493187362065bcaf3b1d86679cc277 moves the tests for this auth provider into their own file, so they are a bit easier to work on. 
- https://github.com/ThePalaceProject/circulation/commit/6c48525e3477ce7de6701027cc56f199b76bb077 contains the actual changes.
  - Use the `SETTINGS` and `LIBRARY_SETTINGS` from the base class.
  - Leave the enforcement of the patron type prefix to the base class library identifier enforcement code.
  - Update the disallowed suffix blocks to not depend on the prefix

I will note, I really don't like this bit of code included in this PR:
https://github.com/ThePalaceProject/circulation/blob/6c48525e3477ce7de6701027cc56f199b76bb077/api/sirsidynix_authentication_provider.py#L88-L97

Any suggestions on cleaning that up would be appreciated. My end goal in this refactoring work is to completely replace the `SETTINGS` and `LIBRARY_SETTINGS` properties with Pydantic dataclasses that contain all the settings for each external integration, so eventually this code will go away. Having the Sirsi auth integration inherit its settings from the base class is just a step in that direction. 

## How Has This Been Tested?

Running test suite locally.

## Checklist


jonathangreen:
@RishiDiwanTT I tagged you for review on this one. Would appreciate if you could take a look at the draft and let me know what you think, before I create the settings migration for this one.

api/sirsidynix_authentication_provider.py
```diff
@@ -64,18 +61,7 @@ class Keys:
...
-            "label": _("Test Identifier"),
-            "description": BasicAuthenticationProvider.TEST_IDENTIFIER_DESCRIPTION_FOR_OPTIONAL_PASSWORD,
-            "required": True,
-        },
-        {
-            "key": BasicAuthenticationProvider.TEST_PASSWORD,
-            "label": _("Test Password"),
-            "description": BasicAuthenticationProvider.TEST_PASSWORD_DESCRIPTION_OPTIONAL,
-        },
-    ]
```
jonathangreen, line 78: These settings are part of `BasicAuthenticationProvider.SETTINGS` so no need for them here.

api/sirsidynix_authentication_provider.py
```diff
@@ -86,15 +72,6 @@ class Keys:
...
         },
-        {
-            "key": Keys.LIBRARY_PREFIX,
-            "label": _("Library Prefix"),
-            "description": _(
-                "This is used to match a member of the Library to their member branch location."
-                " Should be a prefix letter, eg. 'p' or 'co' etc..."
-            ),
-            "required": True,
-        },
```
jonathangreen, line 97: The `LIBRARY_PREFIX` setting is removed, and replaced with the `LIBRARY_IDENTIFIER_RESTRICTION` setting on the base class.

api/sirsidynix_authentication_provider.py
```diff
@@ -179,32 +162,24 @@ def remote_patron_lookup(
         if not data or "fields" not in data:
             return None
 
+        patrondata.complete = True
```
jonathangreen, line 165: Moved this up, so that `complete` always gets set, even if we return early from this function.

api/sirsidynix_authentication_provider.py
```diff
@@ -179,32 +162,24 @@ def remote_patron_lookup(
...
         fields: dict = data["fields"]
         patrondata.personal_name = fields.get("displayName")
         patron_type: str = fields["patronType"].get("key", "")
-
-        patrondata.external_type = patron_type
+        patrondata.library_identifier = patron_type
```
jonathangreen, line 169: Putting the `patron_type` into `library_identifier` lets us use the `library_identifier_restriction` settings on it.

tests/api/test_sirsidynix_auth_provider.py
```diff
@@ -0,0 +1,348 @@
...
+                True,
+            ),
+            (
+                SirsiDynixHorizonAuthenticationProvider.LIBRARY_IDENTIFIER_RESTRICTION_TYPE_PREFIX,
+                "abc",
+                PATRON_OF_ANOTHER_LIBRARY,
+            ),
+        ],
+    )
+    def test_full_auth_request(
```
jonathangreen, line 218: New test to make sure that we can still block users of a different library

api/sirsidynix_authentication_provider.py
```diff
@@ -179,32 +162,24 @@ def remote_patron_lookup(
...
-        patron_suffix = patron_type[len(self.sirsi_library_prefix) :]
-        if patron_suffix in self.sirsi_disallowed_suffixes:
-            patrondata.block_reason = SirsiBlockReasons.PATRON_BLOCKED
-            return patrondata
+        # If the patron type ends with a disallowed suffix the
+        # patron will be authenticated but marked as blocked.
+        for suffix in self.sirsi_disallowed_suffixes:
+            if patron_type.endswith(suffix):
+                patrondata.block_reason = SirsiBlockReasons.PATRON_BLOCKED
+                return patrondata
```
RishiDiwanTT, line 182: An issue with not using the "prefix" is that we may have a prefix of "ca" and a disallowed suffix of "ad" which means "caad" should be disallowed.
Theoretically we may encounter a patron type of "cad" (suffix "d") which should be allowed, but will be blocked through this logic.
If this system is getting confusing we can just change this to DISALLOWED PATRON TYPES, wherein the entire type is explicitly disallowed eg. the i-team would input the entire "caad" as a disallowed type avoiding any confusion.

jonathangreen: That is a good point @RishiDiwanTT. I don't think we have any cases like this currently, but I'll have to go back over the spreadsheet of types to check and be sure.

RishiDiwanTT (approved):
Looks Good :+1:

PR #1033, opened 2023-04-21
Make sure remote_patron_lookup calls are necessary

jonathangreen:
## Description

While writing the tests for #1032, I realized that in the case of a new patron, we were calling `remote_patron_lookup` twice. This is a potentially expensive call, so it would be nice to avoid making it twice. 

This PR is based on #1027, but I wanted to keep it as a separate PR, since it changes the logic in the `authenticate` function, so it should get a good code review before going in.

## Motivation and Context

This reworks the logic in the `authenticate` function, to avoid calling out to `remote_patron_lookup` in the case where we already have a complete `patrondata` and adds a new test, to make sure that we are calling `remote_patron_lookup` the expected number of times.

## How Has This Been Tested?

Running unit tests locally.

## Checklist

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -1905,67 +1905,41 @@ class BasicAuthenticationProvider(AuthenticationProvider, ABC):
 
         # First, try to look up the Patron object in our database.
         patron = self.local_patron_lookup(_db, username, patrondata)
-        if patron and (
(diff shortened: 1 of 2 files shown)


RishiDiwanTT (approved):
Looks Good :+1:

PR #1034, opened 2023-04-21
Drop parameterized and use pytest instead.

jonathangreen:
## Description

We use the parameterized module in some places to parameterize our test functions. Pytest provides this functionality, allowing us to drop this dependency.

## Motivation and Context

We use `@pytest.mark.parametrize` and `@parameterized.expand` other places. Its nice to have the same pattern everywhere and it lets us drop another dependency.

## How Has This Been Tested?

Running tests locally.

## Checklist


tdilauro (approved):
Looks good! 🚀 

I minor comment about consistent formatting in param specification.

io7m:
Glad to see this going in. The test suite refactoring removed a lot of these, but it was always irritating to see that `parameterized` wasn't quite gone yet.

PR #1040, opened 2023-04-21
Enable the authentication document cache by default

jonathangreen:
## Description

This reverts commit a8f65748f2e9a995042205ce690b31a3a80ba84b.

See [notion](https://www.notion.so/lyrasis/Investigate-CM-authentication-document-caching-289f93f5675c4bf6bd529efc24ba150a?pvs=4). 

## Motivation and Context

This re-enables authentication document caching reverting an older commit that disabled it. Authentication documents are expensive to build, so it would be nice to be able to cache them. There is some risk that this will trigger a bug, outlined here https://jira.nypl.org/browse/SIMPLY-3346, which is the reason it was disabled. However a lot has changed since that bug was a problem, so we don't think it will cause any issue here. 

If it does end up causing an issue, the plan is to just revert this PR.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -31,10 +31,6 @@ class Configuration(CoreConfiguration):
     # documents are cached.
     AUTHENTICATION_DOCUMENT_CACHE_TIME = "authentication_document_cache_time"
 
-    # The name of a setting that turns UWSGI debugging information on
-    # or off.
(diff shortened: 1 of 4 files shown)


RishiDiwanTT (approved):
Looks good :+1:

PR #1073, opened 2023-05-03
Switch to using Self when type checking

jonathangreen:
## Description

This just updates our code to use the new [PEP 673](https://peps.python.org/pep-0673/) `typing.Self` style `Self` annotations, instead of the old style ones. It doesn't functionally change the code at all.

## Motivation and Context

We can use the new [`typing.Self`](https://mypy.readthedocs.io/en/stable/generics.html?#automatic-self-types-using-typing-self) in Python versions before 3.11 by importing `typing_extensions`. See the [mypy docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module).

The new `Self` notation is a bit easier at the expense of a slightly more ugly import. It also gives us the ability to use other new type extensions like `ParamSpec`.

## How Has This Been Tested?

Running Mypy.

## Checklist

### DIFF
--- a/api/admin/model/dashboard_statistics.py
+++ b/api/admin/model/dashboard_statistics.py
@@ -1,8 +1,14 @@
 from __future__ import annotations
 
-from typing import List, TypeVar
+import sys
+from typing import List
 
 from pydantic import BaseModel, Extra, Field, NonNegativeInt
+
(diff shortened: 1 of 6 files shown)


tdilauro (approved):
🚀

PR #1075, opened 2023-05-04
Mypy linting mistakes fixed

RishiDiwanTT:
## Description
Not sure how [this github action](https://github.com/ThePalaceProject/circulation/actions/runs/4850946216) ran successfully but the mypy typing was incorrect after the change from using dict/str's to werkzeug.Authorization objects.
The inner function still need to use dict/str types.

## Motivation and Context
Main broke due to these incorrectly typed methods.

## How Has This Been Tested?
Mypy ran successfully.

## Checklist


jonathangreen:
@RishiDiwanTT looks like this is causing some test failures

dbernstein (approved):
🐎

PR #1076, opened 2023-05-04
JWE access token implementation

RishiDiwanTT:
## Description

Token creation endpoint in implemented
Token decode and key generation has been implemented in the auth class
Implemented JWE key rotation scripts and cron
Refactored api.authenticator by moving out the base classes into their own files
since the file size had gotten to >2000 lines and kept growing

This mechanism has only been implemented for the basic authentication type authorization because the SAML authorization is already local.

The provider has an authentication document component

<details>
<summary>Authentication Document</summary>

```
{
    "description": "Library Barcode + Token",
    "type": "http://librarysimplified.org/authtype/basic-token",
    "inputs": {
        "login": {
            "keyboard": "default"
        },
        "password": {
            "keyboard": "default"
        }
    },
    "links": [
        {
            "rel": "authenticate",
            "href": "http://localhost:6500/localtest/patrons/me/token/"
        }
    ],
    "labels": {
        "login": "Barcode",
        "password": "Pin"
    }
}
```
</details>

## Motivation and Context
We needed to implement token based authentication for our API endpoints, so that we do not have to always validate a barcode / pin against the client ILS.

[Notion](https://www.notion.so/lyrasis/Implement-token-based-authentication-f601474812284d449ba5f61ca0b9401e)

## How Has This Been Tested?
Local unit tests have been updated and run.
One can request an access token via `GET /<library>/patrons/me/token` with Basic Auth.
The `Bearer` access token can be used with any authenticated API like `/patrons/me` or `/loans`.

## Checklist


api/admin/model/dashboard_statistics.py
```diff
@@ -2,30 +2,9 @@
...
-        extra = Extra.forbid
-
-    def api_dict(self, *args, by_alias=True, **kwargs):
-        """Return the instance in a form suitable for a web response.
-
-        By default, the properties use their lower camel case aliases,
-        rather than their Python class member names.
-        """
-        return self.dict(*args, by_alias=by_alias, **kwargs)
+from core.util.flask_util import CustomBaseModel
```
RishiDiwanTT, line 7: Moved this to an accessible location

api/authentication/access_token.py
```diff
@@ -0,0 +1,165 @@
...
+    @classmethod
+    def is_access_token(cls, token: str) -> bool:
+        raise NotImplementedError()
+
+    @classmethod
+    def get_integration(cls, _db):
+        raise NotImplementedError()
+
+
+class PatronJWEAccessTokenProvider(PatronAccessTokenProvider):
```
RishiDiwanTT, line 40: This class only provides the JWE token, it does not do authentication

api/authentication/base.py
RishiDiwanTT, line 1: Just refactoring changes

api/authentication/basic.py
RishiDiwanTT, line 1: Refactoring changes

api/authentication/patron_authentication_provider.py
```diff
@@ -0,0 +1,54 @@
...
+
+from api.authentication.access_token import AccessTokenProvider
+from api.authentication.base import AuthenticationProvider
+from api.problem_details import PATRON_AUTH_ACCESS_TOKEN_INVALID
+from core.model import Patron, Session, get_one
+from core.util.problem_detail import ProblemDetail
+
+
+class PatronAccessTokenAuthenticationProvider(AuthenticationProvider):
+    FLOW_TYPE = "http://librarysimplified.org/authtype/bearer"
```
RishiDiwanTT, line 13: This is an imaginary flow type, if there are semantics to this we should change it.

core/util/authentication_for_opds.py
```diff
@@ -74,7 +76,9 @@ def to_dict(self, _db):
         document = dict(id=self.id, title=self.title)
         flow_documents = document.setdefault("authentication", [])
         for flow in self.authentication_flows:
-            flow_documents.append(flow.authentication_flow_document(_db))
+            doc = flow.authentication_flow_document(_db)
+            if doc is not None:
+                flow_documents.append(flow.authentication_flow_document(_db))
```
RishiDiwanTT, line 81: This allows us to skip flows in the authentication document

api/authentication/access_token.py
```diff
@@ -0,0 +1,176 @@
...
+        key = cls.get_current_key(_db)
+        payload = dict(id=patron.id, pwd=password, typ="patron")
+
+        token = jwe.JWE(
+            jwe.json_encode(payload),
+            dict(
+                alg="dir",
+                kid=key.key_id,
+                typ="JWE",
+                enc="A128CBC-HS256",
```
io7m, line 115: Just wanted to confirm here that we're not using a predictable `iv`, given that we're using AES 128 CBC.

core/util/authentication_for_opds.py
io7m: Why do we need to skip flows?

api/authentication/patron_authentication_provider.py
io7m: I think the mobile clients will inspect this value to decide whether a server supports token auth, right?

If so, yeah, we should decide if this is the constant we want. Extremely hard to change afterwards.

core/util/flask_util.py
```diff
@@ -179,3 +181,28 @@ def boolean_value(value: str) -> bool:
     """Convert a string request value into a boolean, used for form encoded requests
     JSON encoded requests will get automatically converted"""
     return True if value in ["true", "True", True, "1"] else False
+
+
+def _snake_to_camel_case(name: str) -> str:
+    """Convert from Python snake case to JavaScript lower camel case."""
+    new_name = "".join(word.title() for word in name.split("_") if word)
+    if not new_name:
```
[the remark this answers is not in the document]
RishiDiwanTT: Hmm.. I didn't write this method, just moved it here, I can add tests for it though

bin/rotate_patron_auth_key
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
+"""Assign subjects to genres."""
```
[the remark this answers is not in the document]
RishiDiwanTT: No SIR! I did not copy from a another file!!!

core/util/authentication_for_opds.py
io7m: I'm kind of unsure about this. The clients used basic authentication to retrieve a token, that part I understand. The problem is that (given that I know a bit about how the Android client is implemented), they're going to need to understand that this "basic authentication" behaves differently to the other kind of "basic authentication" they'll see elsewhere, and that they'll need to store and use a token repeatedly. I feel like this information _should_ be in the authentication document.

core/util/flask_util.py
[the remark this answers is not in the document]
io7m: Oh, I didn't realize that.

bin/rotate_patron_auth_key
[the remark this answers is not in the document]
RishiDiwanTT: I admit guilt, and have changed my ways.

core/util/flask_util.py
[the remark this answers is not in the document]
io7m: Thanks for adding tests! This is a public service. :laughing:

core/util/authentication_for_opds.py
RishiDiwanTT: I don't think there is any documented protocol to define that in the authentication document. I think the prevailing thought here was that this would be written in directly into the code of the Basic Auth implementations and not controlled via the authentication document.

RishiDiwanTT: Eg.
```
"authentication": [
    {
        "description": "Library Barcode",
        "labels": {
            "login": "Barcode",
            "password": "PIN"
        },
        "inputs": {
            "login": {
                "keyboard": "Default"
            },
            "password": {
                "keyboard": "Default"
            }
        },
        "type": "http://opds-spec.org/auth/basic",
        "secondary": {
           "type": "http://opds-spec.org/auth/bearer",
           "href": ".../patrons/me/token",
        }
    }
],
```

io7m: I feel like it should be treated as an entirely separate authentication method, with a different `type`.

Doing so unambiguously signals to the client that they can use a different login workflow to use tokens (and by putting the token option in the authentication document ahead of plain basic authentication, it signals to them that they _should_ use tokens in preference to plain basic authentication). The clients already have separate code paths for basic auth, OAUTH, and SAML, so adding another authentication method shouldn't be a problem.

It also allows us to turn off plain basic authentication eventually, when we've determined that no clients have used it for _insert time period here_.

io7m: As an aside, the Android client is using sealed types for the credentials, so if @nunommts adds a new `PalaceBearerToken` credentials type, he'll get nice errors indicating all of the places in the codebase that now need updating. :slightly_smiling_face:

RishiDiwanTT:
@jonathangreen @io7m Authentication Document implemented

jonathangreen:
To make the conflict on this PR with #1063 less nightmarish, would it make sense to put the refactor of `api.authenticator` into its own module, into its own separate PR?

RishiDiwanTT:
> To make the conflict on this PR with #1063 less nightmarish, would it make sense to put the refactor of `api.authenticator` into its own module, into its own separate PR?

I don't think it'll be easy to revert the refactor, since the JWE implementation and the refactor are in a single commit. Stupid, I know.
It's probably the same amount of work at this point in time, so I'll keep this updated as we go along.

api/authentication/access_token.py
```diff
@@ -0,0 +1,201 @@
...
+    def rotate_key(cls, _db) -> jwk.JWK:
+        """Rotate the current JWK key in the DB"""
+        key = cls.generate_key()
+        integration = cls.get_integration(_db)
+
+        # We must make a copy() because sqlalchemy does not think
+        # a change in the same dict is a change at all.
+        # So the dict must be a new dict to register the change.
+        # https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#sqlalchemy.dialects.postgresql.JSONB
+        settings_dict: dict = cast(dict, integration.settings.copy())
```
RishiDiwanTT, line 88: @jonathangreen The fix was to have a completely new dict object as opposed to changing the current object

RishiDiwanTT: The solution came while I was asking phind.com for random solutions :sweat_smile: . Time to be replaced :robot:

jonathangreen (commented):
Started to review this today. I'll keep going with it tomorrow. One thing that is making this a bit harder to review is the refactor mixed in. It would be nice to make that its own PR, so its easier to read the diff here. 

I attempted to move around the same classes, with no changes in this PR https://github.com/ThePalaceProject/circulation/pull/1138, so we could merge the refactor separately. If that PR isn't helpful feel free to close it and make a new branch, or make any changes there.

api/authentication/patron_authentication_provider.py
```diff
@@ -0,0 +1,103 @@
...
+        return None
+
+    def _authentication_flow_document(self, _db):
+        """This auth type should not have an entry in the authentication document"""
+        token_url = url_for(
+            "patron_auth_token",
+            library_short_name=self.library(_db).short_name,
+            _external=True,
+        )
+        inputs = {"login": {"keyboard": "default"}, "password": {"keyboard": "default"}}
```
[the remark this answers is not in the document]
RishiDiwanTT: Where would we set these values if we are not setting them through the UI? Or do we want to have this provider show up in the `patron_auth_services` ?

api/authentication/access_token.py
```diff
@@ -0,0 +1,201 @@
...
+    @classmethod
+    def is_access_token(cls, token: str | None) -> bool:
+        raise NotImplementedError()
+
+    @classmethod
+    def get_integration(cls, _db):
+        raise NotImplementedError()
+
+
+class PatronAccessTokenAuthenticationSettings(AuthProviderSettings):
```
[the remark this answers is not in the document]
RishiDiwanTT: I've created a new `systemkeys` table, the credentials table is geared towards Patrons.
I wanted to use sitewide settings, but is deprecated now with the configurationsettings table. Which brings to mind a different question of how those will be supported?

api/authentication/patron_authentication_provider.py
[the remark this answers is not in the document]
RishiDiwanTT: Ok, I've changed the `settings` to inherit from the Basic Auth settings, and use those values in the authentication document. However I see no use of these values in the document since the access token provider should not have a UI for the values anyways, nor should it be presented to the user in any way.
It is a secondary authentication mechanism which does not have the features of the basic auth mechnism.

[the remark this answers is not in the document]
RishiDiwanTT: Based on our slack conversation I have changed the Access token provider to shadow the basic auth provider, from inception to authentication document creation.

jonathangreen (commented):
Functionally this looks good, tested and we're generating tokens and can use them. 

Made a few minor comments. My biggest concern is around the class hierarchy structure here. I'm wondering if `PatronAccessTokenAuthenticationProvider` doesn't really make sense to inherit from `AuthenticationProvider`.

api/authentication/patron_authentication_provider.py
```diff
@@ -0,0 +1,131 @@
...
+from api.problem_details import PATRON_AUTH_ACCESS_TOKEN_INVALID
+from core.model import Patron, Session, get_one
+from core.selftest import SelfTestResult
+from core.util.problem_detail import ProblemDetail, ProblemError
+
+if TYPE_CHECKING:
+    from core.model import Library
+
+
+class PatronAccessTokenAuthenticationSettings(AuthProviderSettings):
```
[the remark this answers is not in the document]
RishiDiwanTT: Nope, will remove

tests/api/mock_authentication_provider.py
```diff
@@ -0,0 +1,28 @@
+from api.authentication.basic import BasicAuthenticationProvider
+from core.util.http import RemoteIntegrationException
+
+
+class MockExplodingAuthenticationProvider(BasicAuthenticationProvider):
```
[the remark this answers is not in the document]
RishiDiwanTT: Looks like a merge issue, will remove it

api/authentication/patron_authentication_provider.py
```diff
@@ -0,0 +1,131 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
RishiDiwanTT: I didn't quite like it myself. How about `basic_token.py::BasicTokenAuthenticationProvider`?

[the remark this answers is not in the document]
jonathangreen: That sounds reasonable to me.

RishiDiwanTT:
@jonathangreen I think all comments have been addresses

api/sirsidynix_authentication_provider.py
```diff
@@ -25,6 +25,7 @@
 if TYPE_CHECKING:
     from requests import Response
 
+    from core.analytics import Analytics
```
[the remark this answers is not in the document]
jonathangreen: @RishiDiwanTT I think this one still needs to be addressed.

jonathangreen (approved):
I think this looks good. 

Would be nice to address these two comments before merging though: 
- https://github.com/ThePalaceProject/circulation/pull/1076#discussion_r1210306366
- https://github.com/ThePalaceProject/circulation/pull/1076#discussion_r1210300813

PR #1082, opened 2023-05-04
Remove unnecessary type imports

jonathangreen:
## Description

Removes unused type stubs, these libraries now have types integrated.
- types-flask
- types-freezegun
- types-itsdangerous
- types-mock

Remove some unused ignores:
- oauth2client
- onelogin
- opensearch.*
- parameterized

Add a dep for `itsdangerous` since we use, but didn't have a direct dependency. Bump version to a version of `freezegun` that has types (we were already using this version, just wanted to make it explicit).

## Motivation and Context

Looking at the `# type: ignore` comments added in https://github.com/ThePalaceProject/circulation/pull/1055 I was wondering why we had to ignore the new fields on the `Authorization` object. This lead me on a bit of a journey. 

## How Has This Been Tested?

Running mypy and tests locally.

## Checklist


api/admin/controller/__init__.py
```diff
@@ -259,7 +259,7 @@ def _check_redirect(target: str) -> Tuple[bool, str]:
         return True, ""
 
     @staticmethod
-    def redirect(target: str) -> Response:
+    def redirect(target: str) -> WerkzeugResponse:
```
RishiDiwanTT, line 262: The response type here was a Flask Response and so is the return statement, should it be left as-is?

RishiDiwanTT: I guess it works either way since it's the parent class

api/authenticator.py
```diff
@@ -832,13 +832,13 @@ def authenticated_patron(self, _db, auth: Authorization):
             # The patron wants to authenticate with the
             # BasicAuthenticationProvider.
             provider = self.basic_auth_provider
-            provider_token = auth.parameters  # type: ignore[attr-defined]
+            provider_token = auth.parameters
```
RishiDiwanTT, line 835: When I impremented this code the Authorization class did not have types in the werkzeug library, how is this working now?

api/util/flask.py
```diff
@@ -0,0 +1,46 @@
...
+      █▓╬░     `»` »`»»»»»░░░░░░░░░░▒▒▒▒▒▒▒▒▒Ü░░░░ [╫▌
+      ▓▓╬░      `   `»»»»»»░░░░░░░Ü▒▒▒▒▒▒▒▒▒▒Ü░░░░ [╫▌
+      ▓▓╬░»``  » `  » »»»»░░░░░░░░▒▒▒▒▒▒▒▒▒▒▒Ü░░░░ [╫▌
+      ▓▓╬░»_` ``»`»»»»»»»»»░░░░░░Ü▒▒▒▒▒▒▒▒▒▒▒▒Ü░░░ |╫▌
+      ▓▓╬░»»»»»»»»»»»»»»»»░░░░░░░▒░▒▒▒▒▒▒▒▒▒▒ÜÜ░Ü░_|╟▌
+      ╣▓╬░»»»»»»»»»»»»»»»»»░░░░░░░▒▒▒▒▒▒▒▒▒▒▒▒▒ÜÜ░⌐|╟▌
+      ╝▓╬░__»»»»»»»»»»░»░░░░░░░░░▒▒▒▒▒▒▒▒▒▒▒▒▒▒░░░_|╠╛
+      ╚╚╚╙╙╙╙╙╙╙╙╙╙╙╙╙╙╙╙╙╙╙╙╙╚╚╚╚╚╚╚ÜÜÜÜÜÜÜÜÜÜÜÜÜ╙╚ÜH
+
+    Palace: You're going to need a stiff drink after this.
```
RishiDiwanTT, line 40: :laughing:

api/util/flask.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._db: Session
+        self.manager: Any
```
RishiDiwanTT, line 46: Why not have this as a "CirculationManager"? It would make autocomplete work

api/authenticator.py
jonathangreen: Good question Rishi, I should have made what I was doing here more clear. My PR description was a bit lacking 😅. These ignores are actually what prompted this PR. 

I thought `flask` included type hints, so I was wondering why we had to put in these ignores. What this lead me to is that `flask` does include type hints, but we were still installing `types-flask`, and that is overriding the included type hints, which are out of date. 

Once I removed `types-flask`, this started working, but caused some other type errors, which is what I was addressing in the rest of this PR.

api/util/flask.py
jonathangreen: Thats a good point Rishi. I originally had this annotated as `CirculationManager`, but it caused other type errors, because the CirculationManager object doesn't have type hints for its members. 

The members of CirculationManager actually get setup in [setup_admin_controllers](https://github.com/ThePalaceProject/circulation/blob/f3c8210588107d7e5739160bdbc894b8fca0d4f9/api/admin/controller/__init__.py#L123-L129) instead of the __init__ function, so the type hints aren't implied.

So I was lazy and just made this `Any`. Maybe a better solution is to add type hints to `CirculationManager` so we actually get type info and autocomplete.

I'll see what it will take to do this.

jonathangreen: Okay, this should be addressed now. I added hints to `CirculationManager`.

api/admin/controller/__init__.py
jonathangreen: The error from mypy if this is set to `Response`:
```
api/admin/controller/__init__.py:266: error: Incompatible return value type (got "werkzeug.wrappers.response.Response", expected "flask.wrappers.Response")
```

api/authenticator.py
RishiDiwanTT: Thanks for the detailed answer! :+1:

RishiDiwanTT (approved):
Looks excellent :rocket:

PR #1099, opened 2023-05-11
Type hint the memoize decorator.

jonathangreen:
## Description

Adds type hints to the `memoize` decorator, using `ParamSpec`. Uses the approach to type hinting decorators from the [mypy docs](https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators).

## Motivation and Context

Since the `memoize` decorator didn't have any type hints, any functions that it decorates lost their type hints. This caused me some issues in https://github.com/ThePalaceProject/circulation/pull/1063 so I thought I would add type hints to it.

## How Has This Been Tested?

Running tests, running mypy

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -2090,9 +2090,9 @@ class JSONQuery(Query):
         @staticmethod
         def data_source(value: str) -> int:
             """Transform a datasource name into a datasource id"""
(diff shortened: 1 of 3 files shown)


RishiDiwanTT (approved):
This was so complicated to read :eyeglasses: :+1:

PR #1100, opened 2023-05-11
Adding more logging to SIP2 exception log

pcustic:
## Description

This adds SIP2 response and SIP2 server to the error log when we encounter missing required field.

## Motivation and Context

There has been a lot of [Expected required AO field but did not find it](https://www.notion.so/Expected-required-field-AO-but-did-not-find-it-3da3a58014a54e01b6dbd4397ba345a2?d=de728eeedf11434293b3a45ed18318ab) errors from the SIP client but current logging of that exception is not enough to understand what has really happened. This adds response and server that triggers error so we can understand if the problem is in our parser or in malformed response.
 

## How Has This Been Tested?

Manually written bad test that would trigger this error and checked if the log is in the test output.

## Checklist


RishiDiwanTT (approved):
Looks good :+1:

PR #1102, opened 2023-05-15
Escaping reserved characters in the JSON search type

RishiDiwanTT:
## Description
In case a JSON query is made for a regex/contains match we must escape the regex special characters as documented here
https://www.elastic.co/guide/en/elasticsearch/reference/current/regexp-syntax.html#regexp-reserved-characters

## Motivation and Context
No results are returned when searching for titles that include a plus sign (+) or an ampersand (&) when searching for titles on the list search screen using the contains operator.
[Notion](https://www.notion.so/lyrasis/Investigate-and-Fix-Cannot-search-for-titles-with-ampersand-or-plus-sign-in-list-title-search-using-5942cc480346485988e1c966fb0ccf21)

## How Has This Been Tested?
Manually tested the search endpoint with queries like
http://localhost:6500/localtest/search/?q={%22query%22:{%22value%22:%22Essays%20%26%20Term%22,%22key%22:%22title%22,%22op%22:%22contains%22}}&size=100&search_type=json
Updated unit tests for special characters

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -2015,6 +2015,10 @@ class JSONQuery(Query):
         REGEX = "regex"
         CONTAINS = "contains"
 
+    # Reserved characters and their mapping to escaped characters
+    RESERVED_CHARS = '.?+*|{}[]()"\\#@&<>~'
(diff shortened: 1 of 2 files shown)


io7m (approved):
Seems good to me!

PR #1103, opened 2023-05-15
Revert PR #980 temporarily, while we do further work on it

jonathangreen:
## Description

We're seeing an elevated number of 5XX errors in production due to the auth disable that was added in #980. 

Screenshot 2023-05-15 at 11 03 10 AM

## Motivation and Context

Going to disable this temporarily while we work out how we want to move this forward. 

I did a git revert the PR and then modified it to keep the migration and new models in the code base, since they have already been rolled out to production. This just reverts the code that actually records and processes the auth failures. 

## How Has This Been Tested?

- Running unit tests locally.

## Checklist


tdilauro (approved):
This looks good to me.

jonathangreen:
@RishiDiwanTT I'll wait for you to take a look at this, then get it rolled out tomorrow if it looks okay to you.

RishiDiwanTT:
While this looks fine as a code commit, do we want to keep recording the errors in the DB as a form of downtime analysis of the third party APIs, while keeping the threshold number high (like 100k errors) so it never truly goes down?
This way we can run some statistical analysis on the recorded numbers to know what the right numbers are.

jonathangreen:
I think we can pull the data we need for analysis from our existing logs, so I'm going to merge this one and make a new ticket to do some analysis and revisit.

PR #1110, opened 2023-05-16
Add pytest alembic

jonathangreen:
## Description

Add [pytest alembic](https://pypi.org/project/pytest-alembic/), to let us test our alembic migrations.

## Motivation and Context

For now this just enables the couple tests that ship with pytest alembic. But in the future we should be able to use it to write tests for our migrations. The most useful test there is the one that checks that there aren't multiple migration heads. Otherwise we don't have any CI tests that will fail if that is the case.

I saw @RishiDiwanTT you wrote up some tests in in a `__main__` block in https://github.com/ThePalaceProject/circulation/pull/1101 and I've been trying to figure out the best way to test my migration in https://github.com/ThePalaceProject/circulation/pull/1063. So I think we have the need for this. 

## How Has This Been Tested?

Running CI.

## Checklist

### DIFF
--- a/tests/migration/conftest.py
+++ b/tests/migration/conftest.py
@@ -1,28 +1,51 @@
 from __future__ import annotations
 
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Generator, Union
 
 import pytest
 import pytest_alembic
 from pytest_alembic.config import Config
 
+from core.model import SessionManager
+from tests.fixtures.database import ApplicationFixture, DatabaseFixture
+
 if TYPE_CHECKING:
     from pytest_alembic import MigrationContext
     from sqlalchemy.engine import Engine
 
     import alembic.config
-    from tests.fixtures.database import ApplicationFixture, DatabaseFixture
 
-pytest_plugins = [
-    "tests.fixtures.api_config",
-    "tests.fixtures.database",
-]
+
+@pytest.fixture(scope="function")
+def database() -> Generator[DatabaseFixture, None, None]:
+    # This is very similar to the normal database fixture and uses the same object,
+    # but because these tests are done outside a transaction, we need this fixture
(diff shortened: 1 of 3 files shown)


tests/migration/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+    alembic_config: Union[Dict[str, Any], alembic.config.Config, Config],
+    alembic_engine: Engine,
+) -> Generator[MigrationContext, None, None]:
+    """
+    Override this fixture to make sure that we stamp head. Since this is how out database
+    is initialized. The normal fixtures assume you start from an empty database.
+    """
+    config = Config.from_raw_config(alembic_config)
+    with pytest_alembic.runner(config=config, engine=alembic_engine) as runner:
+        runner.command_executor.stamp("head")
```
RishiDiwanTT, line 43: While using the alembic_runner fixture, according to the docs one of the useful methods for custom tests is the `alembic_runner.migrate_up_to` method, which will not work as intended if we've already stamped the migration state at `head`. I think.
https://pytest-alembic.readthedocs.io/en/latest/api.html#alembic-runner

jonathangreen: @RishiDiwanTT agreed that `migrate_up_to` won't work, but there is a similar function `migrate_down_to` and I think that will work for our testing. 

I wrote a test using pytest alembic for #1063 here: 
https://github.com/ThePalaceProject/circulation/blob/8f339a671cf94b36d53d8fc8359cab1efc1e820d/tests/migration/test_20230510_a9ed3f76d649.py#L59

And it seemed to work okay using `migrate_down_to`,`migrate_down_one` and `migrate_up_one` to do some testing.

RishiDiwanTT: Perfect :+1:

RishiDiwanTT (approved):
:rocket:

PR #1115, opened 2023-05-17
Fixes a bug in the recent millineum alembic migration by ensuring tha…

dbernstein:
…t if the use_post key exists with a null value the database will set the value to true.

## Description
Fixes a bug in the recent millineum alembic migration by ensuring that if the use_post key exists with a null value the database will set the value to true rather than try to insert a new configuration record for that key.

## Motivation and Context
The recently added migration, `alembic/versions/20230512_5a425ebe026c_migrate_millenium_apis_to_post.py` has a minor bug that caused a failure in the cerberus deployment.   The `use_post_requests` configuration setting has a NULL value.  As a result the script mistakenly thinks the setting doesn't exist.  

```
cerberus-circ=> select a.*, b.*  from configurationsettings a, externalintegrations b where a.external_integration_id = b.id and  key = 'use_post_requests' and value is NULL;
  id  | external_integration_id | library_id |        key        | value | id |       protocol       |    goal     |                name                | status | last_status_update
------+-------------------------+------------+-------------------+-------+----+----------------------+-------------+------------------------------------+--------+--------------------
 1203 |                       6 |            | use_post_requests |       |  6 | api.millenium_patron | patron_auth | Virtual Library Card - Staging     | green  |
 1204 |                      21 |            | use_post_requests |       | 21 | api.millenium_patron | patron_auth | Test LION Millenium                | green  |
 1205 |                      20 |            | use_post_requests |       | 20 | api.millenium_patron | patron_auth | Test LCI Sierra                    | green  |
 1206 |                      23 |            | use_post_requests |       | 23 | api.millenium_patron | patron_auth | Virtual Library Card - Development | green  |
```

## Checklist

### DIFF
--- a/alembic/versions/20230512_5a425ebe026c_migrate_millenium_apis_to_post.py
+++ b/alembic/versions/20230512_5a425ebe026c_migrate_millenium_apis_to_post.py
@@ -80,6 +80,8 @@ def upgrade() -> None:
         # This setting may or may not exist
         if config_results and len(post_results) > 0:
             use_post = post_results[0][0]
+            if use_post is None:
+                use_post == "false"
         else:
             use_post = None
 


RishiDiwanTT (approved):
Looks good, thanks for the fix :rocket: 
Weird that this even exists though :thinking: , in what case do we write NULL settings to the DB?

PR #1117, opened 2023-05-18
Fix Axis360API inheriting from Authenticator

jonathangreen:
## Description

`Axis360API` inherits from `Authenticator` which doesn't make a lot of sense. Looking back at git blame, this was done a long time ago, I think it was added by accident in testing. 

Nothing in `Axis360API` calls any methods from `Authenticator` and all the tests pass without it.

## Motivation and Context

I'm following through the class structure of `BaseCirculationAPI` to understand how we can update the storage of settings there, and this stuck out to me as obviously incorrect.

## How Has This Been Tested?

Running tests in CI.

## Checklist

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -49,7 +49,6 @@ from core.util.problem_detail import ProblemDetail
 from core.util.string_helpers import base64
 from core.util.xmlparser import XMLParser
 
-from .authenticator import Authenticator
 from .circulation import (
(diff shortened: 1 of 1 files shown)


pcustic (approved):
Looks good 👍 
Branch is out of sync with main though.
EDIT: github said it was out of sync but when I added the review the message was gone 🤷‍♂️

PR #1122, opened 2023-05-19
Fix bug in millenium alembic migration

jonathangreen:
## Description

Fixes the bug with null settings in #1101. This was fixed in #1115, but there was an issue with that fix, it was doing a comparison rather then an assignment, so the fix didn't work.

## Motivation and Context

This ended up being a one line fix. The rest of the PR adds a test for this case, since I didn't see at first that this was an equality check instead of an assignment.

Since I needed some of the same helpers to set up the test, as I did in `test_20230510_a9ed3f76d649.py`, this PR turns those functions into fixtures, and refactors that test to use the fixtures as well.

## How Has This Been Tested?

Running tests locally. The tests help me find the issue with the migration, so they did their job.

## Checklist

### DIFF
--- a/alembic/versions/20230512_5a425ebe026c_migrate_millenium_apis_to_post.py
+++ b/alembic/versions/20230512_5a425ebe026c_migrate_millenium_apis_to_post.py
@@ -81,7 +81,7 @@ def upgrade() -> None:
         if config_results and len(post_results) > 0:
             use_post = post_results[0][0]
             if use_post is None:
-                use_post == "false"
+                use_post = "false"
         else:
             use_post = None
 
--- a/tests/migration/conftest.py
+++ b/tests/migration/conftest.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+import random
+import string
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, Generator, Union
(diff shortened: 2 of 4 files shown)


alembic/versions/20230512_5a425ebe026c_migrate_millenium_apis_to_post.py
```diff
@@ -81,7 +81,7 @@ def upgrade() -> None:
         if config_results and len(post_results) > 0:
             use_post = post_results[0][0]
             if use_post is None:
-                use_post == "false"
+                use_post = "false"
```
jonathangreen, line 84: This is the actual fix. The rest of this PR is just test setup and refactoring 😅

tests/migration/conftest.py
```diff
@@ -23,7 +23,7 @@
 def database() -> Generator[DatabaseFixture, None, None]:
     # This is very similar to the normal database fixture and uses the same object,
     # but because these tests are done outside a transaction, we need this fixture
-    # to have function scope, so the dataabase schema is completely reset between
```
tdilauro, line 24: Was assuming that this was the Dutch spelling. 😄

tdilauro (commented):
Looks good! 🏁

PR #1126, opened 2023-05-19
Upgrade to SQLAlchemy 1.4

jonathangreen:
## Description

Upgrade to SQLAlchemy 1.4. This is a work in progress. Only one test failing right now though.

## Checklist


jonathangreen:
Mypy is still failing, but I'd like to get a review to see if this looks reasonable.

core/model/listeners.py
```diff
@@ -244,29 +188,96 @@ def last_update_time_change(target, value, oldvalue, initator):
...
-        RecursiveEquivalencyCache(
-            parent_identifier_id=target.id, identifier_id=target.id
-        )
+        RecursiveEquivalencyCache(parent_identifier=instance, identifier=instance)
     )
-    session.commit()
-    session.close()
+
+
+def before_flush_event_listener(
```
RishiDiwanTT, line 212: Does it make sense to create our own decorator `@before_flush(<model>, <action>) ` which does the type checking so we can modularize the implementation of this method.

RishiDiwanTT: In case the commit gets rolled back after the flush I assume the listeners changes will also be rolled back?

core/query/coverage.py
```diff
@@ -33,8 +33,19 @@ def add_coverage_for_identifiers_chain(
             _db, (p[0] for p in parent_ids)
         )
         records = []
-        for equiv in equivs:
 
+        # Make sure we haven't already added this record to the session, but not committed it yet
```
RishiDiwanTT, line 37: Why do we need this, if the method is not recursively calling itself?

jonathangreen: We need this, because this listener gets fired twice, once for Equivalency (input = 1, output = 2) and once for Equivalency (input = 2, output = 1).

This used to happen in the old listeners as well, but in the case of the old listeners, the add_for function wouldn't add a second record, because it had already been flushed, so it just returns the existing row.

In this case, its happening as part of the flush operation, so the new objects are not flushed on creation, but at the end all the listeners firing, so the new object can't be queried from the DB yet, so we happily create multiple copies, and fail later in the flush when we try to add them to the DB.

core/model/listeners.py
jonathangreen: It took wayyyy longer then I expected, because I ran into nightmarish circular imports, but I added our own decorator here.

core/model/listeners.py
```diff
@@ -18,6 +24,128 @@
...
+    def default_instance_filter(_: Base) -> bool:
+        return True
+
+    if instance_filter is None:
+        instance_filter = default_instance_filter
+
+    hooks = {
+        model: listeners
+        for (model, state), listeners in _before_flush_hooks.items()
+        if state == State.ANY or state == listening_for
```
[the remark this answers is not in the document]
RishiDiwanTT: Yea, I think both the loops on Line 98 and 105 are extra.

core/model/listeners.py
```diff
@@ -18,6 +24,128 @@
...
+    """
+
+    def decorator(func: Callable[P, None]) -> Callable[P, None]:
+        _before_flush_hooks[(model, state)].append(func)  # type: ignore[arg-type]
+        return func
+
+    return decorator
+
+
+def before_flush_trigger(
```
RishiDiwanTT, line 69: How is this different from `before_flush(<model>, ANY)` apart from the multiple models parameter?

RishiDiwanTT: Looks like a lot of special code for the same use case

core/query/coverage.py
RishiDiwanTT: Oooh thats an interesting problem to have

RishiDiwanTT (commented):
New comments are mostly to clarify my own confusions with the decorator :bowing_man:

core/model/listeners.py
jonathangreen: It only fires the listener once, if any of the models are changed. Rather then firing it repeatedly for each instance that has changed. This is used for `site_configuration_has_changed`. If we don't have this triggered style decorator, if a flush has updated 100 ConfigurationSettings, we will call `site_configuration_has_changed` for each of these.

jonathangreen: Since `site_configuration_has_changed` acquires a lock, and emits SQL directly, this ends up being a lot of unnecessary work. I don't see exactly how to support this without an additional decorator here. Honestly I think implementing decorators might be more work then its worth here, its taken almost as much time as it did to do the upgrade.

Would it be better to just drop the commit that adds decorators here, and go back to the previous implementation.

core/model/listeners.py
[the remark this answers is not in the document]
jonathangreen: This is only true if we don't want to support subclassed instances. A `type(instance)` check isn't the same as an `isinstance` check, if you want listeners to be able to fire based on subclassed instances.

core/model/listeners.py
```diff
@@ -18,6 +24,128 @@
...
+    session: Session, flush_context: UOWTransaction, instances: Optional[List[object]]
+) -> None:
+    triggers = {
+        tuple(models): Triggers(hook=hook)
+        for hook, models in _before_flush_trigger_hooks.items()
+    }
+
+    _fire_listeners(State.NEW, session, triggers)
+    _fire_listeners(State.DELETED, session, triggers)
+    _fire_listeners(State.DIRTY, session, triggers, before_flush_dirty_filter)
```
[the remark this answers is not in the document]
jonathangreen: I added a lot more documentation to the decorator function. Let me know if this is more clear now.

RishiDiwanTT (approved):
This looks great. SQL Level Up! :arrow_up:

core/model/listeners.py
RishiDiwanTT: I don't know why but this is much easier to read. Thanks for doing that :+1:

jonathangreen: Thanks for the review and feedback! It made this a lot better. 

I rushed the first implementation of decorators and looking back at it, it wasn't very readable. Think it's in a much better place now.

PR #1138, opened 2023-05-24
Refactor authentication providers

jonathangreen:
## Description

Refactored api.authenticator by moving out the base classes into their own files
since the file size had gotten to >2000 lines and kept growing

## Motivation and Context

I've been trying to code review https://github.com/ThePalaceProject/circulation/pull/1076 and its really hard to see what changed there because of the files that are moved around and mixed in. This does the same move, without any changes. 

That should make it easier to see what has changed in that diff.

## How Has This Been Tested?

Running tests in CI.

## Checklist


api/authentication/base.py
```diff
@@ -0,0 +1,532 @@
...
+            and self.external_type == other.external_type
+            and self.fines == other.fines
+            and self.block_reason == other.block_reason
+            and self.library_identifier == other.library_identifier
+            and self.complete == other.complete
+            and self.personal_name == other.personal_name
+            and self.email_address == other.email_address
+            and self.neighborhood == other.neighborhood
+            and self.cached_neighborhood == other.cached_neighborhood
+        )
```
io7m, line 303: Do we have a test that covers this `__eq__`? I couldn't see one.
That's the kind of code that's easy to break without realizing, and causes next-to-impossible-to-locate problems.

io7m (commented):
This looks OK, but there's one comment about something I really think should be tested if it isn't.

api/authentication/base.py
jonathangreen: It doesn't look like we have a test for this. I think it would be worth adding one.

I don't want to do that as part of this PR though. This is literally just a copy and paste to move classes around into other files, and update the import paths, no other changes are included. The code is exactly the same as it was before. 

I wanted to move the files around this way, so its easier to see what actually changed in #1074.

jonathangreen:
@io7m responded to your comment there. I'd like to keep this PR strictly moving things around, rather then introducing any logic or test changes.

io7m (approved):

PR #1144, opened 2023-05-25
Remove tables leftover after reverting #980

jonathangreen:
## Description

In https://github.com/ThePalaceProject/circulation/pull/1103 we reverted #980, but left the tables behind. This PR finishes the removal by removing the tables from the code and adding a migration to remove them as well.

## Motivation and Context

In the sprint planning meeting we discussed the issues with #980 and decided that storing this data in PG may not be the best place for it and perhaps we should implement this using something like reddis where we are able to respond more quickly to errors and connections that succeed. 

Since we are moving away from this implementation, we should cleanup the bits that were leftover after the revert.

## How Has This Been Tested?

Ran revert locally, ran alembic tests

## Checklist


jonathangreen:
Convert this to a draft since some tests are failing.

RishiDiwanTT:
The new IntegrationConfiguration table also has the status column, which needs to be removed.

jonathangreen:
Thanks @RishiDiwanTT, completely forgot about those columns. Updated the PR and migration to remove those as well.

RishiDiwanTT (approved):
Looks Good :+1:

PR #1155, opened 2023-05-30
Build container without nginx

jonathangreen:
## Description

Build our web containers running uwsgi directly, without nginx intermediating, since we are running behind AWS ALB anyway.

## Motivation and Context

Going to get an image building from this branch, so we can test the performance impact of this change. If it is minimal or none, this will help simplify our docker builds.

## How Has This Been Tested?

Build the containers and did some light testing with `docker-compose`.

## Checklist


io7m (approved):
Looks straightforward.

jonathangreen:
Going to close this one for now, since testing showed the image without nginx is slower. Need to do some more investigation there to understand why this is the case, and how we might be able to remove nginx in the future.

PR #1165, opened 2023-06-02
Mark OPDS For Distributors titles as explicitly unlimited access

io7m:
## Description

This switches the OPDS For Distributors collection type to using the new "unlimited access" model introduced with OPDS2.

## Motivation and Context

The OPDS For Distributors collection type was emulating "unlimited access" by setting the number of available licenses to `1` and never decrementing this count in the database. The OPDS2 collection type introduced the notion of explicitly unlimited access, but this OPDS4D collection type was never updated to use it. This means that dashboards and the like always show `1` license available instead of "unlimited".

Affects: https://www.notion.so/lyrasis/Investigate-why-BiblioBoard-licenses-look-like-metered-licenses-but-behave-like-unlimited-use-licen-72ad2d99682e4dc0aa8ad04f522eaec0

## How Has This Been Tested?

I ran the test suite, and I tried borrowing books from my local CM's OPDS feed.

## Checklist


api/opds_for_distributors.py
```diff
@@ -306,12 +306,7 @@ def update_work_for_edition(self, *args, **kwargs):
         pool, work = super().update_work_for_edition(
             *args, is_open_access=False, **kwargs
         )
-        pool.update_availability(
-            new_licenses_owned=1,
-            new_licenses_available=1,
-            new_licenses_reserved=0,
-            new_patrons_in_hold_queue=0,
```
tdilauro, line 313: `pool.unlimited_access = True` sets `licenses_owned` and `licenses_available`, but doesn't reset `licenses_reserved` or `patrons_in_hold_queue`. Because all OPDS for Distributor books are either open access or unlimited access, it seems like this shouldn't be a problem, since these values should never be changed from their default of zero; but, just wanted to ensure that this has been considered.

api/opds_for_distributors.py
```diff
@@ -306,12 +306,7 @@ def update_work_for_edition(self, *args, **kwargs):
...
-        )
+        pool.unlimited_access = True
```
tdilauro, line 309: The docstring for this method is misleading. OPDS for Distributors books can be either Open Access or unlimited access and should be assigned appropriately, depending on the acquisition link in the feed:
- `rel="http://opds-spec.org/acquisition"` for unlimited access and
- `rel="http://opds-spec.org/acquisition/open-access"` for OA

The current `OPDSForDistributorsImporter` appears to respect this, since I can see at least one production collection that contains both. I'm not entirely sure where we handle the OA part, though.

api/opds_for_distributors.py
```diff
@@ -398,7 +393,7 @@ def run_once(self, progress):
             .join(Identifier)
             .filter(LicensePool.collection_id == self.collection.id)
             .filter(~Identifier.id.in_(identifier_ids))
-            .filter(LicensePool.licenses_available > 0)
+            .filter(LicensePool.licenses_available == LicensePool.UNLIMITED_ACCESS)
```
tdilauro, line 396: This won't work for existing content without a database migration.

tdilauro (commented):
This looks pretty good! I think we need a database migration to update the `licenses_owned` and `licenses_available` for each of the already present and active OPDS for Distributors license pools from `1` to `-1`. A few additional comments below.

api/opds_for_distributors.py
io7m: I'm not entirely sure what to do about this. I haven't changed any behaviour with regards to whether or not something is assigned open access status, so it should continue to do whatever it did before with respect to the link relations.

api/opds_for_distributors.py
io7m: Yeah, I left it alone. I think those arguments were only there because _something_ had to be passed in.

api/opds_for_distributors.py
io7m: Working on this now. :+1:

io7m:
Waiting for tests to complete, but I think this is ready.

Nice to have a database migration that's reversible for once. :slightly_smiling_face:

io7m:
There's a single test failing that looks flaky. I don't think it's anything to do with my changes (I merged from `main` about 20 minutes ago).

io7m:
Checking these test failures now. They appear to be new... :thinking:

jonathangreen:
@io7m I think the issue here is that the down revision is now not the most recent revision. The easiest fix is probably to just update the down revision and the comment, so that it points to the latest migration in `main`.

jonathangreen (approved):
Looks good to me

PR #1178, opened 2023-06-07
Faster docker builds

jonathangreen:
## Description

Our docker builds are slow, because we are compiling wheels for packages in the emulated arm64 environment we use to build arm images.

This PR replaces 3 of the 4 packages we were compiling: 
- `uwsgi`
  - Obviously we need `uwsgi`, but there is a second package `pyuwsgi` built from the same source, that ships with wheels we can install directly. We are using this package already in VLC.
- `py-bcrypt`
  - The last release of this package was in 2013 😓, so its not about to start shipping ARM wheels.
  - It looks like this package has been replaced by `bcrypt`, which does have wheels, and is still being developed. 
  - This change took some minor code changes, as the `bcrypt` package operates on `bytes` rather then strings, so we have to do some decoding / encoding to make it work.
  - There is a small change to a test, since the new library creates bcrypt hashes with a newer [version string](https://en.wikipedia.org/wiki/Bcrypt#Versioning_history). Since this new version has existed since 2014, we should probably have updated this library long ago.
- `psycopg2`
  - They have a second package `psycopg2-binary` that ships with built wheels. We were using this in dev, but not production, so we shift over entirely to use this.

The only one that is left is `xmlsec`, but that will take more work since there are no wheels available there. 

## Motivation and Context

While debugging https://github.com/ThePalaceProject/circulation/pull/1155 yesterday, I got frustrated with our super slow docker build times. So I tried to improve them while waiting for my changes to build.

## How Has This Been Tested?

Running tests in CI. Did a little testing with the UI to make sure passwords were fine.

## Checklist


pcustic:
[Docs](https://pypi.org/project/psycopg2-binary/) (at to bottom) for `psycopg2-binary` says: `The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.`

jonathangreen:
Thats a good point @pcustic. Maybe need another approach here. I'd like to get the bcrypt changes merged, because they seem valuable, even if they are not speeding up the docker build. I'll close this and make another PR with just those changes.

PR #1179, opened 2023-06-07
Speed up API tests

jonathangreen:
## Description

This is the same change that was made in this PR: 
https://github.com/ThePalaceProject/circulation/pull/871

From that PR: 
> Doing a bit more profiling of our code, it seems that the single most expensive function we call when running the tests is now the call to set password in AdminControllerTest. This removes the need to hash this password on each test setup, by directly setting the the hashed password.

> This doesn't speed up the tests dramatically, but does shave about a minute off the test run on my machine.

## Motivation and Context

I noticed when working on https://github.com/ThePalaceProject/circulation/pull/1178 that this change was lost when we moved to pytest fixtures. Its not a huge speed increase, but I think its worth adding back in. Its nice to save a bit of time since our tests take so long anyway.

## How Has This Been Tested?

Running tests in CI.

## Checklist

### DIFF
--- a/tests/fixtures/api_admin.py
+++ b/tests/fixtures/api_admin.py
@@ -39,7 +39,11 @@ class AdminControllerFixture:
             Admin,
             email="example@nypl.org",
         )
-        self.admin.password = "password"
+        # This is a hash for 'password', we use the hash directly to avoid the cost
+        # of doing the password hashing during test setup.
+        self.admin.password_hashed = (
+            "$2a$12$Dw74btoAgh49.vtOB56xPuumtcOY9HCZKS3RYImR42lR5IiT7PIOW"
+        )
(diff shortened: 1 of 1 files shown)


dbernstein (approved):
Does this hashing slow things down appreciably?

dbernstein (approved):
Does this hashing slow things down appreciably?

jonathangreen:
@dbernstein on my machine it reduces the test run time by 1 minute or two. Since the tests take 10 minutes to run anyway, its not that much, but still nice to save that minute I think. 

Its hard to tell if its making a difference in CI since the time it takes to run the tests seems to vary by +/- 3 minutes anyway. So any improvement is tough to see among the noise.

PR #1180, opened 2023-06-08
Move from py-bcrypt to bcrypt library

jonathangreen:
## Description

This replaces the `py-bcrypt` package with the `bcrypt` package.

- The last release of `py-bcrypt` was in 2013 😓
- It looks like this package has been replaced by `bcrypt`, which does have wheels, and is still being developed.
- This change took some minor code changes, as the `bcrypt` package operates on `bytes` rather then `str`, so we have to do some decoding / encoding to make it work.
- There is a small change to a test, since the new library creates bcrypt hashes with a newer [version string](https://en.wikipedia.org/wiki/Bcrypt#Versioning_history). Since this new version has existed since 2014, we should probably have updated this library long ago.

## Motivation and Context

This change was originally part of https://github.com/ThePalaceProject/circulation/pull/1178. @pcustic made a good point that that PR needs to be re-worked, so I wanted to pull this piece of the PR out, because I think its a valuable change.

## How Has This Been Tested?

Running CI on change.

## Checklist

### DIFF
--- a/core/model/admin.py
+++ b/core/model/admin.py
@@ -62,11 +62,16 @@ class Admin(Base, HasSessionCache):
         raise NotImplementedError("Password comparison is only with Admin.authenticate")
 
     @password.setter
-    def password(self, value):
-        self.password_hashed = str(bcrypt.hashpw(value, bcrypt.gensalt()))
+    def password(self, value: str) -> None:
+        self.password_hashed = bcrypt.hashpw(value.encode(), bcrypt.gensalt()).decode()
 
-    def has_password(self, password):
-        return self.password_hashed == bcrypt.hashpw(password, self.password_hashed)
+    def has_password(self, password: str) -> bool:
+        if self.password_hashed is None:
+            return False
+        return (
+            self.password_hashed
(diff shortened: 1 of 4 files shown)


jonathangreen:
@pcustic tagging you for review, since you took at look at https://github.com/ThePalaceProject/circulation/pull/1178 already.

core/model/admin.py
```diff
@@ -62,11 +62,16 @@ def password(self):
...
-        return self.password_hashed == bcrypt.hashpw(password, self.password_hashed)
+    def password(self, value: str) -> None:
+        self.password_hashed = bcrypt.hashpw(value.encode(), bcrypt.gensalt()).decode()
+
+    def has_password(self, password: str) -> bool:
+        if self.password_hashed is None:
+            return False
+        return (
+            self.password_hashed
+            == bcrypt.hashpw(password.encode(), self.password_hashed.encode()).decode()
```
pcustic, line 73: Not that important but I think this line could be a bit simplified:
`return bcrypt.checkpw(password.encode(), self.password_hashed.encode())`

pcustic (approved):
Looks good, only a suggestion for a minor simplification.

PR #1188, opened 2023-06-08
Actually faster docker builds - Part 2

jonathangreen:
## Description

This PR updates our main docker build to use the new base image added in #1187. Since most of the logic has been moved into the base image build, it does a bit of refactoring to the main `Dockerfile`, so that all the logic is contained there, instead of being spread over a number of different scripts.

This also updates the build process for the base image slightly, so its easier to make and test changes to the base image in a PR. If the baseimage has been updated in a PR then `test-build.yml` will build the baseimage, and then use that base image for the container builds. If not, then we just use the prebuilt baseimage. 

This should get most of our docker builds to complete in a couple minutes, rather then the ~45 minute builds we currently have.

## Motivation and Context

I was frustrated with our slow docker build times, while trying to build images for nginx.

## How Has This Been Tested?

I've built the images and done some testing, spinning them up with `docker-compose` and doing some basic testing to make sure they are functional. I've also tested the CI part of this in my fork of the circulation repo where I can test pushing to `main` with reckless abandon.

## Checklist


jonathangreen:
@io7m pinging you on this one, since you took a look at https://github.com/ThePalaceProject/circulation/pull/1187.

docker/services/nginx/nginx.conf
RishiDiwanTT, line 1: This is an entirely new configuration, why has it been introduced?

RishiDiwanTT (commented):
Just have the one question about nginx.

docker/services/nginx/nginx.conf
jonathangreen: It's not really related to the speed improvements, but to the refactor of the docker file. 

The script we previously used to setup nginx used to echo a couple lines into the nginx config when we were setting it up the copied over the site config. 

I wanted to be able to just use a copy command to setup the whole nginx configuration, so I took the approach of copying over the configuration rather then echoing lines into it.

RishiDiwanTT (approved):

PR #1189, opened 2023-06-09
PP-35 Migrate license goal integrations :mountain: 

RishiDiwanTT:
## Description
These changes are a refactor of the external integrations system.
The refactor has undergone several refactors itself, every time a unit test would point out a different use case, some changes would have to be made. Due to this there may be some dead code lying around, I will try to clean it up.

Defined pydantic settings for all license goal type APIs.
Migration scripts have been added for the Goal enum and to move the data into IntegrationConfigurations.
The External Integration data has not been deleted, yet.
The Admin CollectionSettings controller defines new methods for integration configurations and uses them appropriately.
All API classes have been modified to utilise IntegrationConfigurations.
All API classes use the modified HasCollectionSelfTests class.
Migrated all collection tests to use IntegrationConfigurations.

:warning: If you create a collection while in this branch, then downgrade the DB migration to move to another branch. The upgrade will not work again until the new collection has been deleted from the database. This is because the new collection would not have the ConfigurationSettings the migration relies on.

[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-35)

## How Has This Been Tested?
Manually imported an OPDS feed.
Randomly ran collection self tests.
Modified the unit tests to work with integration configurations.
Modified existing collections and created new collections from the Admin UI.

## Checklist


api/admin/controller/collection_settings.py
RishiDiwanTT, line 1: This file contains all the Admin UI functionality for the Collections page

core/migration/migrate_external_integration.py
RishiDiwanTT, line 1: Although code coverage complains about this file, the --test-alembic option does test all except 4 lines of the code.

core/model/__init__.py
```diff
@@ -45,15 +56,22 @@ def flush(db):
...
     created = getattr(model, create_method, model)(**kwargs)
     db.add(created)
     flush(db)
     return created, True
 
 
-def get_one(db, model, on_multiple="error", constraint=None, **kwargs):
+def get_one(
+    db: Session, model: Type[T], on_multiple="error", constraint=None, **kwargs
+) -> T | None:
```
RishiDiwanTT, line 74: I added generic typing to these foundational methods because I got tired of added type hints in all the methods that use them, and guess what, I had to add defensive code ALLLL over the place so mypy stays happy now.
This added defensive code is why test coverage has also dropped in many files.
I'm too stupid for my own good sometimes.

api/admin/controller/__init__.py
```diff
@@ -1112,6 +1125,8 @@ def _create_or_update_list(
             list, is_new = create(
                 self._db, CustomList, name=name, data_source=data_source
             )
+            if list is None:
+                return INTERNAL_SERVER_ERROR.detailed("Could not create the list.")
```
RishiDiwanTT, line 1129: I've used internal server error in several cases where the "create" would have failed because that is essentially what it is, the server failed to write to the DB for some reason.

RishiDiwanTT: I some cases to avoid changing the function definition I've simply raised ValueErrors, which would translate to an INTERNAL_SERVER_ERROR higher up in the stack.

api/admin/controller/__init__.py
```diff
@@ -1727,21 +1742,6 @@ class SettingsController(AdminCirculationManagerController):
...
-        OdiloAPI,
-        BibliothecaAPI,
-        Axis360API,
-        EnkiAPI,
-        ODLAPI,
-        ODL2API,
-        SharedODLAPI,
-        LCPAPI,
-    ]
-
```
RishiDiwanTT, line 1744: These all belong in the LicenseProvidersRegistry now

api/admin/controller/__init__.py
```diff
@@ -1796,13 +1796,31 @@ def _set_storage_external_integration_link(
...
         return None
 
-    @classmethod
-    def _get_integration_protocols(cls, provider_apis, protocol_name_attr="__module__"):
+    def _get_protocol_class(self, registry, protocol_name, is_child=False):
+        api_class = registry.get(protocol_name)
+        if not api_class:
+            return None
+        if is_child:
+            if getattr(api_class, "child_settings_class", False):
```
RishiDiwanTT, line 1813: Rather that cls.CHILD_SETTINGS we now use a `def child_settings` implementation in the Overdrive API.

api/admin/controller/__init__.py
```diff
@@ -1825,15 +1843,22 @@ def _get_integration_protocols(cls, provider_apis, protocol_name_attr="__module_
 
             settings = getattr(api, "SETTINGS", [])
             protocol["settings"] = list(settings)
+            if _db and issubclass(api, HasIntegrationConfiguration):
+                protocol["settings"] = api.settings_class().configuration_form(_db)
```
RishiDiwanTT, line 1847: Both cls.SETTINGS and `def settings_class` must coexist while some integrations still use external integrations.

api/admin/controller/collection_self_tests.py
```diff
@@ -58,7 +55,7 @@ def run_tests(self, collection):
                 value, results = self.protocol_class.run_self_tests(
                     self._db, self.protocol_class, self._db, collection, OPDSImporter
                 )
-            elif issubclass(self.protocol_class, HasSelfTests):
+            elif issubclass(self.protocol_class, HasCollectionSelfTests):
```
RishiDiwanTT, line 58: All License integrations now use the updated `HasCollectionSelfTests` because the `HasSelfTests` class is used by other external integration implementations as well.

api/integration/registry/license_providers.py
```diff
@@ -0,0 +1,43 @@
...
+            ExternalIntegration.BIBLIOTHECA: BibliothecaAPI,
+            ExternalIntegration.AXIS_360: Axis360API,
+            EnkiAPI.ENKI_EXTERNAL: EnkiAPI,
+            OPDSForDistributorsAPI.NAME: OPDSForDistributorsAPI,
+            ODLAPI.NAME: ODLAPI,
+            ODL2API.NAME: ODL2API,
+            SharedODLAPI.NAME: SharedODLAPI,
+            LCPAPI.NAME: LCPAPI,
+            OPDSImporter.NAME: OPDSImporter,
+            OPDS2Importer.NAME: OPDS2Importer,
```
RishiDiwanTT, line 39: The bottom 2 aren't technically BaseCirculationAPIs but behave close enough to them. Hence the `cast` below.

api/lcp/factory.py
```diff
@@ -16,13 +15,10 @@ def create(self, integration_association):
...
         :rtype: LCPServer
         """
-        configuration_storage = ConfigurationStorage(integration_association)
-        configuration_factory = ConfigurationFactory()
         hasher_factory = HasherFactory()
         credential_factory = LCPCredentialFactory()
         lcp_server = LCPServer(
-            configuration_storage,
-            configuration_factory,
+            integration_association.configuration,
```
RishiDiwanTT, line 21: This is a Callable.

api/lcp/server.py
```diff
@@ -124,55 +142,39 @@ class LCPServer:
 
     def __init__(
         self,
-        configuration_storage,
-        configuration_factory,
+        get_configuration,
```
RishiDiwanTT, line 145: Rather than a configuration factory and storage we use a get_configuration Callable to ensure the configuration values we pull at runtime are always pulled from the database and not an in-memory object.

api/odl2.py
```diff
@@ -1,80 +1,78 @@
...
-    skipped_license_formats = ConfigurationMetadata(
-        key="odl2_skipped_license_formats",
-        label=_("Skipped license formats"),
-        description=_(
-            "List of license formats that will NOT be imported into Circulation Manager."
-        ),
-        type=ConfigurationAttributeType.LIST,
-        required=False,
+class ODL2Settings(ODLSettings):
+    odl2_skipped_license_formats: Optional[List[str]] = FormField(
```
RishiDiwanTT, line 29: All the odl2 settings need to maintain the configuation key names, hence the `odl2` prefix

api/selftest.py
```diff
@@ -114,70 +115,25 @@ def _determine_self_test_patron(
         raise cls._NoValidLibrarySelfTestPatron(message, detail=detail)
 
 
-class RunSelfTestsScript(LibraryInputScript):
```
RishiDiwanTT, line 117: Moved to core/scripts.py due to import issues

core/integration/settings.py
```diff
@@ -139,6 +149,11 @@ class ConfigurationFormItemType(Enum):
     SELECT = "select"
     LIST = "list"
     MENU = "menu"
+    NUMBER = "number"
+
+    @classmethod
+    def options_from_enum(cls, enum_: Type[Enum]) -> Dict[Enum | str, str]:
```
RishiDiwanTT, line 155: The return type is actually Dict[str, str] but due to type invariance of the `ConfigurationFormItem.options` attribute we must match the typing.

core/model/collection.py
```diff
@@ -243,12 +260,15 @@ def by_datasource(cls, _db, data_source):
...
-                ExternalIntegration,
-                cls.external_integration_id == ExternalIntegration.id,
+                IntegrationConfiguration,
+                cls.integration_configuration_id == IntegrationConfiguration.id,
+            )
+            .filter(
+                IntegrationConfiguration.settings[
+                    Collection.DATA_SOURCE_NAME_SETTING
+                ].astext
+                == data_source
```
RishiDiwanTT, line 270: JSONB filter

core/model/collection.py
```diff
@@ -300,26 +324,30 @@ def default_loan_period(self, library, medium=EditionConstants.BOOK_MEDIUM):
...
         """
         return (
-            self.default_loan_period_setting(library, medium).int_value
+            self.default_loan_period_setting(library, medium)
             or self.STANDARD_DEFAULT_LOAN_PERIOD
         )
 
-    def default_loan_period_setting(self, library, medium=EditionConstants.BOOK_MEDIUM):
+    def default_loan_period_setting(
+        self, library, medium=EditionConstants.BOOK_MEDIUM, set_value=None
```
RishiDiwanTT, line 332: Since the earlier implementation was returning the configurationsettings object which allowed setting of the value, the additional `set_value` argument allows us to do the same.

core/model/collection.py
```diff
@@ -981,55 +1049,3 @@ def collection_external_integration(
         :return: External integration associated with the collection
         """
         raise NotImplementedError()
-
-
-class CollectionConfigurationStorage(BaseConfigurationStorage):
```
RishiDiwanTT, line 986: Collections now use IntegrationConfigurations, don't need this anymore.

alembic/versions/20230531_0af587ff8595_migrate_license_integrations_to_.py
```diff
@@ -0,0 +1,106 @@
...
+            raise RuntimeError(f"Could not find API class for '{protocol}'")
+
+        # Create the settings and library settings dicts from the configurationsettings
+        settings_dict, library_settings, self_test_result = get_configuration_settings(
+            connection, integration
+        )
+
+        if "IGNORED_IDENTIFIER_TYPE" in settings_dict.keys():
+            value = settings_dict.pop("IGNORED_IDENTIFIER_TYPE")
+            settings_dict["ignored_identifier_types"] = value
```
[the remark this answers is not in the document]
RishiDiwanTT: I'll rename the others but I'd like to change this to the lowercase scheme so the data is more consistent in the DB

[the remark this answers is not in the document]
jonathangreen: If you set this as an alias, the data does get stored in the DB as the name of the key. The alias is only used when the data is loaded into the Pydantic model. If the key is missing, Pydantic checks the alias and loads it as the key.

api/admin/controller/collection_settings.py
```diff
@@ -69,33 +75,30 @@ def process_get(self):
...
-            )
-            collection_dict[
-                "marked_for_deletion"
-            ] = collection_object.marked_for_deletion
+                collection_dict["self_test_results"] = self._get_prior_test_results(
+                    collection_object
+                )
+                collection_dict[
+                    "marked_for_deletion"
+                ] = collection_object.marked_for_deletion
```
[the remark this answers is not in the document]
RishiDiwanTT: Hmm seems like an accidental indent.

api/axis.py
```diff
@@ -63,61 +71,72 @@
...
-            "description": _(
+
+class Axis360Settings(BaseSettings):
+    username: str = FormField(
+        form=ConfigurationFormItem(label=_("Username"), required=True)
+    )
+    password: str = FormField(
+        form=ConfigurationFormItem(label=_("Password"), required=True)
+    )
+    external_account_id: Optional[str] = FormField(
```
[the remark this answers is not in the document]
RishiDiwanTT: We actually never use the configuration.external_account_id, we only ever use the collection.external_account_id, so this needs to be `required` in the ConfigurationFormItem but `Optional` in the pydantic model

api/integration/registry/license_providers.py
RishiDiwanTT: Implemented a Protocol class. I think mypy was introduced by the BDFL to prove no dictator is truly benevolent.

core/model/integration.py
```diff
@@ -140,6 +194,19 @@ def for_library_and_goal(
             )
         )
 
+    # Settings model inheritance
+    # If the library configuration has no such value
+    # but the parent does, retrieve the parent value
+    def get(self, key: str, *args: Any) -> Any:
```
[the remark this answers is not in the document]
RishiDiwanTT: Ok I'll remove it, I wasn't sure if it was in use anywhere

core/model/integration.py
```diff
@@ -20,16 +20,37 @@
     from typing_extensions import Self
 
 if TYPE_CHECKING:
-    from core.model import Library
+    from core.model import Collection, Library
 
 
-class IntegrationConfiguration(Base):
+class SettingsModel:
```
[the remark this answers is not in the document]
RishiDiwanTT: As I run through the code I see this will not be entirely possible since the integration settings are sometimes accessed directly from the collection object without context of a settings class.
One example is [here](https://github.com/ThePalaceProject/circulation/pull/1189/files#diff-c2b30eb26660a26a6fb7181ed75404502f9c840b3798aedc6dae772c8de17464R361).

core/configuration/ignored_identifier.py
```diff
@@ -1,57 +1,67 @@
...
 from core.model.constants import IdentifierType
+from core.model.integration import IntegrationConfiguration
+
+ALL_IGNORED_IDENTIFIER_TYPES = {
+    identifier_type.value for identifier_type in IdentifierType
+}
+
+
+class IgnoredIdentifierSettings(BaseSettings):
+    ignored_identifier_types: Optional[list] = FormField(
```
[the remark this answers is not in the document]
RishiDiwanTT: I'd like to keep this change since lowercase is the norm and this is the only uppercase key.
The change will keep the keys somewhat standardized.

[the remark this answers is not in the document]
RishiDiwanTT: Changed it to `Optional[List[IdentifierType]]`

core/integration/settings.py
```diff
@@ -139,6 +149,11 @@ class ConfigurationFormItemType(Enum):
     SELECT = "select"
     LIST = "list"
     MENU = "menu"
+    NUMBER = "number"
```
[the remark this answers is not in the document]
RishiDiwanTT: I'll change the usage to `PositiveInt` types but will keep the `ConfigurationFormItemType.NUMBER` as the `type` since this changes the Admin UIs behaviour to only allow numbers in the text field it generates.

core/integration/settings.py
```diff
@@ -182,6 +197,9 @@ class ConfigurationFormItem:
     # weight will be displayed in the order they were added.
     weight: int = 0
 
+    # Allowed values for any field
+    allowed: List[Any] | None = None
```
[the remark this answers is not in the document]
RishiDiwanTT: The weird nickname as urls confused me for a good 10 minutes here, added a validator for it.

core/migration/migrate_external_integration.py
```diff
@@ -0,0 +1,137 @@
...
+    BaseSettings,
+    ConfigurationFormItemType,
+    FormFieldInfo,
+)
+from core.model import json_serializer
+
+T = TypeVar("T", bound=BaseSettings)
+
+
+def _validate_and_load_settings(
```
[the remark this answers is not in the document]
RishiDiwanTT: It's not an exact copy, I've broken up the code to be more modular so it can be used for the other kinds of integration migrations. 
The Auth migration did some special things within these methods specially for Auth which I've removed.

alembic/versions/20230531_0af587ff8595_migrate_license_integrations_to_.py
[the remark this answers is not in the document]
RishiDiwanTT: Currently in the ConfigurationSettings the key name is uppercase, so in the data-driven approach it is a one to one migration for key naming. In order to move the name to lowercase in the DB I have to have this code in place.

core/model/integration.py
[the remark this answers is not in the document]
RishiDiwanTT: I've dropped the `setitem` interface and kept the `getitem` interface. Which means there are some places where some shoddy dict.copy() occurs.

api/circulation.py
```diff
@@ -414,56 +425,147 @@ def __repr__(self):
...
-    # distributor which includes ebooks and allows clients to specify
-    # their own loan lengths.
-    EBOOK_LOAN_DURATION_SETTING = {
-        "key": Collection.EBOOK_LOAN_DURATION_KEY,
-        "label": _("Ebook Loan Duration (in Days)"),
-        "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
-        "type": "number",
-        "description": _(
-            "When a patron uses SimplyE to borrow an ebook from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor."
+    ebook_loan_duration: Optional[PositiveInt] = FormField(
```
jonathangreen, line 431: It looks like this is used in two different files: 
- axis.py 
- bibliotecha.py

But the type we use in each of those cases is different then what is specified on the base class. Instead of implementing it like this, maybe we could implement this as a couple real base classes, and use inheritance. That way there won't be the possibility of having different types on the subclasses, and we can use type checks to determine if a class has this setting or not.

api/circulation.py
```diff
@@ -414,56 +425,147 @@ def __repr__(self):
...
     # distributor which includes audiobooks and allows clients to
     # specify their own loan lengths.
-    AUDIOBOOK_LOAN_DURATION_SETTING = {
-        "key": Collection.AUDIOBOOK_LOAN_DURATION_KEY,
-        "label": _("Audiobook Loan Duration (in Days)"),
-        "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
-        "type": "number",
-        "description": _(
-            "When a patron uses SimplyE to borrow an audiobook from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor."
+    audiobook_loan_duration: Optional[PositiveInt] = FormField(
```
jonathangreen, line 445: This appears unused, maybe it can just be dropped?

api/circulation.py
```diff
@@ -414,56 +425,147 @@ def __repr__(self):
...
     # such that the circulation manager cannot _specify_ the length of
     # a loan.
-    DEFAULT_LOAN_DURATION_SETTING = {
-        "key": Collection.EBOOK_LOAN_DURATION_KEY,
-        "label": _("Default Loan Period (in Days)"),
-        "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
-        "type": "number",
-        "description": _(
-            "Until it hears otherwise from the distributor, this server will assume that any given loan for this library from this collection will last this number of days. This number is usually a negotiated value between the library and the distributor. This only affects estimates&mdash;it cannot affect the actual length of loans."
+    default_loan_duration: Optional[PositiveInt] = FormField(
```
jonathangreen, line 460: Same as above, the uses of this are defining a different type then specified here. Could we use inheritance for this instead?

api/enki.py
```diff
@@ -37,41 +41,52 @@
...
-    LIBRARY_SETTINGS = BaseCirculationAPI.LIBRARY_SETTINGS + [
-        {"key": ENKI_LIBRARY_ID_KEY, "label": _("Library ID"), "required": True},
-        {
-            "key": ExternalIntegration.DISPLAY_RESERVES,
-            "label": _("Show/Hide Titles with No Available Loans"),
-            "required": False,
-            "description": _(
+
+class EnkiSettings(BaseSettings):
+    url: Optional[str] = FormField(
```
jonathangreen, line 52: Does this need to be optional? And can we do validation with the `HttpUrl` type rather then adding `format="url"` below?

api/integration/registry/license_providers.py
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        for name, api in apis.items():
+            self.register(cast(Type[BaseCirculationAPIProtocol], api), canonical=name)
```
jonathangreen, line 43: Since we have a Protocol, this cast should no longer be needed.

api/lcp/encrypt.py
```diff
@@ -4,49 +4,56 @@
...
 
-    lcpencrypt_location = ConfigurationMetadata(
-        key="lcpencrypt_location",
-        label=_("lcpencrypt's location"),
-        description=_(
-            "Full path to the local lcpencrypt binary. "
-            "The default value is {}".format(DEFAULT_LCPENCRYPT_LOCATION)
+
+class LCPEncryptionSettings(BaseSettings):
+    lcpencrypt_location: Optional[str] = FormField(
```
jonathangreen, line 32: Should this be optional? It has a default, so we probably don't want it to be optional, and instead want to error if there is no data in the field.

alembic/versions/20230531_0af587ff8595_migrate_license_integrations_to_.py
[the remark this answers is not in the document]
jonathangreen: I don't understand what you are saying here @RishiDiwanTT. I agree this should be lowercase, but I don't understand why this can't be handled using an alias, like the rest of the settings that need to have their names changed as part of this. Pydantic is case sensitive, so it doesn't seem that we need a special path for this name change.

core/configuration/ignored_identifier.py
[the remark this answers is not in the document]
jonathangreen: I agree that this key should be lowercase, but since Pydantic is case sensitive, I don't understand why this can't just use an alias to do the case conversion. 

Is there some reason that this wouldn't work?

```python
    ignored_identifier_types: Optional[List[IdentifierType]] = FormField(
        default=[],
        alias="IGNORED_IDENTIFIER_TYPES",
        form=ConfigurationFormItem(
            label=_("List of identifiers that will be skipped"),
            description=_(
                "Circulation Manager will not be importing publications with identifiers having one of the selected types."
            ),
            type=ConfigurationFormItemType.MENU,
            required=False,
            options={
                identifier_type: identifier_type
                for identifier_type in ALL_IGNORED_IDENTIFIER_TYPES
            },
            format="narrow",
        ),
    )
```

core/model/integration.py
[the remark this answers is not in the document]
jonathangreen: I'd really like to be able to remove the `getitem` interface here as well, but I can make a follow up ticket to handle that if we don't want to remove it as part of this ticket.

alembic/versions/20230531_0af587ff8595_migrate_license_integrations_to_.py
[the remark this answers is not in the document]
RishiDiwanTT: Ok, I was running with some misconceptions about alias, I've cleared those out now :+1:

api/axis.py
```diff
@@ -63,61 +71,72 @@
...
+        form=ConfigurationFormItem(
+            label=_("Library ID"),
+            required=True,
+        )
+    )
+    url: str = FormField(
```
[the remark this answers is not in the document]
RishiDiwanTT: I've changed these everywhere now, no more format="url"

api/circulation.py
RishiDiwanTT: That makes sense :+1:

core/model/integration.py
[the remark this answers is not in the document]
RishiDiwanTT: This will not remove the need to access the `.settings` attribute in many places throughout the application code because the pydantic models are only known to the API classes.

api/odl.py
```diff
@@ -1400,12 +1448,10 @@ def _get(
...
         _db = Session.object_session(patron)
         collection = self.collection(_db)
-        shared_secret = ConfigurationSetting.for_library_and_externalintegration(
-            _db,
-            ExternalIntegration.PASSWORD,
-            patron.library,
-            collection.external_integration,
-        ).value
+        config = collection.integration_configuration
+        shared_secret = config.for_library(patron.library.id).get(
```
RishiDiwanTT, line 1450: While removing the getitem interface I noticed this value should never exist since the SharedODLAPI has no library settings. Is this an instance of settings inheritance?

jonathangreen:
I think the code here is looking good @RishiDiwanTT. Before merging, I wanted to make sure that we did a smoke test of the migration with an existing database, just to make sure it looked okay. 

I tried with the database from minotaur, and I'm seeing an error: 
```
INFO  [alembic.runtime.migration] Running upgrade a9ed3f76d649 -> b883671b7bc5, Add the license type goal
INFO  [alembic.runtime.migration] Running upgrade b883671b7bc5 -> 0af587ff8595, Migrate license integrations to configuration settings
  File "/Users/jgreen/.pyenv/versions/3.9.16/envs/circulation/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1905, in _execute_context
    self.dialect.do_execute(
  File "/Users/jgreen/.pyenv/versions/3.9.16/envs/circulation/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
psycopg2.errors.UndefinedColumn: column "status" of relation "integration_configurations" does not exist
```

I think that may be an issue that came in during the rebase. Can you take a look? If you want a copy of the database, ping me on slack and I can send it to you.

jonathangreen (approved):
Looks good Rishi!

alembic/versions/20230525_0a1c9c3f5dd2_revert_pr_980.py
```diff
@@ -90,6 +90,7 @@ def downgrade() -> None:
             postgresql.ENUM("RED", "GREEN", name="status"),
             autoincrement=False,
             nullable=False,
+            server_default=sa.text("'GREEN'::status"),
```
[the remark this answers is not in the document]
RishiDiwanTT: This was required because NULL is not allowed in the column, so rollbacks were failing.

PR #1196, opened 2023-06-14
Allow FCM credential via JSON string in env var.

tdilauro:
## Description

- Adds ability to configure FCM credentials via a JSON string as an alternative to configuring with a file.

## Motivation and Context

Support configuration via AWS parameter store secret for ECS instances as part of [PP-41](https://ebce-lyrasis.atlassian.net/browse/PP-41).

## How Has This Been Tested?

- Manually verified that dictionary could be used to configure `firebase_admin` credentials.
- Updated tests for new test cases.
- CI passes for the PR's branch.

## Checklist

[PP-41]: https://ebce-lyrasis.atlassian.net/browse/PP-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os
 from typing import Dict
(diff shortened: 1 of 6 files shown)


core/config.py
```diff
@@ -359,15 +361,47 @@ def database_url(cls):
...
-        fcm_file = os.environ.get(cls.FCM_CREDENTIALS_FILE_ENVIRONMENT_VARIABLE)
-        if not fcm_file:
+    def fcm_credentials(cls) -> Dict[str, str]:
+        """Returns a dictionary containing Firebase Cloud Messaging credentials.
+
+        Credentials are provided as a JSON string, either (1) directly in an environment
+        variable or (2) in a file that is specified in another environment variable.
+        """
+        config_json = os.environ.get(cls.FCM_CREDENTIALS_JSON_ENVIRONMENT_VARIABLE, "")
+        config_file = os.environ.get(cls.FCM_CREDENTIALS_FILE_ENVIRONMENT_VARIABLE, "")
```
tdilauro, line 371: Using default empty string here since it is both falsy and implicitly types these variables as `str`, rather than `Optional[str]` (`str | None`), which matters to `os.path.exists` and `open`, later in this method. Okay, really it's just that `mypy` has some objections on their behalf.

tdilauro:
@RishiDiwanTT requesting your review here, since you've done the bulk of the notification work.

RishiDiwanTT (commented):
Only have the one comment about nginx

RishiDiwanTT (approved):

PR #1198, opened 2023-06-14
Fix flaky test - TestIndividualAdmins.test_individual_admins_get

jonathangreen:
## Description

Seems like its just a flaky test, since its depending on the ordering of `roles`, and the order there isn't guaranteed.

## Motivation and Context

Saw this test fail in CI here: 
https://github.com/ThePalaceProject/circulation/actions/runs/5269226097/jobs/9527046094?pr=1197#step:8:427

Trying to make sure we clean these up as we see them, so we have less test failures happening.

## How Has This Been Tested?

Ran tests locally and in CI.

## Checklist


RishiDiwanTT (approved):
Looks good :+1:

PR #1203, opened 2023-06-15
Update poetry version to match what dependabot is using

jonathangreen:
## Description

Dependabot [bumped the version](https://github.com/dependabot/dependabot-core/pull/7350) of poetry they use up to 1.5.1. 

## Motivation and Context

We are seeing large diffs in the lock file, since 1.5.1 drop some keys in the lock file. Good to keep our version of Poetry roughly aligned with what Dependabot is using.

## How Has This Been Tested?

Running CI.

## Checklist

### DIFF
--- a/.github/actions/poetry/action.yml
+++ b/.github/actions/poetry/action.yml
@@ -5,7 +5,7 @@ inputs:
   version:
     description: "Version of poetry to install"
     required: false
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
👍🏽

PR #1206, opened 2023-06-19
Don't automatically assign books to children's lanes without explicit target age range

pcustic:
## Description
This PR makes changes to external_search module so that when search index is used to build lanes with audience="Children" it only includes works with explicit target age range.

## Motivation and Context
Books for children without explicit age range `(,)` were included in all lanes for children. That would result in situations were books for teens were added to lanes for very young kids (e.g. "Picture books", "Early readers", etc.). This makes necessary for books to have explicit target age to be put into children's lanes 

[Jira](https://ebce-lyrasis.atlassian.net/browse/PP-29)

## How Has This Been Tested?

Checked in CM admin and added test for lanes groups.

## Checklist


jonathangreen (approved):
Looks reasonable to me

PR #1143, opened 2023-05-25
Revert: initialize outside of first request

jonathangreen:
## Description

This PR reverts 3 previous PRs:
- #898 
- #1083 
- #1104 

## Motivation and Context

Once #898 was deployed out to dev we started seeing: 
```
minotaur-web    | Traceback (most recent call last):
minotaur-web    |   File "/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
minotaur-web    |     self.dialect.do_execute(
minotaur-web    |   File "/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute
minotaur-web    |     cursor.execute(statement, parameters)
minotaur-web    | psycopg2.errors.DeadlockDetected: deadlock detected
minotaur-web    | DETAIL:  Process 10517 waits for AccessExclusiveLock on relation 16897 of database 16409; blocked by process 10516.
minotaur-web    | Process 10516 waits for AccessExclusiveLock on relation 16897 of database 16409; blocked by process 10517.
minotaur-web    | HINT:  See server log for query details.
minotaur-web    |
minotaur-web    |
minotaur-web    | The above exception was the direct cause of the following exception:
minotaur-web    |
minotaur-web    | Traceback (most recent call last):
minotaur-web    |   File "/var/www/circulation/api/app.py", line 55, in initialize_application
minotaur-web    |     initialize_admin()
minotaur-web    |   File "/var/www/circulation/api/app.py", line 81, in initialize_admin
minotaur-web    |     with access_exclusive_lock_configurationsettings(_db) as db_session:
minotaur-web    |   File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
minotaur-web    |     return next(self.gen)
minotaur-web    |   File "/var/www/circulation/core/model/configuration.py", line 43, in access_exclusive_lock_configurationsettings
minotaur-web    |     session.execute("LOCK TABLE configurationsettings IN ACCESS EXCLUSIVE MODE;")
minotaur-web    |   File "/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/orm/scoping.py", line 163, in do
minotaur-web    |     return getattr(self.registry(), name)(*args, **kwargs)
minotaur-web    |   File "/var/www/circulation/env/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 1295, in execute
[...]

PR #1212, opened 2023-06-21
Move instance initialization out of before_first_request (PP-92)

jonathangreen:
## Description

This PR adds back in the code reverted in https://github.com/ThePalaceProject/circulation/pull/1143.

It makes some changes to our instance initialization to try to reduce issues we've had with locking: 
- The application does not initialize the DB on its own, this is done via a script run at startup of the docker container
- The instance initialization and migration scripts were merged together
- Instance initialization is done while holding a PG advisory lock, so only a single instance does the initialization / migration at a time.

## Motivation and Context

JIRA: PP-92

This should allow us to: 
- Upgrade flask
- Make sure the webserver is initialized before responding to requests. This will help our load balancer configuration for fargate.

## How Has This Been Tested?

Tried testing the following locally: 
- Running development server
  - Fresh DB
  - Existing DB
- Docker containers
  - Fresh DB
  - Existing DB

There are a lot of different cases here though, so its possible I've missed some cases that should be tested.

## Checklist


RishiDiwanTT:
@jonathangreen what had failed previously with this change, before the revert?

jonathangreen:
@RishiDiwanTT Advisory locks were in use at that time, but I think it was the exclusive lock on the `configurationsettings` table that was causing the database deadlocks, according the logs we saw. So I'm trying to avoid that lock this time around, and just use an advisory lock. 

In the end until we get rid of all the configuration settings, I think we need two advisory locks. One for setting up the DB, and one for initializing the application. I put a TODO in the code, because the one for initializing the application can go away once we move all the configuration settings over.

tests/migration/test_instance_init_script.py
```diff
@@ -0,0 +1,111 @@
...
+
+def _run_script() -> None:
+    try:
+        # Run the script, capturing the log output
+        script = InstanceInitializationScript()
+        stream = StringIO()
+        logging.basicConfig(stream=stream, level=logging.INFO, force=True)
+        script.run()
+
+        # Set out exit code to the number of upgrades we ran
```
tdilauro, line 24: Minor - typo: "Set [out]..." -> "Set [our]..."

tests/migration/test_instance_init_script.py
```diff
@@ -0,0 +1,111 @@
...
+    inspector = inspect(engine)
+    assert "alembic_version" in inspector.get_table_names()
+    assert "libraries" in inspector.get_table_names()
+    assert len(inspector.get_table_names()) > 2
+
+    assert script.initialize_database.call_count == 1
+    assert script.migrate_database.call_count == 0
+
+    # Run the script again and make sure we don't call initialize_database,
+    # but we do call migrate_database since the schema already exists.
```
tdilauro, line 87: Minor wording suggestion for clarity:
```
    # Run the script again. Ensure we don't call initialize_database again,
    # but that we do call migrate_database, since the schema already exists.
```

tdilauro (approved):
Looks good! 🎉

A couple of minor suggestions, but no show stoppers.

PR #1215, opened 2023-06-22
PP-61 opds2 feed fixes

RishiDiwanTT:
## Description
Minor changes to the OPDS2 publication feed 
- Identifiers as urns 
- Published and Modified as dates not datetimes

## Motivation and Context
The OPDS2 feed has some issues that are currently being ironed out, these are a couple of minor fixes.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-61)

## How Has This Been Tested?
The OPDS2 feed was tested manually.
All unit tests are passing.

## Checklist

### DIFF
--- a/core/opds2.py
+++ b/core/opds2.py
@@ -43,7 +43,7 @@ class OPDS2Annotator:
         result["@type"] = Edition.medium_to_additional_type.get(str(edition.medium))
         result["title"] = edition.title
         result["subtitle"] = edition.subtitle
-        result["identifier"] = edition.primary_identifier.identifier
+        result["identifier"] = edition.primary_identifier.urn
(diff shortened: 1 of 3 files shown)


core/opds2.py
```diff
@@ -69,9 +69,9 @@ def metadata_for_work(self, work: Work) -> Optional[Dict[str, Any]]:
         if edition.imprint:
             result["imprint"] = {"name": edition.imprint}
         if work.last_update_time:
-            result["modified"] = work.last_update_time.isoformat()
+            result["modified"] = work.last_update_time.date().isoformat()
```
jonathangreen, line 72: I don't think we want the modified date changed here, only the published date.

jonathangreen (approved):

PR #1216, opened 2023-06-22
PP-63 Made SIP2 Institution ID optional

RishiDiwanTT:
## Description
Simply made the SIP field "AO" for Instiution ID an optional response field, to avoid constant logging of the field being missing.

## Motivation and Context
Certain SIP2 servers would not respond with an "AO" field  if the CM sent a blank value for the same field in its request.
This would cause unnecessary logging in the CM.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-63)

## How Has This Been Tested?
This has not been tested, the self-tests for the SIP2 servers I tried all returned an "AO" field even if one was not sent accross.

## Checklist


jonathangreen (approved):
Looks good

PR #1217, opened 2023-06-22
PP-68 OPDS2 feeds with relative "next" urls are supported now

RishiDiwanTT:
## Description
Any url which uses the OPDSImportMonitor's _get method will be supported for relative urls

## Motivation and Context
In trying to import a feed containing relative URLs, the OPDS feed importer fails with an error log suggesting it requires an absolute URL reference in the links array
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-68)

## How Has This Been Tested?
A pseudo feed was created locally to test relative "next" urls.
Unit tests were created for the change.

## Checklist

### DIFF
--- a/core/opds_import.py
+++ b/core/opds_import.py
@@ -4,7 +4,7 @@ import logging
 import traceback
 from io import BytesIO
 from typing import TYPE_CHECKING, Optional
-from urllib.parse import urljoin, urlparse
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
Looks good

PR #1218, opened 2023-06-22
PP-97 Include excluded api/admin tests from mypy type checking

pcustic:
## Description
Adding tests from tests/api/admin module that were excluded to mypy type checking.

## Motivation and Context
During tests refactor some files were excluded from type checking. This is part of the effort to have full CM type checking coverage.

[Jira](https://ebce-lyrasis.atlassian.net/browse/PP-97)

## Checklist


core/lane.py
```diff
@@ -3211,12 +3213,6 @@ def explain(self):
         return lines
 
 
-Library.lanes = relationship(
-    "Lane",
-    backref="library",
-    foreign_keys=Lane.library_id,
-    cascade="all, delete-orphan",
-)
 DataSource.list_lanes = relationship(
```
RishiDiwanTT, line 3216: Since we removed the above bit of nightmare surprise database modelling, could we remove this nightmare surprise database modelling as well?

api/admin/controller/__init__.py
```diff
@@ -1109,7 +1109,7 @@ def _create_or_update_list(
         deleted_entries: Optional[List[Dict]] = None,
         id: Optional[int] = None,
         auto_update: Optional[bool] = None,
-        auto_update_query: Optional[str] = None,
+        auto_update_query: Optional[dict[str, str]] = None,
         auto_update_facets: Optional[str] = None,
```
RishiDiwanTT, line 1113: auto_update_facets is also the wrong type here

core/model/classification.py
```diff
@@ -348,9 +350,11 @@ class Classification(Base):
     __tablename__ = "classifications"
     id = Column(Integer, primary_key=True)
     identifier_id = Column(Integer, ForeignKey("identifiers.id"), index=True)
+    identifier: Identifier
     subject_id = Column(Integer, ForeignKey("subjects.id"), index=True)
     subject: Mapped[Subject] = relationship("Subject", back_populates="classifications")
     data_source_id = Column(Integer, ForeignKey("datasources.id"), index=True)
+    data_source: DataSource
```
RishiDiwanTT, line 357: Are data_source and identifier `Optional`?

RishiDiwanTT: Would it make sense for them to be `Mapped[<type>]` ?

core/model/configuration.py
```diff
@@ -584,7 +584,13 @@ class ConfigurationSetting(Base, HasSessionCache):
     external_integration_id = Column(
         Integer, ForeignKey("externalintegrations.id"), index=True
     )
+    external_integration: ExternalIntegration = relationship(
+        "ExternalIntegration", back_populates="settings"
+    )
+
     library_id = Column(Integer, ForeignKey("libraries.id"), index=True)
+    library: Library = relationship("Library", back_populates="settings")
```
RishiDiwanTT, line 592: Should this be `Mapped[Library]`

core/lane.py
pcustic: Removed 👍

RishiDiwanTT (approved):
Looks good :+1:

PR #1222, opened 2023-06-26
PP-57 Unlimited loan for unlimited access

RishiDiwanTT:
## Description
For the AcquisitionFeed, in case of an unlimited access pool, we don't enforce an end date to a Loan anymore.
The LoanReaperMonitor will no longer reap unlimited access loans

## Motivation and Context
Unlimited access loans (e.g. from BiblioBoard) have no defined end date, but the CM currently imposes a default 21 day loan period for such loans (see related issue). As a patron, I don’t want to be forced to return such loans when their use is unlimited.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-57)

## How Has This Been Tested?
Unit tests have been updated to match the expected behaviour.

## Checklist

### DIFF
--- a/api/monitor.py
+++ b/api/monitor.py
@@ -1,3 +1,5 @@
+from typing import Type
+
 from sqlalchemy import and_, or_
 
 from core.model import (
@@ -15,7 +17,6 @@ from .odl import ODLAPI, SharedODLAPI
 
 
 class LoanlikeReaperMonitor(ReaperMonitor):
-
     SOURCE_OF_TRUTH_PROTOCOLS = [
         ODLAPI.NAME,
         SharedODLAPI.NAME,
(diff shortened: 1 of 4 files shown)


api/monitor.py
```diff
@@ -72,7 +73,16 @@ def where_clause(self):
...
         )
-        return and_(superclause, or_(expired, very_old_with_no_clear_end_date))
+        not_unlimited_access = LicensePool.unlimited_access == False
+        return and_(
+            superclause,
+            not_unlimited_access,
+            or_(expired, very_old_with_no_clear_end_date),
+        )
+
+    def query(self):
```
pcustic, line 83: For what is this `query` method used?

pcustic (commented):
Seems fine, I have just one question.

api/monitor.py
RishiDiwanTT: The query() method is an overridden method from `core.monitor.ReaperMonitor`. We override it here to add the additional `join(licensepool)` to the SQL.

pcustic (approved):

PR #1230, opened 2023-06-28
Remove unused Adobe vendor ID code 🔥 (PP-135)

jonathangreen:
## Description

NYPL used to serve Adobe vendor ID directly from the CM, however this is no longer the case. Vendor ID are served via the library registry, however there is still a lot of code in the CM to support this unused use case. This PR cleans up that code.

JIRA: PP-135

## Motivation and Context

In the work on PP-11, i noticed that adobe vendor ID code requires refactoring to use the new library settings. However since this code is unused anyway, it makes more sense to just remove it entirely, rather then refactor it. 

If you assume that there are on average 15 bugs per kloc, this PR fixes ~35 statistical bugs 🐛🥾  😄  

## How Has This Been Tested?

- Running tests locally and in CI
- Did some light testing locally with a copy of the Minotaur database, to make sure that we still see the correct drm tokens in the feed, and we do. 

This will need some more testing when it hits dev, with the mobile apps to make sure it doesn't cause any issues with adobe loans.

## Checklist


RishiDiwanTT (approved):
I'm unaware of the use-case of the changed test case, but the rest of the changes seem all good :+1:

PR #1236, opened 2023-06-29
Remove second make_default_library function from fixtures (PP-11)

jonathangreen:
## Description

Remove deprecated `make_default_library` function, so that we don't have two copies of this function in our fixtures.

@RishiDiwanTT tagging you for a review, because I think you looked at this code while moving the mocks into the test suite.

## Motivation and Context

While working on PP-11, I ran into issues in the tests with library creation because there are two places in the code we can create a "default" library. 

## How Has This Been Tested?

Running unit tests.

## Checklist


RishiDiwanTT (approved):
Now why didn't I think of just passing the library in :shrug:

PR #1237, opened 2023-06-30
PP-101 Add excluded tests/core/* files to mypy type check

pcustic:
## Description
Adding tests from tests/core module that were excluded to mypy type checking.

## Motivation and Context
During tests refactor some files were excluded from type checking. This is part of the effort to have full CM type checking coverage.

[Jira](https://ebce-lyrasis.atlassian.net/browse/PP-101)

## How Has This Been Tested?

Ran mypy type checker and existing tests.

## Checklist


jonathangreen:
I rebased this to the `main` branch, and resolved the couple test failures, so I think this is good to go. We can remove the exclude for the last two files after @io7m is done his elasticsearch refactor.

RishiDiwanTT (approved):
All looks good :+1:

PR #1242, opened 2023-07-04
PP-120 Fix the test_groups method of the lanes test cases

RishiDiwanTT:
## Description
The model seems to have moved on since the tests, so some changes were required
The sorting of the 2 databases seems to have drifted in some way so similarly named works had to be named a certain way.

## Motivation and Context
While working on tests for different ticket I noticed we have one test in `tests/core/test_lane.py → TestWorkListGroupsEndToEnd::test_groups` that is occasionally failing but it is marked with pytest.mark.xfail so it is not failing our CI system.

## How Has This Been Tested?
The unit test now passes. No changes were made in the application.

## Checklist

### DIFF
--- a/tests/core/test_lane.py
+++ b/tests/core/test_lane.py
@@ -4094,13 +4094,13 @@ class TestWorkListGroupsEndToEnd:
 
         # Create children works.
         result.children_with_age = _w(
-            title="Children work with set target age",
+            title="Children work with target age",
             audience=Classifier.AUDIENCE_CHILDREN,
         )
-        result.children_with_age.target_age = (0, 3)
+        result.children_with_age.target_age = tuple_to_numericrange((0, 3))
 
         result.children_without_age = _w(
-            title="Children work without target age",
+            title="Children work with out target age",
             audience=Classifier.AUDIENCE_CHILDREN,
         )
 
@@ -4136,9 +4136,6 @@ class TestWorkListGroupsEndToEnd:
         result.staff_picks_list.add_entry(result.mq_sf)
         return result
 
-    # TODO: This test needs to be fixed. It fails roughly one out of five runs. For
-    #  now I'm just marking it so it doesn't fail our CI every time it fails.
-    @pytest.mark.xfail()
     def test_groups(
         self,
         end_to_end_search_fixture: EndToEndSearchFixture,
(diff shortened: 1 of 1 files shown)


tests/core/test_lane.py
```diff
@@ -4094,13 +4094,13 @@ def _w(**kwargs):
 
         # Create children works.
         result.children_with_age = _w(
-            title="Children work with set target age",
+            title="Children work with target age",
```
RishiDiwanTT, line 4097: SQL vs OpenSearch alphabetical sort seems to have some differences that sort the previous title differently. This change works around the difference.

tests/core/test_lane.py
```diff
@@ -4094,13 +4094,13 @@ def _w(**kwargs):
...
             audience=Classifier.AUDIENCE_CHILDREN,
         )
-        result.children_with_age.target_age = (0, 3)
+        result.children_with_age.target_age = tuple_to_numericrange((0, 3))
 
         result.children_without_age = _w(
-            title="Children work without target age",
+            title="Children work with out target age",
```
RishiDiwanTT, line 4103: Same as above, making sure the sorting happens in the same way with both back-ends

tests/core/test_lane.py
```diff
@@ -4183,7 +4178,8 @@ def test_groups(
 
         # "Children", which will contain one book, the one with audience children and defined target age.
         children = db.lane("Children")
-        children.audience = Classifier.AUDIENCE_CHILDREN
+        children.audiences = Classifier.AUDIENCE_CHILDREN
+        children.target_age = (0, 4)
```
RishiDiwanTT, line 4182: The filter object seems to now __need__ a target age range to apply age ranges now.

tests/core/test_lane.py
```diff
@@ -4183,7 +4178,8 @@ def test_groups(
 
         # "Children", which will contain one book, the one with audience children and defined target age.
         children = db.lane("Children")
-        children.audience = Classifier.AUDIENCE_CHILDREN
+        children.audiences = Classifier.AUDIENCE_CHILDREN
```
RishiDiwanTT, line 4181: The Lanes model attribute has changed since this test was last run, it seems.

jonathangreen (approved):
Looks good. Nice to have this test fixed.

PR #1243, opened 2023-07-04
Fix RMS value for histogram of images.

jonathangreen:
## Description

Replace the function we are using to calculate the RMS differences between images in our tests with this one from stackoverflow: https://stackoverflow.com/a/40176818. 

I didn't spend a lot of time debugging what was happening with the old RMS function, but the new one is returning 0 as the RMS value, when the old one would return 13.8 on my local system. And if I actually change the image (say by changing font size, it gives reasonable RMS value). 

I think this will reduce the number of test failures we see locally.

## Motivation and Context

The `test_process_cover_image` test in `test_work_editor.py` is constantly failing for me. We've bumped this value up several times, but it still seems to fail randomly on my local system.

Pinging you @io7m for a review, since you've spent some time troubleshooting this test before.

## How Has This Been Tested?

Running tests locally, running tests in CI.

## Checklist


tdilauro (approved):
Looks good!

PR #1245, opened 2023-07-05
Possibly fix the flakiness of core.test_lanes

RishiDiwanTT:
## Description
Dropping the genre quality of one work in test_groups to ensure test reproducibility
hq_litfic, hq_sf and hq_romance all have quality 0.8, which means it's possible that the opensearch index may return any 2 of the 3 works we ask for.
By dropping the quality of hq_romance we ensure only the first 2 show up.

## Motivation and Context
The test_lanes test randomly fails in the CI, but not locally, this is a hypotheses, and should be run multiple times in the CI.

## How Has This Been Tested?
This has only been tested locally, where it isn't an issue.
:warning:  The CI must run this test multiple times to ensure the test is passing.

## Checklist


jonathangreen (approved):
✅ Looks good to me

PR #1250, opened 2023-07-06
PP-159 CDN feed urls

RishiDiwanTT:
## Description
CDN urls options are available for OPDS1 and OPDS2 feeds

The feed only cdnify's urls if explicitly asked to, like the links to other feeds
- search url
- contributor url
- facet url
- etc..
Any action oriented url is not cdn'd, eg. borrow links

CDN options are picked up from the environment
The Readme has been updated to reflect the new environment variables

## Motivation and Context
We should have CloudFront CDN caching for the OPDS 2 crawlable feeds so our systems are not negatively impacted when the feeds are harvested.

## How Has This Been Tested?
Unit tests have been updated for the new functionality
OPDS1 and 2 feeds were manually tested locally

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,18 @@ export SIMPLIFIED_FCM_CREDENTIALS_FILE="/opt/credentials/fcm_credentials.json"
 The FCM credentials can be downloaded once a Google Service account has been created.
 More details in the [FCM documentation](https://firebase.google.com/docs/admin/setup#set-up-project-and-service-account)
 
+##### CDN caching
+
+OPDS1 and OPDS2 feeds can have cacheable links using the following variables
+
+```sh
+export SIMPLIFIED_CDN_BASE_URL=https://cdn.host.tld
+export SIMPLIFIED_CDN_OPDS1_ENABLED=TRUE
+export SIMPLIFIED_CDN_OPDS2_ENABLED=TRUE
+```
+
+Any value that is not a case-variant of "TRUE" indicates a False to the application.
+
 ### Email sending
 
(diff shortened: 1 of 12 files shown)


jonathangreen:
@RishiDiwanTT I'd like to explore the possibility of a different approach here.

Ideally, I'd like the application not to have to re-write URLs like this, but to make the CDN caching transparent to the application. 

We can point cloudfront at our ALB, then serve the application via cloudfront. We need to do some investigation about how we can configure the cloud-front distribution, ideally we would only cache a few feeds at first. But the idea would be to eventually serve all non-authenticated requests via cloudfront.

I've not used cloudfront as a cache like this before, I've used fastly, so I'm not sure how much configuration cloudfront offers us, but I think we should do a bit of investigation here.

jonathangreen:
Going with the approach of serving all our application URLs via CDN cache gives us a lot of flexibility to serve as many of our static URLs via a cache though.

RishiDiwanTT:
> Going with the approach of serving all our application URLs via CDN cache gives us a lot of flexibility to serve as many of our static URLs via a cache though.

Going through cloudfront in such a manner would mean we need no changes in the application code.

jonathangreen:
Agreed @RishiDiwanTT, the only app side changes we may need, are to change the caching headers we send out to cloudfront, to make sure the correct things are being cached. Thats why I was surprised to see this PR 😄

RishiDiwanTT:
Closing this PR in favour of having this solved through a terraform deployment

PR #1251, opened 2023-07-06
Store library key pair in library table (PP-11)

jonathangreen:
## Description

Move the libraries key pair into two columns on the database: `public_key` and `private_key` rather then storing them in configuration settings. Instead of initializing these values when we load the authenticators, we create them at library creation time. That way we know each library always has a key pair.

## Motivation and Context

This work is part of PP-11. As part of moving libraries over to `IntegrationSettings`, we want to stop storing settings that are not user configurable in `ConfigurationSettings`. There are several such settings to be moved around for libraries. 

## How Has This Been Tested?

- Did some light local testing, creating a library, running migration, etc
- Ran tests in CI

## Checklist


jonathangreen:
Most of the test changes are just changes to make sure we use our fixtures to create a library, instead of directly creating one in the database. This is necessary because we added two new columns for library that are not nullable.

alembic/versions/20230706_04bbd03bf9f1_migrate_library_key_pair.py
```diff
@@ -0,0 +1,69 @@
...
+        sa.Column("private_key", sa.LargeBinary(), nullable=True),
+    )
+
+    # Now we update the value stored for the key pair
+    connection = op.get_bind()
+    libraries = connection.execute("select id, short_name from libraries")
+    for library in libraries:
+        setting = connection.execute(
+            "select cs.value from configurationsettings cs "
+            "where cs.library_id = (%s) and cs.key = 'key-pair'",
```
tdilauro, line 44: For a per-library setting, the `where` clause should include a null external integration:
```
"where cs.library_id = (%s) and cs.external_integration_id IS NULL and cs.key = 'key-pair'",
```

api/authenticator.py
```diff
@@ -705,7 +699,8 @@ def create_authentication_document(self) -> str:
             doc["service_area"] = service_area
 
         # Add the library's public key.
-        doc["public_key"] = dict(type="RSA", value=self.public_key)
+        if self.library and self.library.public_key:
```
tdilauro, line 702: Should the absence of a public key here be an error?

core/model/library.py
```diff
@@ -155,6 +157,12 @@ class Library(Base, HasSessionCache):
         cascade="all, delete-orphan",
     )
 
+    # The libraries public / private key-pair.
```
tdilauro, line 160: Minor typo?: "The library's..."

tdilauro (approved):
Looks good! Just a couple of small suggestions and a question.

api/authenticator.py
jonathangreen: I think missing a public key can arise normally in the current code. 

Since we initialize the key-pair when we load a `LibraryAuthenticator` for the first time, any library without an auth mechanism won't have a keypair yet. I don't think this warrents an error, since we would have just created a key pair once an auth mechanism is added. I don't think we'll see many libraries in this state, but we will probably see a few on our dev and staging servers.

tdilauro: That sounds fine. Just wanted to check, since the keys can't be null in the database and the migration should take care of any that are.

jonathangreen: I was looking at the wrong code for this comment. 🤦🏻 

This is mostly for mypy, since both `self.library` and `self.library.public_key` could theoretically potentially both be `None`. I'll add an error log statement in the `else` condition here, so that we get some indication if this is ever happening anywhere.

PR #1258, opened 2023-07-06
Store library logo in a seperate table (PP-11)

jonathangreen:
## Description

Create a new table `libraries_logos` that stores the logos for libraries. I chose to put this in a new table rather then add it to the libraries table because the images seemed like they might be large enough that we would want to be able to lazy load them, instead of having to load them every time we load a library.

This PR is build off of #1251 and should go in after that one does.

## Motivation and Context

This work is part of PP-11. As part of moving libraries over to `IntegrationSettings`, it doesn't seem reasonable to be storing a base-64 encoded PNG, inside a json object, inside the database. So I took the approach of moving the logos out of configuration settings all together, and putting them in their own table.

## How Has This Been Tested?

- Light local testing, creating a library, editing a library, deleting a library, and checking that the database is updated as I would expect.
- Running tests locally and in CI.

## Checklist


alembic/versions/20230706_c471f553249b_migrate_library_logo.py
```diff
@@ -0,0 +1,66 @@
...
+        setting = connection.execute(
+            "select cs.value from configurationsettings cs "
+            "where cs.library_id = (%s) and cs.key = 'logo'",
+            (library.id,),
+        ).first()
+        if setting and setting.value:
+            log.info("Library {library.short_name} has a logo, migrating it.")
+            logo_str = setting.value
+
+            # We stored the logo with a data:image prefix before, but we
```
RishiDiwanTT, line 51: Are we always just assuming it's png forevermore?

RishiDiwanTT (approved):
Minor comment. All looks good :+1:

alembic/versions/20230706_c471f553249b_migrate_library_logo.py
jonathangreen: Yes, I suppose we are. Since we convert everything to PNG, I figured this was okay. I could add a content_type column to the LibraryImages table if you think that would be a good change.

RishiDiwanTT:
No, if everything is expected to be png then that's fine.

PR #1259, opened 2023-07-07
PP-156 Switched the auto update lists to use sort key pagination

RishiDiwanTT:
## Description
This allows extremely large lists to be populated via auto update.

## Motivation and Context
The CustomListQueries.populate_query_pages method uses offset based pagination which will be limited to 10000 search results by opensearch.

[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-156)

## How Has This Been Tested?
Unit tests have been updated to expect the changed pagination
Manually tested a list of > 100K size, which is painfully slow to populate.

## Checklist


jonathangreen (commented):
🚀

jonathangreen (approved):

PR #1267, opened 2023-07-11
Move announcements out of configurationsettings into own table (PP-177)

jonathangreen:
## Description

Move the global and library announcements out of configuration settings, and into their own table.

## Motivation and Context

This work is an offshoot of the work in PP-11, but was broken out to its own ticket in PP-177. Since these settings used to be stored in configuration settings, they had to be converted. However the nature of announcements, having both library specific and global announcements, made them a much better fit for having their own table, then being stored as a configuration setting.

## How Has This Been Tested?

- Local CM testing
- Running tests

## Checklist


RishiDiwanTT (approved):
This looks good :+1:

PR #1268, opened 2023-07-12
PP65: Opensearch migrations

io7m:
## Description

This is the bulk of the herculean task required to perform safe Opensearch migrations. The general outline is detailed in this document:

https://ataxia.io7m.com/2023/06/08/search/

The main changes here:

* We now treat schemas ("mappings" in Opensearch terminology) as immutable. If a field needs to change type, or if a new field needs to be added, that requires a schema upgrade.
* Schema revisions are represented by (subclasses of) the `SearchSchemaRevision` class. All of the code required to set up mappings for a single schema version are in the class for that version. Currently, the only defined version is version `5` (`SearchV5`).
* Schema version classes are aggregated into a directory (`SearchRevisionDirectory`) with the system picking the "best" (read: highest) available version as the version to which the system will attempt to migrate when instructed.
* Atomic read/write pointers are used to ensure that we can always read search documents from old indexes if the migration to the new index didn't succeed, or is in progress and is taking all night.
* Migration is performed by the `SearchMigrator` class.
* ~~Code no longer mocks the `ExternalSearchIndex`. There was a lot of confusing code around this class, including lots of injection of mock classes into it, layering violations, code reaching into the internals, and some very confused tests that appeared to be testing the behaviour of mocks alone. :shrug:~~ We mock the search index, but have no special tests for the mocks, we only expect the mocked search index to provide specific data when testing some other functionality
* All communication with Opensearch is encapsulated in a `SearchServiceOpensearch1` class, which is a subclass of an abstract `SearchService` class. If behaviour needs to be mocked, the correct approach is to produce mock implementations of the `SearchService` class as that encapsulates all of the actual side effects that need to be mocked.
* The `SearchServiceOpensearch1` class does not do any retries if an entire batch of documents fails to be submitted (perhaps due to a timeout). This is something that will need to be added back in if we need it. Thankfully, there's only a single method that would require updates, as there's no longer a free-for-all all over the codebase of submitting search documents through the bulk API. :angry: 

This required quite an obscene amount of debugging and untangling of code.

~~There are currently 12 test failures that I don't understand. One existing test had to be marked as skipped, because despite consisting mostly of mocks, the test itself would hang somewhere inside the bowels of `pytest` and require a `SIGKILL` to stop.~~

## Motivation and Context

We have no way to safely and atomically upgrade (or downgrade) our search indexes. If a search index fails for any reason, we have an outage until the new index is populated.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-65

## How Has This Been Tested?

This has been tested extensively locally, and lots of new tests were written.

## Checklist

### DIFF
--- a/api/admin/controller/custom_lists.py
+++ b/api/admin/controller/custom_lists.py
@@ -283,7 +283,12 @@ class CustomListsController(
         if membership_change:
             # We need to update the search index entries for works that caused a membership change,
             # so the upstream counts can be calculated correctly.
-            self.search_engine.bulk_update(works_to_update_in_search)
+            documents = self.search_engine.create_search_documents_from_works(
+                works_to_update_in_search
+            )
+            self.search_engine.search_service().index_submit_documents(
+                self.search_engine._search_write_pointer, documents
+            )
 
             # If this list was used to populate any lanes, those lanes need to have their counts updated.
             for lane in Lane.affected_by_customlist(list):
--- a/api/admin/controller/custom_lists.py
+++ b/api/admin/controller/custom_lists.py
@@ -286,9 +286,9 @@ class CustomListsController(
             documents = self.search_engine.create_search_documents_from_works(
                 works_to_update_in_search
             )
-            self.search_engine.search_service().index_submit_documents(
-                self.search_engine._search_write_pointer, documents
-            )
+            index = self.search_engine.start_updating_search_documents()
+            index.add_documents(documents)
+            index.finish()
 
             # If this list was used to populate any lanes, those lanes need to have their counts updated.
             for lane in Lane.affected_by_customlist(list):
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -514,6 +514,28 @@ class ExternalSearchIndex(HasSelfTests):
 
         yield self.run_test("Total number of documents per collection:", _collections)
 
+    def initialize_indices(self) -> bool:
+        """Attempt to initialize the indices and pointers for a first time run"""
+        service = self.search_service()
+        read_pointer = service.read_pointer(self._revision_base_name)
+        if not read_pointer or service.is_pointer_empty(
+            self._revision_base_name, read_pointer
+        ):
+            # A read pointer does not exist, or points to the empty index
+            # This means either this is a new deployment or the first time
+            # the new opensearch code was deployed.
+            # In both cases doing a migration to the latest version is safe.
+            migration = self.start_migration()
+            if migration is not None:
+                migration.finish()
+            else:
+                self.log.warning(
+                    "Read pointer was set to empty, but no migration was available."
+                )
+                return False
+
+        return True
+
 
 class SearchBase:
     """A superclass containing helper methods for creating and modifying
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -181,12 +181,6 @@ class ExternalSearchIndex(HasSelfTests):
         if not integration:
             raise CannotLoadConfiguration("No search integration configured.")
 
-        # Determine the URL that is going to be used to connect to the search server.
-        if isinstance(url, ExternalIntegration):
-            # This is how the self-test initializes this object.
-            integration = url
-            url = integration.url
-
         if not url:
             url = url or integration.url
             test_search_term = integration.setting(self.TEST_SEARCH_TERM_KEY).value
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -2802,10 +2802,16 @@ class SearchIndexCoverageProvider(RemovesSearchCoverage, WorkPresentationProvide
             self.remove_search_coverage_records()
 
     def on_completely_finished(self):
-        super().on_completely_finished()
         # Tell the search migrator that no more documents are going to show up.
         target: SearchDocumentReceiverType = self.migration or self.receiver
         target.finish()
+
+    def run_once_and_update_timestamp(self):
+        try:
+            result = super().run_once_and_update_timestamp()
+            return result
+        finally:
+            self.on_completely_finished()
 
     def process_batch(self, works) -> List[Work | CoverageFailure]:
         target: SearchDocumentReceiverType = self.migration or self.receiver
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -191,6 +191,11 @@ class ExternalSearchIndex(HasSelfTests):
         if not url:
             raise CannotLoadConfiguration("No URL configured to the search server.")
 
+        # Determine the base name we're going to use for storing revisions.
+        self._revision_base_name = integration.setting(
+            ExternalSearchIndex.WORKS_INDEX_PREFIX_KEY
+        ).value
+
         # Create the necessary search client, and the service used by the schema migrator.
         if custom_client_service:
             self._search_service = custom_client_service
@@ -198,7 +203,9 @@ class ExternalSearchIndex(HasSelfTests):
             use_ssl = url.startswith("https://")
             self.log.info("Connecting to the search cluster at %s", url)
             new_client = OpenSearch(url, use_ssl=use_ssl, timeout=20, maxsize=25)
-            self._search_service = SearchServiceOpensearch1(new_client)
+            self._search_service = SearchServiceOpensearch1(
+                new_client, self._revision_base_name
+            )
 
         # Locate the revision of the search index that we're going to use.
         # This will fail fast if the requested version isn't available.
@@ -209,11 +216,6 @@ class ExternalSearchIndex(HasSelfTests):
             self._revision = self._revision_directory.find(version)
         else:
             self._revision = self._revision_directory.highest()
-
-        # Determine the base name we're going to use for storing revisions.
-        self._revision_base_name = integration.setting(
-            ExternalSearchIndex.WORKS_INDEX_PREFIX_KEY
-        ).value
 
         # initialize the cached data if not already done so
         CachedData.initialize(_db)
@@ -226,12 +228,8 @@ class ExternalSearchIndex(HasSelfTests):
 
         try:
             # Get references to the read and write pointers.
-            self._search_read_pointer = self._search_service.read_pointer_name(
-                self._revision_base_name
-            )
-            self._search_write_pointer = self._search_service.write_pointer_name(
-                self._revision_base_name
-            )
+            self._search_read_pointer = self._search_service.read_pointer_name()
+            self._search_write_pointer = self._search_service.write_pointer_name()
             self._search_migrator = SearchMigrator(
                 revisions=self._revision_directory,
                 service=self._search_service,
@@ -281,9 +279,7 @@ class ExternalSearchIndex(HasSelfTests):
         else:
             query = Query(query_string, filter)
 
-        search = query.build(
-            self._search_service.search_client(self._search_read_pointer), pagination
-        )
+        search = query.build(self._search_service.search_client(), pagination)
         if debug:
             search = search.extra(explain=True)
 
@@ -360,7 +356,7 @@ class ExternalSearchIndex(HasSelfTests):
             (query string, Filter, Pagination) 3-tuple.
         """
         # Create a MultiSearch.
-        multi = self._search_service.search_multi_client(self._search_read_pointer)
+        multi = self._search_service.search_multi_client()
 
         # Give it a Search object for every query definition passed in
         # as part of `queries`.
@@ -488,7 +484,7 @@ class ExternalSearchIndex(HasSelfTests):
 
         def _count_docs():
             service = self.search_service()
-            client = service.search_client(self._search_read_pointer)
+            client = service.search_client()
             return str(client.count())
 
         yield self.run_test(
@@ -518,10 +514,8 @@ class ExternalSearchIndex(HasSelfTests):
     def initialize_indices(self) -> bool:
         """Attempt to initialize the indices and pointers for a first time run"""
         service = self.search_service()
-        read_pointer = service.read_pointer(self._revision_base_name)
-        if not read_pointer or service.is_pointer_empty(
-            self._revision_base_name, read_pointer
-        ):
+        read_pointer = service.read_pointer()
+        if not read_pointer or service.is_pointer_empty(read_pointer):
             # A read pointer does not exist, or points to the empty index
             # This means either this is a new deployment or the first time
             # the new opensearch code was deployed.
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -339,8 +339,15 @@ class ExternalSearchIndex(HasSelfTests):
 
         pagination = pagination or Pagination.default()
         query_data = (query_string, filter, pagination)
-        [result] = self.query_works_multi([query_data], debug)
-        return result
+        query_hits = self.query_works_multi([query_data], debug)
+        if not query_hits:
+            return []
+
+        result_list = list(query_hits)
+        if not result_list:
+            return []
+
+        return result_list[0]
 
     def query_works_multi(self, queries, debug=False):
         """Run several queries simultaneously and return the results
@@ -441,7 +448,7 @@ class ExternalSearchIndex(HasSelfTests):
             pointer=self._search_read_pointer, id=work.id
         )
 
-    def _run_self_tests(self, _db, in_testing=False):
+    def _run_self_tests(self, _db):
         # Helper methods for setting up the self-tests:
 
         def _search():
@@ -466,10 +473,6 @@ class ExternalSearchIndex(HasSelfTests):
 
         def _get_raw_doc():
             search = _search()
-            if in_testing:
-                if not len(search):
-                    return str(search)
-                search = search[0]
             return json.dumps(search.to_dict(), indent=1)
 
         yield self.run_test(
@@ -485,10 +488,9 @@ class ExternalSearchIndex(HasSelfTests):
         )
 
         def _count_docs():
-            # The mock methods used in testing return a list, so we have to call len() rather than count().
-            if in_testing:
-                return str(len(self._search))
-            return str(self._search.count())
+            service = self.search_service()
+            client = service.search_client()
+            return str(client.count())
 
         yield self.run_test(
             ("Total number of search results for '%s':" % (self._test_search_term)),
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -2807,11 +2807,11 @@ class SearchIndexCoverageProvider(RemovesSearchCoverage, WorkPresentationProvide
         target.finish()
 
     def run_once_and_update_timestamp(self):
-        try:
-            result = super().run_once_and_update_timestamp()
-            return result
-        finally:
-            self.on_completely_finished()
+        # We do not catch exceptions here, so that the on_completely finished should not run
+        # if there was a runtime error
+        result = super().run_once_and_update_timestamp()
+        self.on_completely_finished()
+        return result
 
     def process_batch(self, works) -> List[Work | CoverageFailure]:
         target: SearchDocumentReceiverType = self.migration or self.receiver
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -96,7 +96,7 @@ class ExternalSearchIndex(HasSelfTests):
     WORKS_INDEX_PREFIX_KEY = "works_index_prefix"
     DEFAULT_WORKS_INDEX_PREFIX = "circulation-works"
 
-    TEST_SEARCH_TERM_KEY = "test_search_term"
+    TEST_SEARCH_TERM_KEY = "a search term"
     DEFAULT_TEST_SEARCH_TERM = "test"
     CURRENT_ALIAS_SUFFIX = "current"
 
@@ -189,6 +189,8 @@ class ExternalSearchIndex(HasSelfTests):
             url = url or integration.url
             test_search_term = integration.setting(self.TEST_SEARCH_TERM_KEY).value
 
+        self._test_search_term = test_search_term or self.DEFAULT_TEST_SEARCH_TERM
+
         if not url:
             raise CannotLoadConfiguration("No URL configured to the search server.")
 
@@ -216,18 +218,8 @@ class ExternalSearchIndex(HasSelfTests):
             ExternalSearchIndex.WORKS_INDEX_PREFIX_KEY
         ).value
 
-        # Get references to the read and write pointers.
-        self._search_read_pointer = self._search_service.read_pointer_name(
-            self._revision_base_name
-        )
-        self._search_write_pointer = self._search_service.write_pointer_name(
-            self._revision_base_name
-        )
-
         # initialize the cached data if not already done so
         CachedData.initialize(_db)
-
-        self._test_search_term = test_search_term or self.DEFAULT_TEST_SEARCH_TERM
 
         # Try to perform a migration to the requested schema version. This might be a no-op if everything
         # is already up-to-date, or if we're downgrading to an already-populated index. It also might fail
@@ -235,12 +227,18 @@ class ExternalSearchIndex(HasSelfTests):
         # a good state, but we'll have to submit new search documents and re-run the migration later to fully
         # upgrade.
 
-        self._search_migrator = SearchMigrator(
-            revisions=self._revision_directory,
-            service=self._search_service,
-        )
-
         try:
+            # Get references to the read and write pointers.
+            self._search_read_pointer = self._search_service.read_pointer_name(
+                self._revision_base_name
+            )
+            self._search_write_pointer = self._search_service.write_pointer_name(
+                self._revision_base_name
+            )
+            self._search_migrator = SearchMigrator(
+                revisions=self._revision_directory,
+                service=self._search_service,
+            )
             migration = self._search_migrator.migrate(
                 base_name=self._revision_base_name, version=self._revision.version
             )
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import contextlib
 import datetime
 import json
@@ -284,7 +286,9 @@ class ExternalSearchIndex(HasSelfTests):
         else:
             query = Query(query_string, filter)
 
-        search = query.build(self._search_service.search_client(), pagination)
+        search = query.build(
+            self._search_service.search_client(self._search_read_pointer), pagination
+        )
         if debug:
             search = search.extra(explain=True)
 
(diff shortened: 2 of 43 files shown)


core/search/migrator.py
```diff
@@ -0,0 +1,171 @@
...
+        self, documents: Iterable[dict]
+    ) -> List[SearchServiceFailedDocument]:
+        """Submit documents to be indexed."""
+
+    @abstractmethod
+    def finish(self) -> None:
+        """Make sure all changes are committed."""
+
+
+class SearchDocumentReceiver(SearchDocumentReceiverType):
```
RishiDiwanTT, line 32: Is this an extra step, these behaviours seem to be all part of the SearchService interface

core/search/revision_directory.py
```diff
@@ -0,0 +1,41 @@
+from typing import Mapping
+
+from core.config import CannotLoadConfiguration
+from core.search.revision import SearchSchemaRevision
+from core.search.v5 import SearchV5
+
+
+class SearchRevisionDirectory:
```
RishiDiwanTT, line 8: Do we need anything more than the current and next expected index?

core/search/service.py
```diff
@@ -0,0 +1,330 @@
...
+            error_status = error_indexed["status"]
+            error_reason = error_indexed["error"]["reason"]
+            return SearchServiceFailedDocument(
+                id=error_id,
+                error_message=error_reason,
+                error_status=error_status,
+                error_exception="<unavailable>",
+            )
+        else:
+            raise NotImplementedError
```
RishiDiwanTT, line 61: Minor: raise NotImplementedError`()`

RishiDiwanTT (commented):
Still reviewing this, only done 20%

core/search/migrator.py
io7m: The `SearchService` is supposed to be an absolute bare minimum stronger-typed interface to the Opensearch client. I've tried to separate mechanism and policy. The `SearchDocumentReceiver` and `SearchMigrationInProgress` classes are wrappers to give the code that needs to use them a much smaller interface to deal with ("start", "add documents", "finish").

core/search/revision_directory.py
io7m: Hopefully not, no. Replace `current` and `next` with their numeric values, and we get the same thing. :slightly_smiling_face:

core/search/service.py
io7m: :+1: 

Oversight. Didn't mean to leave this one in.

core/search/service.py
```diff
@@ -0,0 +1,330 @@
...
+        self._logger.debug(f"setting write pointer {alias_name} to {target_index}")
+        self._client.indices.update_aliases(body=action)
+
+    def read_pointer(self, base_name: str) -> Optional[str]:
+        try:
+            result: dict = self._client.indices.get_alias(
+                name=self.read_pointer_name(base_name)
+            )
+            for name in result.keys():
+                if name.startswith(f"{base_name}-"):
```
RishiDiwanTT, line 307: What is the difference between the 2 checks for write and read pointers wherein one is a regex and this is a "startswith"?

tests/core/search/test_documents.py
```diff
@@ -0,0 +1,162 @@
...
+    date,
+    icu_collation_keyword,
+    keyword,
+    nested,
+    sort_author_keyword,
+)
+
+
+class TestDocuments:
+    def test_binary(self):
```
RishiDiwanTT, line 25: Being a pedant again: All the type checking tests could be a single parameterized test

core/search/revision_directory.py
RishiDiwanTT: In case a downgrade is required, will deleting the previous highest schema work? (Ensuring the new highest is the rollback version)

io7m: Downgrades definitely work as they're just an atomic alias change.

I've gone back and forth on adding a configuration option that allows for setting the schema version explicitly, but I'm still unsure about it. If an explicit schema version can be set in a configuration, that also means that at some point it needs to be unset. It's yet another manual bit of configuration for deployments. :thinking:

io7m: Er, to be clear, I don't think deleting the highest schema is a good idea. It shouldn't break things, but it's a bit conceptually unsound; it essentially makes the system pretend that it doesn't know about a version.

core/search/service.py
```diff
@@ -0,0 +1,330 @@
...
+    def indexes_created(self) -> List[str]:
+        return self._indexes_created
+
+    def write_pointer(self, base_name: str) -> Optional[SearchWritePointer]:
+        try:
+            result: dict = self._client.indices.get_alias(
+                name=self.write_pointer_name(base_name)
+            )
+            for name in result.keys():
+                match = re.search(f"{base_name}-v([0-9]+)", string=name)
```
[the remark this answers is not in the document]
io7m: I've been so buried in test failures that I've forgotten why it was like this. I'm going to adjust these patterns today as I think they need to be more explicit. Our read/write pointers should only ever point to the versioned indices, or the empty index. They're not user-serviceable parts and should never point to anything else.

tests/core/search/test_documents.py
io7m: Is there any benefit to doing this?

core/search/service.py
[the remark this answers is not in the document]
io7m: Actually, this is correct:

The asymmetry is because it's always safe to _read_ from an index (even the empty index), but it's not safe to _write_ to an index (as in, submit documents and/or update the mappings) unless you know the structure of the mappings. Hence, the `write_pointer` method either returns the versioned index name, or it returns nothing at all so that you can't write to an index.

core/search/service.py
io7m: See above. :arrow_up:

tests/core/search/test_documents.py
RishiDiwanTT: Reusability if you want to make a change in the test code everywhere

core/search/service.py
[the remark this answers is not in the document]
RishiDiwanTT: Cool. That's some implicit behaviour, maybe we can add some comments explaining it out in the method.

core/search/revision_directory.py
RishiDiwanTT: How would a downgrade work with it knowing all the versions, including the highest version?

core/search/service.py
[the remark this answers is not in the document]
io7m: Will do!

core/search/revision_directory.py
io7m: If the currently deployed version of the app knows about some set of versions `{ ... N, M}` and you're at version `M` and want to downgrade to `N`, it's safe to set `{base_name}-search-read` to `{base_name}-vN` as long as there exists a `{base_name}-vN-indexed` alias (that alias tells you that at some point index `N` was fully populated and working).

There's no explicit in-application way to do this right now as we actually never discussed one. I assumed it'd be a temporary emergency measure if some version turns out to be broken, so I assumed someone would be setting the value of the alias manually in opensearch.

If no `{base_name}-vN-indexed` alias exists, then that's going to involve repopulating the index. If we do need some kind of explicit in-app way to do this (explicitly set the version and repopulate the index) then I think that work is going to have to happen in a separate ticket. It's trivial to do, it's just that there's no exposed way to do it in the app right now.

io7m: > It's trivial to do, it's just that there's no exposed way to do it in the app right now.

Trivial to do in that there's an API for this, and tests already that show how it works, but there's no exposed configuration setting to do it.

io7m: I forgot to point out that deleting versions as a downgrade runs the risk of accidentally reusing a version number, and doing that makes the system unsound.

To illustrate:

1. I create some code for a new version `M`.
2. I run the CM and tell it to upgrade to version `M`.
3. I populate `M` with search documents.
4. I find out a few hours later that `M` is broken in some way, so I downgrade by setting the read/write pointers and delete the code for `M`.
5. I then write _new_ code and I reuse the version number `M`.
6. I deploy a new version of the CM and tell it to upgrade to `M`.
7. The CM sees that there's a populated index with version `M` and does nothing.

If instead, at step `5`, I used a new version number `P` where `P > M`, everything would have worked.

Version numbers cannot be reused!

core/external_search.py
```diff
@@ -123,369 +124,165 @@ class ExternalSearchIndex(HasSelfTests):
...
-                (
-                    "Index '%s' is not in series with current index '%s'. "
-                    "Confirm the base name (without version number) of both indices"
-                    "is the same."
-                )
-                % (new_index, self.works_index)
+            self._search_write_pointer = self._search_service.write_pointer_name(
+                self._revision_base_name
             )
+            self._search_migrator = SearchMigrator(
```
RishiDiwanTT, line 234: @io7m what is the purpose of having a migrator then cancelling the migration in the init method?

RishiDiwanTT: @io7m :point_up:

core/external_search.py
```diff
@@ -3627,16 +2784,49 @@ def __init__(self, *args, **kwargs):
...
+            )
+        else:
+            # We do have a migration, we must clear out the index and repopulate the index
+            self.remove_search_coverage_records()
+
+    def on_completely_finished(self):
+        super().on_completely_finished()
+        # Tell the search migrator that no more documents are going to show up.
+        target: SearchDocumentReceiverType = self.migration or self.receiver
+        target.finish()
```
tdilauro, line 2808: It doesn't look like this `on_completely_finished` will get called, because `Scripts.run`, rather than `BaseCoverageProvider.run`, is the runner here.

tdilauro (commented):
I've taken a pass through this and have added some questions and comments. 

My biggest concern is around whether we would successfully finalize a migration after a new write index was populated, since it doesn't seem like the `on_completely_finished` method would be run.

It would also be helpful to have some documentation on how a new migration would be created and performed. Is it just a matter of creating an additional revision and adding it to the `revision_directory` to create it and then running `bin/search_index_refresh` to perform the migration?

api/admin/controller/custom_lists.py
```diff
@@ -281,7 +281,12 @@ def _create_or_update_list(
         if membership_change:
             # We need to update the search index entries for works that caused a membership change,
             # so the upstream counts can be calculated correctly.
-            self.search_engine.bulk_update(works_to_update_in_search)
+            documents = self.search_engine.create_search_documents_from_works(
+                works_to_update_in_search
+            )
+            index = self.search_engine.start_updating_search_documents()
+            index.add_documents(documents)
+            index.finish()
```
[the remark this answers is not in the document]
RishiDiwanTT: The purpose of `finish()` is to indicate a migration or update of documents has occurred without error.  
There are 2 cases
- No Migration: This is just an update of documents, in which case the `refresh()` is just a speed up of an internal process that occurs [every second anyways](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html#refresh-api-desc).
- Yes Migration: In this case the read and write pointers update to point to the latest search index, if there was a problem during the migration then we wouldn't want that to occur anyways.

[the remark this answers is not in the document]
RishiDiwanTT: @io7m Do you have any additional views?

core/search/migrator.py
```diff
@@ -0,0 +1,178 @@
...
+        """Finish the migration."""
+        self._logger.info(f"Completing migration to {self._revision.version}")
+        # Make sure all changes are committed.
+        self._receiver.finish()
+        # Create the "indexed" alias.
+        self._service.index_set_populated(self._revision)
+        # Update the write pointer to point to the now-populated index.
+        self._service.write_pointer_set(self._revision)
+        # Set the read pointer to point at the now-populated index
+        self._service.read_pointer_set(self._revision)
```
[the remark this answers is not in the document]
RishiDiwanTT: Kind of inconsequential since the `refresh()` method affects indices and the `read_pointer_set` affects aliases. The index that had new documents added would get affected by the `refresh()` regardless of the alias, since the API targets the entire cluster.

core/search/migrator.py
```diff
@@ -0,0 +1,178 @@
...
+        self._service.refresh()
+        self._logger.info(f"Completed migration to {self._revision.version}")
+
+    def cancel(self) -> None:
+        """Cancel the migration, leaving the read and write pointers untouched."""
```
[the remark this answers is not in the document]
RishiDiwanTT: I've no idea why `cancel()` exists :sweat_smile: 
The one place it exists has me really confused anyways :sweat_smile: :sweat_smile: 
@io7m help?

core/search/revision_directory.py
```diff
@@ -0,0 +1,41 @@
...
+    """A directory of the supported search index schemas."""
+
+    @staticmethod
+    def _create_revisions() -> Mapping[int, SearchSchemaRevision]:
+        return dict(map(lambda r: (r.version, r), [SearchV5()]))
```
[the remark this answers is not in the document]
RishiDiwanTT: Nope, was just code that existed, will update.

core/search/document.py
```diff
@@ -0,0 +1,259 @@
...
+
+# See: https://opensearch.org/docs/latest/field-types/supported-field-types/date/
+def date() -> SearchMappingFieldTypeParameterized:
+    return SearchMappingFieldTypeParameterized("date")
+
+
+# See: https://opensearch.org/docs/latest/field-types/supported-field-types/keyword/
+def keyword() -> SearchMappingFieldTypeParameterized:
+    mapping = SearchMappingFieldTypeParameterized("keyword")
+    mapping._parameters["normalizer"] = "filterable_string"
```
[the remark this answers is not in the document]
RishiDiwanTT: Subconsciously judged the parameters property getter as being unable to update the dict values, which is of course incorrect.

core/search/document.py
```diff
@@ -0,0 +1,259 @@
...
+    t.parameters["fielddata"] = "true"
+    return t
+
+
+class SearchMappingFieldTypeObject(SearchMappingFieldType):
+    """See: https://opensearch.org/docs/latest/field-types/supported-field-types/object/"""
+
+    _properties: Dict[str, SearchMappingFieldType]
+
+    def __init__(self, type: str):
```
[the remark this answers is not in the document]
RishiDiwanTT: @io7m was there a code style reason for this?
This is [where and how](https://github.com/ThePalaceProject/circulation/blob/feature/opensearch-migrations/core/search/v5.py#L249) it is used.

core/search/v5.py
```diff
@@ -0,0 +1,299 @@
...
+        ("collapse_two_initials", r" ([A-Z]) ([A-Z])$", " $1$2"),
+    ]:
+        normalizer = dict(
+            type="pattern_replace", pattern=pattern, replacement=replacement
+        )
+        CHAR_FILTERS[name] = normalizer
+        AUTHOR_CHAR_FILTER_NAMES.append(name)
+
+    def __init__(self):
+        super().__init__(5)
```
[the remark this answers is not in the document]
RishiDiwanTT: I changed this up to not pass the `SEARCH_VERSION` as an argument, since it is expected they all define the vairable.
The parent class just does this instead.

```python
...
def __init__():
    self._version = self.SEARCH_VERSION
...
```

core/external_search.py
```diff
@@ -123,369 +125,161 @@ class ExternalSearchIndex(HasSelfTests):
...
-        if new_base_name != current_base_name:
-            raise ValueError(
-                (
-                    "Index '%s' is not in series with current index '%s'. "
-                    "Confirm the base name (without version number) of both indices"
-                    "is the same."
-                )
-                % (new_index, self.works_index)
+            migration = self._search_migrator.migrate(
+                base_name=self._revision_base_name, version=self._revision.version
```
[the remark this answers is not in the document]
RishiDiwanTT: I don't know why this is the case myself, I have the same question for @io7m [here](https://github.com/ThePalaceProject/circulation/pull/1268#discussion_r1281504329).

core/external_search.py
RishiDiwanTT: I just assumed this was in working order :sweat_smile: 
I've updated the code, and written a test for it.

core/external_search.py
```diff
@@ -795,461 +511,25 @@ def _collections():
...
+        service = self.search_service()
+        read_pointer = service.read_pointer()
+        if not read_pointer or service.is_pointer_empty(read_pointer):
+            # A read pointer does not exist, or points to the empty index
+            # This means either this is a new deployment or the first time
+            # the new opensearch code was deployed.
+            # In both cases doing a migration to the latest version is safe.
+            migration = self.start_migration()
+            if migration is not None:
+                migration.finish()
```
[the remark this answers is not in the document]
RishiDiwanTT: The coupled statement is used in a few places, but not repeated in the `ExternalSearchIndex` again.

core/external_search.py
io7m: Sorry, wasn't getting notifications for some reason. Checking!

core/search/migrator.py
```diff
@@ -0,0 +1,178 @@
...
+        self._revisions = revisions
+        self._service = service
+
+    def migrate(
+        self, base_name: str, version: int
+    ) -> Optional[SearchMigrationInProgress]:
+        """
+        Migrate to the given version using the given base name (such as 'circulation-works'). The function returns
+        an object that expects to receive batches of search documents used to populate any new index. When all
+        the batches of documents have been sent to the object, callers must call 'finish' to indicate to the search
```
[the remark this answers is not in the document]
io7m: `finish` indicates to the underlying search service that everything succeeded and all of the various pointers should be updated to point to their new values. `finish` absolutely shouldn't be called if an exception is raised at some point during the process... Won't a context manager cause `finish` to be called unconditonally?

core/external_search.py
io7m: This one is on me. The scripts and coverage providers are a complete nightmare of subclassing with random implementation methods being overridden. I couldn't tell what was calling what...

core/external_search.py
io7m: Creating a migration has side effects. We want the side effects. However, we don't actually have all of the works and search documents we'll need at this point, so we cancel the rest of the migration.

I'm not completely happy with it either, but the structure of the existing CM code meant that there had to be _some_ initialization work done up-front to allow the system to work, with the actual migration (creation and submission of search documents and the like) happening a bit later.

core/search/document.py
```diff
@@ -0,0 +1,259 @@
...
+
+class SearchMappingDocument:
+    """
+    A top-level Opensearch mapping document.
+
+    See: https://opensearch.org/docs/latest/field-types/index/
+    """
+
+    def __init__(self):
+        super().__init__()
```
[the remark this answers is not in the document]
io7m: Without it, won't the code silently become wrong as soon as someone adds some kind of side effect or code to one of the superclass constructors?

core/search/migrator.py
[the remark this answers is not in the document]
io7m: I suppose it could elect not to if there was some kind of error. I chose to be explicit.

core/search/v5.py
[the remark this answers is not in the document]
tdilauro: @io7m Yeah, that's a good point. I'd prefer a mandatory keyword here, as well. The keyword can make it more clear what this value is and the superclass can ensure that a value was passed.

core/external_search.py
tdilauro: Could we have a method -- with a descriptive name -- that just evokes those side effects and handles the `cancel` itself?

io7m: I think `initialize_versioning()` is the most literal description of what the method would actually be doing.

core/search/v5.py
```diff
@@ -18,6 +18,7 @@
 
 
 class SearchV5(SearchSchemaRevision):
+    SEARCH_VERSION = 5
```
tdilauro, line 21: Minor: Move class var after docstring.

core/search/revision_directory.py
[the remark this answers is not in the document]
tdilauro: I screwed up the code in my previous comment. I meant to use `REVISIONS` in that last `return` statement:
```python
...
from core.search.v5 import SearchV5
from core.search.v6 import SearchV6

REVISIONS = [SearchV5(), SearchV6()]

class SearchRevisionDirectory:
    """A directory of the supported search index schemas."""

    @staticmethod
    def _create_revisions() -> Mapping[int, SearchSchemaRevision]:
        return {r.version: r for r in REVISIONS}
...
```

core/external_search.py
tdilauro: @RishiDiwanTT Your changes on [Fixed the migration runner script](https://github.com/ThePalaceProject/circulation/pull/1268/commits/2984266147247972ff8d265c6f96ffd0869736c1) look like they resolve this problem.

core/search/revision_directory.py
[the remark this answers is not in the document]
RishiDiwanTT: Aah I see, you want a more verbose method of listing versions.

core/search/v5.py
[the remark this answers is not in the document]
RishiDiwanTT: How about I raise an error if it is not defined correctly.

api/admin/controller/custom_lists.py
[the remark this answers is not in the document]
RishiDiwanTT: Hmm.. so I should remove the try: finally: clause [in here](https://github.com/ThePalaceProject/circulation/blob/feature/opensearch-migrations/core/external_search.py#L2813) then.

core/search/document.py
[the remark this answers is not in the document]
RishiDiwanTT: There is no superclass for [this class](https://github.com/ThePalaceProject/circulation/pull/1268/files/1a49dd5e1654bb5fc319528b2ae2b83aceb1f7d5#diff-e9f234a9e61a28f6d7e7bab71ebeb354ff8e701fbe5c7e49a2dd6b7a3870bebbR221)

core/external_search.py
RishiDiwanTT: Is the side-effect we're looking for just the creation of the empty index? Because even if we create the versioned index, we are not updating the read and write _pointers_ unless we do a `finish()` so the `ExternalSearchIndex` object cannot use the index anyways.

api/admin/controller/custom_lists.py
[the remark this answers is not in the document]
tdilauro: Looks like this is done, so I think we can resolve this one.

tdilauro (approved):
Functionally, this seems to be in pretty good shape and I think it is safe to merge in and deploy.

One concern I have in the longer term is related the messages logged in in the various phases of activity. 

When we are initializing with the latest mapping version fully indexed, we see messages like this:
- starting migration to <library-short-name> <search-version>
- Completing migration to <search-version>
- Finishing search documents
- Finished search documents
- Completed migration to <search-version>

This is confusing, because we're not really doing a migration. These messages are just a side effect of determining whether a migration is needed.

When a new search revision is introduced, we see messages like the following when initializing:
- starting migration to <library-short-name> <new-version>
- Write pointer does not point to the desired version: SearchWritePointer(base_name='<library-short-name>', version=<old-version>) != <new-version>.
- creating index <library-short-name>-v<new-version>
- Write index is not populated.
- Cancelling migration to <new-version>

The "Cancelling migration..." message is confusing here, since some setup -- importantly, creating an empty index for the new version -- has, in fact occurred.

And when `search_index_refresh` is run, we see the following series of messages:
- starting migration to <library-short-name> <new-version>
- Write pointer does not point to the desired version: SearchWritePointer(base_name='<library-short-name>', version=<old-version>) != <new-version>.
- creating index <library-short-name>-v<new-version>
- PUT http://localhost:9200/<library-short-name>-v<new-version> [status:400 request:0.006s]
- Write index is not populated.
- Cancelling migration to <new-version>
- starting migration to <library-short-name> <new-version>
- Write pointer does not point to the desired version: SearchWritePointer(base_name='<library-short-name>', version=<old-version>) != <new-version>.
- creating index <library-short-name>-v<new-version>
- PUT http://localhost:9200/<library-short-name>-v<new-version> [status:400 request:0.006s]
- Write index is not populated.
... indexing happens ...
- Completing migration to <new-version>
- Finishing search documents.
- Finished search documents.
- Completed migration to <new-version>

core/external_search.py
RishiDiwanTT: @io7m @tdilauro before merging this in, I would like to resolve this conversation,

RishiDiwanTT: @io7m :point_up:

io7m: That's a good point. The effect was supposed to be to get things to `S6` and then stop:

https://ataxia.io7m.com/2023/06/08/search/

So the read pointer should be created if it doesn't exist. If that's not happening, then this is a bug. :sweat:

tests/core/search/test_migration_states.py
tdilauro, line 1: This is really helpful!

tests/api/test_scripts.py
```diff
@@ -407,6 +407,7 @@ def test_do_generate(
         external_search_fixture: ExternalSearchFixture,
     ):
         db = lane_script_fixture.db
+        external_search_fixture.init_indices()
```
tdilauro, line 410: Minor: space before comment, since it belongs to `class MockAcquisitionFeed`.

tdilauro (approved):
This looks good! 🎉 

Nice to have the extra side effects out and I'm happy to have the coherent set of tests.

I haven't had a chance to manually test, but this most recent round of changes should make the sequence of log messages less confusing.

I had one minor comment, but it's no big deal if it doesn't make it in.

RishiDiwanTT:
Before merging I ran the following migration scenario
- Created a V6() schema for opensearch in the codebase
- Observed starting up the app, search runs fine, on the previous version.
- Run search_index_refresh, migration starts up, some documents get indexed
- Search runs on the new index (by checking the read-pointer alias)

I then deleted the V6() schema from the codebase, now we're back on V5()
- CM started up, and kept the pointers on the old index (as expected)
- Running search_index_refresh flipped the pointers back
- Search continues to run without issue

To mimic the upgrade deployment to the new migration system
- I deleted the red/write aliases
- This broke the search functionality (as expected)
- Restarted the CM fixes this, since init script manages the creation of these aliases
- Search ran without issue thereafter

tdilauro:
🚀

PR #1269, opened 2023-07-12
Remove AdobeAuth/devices links from feeds (PP-176) 

jonathangreen:
## Description

Remove the `http://librarysimplified.org/terms/drm/rel/devices` links from the CM `/loans` and `/patrons/me` feeds. This link is no longer needed, because we removed the Adobe device activation code in https://github.com/ThePalaceProject/circulation/pull/1230.

We preserve the `drm` tag, that communicates the adobe vendor ID details.

## Motivation and Context

From JIRA: 
> I am unable to sign out of the Minotaur Test Library on either iOS or Android. iOS screenshot attached. On Android, the error manifests as a 500 error.

Looking at server logs, this is because we are getting a 500 response when requesting `/patrons/me` because the CM is trying to create the link for the `adobe_drm_devices` endpoint. 

```json
{"type": "http://librarysimplified.org/terms/problem/internal-server-error", "title": "Internal server error.", "status": 500, "detail": "Internal server error", "debug_message": "Could not build url for endpoint 'adobe_drm_devices' with values ['library_short_name']. Did you mean 'patron_devices' instead?"}
```

## How Has This Been Tested?

Tested locally, and ran CM tests. The tricky part of getting this fixed is to actually test the full workflow with the mobile apps. Once this gets merged, I'll do some testing with the mobile apps to see if they have any issues with these links being removed from the feeds.

## Checklist

- [X] I have updated the documentation accordingly.
- [X] All new and existing tests passed.

PR #1270, opened 2023-07-12
Fix log messages for logo table migration

jonathangreen:
## Description

Fix the log message in `20230706_c471f553249b_migrate_library_logo.py ` migration. One of the log messages forgot to use a fstring, so log messages were not being correctly interpolated.

## Motivation and Context

When debugging https://github.com/ThePalaceProject/circulation/pull/1269 I ran this migration, and noticed that the log messages were not being formatted as I expected.

## How Has This Been Tested?

Running migration locally.

## Checklist


tdilauro (approved):
🚀

PR #1276, opened 2023-07-17
Remove shared ODL API 🔥

jonathangreen:
## Description

This API intersects with the work in PP-11, because it uses library configuration settings to store the shared ODL configuration. We decided not to configure shared ODL on any of our circulation managers, instead relying on having any CM that needs to share a marketplace collection be on the same CM. 

## Motivation and Context

Since we are not using this code anywhere. It seems to make sense to remove it, instead of migrating it to use integration configurations. 

## How Has This Been Tested?

Running tests in CI.

## Checklist


tdilauro (approved):
This looks good! 🔥

PR #1281, opened 2023-07-19
Convert Library to use integration settings (PP-11)

jonathangreen:
## Description

Convert libraries to store their settings in a json column on the libraries table, rather then using configuration settings. See jira ticket PP-11.

Note: This PR is based on https://github.com/ThePalaceProject/circulation/pull/1276 so that one should go in first.

## Motivation and Context

This is part of the larger CM Config Settings epic: PP-4

## How Has This Been Tested?

- Running unit tests

## Checklist


RishiDiwanTT (approved):
I can't really find anything at fault here. Just wondering why are the settings added to the Library model as opposed to using the IntegrationLibraryConfiguration Model without a parent.

jonathangreen:
I decided to add the settings directly to the library table because I didn't want to overload the `IntegrationLibraryConfiguration` table too much. Since these settings are not really configurations for an integration, they didn't seem like they fit in the `IntegrationLibraryConfiguration` table.

PR #1287, opened 2023-07-20
Allow CM to start when running older CM version against new database

jonathangreen:
## Description

This adds a try block that catches the `CommandError`, logs an error, and allows the startup to continue.

## Motivation and Context

When doing some testing today, I ran into the case where the CM fails to start with a `CommandError` when trying to run against a DB that has been updated to a newer schema version. 

It is likely that the CM will fail to run in this case, but I think the right thing to do it to log an error and continue, as depending on the migration, we may still be able to run the CM. If we can't, the CM will fail somewhere else when it can't find the table it is expecting etc. 

## How Has This Been Tested?

Running unit tests.

## Checklist

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -12,6 +12,7 @@ from sqlalchemy.engine import Connection
 from sqlalchemy.orm import Session
 
 from alembic import command, config
+from alembic.util import CommandError
 from api.adobe_vendor_id import AuthdataUtility
 from api.authenticator import LibraryAuthenticator
 from api.axis import Axis360BibliographicCoverageProvider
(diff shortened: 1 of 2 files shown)


RishiDiwanTT (approved):
Code looks good :+1:
Unsure of runtime side effects for a CM~DB mismatch though.
Is it better to fail fast or run handicapped?

jonathangreen:
Started a discussion on slack to get some more feedback on this one before merging: 
https://lyrasis.slack.com/archives/C03E3JHEFK8/p1691721172507589

PR #1288, opened 2023-07-21
Audiobook time tracking

RishiDiwanTT:
## Description
New tables IdentifierPlaytimeEntries and IdentifierPlaytimes have been added.
The time tracking API has been added as `POST /playtimes/<type>/<identifier>/` .
A new environment variable `SIMPLIFIED_REPORTING_EMAIL` has been added for the temporary requirement of emailing reports via a cron job, this will require a deployment change.
Playtime aggregations run every 12 hours.
Playtime reporting occurs on the 2nd of every month.

Cannot add the api spec for now since HTTP 207 is not supported by `flask_pydantic_spec`. Bug ticket [here](https://github.com/turner-townsend/flask-pydantic-spec/issues/58).

## Motivation and Context
We need to track the total amount of time an audiobook is playing on a user’s device. The apps should send this information to a remote server every 1 minute.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-157)

## How Has This Been Tested?
Manually run the APIs and summation jobs.
Emailed a local smtp server with the CSV report.

## Checklist


core/jobs/playtime_entries.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class PlaytimeEntriesEmailReportsScript(Script):
+    def do_run(self):
+        """Send a quarterly report with aggregated playtimes via email"""
+        # 3 months prior, shifted to the 1st of the month
+        cutoff = utc_now() - timedelta(days=90)
+        cutoff = cutoff.replace(day=1).date()
+        # Until the 1st of the current month
+        until = date.today().replace(day=1)
```
tdilauro, line 66: `date.today()` returns the date for local -- rather than UTC -- time, but I think we need UTC time here.

core/jobs/playtime_entries.py
```diff
@@ -0,0 +1,116 @@
...
+            if recipient:
+                EmailManager.send_email(
+                    f"Playtime Summaries {cutoff} - {until}",
+                    receivers=[recipient],
+                    text="",
+                    attachments={f"playtime-summary-{cutoff}-{until}": temp.read()},
+                )
+            else:
+                self.log.error("No reporting email found, logging complete report.")
+                self.log.error(temp.read())
```
tdilauro, line 116: `codecov` is noting that the else case here isn't tested.

core/model/time_tracking.py
```diff
@@ -0,0 +1,110 @@
...
+
+from core.model import Base, get_one_or_create
+
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Mapped
+
+    from core.model import Identifier
+
+
+class IdentifierPlaytimeEntry(Base):
```
tdilauro, line 26: Suggestion: Since I think these entries are "identified" by more than the identifier, I think we should rename this table to `playtime_entries`.

core/model/time_tracking.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    tracking_id = Column(String(64), nullable=False)
+    processed = Column(Boolean, default=False)
+
+    identifier: Mapped[Identifier] = relationship("Identifier", uselist=False)
+
+    __table_args__ = (UniqueConstraint("identifier_id", "tracking_id"),)
+
+
+class IdentifierPlaytime(Base):
```
tdilauro, line 54: Suggestion: Since I think these entries are "identified" by more than the identifier, I think we should rename this table to `playtime_summaries`.

tdilauro (commented):
I've taken a first pass through this one, but have NOT really looked at the tests yet. More detail below, but here are a few high-level thoughts:

- We need to be able to report at the collection and library level for the CSV report, so we need that information in the database tables.
- Since the apps will use the presence of a item entry-level link with the time tracking `rel` to decide whether to perform/report time tracking, we want link to be present for only those books needing tracking to have one. So we need some collection or library/collection-level config to indicate whether time tracking should be used.

- So that we can more easily tune them over time, it would be useful to be able to set (or override) the following with script options:
  - How old a time entry timestamp should be before it is `processed` into the summary table.
  - How old a `processed==True` entry should be (i.e., how long we want to actively avoid duplicates) before it is removed from the `entries` table.

jonathangreen:
Just throwing a note on here to remind us that since https://github.com/ThePalaceProject/circulation/pull/1281 went in, and also had a DB migration, that we will need to fix the DB migration here before merging.

core/model/time_tracking.py
RishiDiwanTT: Sure

core/model/time_tracking.py
```diff
@@ -0,0 +1,110 @@
...
+    __tablename__ = "identifier_playtimes"
+
+    id = Column(Integer, autoincrement=True, primary_key=True)
+
```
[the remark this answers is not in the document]
RishiDiwanTT: For slicing reports by library and collection ok :+1:

core/jobs/playtime_entries.py
RishiDiwanTT: Oh right, silly

core/jobs/playtime_entries.py
```diff
@@ -0,0 +1,116 @@
...
+from core.util.email import EmailManager
+from scripts import Script
+
+
+class PlaytimeEntriesSummationScript(Script):
+    def do_run(self):
+        # Collect everything from one hour ago, reducing entries still in flux
+        cuttoff = utc_now() - timedelta(hours=1)
+
+        # Reap older processed entries
```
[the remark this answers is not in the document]
RishiDiwanTT: Like the cutoff time for reaping?

core/jobs/playtime_entries.py
```diff
@@ -0,0 +1,116 @@
...
+from core.model.time_tracking import IdentifierPlaytime, IdentifierPlaytimeEntry
+from core.util.datetime_helpers import utc_now
+from core.util.email import EmailManager
+from scripts import Script
+
+
+class PlaytimeEntriesSummationScript(Script):
+    def do_run(self):
+        # Collect everything from one hour ago, reducing entries still in flux
+        cuttoff = utc_now() - timedelta(hours=1)
```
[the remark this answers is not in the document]
RishiDiwanTT: cutoff time is just for setting the "newest timestamp" boundary, if a newer entry comes in for a month old timestamp we will still process it since it's processed=False.
Should we also have an "oldest timestamp" cutoff? So we are not adding data to a year old report, if such data comes in?

api/model/time_tracking.py
```diff
@@ -0,0 +1,57 @@
...
+                f"An incorrect timezone was received for a playtime ({value.tzname()})."
+            )
+            raise ValueError("Timezone MUST be UTC always")
+        value = value.replace(second=0, microsecond=0)
+        return value
+
+    @validator("seconds_played")
+    def validate_seconds_played(cls, value: int):
+        """Coerce the seconds played to a max of 60 seconds"""
+        if value > 60:
```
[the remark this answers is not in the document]
RishiDiwanTT: To keep things simple I would constrain it to 60, else we would have to add leap second validations here as well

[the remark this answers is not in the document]
tdilauro: I added an [update to the API spec](https://www.notion.so/lyrasis/Audiobook-Time-Tracking-API-Proposal-d6b9ca3cb1b849cc879950e562d6e981?pvs=4#9a59951f61af4f929387b6ee72b53975):

> The secondsPlayed value in a timeEntry should be limited to 60 seconds, even if a leap second is encountered during a tracking period.

Leap seconds rarely occur and it does not seem worthwhile to either loosen validation or add validation complexity to accommodate it.

api/opds.py
```diff
@@ -831,6 +832,19 @@ def annotate_work_entry(
                 ),
             )
 
+        if edition.medium == EditionConstants.AUDIO_MEDIUM:
```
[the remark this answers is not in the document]
RishiDiwanTT: Just to be clear I remembered this correctly, we want to hardcode this as only trigering for OPDS for Distributors for now, right?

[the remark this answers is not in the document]
tdilauro: That is correct.

docker/services/cron/cron.d/circulation
```diff
@@ -115,3 +115,9 @@ HOME=/var/www/circulation
 10 12 * * * root core/bin/run loan_notifications >> /var/log/cron.log 2>&1
 15 12 * * * root core/bin/run hold_notifications >> /var/log/cron.log 2>&1
 0 1 * * * root core/bin/run patron_activity_sync_notifications >> /var/log/cron.log 2>&1
+
+# Audiobook playtimes
+# Every 12 hours, but spaced after hour 8 to reduce job cluttering
+0 8/12 * * * root core/bin/run playtime_summation >> /var/log/cron.log 2>&1
+# On the 2nd of every month
+0 4 2 * * root core/bin/run playtime_reporting >> /var/log/cron.log 2>&1
```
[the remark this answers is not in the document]
RishiDiwanTT: The report will have overlapping boundaries, the count will be for the boundaries stated in each row of the csv.
So technically, yes, overlapping counts, but the time frame will clearly show overlapping dates.
The monthly report was an arbitrary choice by me, and should change based on actual requirements.

core/jobs/playtime_entries.py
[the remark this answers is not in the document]
RishiDiwanTT: I have added the 401 status to the API responses

core/jobs/playtime_entries.py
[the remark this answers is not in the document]
RishiDiwanTT: I've set the reap cut-off time to 1 month for now, we can change it easily whenever required.

api/opds.py
```diff
@@ -786,6 +788,25 @@ def annotate_work_entry(
                 ),
             )
 
+        # Only OPDS for Distributors should get the time tracking link
+        if (
+            edition.medium == EditionConstants.AUDIO_MEDIUM
+            and active_license_pool
+            and active_license_pool.collection.protocol
+            == ExternalIntegration.OPDS_FOR_DISTRIBUTORS
```
tdilauro, line 796: We want these entries only for books in the loan feed -- and the single entry feed we get after a borrow -- and only for loan (not hold) entries.

core/jobs/playtime_entries.py
```diff
@@ -0,0 +1,134 @@
...
+        # Fetch the unprocessed entries
+        result = self._db.query(PlaytimeEntry).filter(
+            PlaytimeEntry.processed == False,
+            PlaytimeEntry.timestamp <= cuttoff,
+        )
+        by_identifier = defaultdict(int)
+
+        # Aggregate entries per identifier-timestamp tuple
+        # Since timestamps should be on minute-boundaries the aggregation
+        # can be written to IdentifierPlaytimes directly
```
tdilauro, line 47: I think this comment needs to be updated to reflect the changes in db table names and facets.

core/util/datetime_helpers.py
```diff
@@ -60,3 +61,14 @@ def strptime_utc(date_string, format):
     if "%Z" in format or "%z" in format:
         raise ValueError(f"Cannot use strptime_utc with timezone-aware format {format}")
     return to_utc(datetime.datetime.strptime(date_string, format))
+
+
+def previous_months(number_of_months=3) -> Tuple[datetime.date, datetime.date]:
```
tdilauro, line 66: Copypasta? The `number_of_months` parameter should probably not have a default value.

core/query/playtime_entries.py
```diff
@@ -0,0 +1,77 @@
...
+from core.model.collection import Collection
+from core.model.identifier import Identifier
+from core.model.library import Library
+from core.model.time_tracking import PlaytimeEntry
+from core.util.datetime_helpers import utc_now
+
+
+class PlaytimeEntries:
+    # The oldest entry acceptable by the insert API
+    # If anything earlier arrives, we ignore it with a 401 response
```
tdilauro, line 18: I think you already caught this one, but should be 410.

core/query/playtime_entries.py
```diff
@@ -0,0 +1,77 @@
...
+            status_code = 201
+            message = "Created"
+            transaction = _db.begin_nested()
+            try:
+                if (
+                    today - entry.during_minute.date()
+                ).days > cls.OLDEST_ACCEPTABLE_ENTRY_DAYS:
+                    # This will count as a success, since we don't want to repeat the entry
+                    status_code = 401
+                    message = "Gone"
```
tdilauro, line 44: This, too, should be 410. Could we make the message this (or something similarly descriptive)?
```
"Time entry too old and can no longer be processed."
```

tests/api/test_controller_playtime_entries.py
```diff
@@ -0,0 +1,265 @@
...
+                    "seconds_played": 12,
+                },
+                {
+                    "id": "tracking-id-1",
+                    "during_minute": date_string(hour=12, minute=1),
+                    "seconds_played": 17,
+                },
+                {
+                    "id": "tracking-id-2",
+                    "during_minute": "2000-01-01T12:00Z",  # A very old entry, 401
```
tdilauro, line 50: 410 again. 😃

tests/api/test_controller_playtime_entries.py
```diff
@@ -0,0 +1,265 @@
...
+            assert len(data["responses"]) == 3
+            assert data["responses"][0] == dict(
+                id="tracking-id-0", status=201, message="Created"
+            )
+            assert data["responses"][1] == dict(
+                id="tracking-id-1", status=201, message="Created"
+            )
+            assert data["responses"][2] == dict(
+                id="tracking-id-2", status=401, message="Gone"
+            )
```
tdilauro, line 75: 410 again. (Please don't hurt me!)

tdilauro (commented):
I resolved some conversations and added a few more comments.

core/jobs/playtime_entries.py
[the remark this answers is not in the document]
tdilauro: As you noted today, this should be 410. I flagged the places where I saw 401 and also suggested that a more informative message to be associated with it (I suggested the one that I put in the API spec, but am not wedded to it).

tests/core/jobs/test_playtime_entries.py
```diff
@@ -163,15 +163,14 @@ def test_do_run(self, db: DatabaseTransactionFixture):
...
-        with (
-            patch("core.jobs.playtime_entries.csv.writer") as writer,
-            patch("core.jobs.playtime_entries.EmailManager") as email,
-            patch(
-                "core.jobs.playtime_entries.os.environ",
-                new={
-                    Configuration.REPORTING_EMAIL_ENVIRONMENT_VARIABLE: "reporting@test.email"
-                },
-            ),
+        # Horrible unbracketted syntax for python 3.8
```
tdilauro, line 245: So glad this is fixed in more recent Python versions! Absolutely hideous and confusing.

core/util/datetime_helpers.py
RishiDiwanTT: Copypasta is a synonym for coding

api/opds.py
RishiDiwanTT: Moved this to LibraryLoanAndHoldAnnotator

core/query/playtime_entries.py
```diff
@@ -0,0 +1,77 @@
...
+        today = utc_now().date()
+        for entry in data.time_entries:
+            status_code = 201
+            message = "Created"
+            transaction = _db.begin_nested()
+            try:
+                if (
+                    today - entry.during_minute.date()
+                ).days > cls.OLDEST_ACCEPTABLE_ENTRY_DAYS:
+                    # This will count as a success, since we don't want to repeat the entry
```
tdilauro, line 42: This is reported as a failure, but one that means the client should discard the corresponding entry.

RishiDiwanTT: Changed this to count as a failure

api/opds.py
```diff
@@ -1813,6 +1815,33 @@ def annotate_feed(self, feed, lane):
...
+        ):
+            feed.add_link_to_entry(
+                entry,
+                rel=LinkRelations.TIME_TRACKING,
+                href=self.url_for(
+                    "track_playtime_events",
+                    identifier_type=identifier.type,
+                    identifier=identifier.identifier,
+                    library_short_name=self.library.short_name,
+                    _external=True,
```
tdilauro, line 1841: The route we're getting the URL for here is `"/playtimes/<collection_name>/<identifier_type>/<path:identifier>"`, so the `href` should be something like:
```
                href=self.url_for(
                    "track_playtime_events",
                    collection_name=active_license_pool.collection.name,
                    identifier_type=identifier.type,
                    identifier=identifier.identifier,
                    _external=True,
                ),
```

But I think it might be better -- for a couple of reasons -- if we used the `id` of the collection, rather than the name:
- A collection name can have spaces, which add to the messiness of the URLs (though, this is also the case with identifiers, especially the id type, as well).
- A collection name can be changed at any time (for example, when we notice a typo or extra spaces in the name).

api/routes.py
```diff
@@ -672,6 +672,18 @@ def track_analytics_event(identifier_type, identifier, event_type):
     )
 
 
+@library_route(
+    "/playtimes/<collection_name>/<identifier_type>/<path:identifier>", methods=["POST"]
```
tdilauro, line 676: I might be better to use `collection_id`, rather than `collection_name`, for the first component of the path, since it is less likely to change.

RishiDiwanTT:
I did originally use collection id, but found that every other collection
based route on the CM, which are a lot, uses collection name. To keep it
consistent I also used the name.
I'm not privy as to why that's the standard though.

tdilauro:
> I did originally use collection id, but found that every other collection based route on the CM, which are a lot, uses collection name. To keep it consistent I also used the name. I'm not privy as to why that's the standard though.

I'd still suggest using the `id` here, since these URLs might be cached for a long time on the client apps and a name change might cause all of them to break. That said, the client apps should update these links in their registries whenever they fetch the `loans` feed, so maybe that's good enough.

api/opds.py
RishiDiwanTT: I'm sticking to "collection_name" for now, only because there are 14 other routes with collection_name and none with collection_id. I'm going to ask the mobile team if having the ID is a problem in any way.

RishiDiwanTT: Also this test passed because we have special `test_mode` in annotators, which is clearly flawed. I ave fixed the test, and the code.

tdilauro: Sounds good! We can also check with @jonathangreen about the collection `nsme` vs. `id` in the route. I don’t think the mobile apps will have a problem with the `id` there, since, according to the spec, they are meant to simply cache the link’s `href` for reporting, rather than parse the `href` in any way.

tdilauro (commented):
I think this is looking pretty good. I’m going to do some more testing once I wake up. And, of course, we’ll need to resolve the collection name vs. id issue.

One more fairly minor comment below.

api/opds.py
tdilauro: @RishiDiwanTT Adding an update from one of my comments in [the Slack thread you started](https://lyrasis.slack.com/archives/CCY3PH8JD/p1691042621570839)...

> This could be especially problematic when a user has been listening offline and the loan expires before they are back online. In that case, they would not get a tracking link update because that book would no longer be in the loans feed. So, any cached timeEntrys would be stuck in limbo.

api/model/time_tracking.py
```diff
@@ -0,0 +1,62 @@
...
+        elif value < 0:
+            logging.getLogger("TimeTracking").warning(
+                "Less than 0 seconds was received for a minute playtime."
+            )
+            value = 0
+        return value
+
+
+class PlaytimeEntriesPost(CustomBaseModel):
+    time_entries: List[PlaytimeTimeEntry] = Field(description="A List of time entries")
```
tdilauro, line 47: The `book_id` and `library_id` from the spec are missing here.

api/routes.py
```diff
@@ -675,10 +676,16 @@ def track_analytics_event(identifier_type, identifier, event_type):
...
+@api_spec.validate(
+    resp=SpecResponse(HTTP_200=PlaytimeEntriesPostResponse),
+    body=PlaytimeEntriesPost,
+    tags=["analytics"],
+)
 @has_library
 @allows_auth
 @returns_problem_detail
 def track_playtime_events(collection_name, identifier_type, identifier):
+    """The actual response type is 207, but due to a bug in flask-pydantic-spec we must document it as a 200"""
```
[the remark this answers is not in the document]
tdilauro: Never mind. I see this is needed to get this info into the api docs.

api/model/time_tracking.py
tdilauro: I pushed a commit to fix this.

tdilauro (approved):
This looks good to go! 🎈🎉 

It'll be good to start testing against this with the apps!

tdilauro:
@RishiDiwanTT I'm going to go ahead and merge this, so that we can get it deployed out for testing. Thanks for all your work on this!

PR #1302, opened 2023-08-01
Refactor CirculationApi to use IntegrationRegistry for lookups (PP-95)

jonathangreen:
## Description

Refactor `CirculationApi` so we don't have two sources API implementation look-ups. This also includes some class hierarchy changes to `BaseCirculationAPI` to simplify the class structure there.

## Motivation and Context

This was done as the start of the work for PP-95.

## How Has This Been Tested?

Running tests locally.

## Checklist


RishiDiwanTT (approved):
Looks Good :+1:

PR #1305, opened 2023-08-03
Bump cryptography to 41.0.3

jonathangreen:
## Description

Update cryptography to 41.0.3.

## Motivation and Context

Fixes: https://github.com/ThePalaceProject/circulation/security/dependabot/22

It appears that dependabot cannot do this update itself because we are running into this issue https://github.com/dependabot/dependabot-core/issues/7389. 

So this PR updates this dependency manually for now.

## How Has This Been Tested?

Running CI

## Checklist


RishiDiwanTT (approved):
Looks good :+1:

PR #1306, opened 2023-08-03
Resolve admin UI tag to a version via jsdelivr api (PP-164)

jonathangreen:
## Description

Instead of just using the tag that we pass via the environment, we resolve that tag to a particular version. This has two benefits:
- The `version.json` endpoint can correctly report the version
- Browser cache knows there is a change, because the URL changes

The result of this lookup is cached, so it will only happen on startup, and it will only happen if a tag is specified via ENV, otherwise we just directly use the version.

## Motivation and Context

Eventually it would be nice to publish a 'next' tag for the admin UI so changes are deployed on merge.

This change helps set us up for PP-164

## How Has This Been Tested?

Running tests

## Checklist

### DIFF
--- a/api/admin/config.py
+++ b/api/admin/config.py
@@ -1,7 +1,12 @@
+import logging
 import os
 from enum import Enum
 from typing import Optional
 from urllib.parse import urljoin
+
+from requests import RequestException
+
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 🎸

PR #1310, opened 2023-08-07
Refactor AdminCirculationManagerController into mixin class (PP-94)

jonathangreen:
## Description

The PR breaks out `AdminPermissionsControllerMixin`, which replaces `AdminCirculationManagerController`. It updates existing classes that depend on `AdminCirculationManagerController` to use the new mixin class, and updates classes that don't need the functionalities of the base class to just use the mixin class.

## Motivation and Context

`AdminCirculationManagerController` sits in the middle of a fairly nested class hierarchy:
`AdminCirculationManagerController` -> `CirculationManagerController` -> `BaseCirculationManagerController`

The refactored controller classes that implement integration configuration don't need all the functionality of this class hierarchy, they only need a couple methods to check permissions that are part of `AdminCirculationManagerController`.

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/api/admin/controller/base.py
+++ b/api/admin/controller/base.py
@@ -17,7 +17,6 @@ from api.admin.problem_details import (
     INVALID_CSRF_TOKEN,
 )
 from api.config import Configuration
-from api.controller import CirculationManagerController
 from core.model import (
     Admin,
     AdminRole,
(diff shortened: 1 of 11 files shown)


RishiDiwanTT (approved):
Looks good :+1:

PR #1315, opened 2023-08-08
Truncate edition.author and edition.sort_author values to prevent fro…

dbernstein:
…m breaking the postgresql indices using this field.

Resolves: https://ebce-lyrasis.atlassian.net/browse/PP-189

## Description
This update ensures that the edition.author and edition.sort_author fields do cause the postgresql index row size to be exceeded by truncating them at 1024 characters. 

I considered changing the index to full text or storing author information in a separately indexed table.  However it seemed overkill considering 1) long author names are rare, 2) when the author names are extremely long, it is usually due to the fact that the book is some kind of compilation of works that contains many many authors and therefore is not very readable for end users, and 3) the full list of contributors is still saved in the database,  so we are not losing information.  The author and sort_author fields appear to be used for indexing and display purposes.

I'm open to approaching this issue in another way if this solution is not acceptable.

## Motivation and Context

## How Has This Been Tested?
Unit test has been added that confirms that the issue has been addressed.  When I ran the test before fixing the code, I was able to see the same index related error that we were seeing in the ticket.  When I implemented the fix, the test passed.

## Checklist


core/model/edition.py
```diff
@@ -724,6 +724,17 @@ def calculate_author(self):
             sort_author = " ; ".join(sorted(sort_names))
         else:
             sort_author = self.UNKNOWN_AUTHOR
+
+        def truncate_string(mystr: str):
+            if len(mystr) > 1024:
```
tdilauro, line 729: It would helpful to use a constant here instead of hard coding 1024 in multiple places. We could do that either on the `Edition` class or a higher-level constant associated with PostgreSQL. This value is really tied to a limit on PG B-tree index row max length, so should probably be named accordingly.

tdilauro, line 729: How did we come to the value `1024` here. The error message mentions `2704`.

```
This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (
raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(psycopg2.errors.ProgramLimitExceeded) index row size 3208 exceeds btree version 4 maximum 2704 for index "ix_editions_author"
```

core/model/edition.py
```diff
@@ -724,6 +724,17 @@ def calculate_author(self):
...
+                return mystr[: (1024 - 3)] + "..."
+            else:
+                return mystr
+
+        # Very long author and sort_author strings can cause the limits of one or more of the
+        # underlying postgres indices.  Therefore,  we limit their size here.
+        author = truncate_string(author)
```
tdilauro, line 736: Do we need to truncate the `author` value here? Or just the `sort_author`?

core/model/edition.py
```diff
@@ -724,6 +724,17 @@ def calculate_author(self):
...
             sort_author = self.UNKNOWN_AUTHOR
+
+        def truncate_string(mystr: str):
+            if len(mystr) > 1024:
+                return mystr[: (1024 - 3)] + "..."
+            else:
+                return mystr
+
+        # Very long author and sort_author strings can cause the limits of one or more of the
+        # underlying postgres indices.  Therefore,  we limit their size here.
```
jonathangreen, line 735: Can you clean up this comment a bit? I think maybe you missed a word or two...

Maybe add a bit of info about the error we were seeing from PG.

core/model/edition.py
jonathangreen, line 729: Is there a rationale for choosing 1024 here? Is this the max allowed by a index?

tests/core/models/test_edition.py
```diff
@@ -325,6 +325,20 @@ def test_calculate_presentation_author(self, db: DatabaseTransactionFixture):
...
+    def test_calculate_presentation_very_long_author(
+        self, db: DatabaseTransactionFixture
+    ):
+        authors = []
+        for i in range(0, 500):
+            author, ignore = db.contributor(sort_name=f"AuthorLast{i}, AuthorFirst{i}")
+            authors.append(author.sort_name)
+        wr = db.edition(authors=authors)
+        author, sort_author = wr.calculate_author()
+        wr.calculate_presentation()
```
tdilauro, line 337: To ensure that this test is valid, I think we need an assertion on the lengths before they are truncated, in addition to the ones after. Probably should check a couple of different scenarios, as well:
- length > max length: assert truncated to max length and ends in "..."
- length = max length (and doesn't end in "..."): assert truncated to max length and doesn't end in ...
- length < max length: assert truncated == original

tests/core/models/test_edition.py
```diff
@@ -325,6 +325,20 @@ def test_calculate_presentation_author(self, db: DatabaseTransactionFixture):
...
+        db.session.commit()
+        assert len(wr.author) == 1024
+        assert len(wr.sort_author) == 1024
```
tdilauro, line 340: These could use the aforementioned constant, as well.

tdilauro (commented):
I think this will work fine, but I have a couple of suggestions.

Also, I think this PR is really addressing the issue on [PP-188](https://ebce-lyrasis.atlassian.net/browse/PP-188), rather than the one currently listed in the main PR comment.

core/model/edition.py
dbernstein: I wasn't sure if the 2704 covered just the author (or sort_author) or included other metadata.  So I chose 1024 as a size large enough to include the complete author name for almost all the books, but a safely limited value in case there are other components of the index size calculation that I'm not aware of.

dbernstein: see comment below.

core/model/edition.py
dbernstein: both author and sort_author are indexed so we'll run into the same problem.

tests/core/models/test_edition.py
dbernstein: I will fix.

dbernstein:
@jonathangreen  and @tdilauro : I think this one should be good to go.  We could truncate less, but I think 1024 is probably more than enough for all practical purposes.

core/model/edition.py
```diff
@@ -49,6 +49,22 @@ class Edition(Base, EditionConstants):
...
+    # DETAIL:  Index row references tuple (48187,9) in relation "editions".
+    # HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
+    #
+    # On rare occasions the author (and sort_author) fields can contain a concatenated list of a
+    # large number of authors which breaks the index and causes failures.  What exactly that threshold is
+    # I am not entirely certain.  It appears that 2704 is the size that broke the 1/3 of a buffer page
+    # limit. However, I'm not entirely sure if the buffer page can always be assumed to be a fixed length.
+    # Also, it is not clear to me what other bytes are included in the index row.  Therefore, I have chosen
+    # a value that is both large enough to ensure little or no information is hidden from the end user in most
+    # cases and small enough that we never have to worry about this causing indexing failures in the future.
```
tdilauro, line 65: It seems like we should be able to validate this with a test case. We could use the constant below to construct an author of the appropriate length.

tdilauro: Also, what are we losing when truncate here? If we have a long list of authors, it seems like we might lose the ability to display any beyond the truncation length. If that's the case, it seems like we should do the least harm.

core/model/edition.py
tdilauro: I wonder why we index author. It would be nice if we didn't. That way we'd need only truncate the field that's built specifically for sorting.

tests/core/models/test_edition.py
tdilauro: Do we have a test to ensure that the ellipses does not get added to strings with length less than or equal to the `Edition.SAFE_AUTHOR_FIELD_LENGTH_TO_AVOID_PG_INDEX_ERROR`?

dbernstein:
I'll add that.

tests/core/models/test_edition.py
```diff
@@ -332,12 +332,48 @@ def test_calculate_presentation_very_long_author(
...
+        author, sort_author = wr.calculate_author()
+        wr.calculate_presentation()
+        db.session.commit()
+
+        def do_check(original_str: str, truncated_str: str):
+            assert truncated_str == original_str
+            assert not truncated_str.endswith("...")
+            assert (
+                len(original_str)
+                < Edition.SAFE_AUTHOR_FIELD_LENGTH_TO_AVOID_PG_INDEX_ERROR
```
tdilauro, line 373: Should this be `<=` on l.372?

Probably good to also ensure that the original does not already end in ellipses, as we did above:
```
    assert not original_str.endswith("...")
```

tests/core/models/test_edition.py
dbernstein: will add.

dbernstein:
@tdilauro :  I did some experiments and realized that the index row is related to the author field length but indirectly.  So an author field of 13500 characters produces an index size of 2800.  So I adjusted the max safe size to 13000 which should be plenty to all intents and purposes.

tdilauro:
> @tdilauro : I did some experiments and realized that the index row is related to the author field length but indirectly. So an author field of 13500 characters produces an index size of 2800. So I adjusted the max safe size to 13000 which should be plenty to all intents and purposes.

Oh, that's interesting.

> I was reluctant to take the index off the author field because I am unsure of what side effects that might cause.

Yeah, that makes sense. To be clear, I was not proposing that you remove it (for the same reasons you cite). I was just wondering why we needed to index it.

tdilauro:
@dbernstein Have you had a chance to test this fix against the collection on the Jira ticket? It would be good to know if that collection can be successfully imported.

dbernstein:
@tdilauro: I will test it - I did not try that yet.

dbernstein:
I'm running it now - we'll see how it goes.

dbernstein:
I tried the single item feed Michael suggested:  https://catalog.biblioboard.com/opds/items/6132cc3e-1945-11ee-bd01-0a9b31268bf5

It appears that it breaks somewhere between 2600 and 2700 characters in the author field.  Which is strange because the unit tests allow for a much larger number of characters.  But it may have something to do with the algorithm for generating the index row.  I'll update it accordingly.

dbernstein:
@tdilauro :  I updated the tests to mirror the conditions we were seeing in the feed data that produced the error.  The safe length has been updated accordingly.

tests/core/models/test_edition.py
```diff
@@ -325,6 +328,67 @@ def test_calculate_presentation_author(self, db: DatabaseTransactionFixture):
...
         assert "Accumulator, Kelly ; Bitshifter, Bob" == wr.sort_author
 
+    def test_calculate_presentation_very_long_author(
+        self, db: DatabaseTransactionFixture
+    ):
+        authors = []
+
+        # author names should be unique and not similar to ensure that the
+        # test mirrors the types of long author lists we'd expect in real data.
+        def generate_random_random_author():
```
tdilauro, line 338: Did you mean "...random_random..." in this function name?

tests/core/models/test_edition.py
```diff
@@ -325,6 +328,67 @@ def test_calculate_presentation_author(self, db: DatabaseTransactionFixture):
...
+            return "".join(
+                random.choices(
+                    string.ascii_uppercase + string.ascii_lowercase + string.digits,
+                    k=25,
+                )
+            )
+
+        for i in range(0, 500):
+            author, ignore = db.contributor(sort_name=generate_random_random_author())
```
tdilauro, line 347: Authors usually have at least a given name and a family name. If we're trying to hew closely to what we see in the real work, should this be:

```
author, ignore = db.contributor(
    ", ".join(
        sort_name=generate_random_random_author(),
        sort_name=generate_random_random_author(),
    )
)

File: tests/core/models/test_edition.py [hunk 10]
```diff
@@ -325,6 +328,67 @@ def test_calculate_presentation_author(self, db: DatabaseTransactionFixture):
...
+        self, db: DatabaseTransactionFixture
+    ):
+        authors = []
+
+        # author names should be unique and not similar to ensure that the
+        # test mirrors the types of long author lists we'd expect in real data.
+        def generate_random_random_author():
+            return "".join(
+                random.choices(
+                    string.ascii_uppercase + string.ascii_lowercase + string.digits,
```
tdilauro, line 341: Minor, but do we want digits in the random author names? Or are we just adding some more randomness here?

tests/core/models/test_edition.py
```diff
@@ -325,6 +328,67 @@ def test_calculate_presentation_author(self, db: DatabaseTransactionFixture):
...
+            )
+            assert truncated_str.endswith("...")
+            assert not original_str.endswith("...")
+            assert (
+                len(original_str)
+                > Edition.SAFE_AUTHOR_FIELD_LENGTH_TO_AVOID_PG_INDEX_ERROR
+            )
+
+        do_check(untruncated_authors, wr.author)
+        do_check(untruncated_authors, wr.sort_author)
```
tdilauro, line 368: On l. 368, I wouldn't expect this test to pass because it's comparing authors to sort authors: `do_check(untruncated_authors, wr.sort_author)`. But maybe that's got something to do with my comment on l. 347, above.

tdilauro (commented):
A couple of issues with the main PR comment: 
- Mentions 13,000 characters as the safe length, but the code itself uses 2650.
- Still mentions two different Jira issues, one of which it does not resolve. PP-188 is the issue this one resolves.

A few more comments below, some of which I think are minor.

tests/core/models/test_edition.py
dbernstein: It's not that.  It passes because the author name is set to the sort_author value if it is not specified when creating the contributor (https://github.com/ThePalaceProject/circulation/blob/main/tests/fixtures/database.py#L431).

tests/core/models/test_edition.py
dbernstein: I wanted to add maximal randomness.

tests/core/models/test_edition.py
dbernstein: Fixed.

tests/core/models/test_edition.py
dbernstein: Fixed.

core/model/edition.py
```diff
@@ -724,6 +739,20 @@ def calculate_author(self):
...
         else:
             sort_author = self.UNKNOWN_AUTHOR
+
+        def truncate_string(mystr: str):
+            if len(mystr) > self.SAFE_AUTHOR_FIELD_LENGTH_TO_AVOID_PG_INDEX_ERROR:
+                return (
+                    mystr[: (self.SAFE_AUTHOR_FIELD_LENGTH_TO_AVOID_PG_INDEX_ERROR - 3)]
+                    + "..."
+                )
+            else:
```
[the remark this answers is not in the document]
dbernstein: Fixed.

tests/core/models/test_edition.py
dbernstein: No - I'll fix.

tests/core/models/test_edition.py
tdilauro: That can be confusing when trying to understand the test.

tests/core/models/test_edition.py
tdilauro: Could be handy to add a comment to that effect.

tests/core/models/test_edition.py
tdilauro: You might consider using a separate variable (or a shared variable that is more clearly general between sorted and non-sorted) or add a comment to clarify what's happening.

tdilauro (approved):
A couple more minor comments in existing threads, but otherwise this looks good to go. 🚀

PR #1317, opened 2023-08-09
Disable mypy "method-assign" errors in tests module

jonathangreen:
## Description

Add mypy override for our tests module, that disables the "method-assign" error, and remove existing comments that are disabling this check.

## Motivation and Context

In our tests, we often overwrite methods on classes with MagicMock to mock out some behaviour. Mypy doesn't like this and throws a "method-assign" error. Until now, we've been adding `# type: ignore[method-assign]` comments to disable this where we need to use mocks. 

This PR instead removes those existing comments, and disables this check for the tests package. It is still enabled for our normal code, where we should not be mocking like this.

This is the recommended work-around from the mypy issue here: https://github.com/python/mypy/issues/2427

## How Has This Been Tested?

- Running mypy locally

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,15 @@ warn_unused_configs = true
 warn_unused_ignores = true
 
 [[tool.mypy.overrides]]
+# In our tests, we often overwrite methods on classes to mock out behavior.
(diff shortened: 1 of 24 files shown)


tdilauro (approved):
#️⃣🔥

PR #1318, opened 2023-08-09
Ensures that OPDS For Distributors Monitor results are displayed in t…

dbernstein:
…he CM Admin.

## Description
This update overrides the SERVICE_NAME property of the OPDS for Distributors Import Monitor.  With a unique service name, the results will show up in their own "OPDS For Distributors Import Monitor" sublist.  I believe the filtering logic of the monitor reports was preventing the OPDS for Distributors from showing up. 

## Motivation and Context

## How Has This Been Tested?
I tested this locally by running ./bin/opds_for_distributors_import_monitor and verifying that the results appeared in the UI as expected. It wasn't clear to me how to add a unit test to cover the fix.

## Checklist

- [ x] I have updated the documentation accordingly.

### DIFF
--- a/api/opds_for_distributors.py
+++ b/api/opds_for_distributors.py
@@ -371,6 +371,7 @@ class OPDSForDistributorsImportMonitor(OPDSImportMonitor):
     """
 
     PROTOCOL = OPDSForDistributorsImporter.NAME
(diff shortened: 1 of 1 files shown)


dbernstein:
I'm closing this one because I don't believe it actually solves the problem.

PR #1319, opened 2023-08-10
Ensures that OPDS For Distributors Monitor results are captured even …

dbernstein:
…when there are database

failures during import.

## Description

This update ensures that a timestamp record is saved even if there is a SQL failure during the import.

## Motivation and Context

## How Has This Been Tested?
There is a unit test that verifies that a timestamp record is created even when a sql error occurs.

## Checklist


core/monitor.py
```diff
@@ -176,6 +176,8 @@ def run(self):
         # to this point. It's the job of the subclass to decide where
         # to go from here.
         timestamp_obj = self.timestamp()
+        # commit it now in case something fails during the run.
+        self._db.commit()
```
RishiDiwanTT, line 180: Is this commit  required if we are forcing the timestamp object in the `except` block?

RishiDiwanTT: If there is a failure expected later on, do we want partial data saved?

RishiDiwanTT (commented):
Just one query, looks good otherwise :+1:

core/monitor.py
dbernstein: That's a good point.  Good catch @RishiDiwanTT .  I'll remove the commit.

dbernstein:
@RishiDiwanTT :  this one  should be good to go pending your approval.

RishiDiwanTT (approved):
:+1:

PR #1320, opened 2023-08-10
Fix tests that are failing at certain times of the day

jonathangreen:
## Description

Convert two tests to use local time, rather then UTC, since the production code compares the dates with local time.

## Motivation and Context

Tonight when working on the CM, suddenly a test started failing for me. I've seen this when working later in the evening before. After doing some investigation, it was caused by using UTC in the dates in tests. This means that every day this test will fail when UTC and the local time zone are a different day 😓

## How Has This Been Tested?

- Running tests in CI

## Checklist


RishiDiwanTT (approved):
:+1:

PR #1321, opened 2023-08-10
Migrate discovery services to integration config (PP-94)

jonathangreen:
## Description

Convert the library registry to use `IntegrationConfiguration` rather then `ExternalIntegration`. The library registration data wasn't a great fit for LibraryIntegrationConfiguration, so that data is stored in its own table `discovery_service_registrations`.

As part of the conversion, I moved the files for the registry into a `discovery` module and fully type hinted them, so it was a bit easier to sort out what was happening. As part of the type hinting process, I added hints to `core/util/problem_detail.py`, which necessitated some changes in files unrelated directly to this PR.

## Motivation and Context

See JIRA ticket: PP-94.
This is part of the larger CM Config Settings epic: PP-4

## How Has This Been Tested?

- Running tests locally and in CI
- Light testing with local CM instance. Loading UI, testing registration functionality.
- Tested migration against some of our production DB, to make sure that it runs correctly.

## Checklist


core/util/http.py
```diff
@@ -430,52 +434,55 @@ def debuggable_request(
...
-                % (response_content)
-            )
-            problem.debug_message = response_content
-            return problem
+            try:
+                problem_detail = INTEGRATION_ERROR.detailed(
+                    f"Remote service returned a problem detail document: '{response.text}'"
+                )
+                problem_detail.debug_message = response.text
+                raise ProblemError(problem_detail=problem_detail)
```
RishiDiwanTT, line 476: This class is used pretty much everywhere, but I don't think all the places where it is used handle ProblemErrors. Would this be problematic?

api/admin/controller/discovery_services.py
```diff
@@ -1,153 +1,147 @@
...
+        if not integration:
             return MISSING_SERVICE
-        service = registry.integration
-        if protocol != service.protocol:
-            return CANNOT_CHANGE_PROTOCOL
-        return service
-
-    def process_delete(self, service_id):
-        return self._delete_integration(service_id, ExternalIntegration.DISCOVERY_GOAL)
+        self._db.delete(integration)
```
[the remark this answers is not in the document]
jonathangreen: Good idea, thanks @RishiDiwanTT. I added `IntegrationSettingsController.delete_service` for this case and updated `DiscoveryServicesController` and `PatronAuthServicesController` to use it.

core/util/http.py
jonathangreen: The HTTP class is used pretty much everywhere, but from what I can tell, the function modified here to raise a `ProblemError` `process_debuggable_response` is only called from `debuggable_get` and `debuggable_post`. And those functions are only called in the `OpdsRegistrationService` code, which does handle the resulting `ProblemError`.

So I don't think these changes will be problematic, but let me know if you see uses of the `debuggable_*` functions I missed.

RishiDiwanTT (approved):
Looks good :+1:

PR #1322, opened 2023-08-10
Fix bug with settings in MARCExporter (PP-306)

jonathangreen:
## Description

In https://github.com/ThePalaceProject/circulation/pull/1303 I mistakenly updated a call to `settings` on an `ExternalIntegration` to a call to `settings_dict`.

This PR fixes the issue, and adds a test for the function that the issue occurred in, since it wasn't tested before.

## Motivation and Context

Minotaur failed to start after updating overnight, due to https://github.com/ThePalaceProject/circulation/pull/1303 being merged.

## How Has This Been Tested?

Local testing with Minotaur DB.

## Checklist

### DIFF
--- a/core/marc.py
+++ b/core/marc.py
@@ -616,7 +616,7 @@ class MARCExporter:
             # Only add an integration to choose from if it has a
             # MARC File Bucket field in its settings.
             configuration_settings = [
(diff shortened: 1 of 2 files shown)


RishiDiwanTT (approved):
:+1: Looks good

PR #1325, opened 2023-08-15
Allow "is not equal" operator in distributor query clauses.

dbernstein:
## Description
This PR enables the use of the "is not equal to" or `neq` operator when constructing query clauses with the `distributor` field in the UI.  Under the covers the distributor maps to the data_source field.  I've added a test to verify that this works as well as modified a test that broke as a result of my changes.

I tested the functionality manually and verified that it works.  It wasn't clear to me from the code why data_source was restricted to the 'eq' operator only.   

So while the fix appears to work, I'm not 100% confident that there wasn't a good reason to restrict the `neq` operator.

@RishiDiwanTT :  Perhaps you have some insight into why datasource was heretofore restricted to `eq`?

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-220

## How Has This Been Tested?
See above.

## Checklist

### DIFF
--- a/core/external_search.py
+++ b/core/external_search.py
@@ -2050,7 +2050,7 @@ class JSONQuery(Query):
         "licensepools.availability_time": dict(path="licensepools", **_LONG_TYPE),
         "licensepools.collection_id": dict(path="licensepools", **_LONG_TYPE),
         "licensepools.data_source_id": dict(
-            path="licensepools", ops=[Operators.EQ], **_LONG_TYPE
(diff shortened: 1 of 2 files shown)


RishiDiwanTT (approved):
The reason was that I didn't think of it :sweat_smile: 
The restriction is actually for "exact" operators like eq and neq because we test against a numerical ID, where value of the ID has no meaning on the number line.
So the other comparative operators, like gt and lt, the operators don't have any semantic meaning for the DB generated ID.

PR #1336, opened 2023-08-22
Correctly handle list and menu collection settings.

tdilauro:
## Description

- Collection `list` and `menu` setting types are correctly handled.
- If a value for an existing setting is not present in the submitted form, then the previous value of the setting is used. This should improve the chances that a setting is not lost if there is a programming error.

## Motivation and Context

- `list` and `menu` type settings were not being captured from the form data.

[Jira [PP-354](https://ebce-lyrasis.atlassian.net/browse/PP-354)]

## How Has This Been Tested?

- ["Test & Build" CI checks](https://github.com/ThePalaceProject/circulation/actions/runs/5927339898) passed for this branch.

## Checklist

[PP-354]: https://ebce-lyrasis.atlassian.net/browse/PP-354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/admin/controller/collection_settings.py
+++ b/api/admin/controller/collection_settings.py
@@ -1,5 +1,5 @@
 import json
-from typing import Dict, List
+from typing import Any, Dict, List, Optional
 
 import flask
 from flask import Response
@@ -20,6 +20,7 @@ from api.admin.problem_details import (
     NO_PROTOCOL_FOR_NEW_SERVICE,
     NO_SUCH_LIBRARY,
     PROTOCOL_DOES_NOT_SUPPORT_PARENTS,
+    PROTOCOL_DOES_NOT_SUPPORT_SETTINGS,
     UNKNOWN_PROTOCOL,
 )
 from api.integration.registry.license_providers import LicenseProvidersRegistry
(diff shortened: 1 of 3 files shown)


api/admin/controller/collection_settings.py
```diff
@@ -315,10 +320,14 @@ def process_settings(self, settings, collection):
...
+            return settings_class
+        if settings_class is None:
+            return PROTOCOL_DOES_NOT_SUPPORT_SETTINGS
         collection_settings = {}
         for setting in settings:
-            key = setting.get("key")
-            value = flask.request.form.get(key)
+            key = setting["key"]
+            value = self._extract_form_setting_value(setting, flask.request.form)
             if key == "external_account_id":
```
tdilauro, line 331: @RishiDiwanTT A question about the `external_account_id`: it is a field on `ODLSettings`/`ODL2Settings`, but is being set only on the collection, not the integration settings. Is that the intention?

api/admin/controller/collection_settings.py
```diff
@@ -334,16 +343,24 @@ def process_settings(self, settings, collection):
...
+            elif value is not None:
                 # Only if the key was present in the request should we add it
                 collection_settings[key] = value
+            else:
+                # Keep existing setting value, when present, if a value is not specified.
+                # This can help prevent accidental loss of settings due to some programming errors.
+                if key in collection.integration_configuration.settings_dict:
+                    collection_settings[
+                        key
+                    ] = collection.integration_configuration.settings_dict[key]
```
jonathangreen, line 355: Are we reasonably sure that the admin UI always supplies a value for a blank setting? Want to make sure that we are not introducing a bug here where we will refuse the clear empty fields.

jonathangreen (approved):
This looks good to me. 

I'd like to replace the settings logic here with the `ProcessFormData` class, but that is more work and can be handled once we get to PP-313.

api/admin/controller/collection_settings.py
tdilauro: Good point! I will verify and remove this case, if it does not behave correctly.

tdilauro: Verified.

PR #1340, opened 2023-08-23
PP-129 OPDS 2 with acquisition links

RishiDiwanTT:
This PR should be reviewed in tandem with the [Feed Refactor PR](https://github.com/ThePalaceProject/circulation/pull/1308).
## Description
Acquisition links for the OPDS 2 feeds as per the [De Marque document](https://docs.google.com/document/d/1tRy8uh4ckC9WDpjKXLedg87FDwsT4syv-Jvbp5dFFMk/edit).
This PR also adds the ability for certain feeds to be served as either XML or JSON based on the requests `content-type`.
This PR was also used to complete the OPDS2 serializer which was left out of #1308 .

Example
```
      {
          "href": "http://localhost:6500/VT0014/works/Overdrive%20ID/0e20dfec-501f-4b5d-926f-dd712aa39160/borrow",
          "rel": "http://opds-spec.org/acquisition/borrow",
          "type": "application/atom+xml;type=entry;profile=opds-catalog",
          "properties": {
            "availability": {
              "state": "available"
            },
            "indirectAcquisition": [
              {
                "type": "application/vnd.adobe.adept+xml",
                "child": [
                  {
                    "type": "application/pdf"
                  }
                ]
              },
              {
                "type": "application/vnd.overdrive.circulation.api+json;profile=ebook"
              },
              {
                "type": "application/atom+xml;type=entry;profile=opds-catalog",
                "child": [
                  {
                    "type": "text/html;profile=\"http://librarysimplified.org/terms/profiles/streaming-media\""
                  }
                ]
              }
            ]
          }
        }
```

#### Missing
The `authenticate` property has not been added, since we are currently providing the authentication document link as part of the feed links. Wasn't sure is we wanted to change that behaviour.

## Motivation and Context
[JIRA-129](https://ebce-lyrasis.atlassian.net/browse/PP-129) and [JIRA-78](https://ebce-lyrasis.atlassian.net/browse/PP-78)

## How Has This Been Tested?
Manually checked 

## Checklist

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -961,6 +961,7 @@ class OPDSFeedController(CirculationManagerController):
         )
         return feed.as_response(
             max_age=int(max_age) if max_age else None,
+            requested_content_type=flask.request.content_type,
         )
 
     def navigation(self, lane_identifier):
@@ -1464,7 +1465,15 @@ class LoanController(CirculationManagerController):
                 )
 
         # Then make the feed.
-        return OPDSAcquisitionFeed.active_loans_for(self.circulation, patron)
+        feed = OPDSAcquisitionFeed.active_loans_for(self.circulation, patron)
+        response = feed.as_response(
+            max_age=0, private=True, requested_content_type=flask.request.content_type
+        )
+
+        last_modified = patron.last_loan_activity_sync
+        if last_modified:
+            response.last_modified = last_modified
+        return response
 
     def borrow(self, identifier_type, identifier, mechanism_id=None):
         """Create a new loan or hold for a book.
--- a/api/controller.py
+++ b/api/controller.py
@@ -25,7 +25,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from api.authentication.access_token import AccessTokenProvider
 from api.model.patron_auth import PatronAuthAccessToken
 from api.model.time_tracking import PlaytimeEntriesPost, PlaytimeEntriesPostResponse
-from api.opds2 import OPDS2NavigationsAnnotator, OPDS2PublicationsAnnotator
+from api.opds2 import OPDS2NavigationsAnnotator
 from api.saml.controller import SAMLController
 from core.analytics import Analytics
 from core.app_server import ApplicationVersionController
@@ -1306,23 +1306,22 @@ class OPDS2FeedController(CirculationManagerController):
         params: FeedRequestParameters = self._parse_feed_request()
         if params.problem:
             return params.problem
-        annotator = OPDS2PublicationsAnnotator(
-            flask.request.url, params.facets, params.pagination, params.library
-        )
         lane = self.load_lane(None)
+        annotator = self.manager.annotator(lane, params.facets)
         max_age = flask.request.args.get("max_age")
-        feed = AcquisitonFeedOPDS2.publications(
+        feed = OPDSAcquisitionFeed.page(
             self._db,
+            lane.display_name,
+            flask.request.url,
             lane,
+            annotator,
             params.facets,
             params.pagination,
             self.search_engine,
-            annotator,
-            max_age=int(max_age) if max_age is not None else None,
         )
-
-        return Response(
-            str(feed), status=200, headers={"Content-Type": annotator.OPDS2_TYPE}
+        return feed.as_response(
+            requested_content_type="json",
+            max_age=int(max_age) if max_age is not None else None,
         )
 
     def navigation(self):
--- a/api/controller.py
+++ b/api/controller.py
@@ -1323,7 +1323,7 @@ class OPDS2FeedController(CirculationManagerController):
             self.search_engine,
         )
         return feed.as_response(
-            mime_types=["application/opds+json"],  # Force the type
+            mime_types=[("application/opds+json", 1)],  # Force the type
             max_age=int(max_age) if max_age is not None else None,
         )
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -961,7 +961,7 @@ class OPDSFeedController(CirculationManagerController):
         )
         return feed.as_response(
             max_age=int(max_age) if max_age else None,
-            requested_content_type=flask.request.content_type,
+            requested_content_type=flask.request.headers.get("Accept"),
         )
 
     def navigation(self, lane_identifier):
@@ -1175,7 +1175,9 @@ class OPDSFeedController(CirculationManagerController):
         )
         if isinstance(response, ProblemDetail):
             return response
-        return response.as_response(requested_content_type=flask.request.content_type)
+        return response.as_response(
+            requested_content_type=flask.request.headers.get("Accept")
+        )
 
     def _qa_feed(
         self, feed_factory, feed_title, controller_name, facet_class, worklist_factory
@@ -1469,7 +1471,9 @@ class LoanController(CirculationManagerController):
         # Then make the feed.
         feed = OPDSAcquisitionFeed.active_loans_for(self.circulation, patron)
         response = feed.as_response(
-            max_age=0, private=True, requested_content_type=flask.request.content_type
+            max_age=0,
+            private=True,
+            requested_content_type=flask.request.headers.get("Accept"),
         )
 
         last_modified = patron.last_loan_activity_sync
--- a/api/controller.py
+++ b/api/controller.py
@@ -21,6 +21,7 @@ from pydantic import ValidationError
 from sqlalchemy import select
 from sqlalchemy.orm import eagerload
 from sqlalchemy.orm.exc import NoResultFound
+from werkzeug.datastructures import MIMEAccept
 
 from api.authentication.access_token import AccessTokenProvider
 from api.model.patron_auth import PatronAuthAccessToken
@@ -1323,7 +1324,7 @@ class OPDS2FeedController(CirculationManagerController):
             self.search_engine,
         )
         return feed.as_response(
-            mime_types=[("application/opds+json", 1)],  # Force the type
+            mime_types=MIMEAccept([("application/opds+json", 1)]),  # Force the type
             max_age=int(max_age) if max_age is not None else None,
         )
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -961,7 +961,7 @@ class OPDSFeedController(CirculationManagerController):
         )
         return feed.as_response(
             max_age=int(max_age) if max_age else None,
-            requested_content_type=flask.request.headers.get("Accept"),
+            mime_types=flask.request.accept_mimetypes,
         )
 
     def navigation(self, lane_identifier):
@@ -1175,9 +1175,7 @@ class OPDSFeedController(CirculationManagerController):
         )
         if isinstance(response, ProblemDetail):
             return response
-        return response.as_response(
-            requested_content_type=flask.request.headers.get("Accept")
-        )
+        return response.as_response(mime_types=flask.request.accept_mimetypes)
 
     def _qa_feed(
         self, feed_factory, feed_title, controller_name, facet_class, worklist_factory
@@ -1325,7 +1323,7 @@ class OPDS2FeedController(CirculationManagerController):
             self.search_engine,
         )
         return feed.as_response(
-            requested_content_type="json",
+            mime_types=["application/opds+json"],  # Force the type
             max_age=int(max_age) if max_age is not None else None,
         )
 
@@ -1473,7 +1471,7 @@ class LoanController(CirculationManagerController):
         response = feed.as_response(
             max_age=0,
             private=True,
-            requested_content_type=flask.request.headers.get("Accept"),
+            mime_types=flask.request.accept_mimetypes,
         )
 
         last_modified = patron.last_loan_activity_sync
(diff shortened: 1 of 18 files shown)


jonathangreen:
@RishiDiwanTT The first thing I was trying to do with this PR is to load the OPDS2 version of the feed in my test environment to see what the serialized OPDS2 looks like. Can you give me some pointers how I can get an OPDS2 response?

RishiDiwanTT:
Currently only the /loans, /borrow and /feed APIs are activated for opds2.
They can be seen by adding a content type : application/opds+json header to
the GET request.

jonathangreen:
Typically the [content-type header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) indicates the type of the data included in the request. I'd expect to do content negotiation like this via an [accept header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) included on the request, which would cause the content-type header to be included in the response.

RishiDiwanTT:
That makes sense, and is a very small change.

jonathangreen:
Okay. So I pointed this branch at an existing DB with some books in it, and then I ran the resulting OPDS2 feed through our OPDS2 validator `/bin/opds2_schema_validate` and our feed is failing our own validation. So i think we need to get that sorted out to start with.

RishiDiwanTT:
> Okay. So I pointed this branch at an existing DB with some books in it, and then I ran the resulting OPDS2 feed through our OPDS2 validator `/bin/opds2_schema_validate` and our feed is failing our own validation. So i think we need to get that sorted out to start with.

This ended up being trivial, changes required were 2
- Use the "Accept" header instead of the "Content-Type" header
- The "audience" scheme should be an URL always

Apart from that the actual validator code was broken, I guess somewhere along the way the OPDSImporter code changed a bit and the validator did not reflect those changes.

jonathangreen:
From the [comment](https://github.com/ThePalaceProject/circulation/pull/1308#discussion_r1310613869) on #1308:

> [jonathangreen](https://github.com/jonathangreen) [Aug 30, 2023](https://github.com/ThePalaceProject/circulation/pull/1308#discussion_r1310613869)
Can we rename the package (directory) for this code as well? Right now its feed_protocol, I'd suggest feed.

> [RishiDiwanTT](https://github.com/RishiDiwanTT) [Sep 1, 2023](https://github.com/ThePalaceProject/circulation/pull/1308#discussion_r1312717382)
This will cause all kinds of unpleasantness while rebasing with https://github.com/ThePalaceProject/circulation/pull/1340 . I will do this once that PR has been merged with this.

Moving this to a comment here so we don't forget the package needs to be renamed.

RishiDiwanTT:
> From the [comment](https://github.com/ThePalaceProject/circulation/pull/1308#discussion_r1310613869) on #1308:
> 
> > [jonathangreen](https://github.com/jonathangreen) [Aug 30, 2023](https://github.com/ThePalaceProject/circulation/pull/1308#discussion_r1310613869)
> > Can we rename the package (directory) for this code as well? Right now its feed_protocol, I'd suggest feed.
> 
> > [RishiDiwanTT](https://github.com/RishiDiwanTT) [Sep 1, 2023](https://github.com/ThePalaceProject/circulation/pull/1308#discussion_r1312717382)
> > This will cause all kinds of unpleasantness while rebasing with #1340 . I will do this once that PR has been merged with this.
> 
> Moving this to a comment here so we don't forget the package needs to be renamed.

Do we have anymore feedback for this PR or should I merge it into the refactor branch?

jonathangreen:
@RishiDiwanTT I'm taking another look over this PR now. I'll post a comment once I'm done.

tests/api/feed_protocol/test_opds_acquisition_feed.py
```diff
@@ -20,21 +21,53 @@
...
+        assert str(raised.value) == "Entry data has not been generated"
+
+        entry.computed = WorkEntryData()
+
+        response = BaseOPDSFeed.entry_as_response(entry)
+        assert isinstance(response, OPDSEntryResponse)
+        # default content type is XML
+        assert response.content_type == OPDSEntryResponse().content_type
+
+        # Specifiically asking for a json type
```
jonathangreen, line 63: Typo in comment

api/controller.py
```diff
@@ -961,6 +961,7 @@ def feed(self, lane_identifier, feed_class=OPDSAcquisitionFeed):
         )
         return feed.as_response(
             max_age=int(max_age) if max_age else None,
+            requested_content_type=flask.request.headers.get("Accept"),
```
jonathangreen, line 964: I think we need to do something a bit more sophisticated then this to properly support [content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation). Some more detail here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

This is a valid accept header:
`Accept: application/atom+xml;profile=opds-catalog;kind=acquisition, application/opds+json;q=0.9, */*;q=0.8`

Which would indicate that the client wants OPDS 1.x response. But we would return an OPDS 2.x response.

It seems like flask provides some parsing of accept headers that we should be using in [Request.accept_mimetypes](https://flask.palletsprojects.com/en/2.3.x/api/#flask.Request.accept_mimetypes), rather then using the header directly.

tests/api/feed_protocol/test_library_annotator.py
```diff
@@ -914,25 +914,12 @@ def test_active_loan_feed(
...
-        # the vendor APIs.
-        #
-        # (The timestamps aren't exactly the same because
-        # last_loan_activity_sync is tracked at the millisecond level
-        # and Last-Modified is tracked at the second level.)
-
-        assert response.last_modified is not None
-        assert (
-            patron.last_loan_activity_sync - response.last_modified
-        ).total_seconds() < 1
```
jonathangreen, line 935: Why does this test come out?

jonathangreen (commented):
@RishiDiwanTT I made a few more comments here

core/feed_protocol/acquisition.py
```diff
@@ -592,11 +588,13 @@ def single_entry_loans_feed(
...
 
-        # TODO: max_age and private response kwargs
         if isinstance(entry, WorkEntry) and entry.computed:
             return cls.entry_as_response(entry, **response_kwargs)
+        elif entry is None:
+            return None
 
-        return None
+        # This is probably an error message
+        return cast(ProblemDetail, entry)
```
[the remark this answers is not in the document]
RishiDiwanTT: I thought of doing that, but I got lazy... told you it wasn't fixed.

tests/api/feed_protocol/test_library_annotator.py
RishiDiwanTT: This code was removed from the annotator, and [added into the controller](https://github.com/ThePalaceProject/circulation/pull/1340/files#diff-944bc0ea08885efcf096971a642422c2f5ea80605fd0b10618c9e6a738d7a837R1470), since the annotator no longer creates the serialized response.
The tests in `tests/api/test_controller_loan.py` already test for that functionality.

core/feed_protocol/acquisition.py
```diff
@@ -590,11 +589,12 @@ def single_entry_loans_feed(
 
         if isinstance(entry, WorkEntry) and entry.computed:
             return cls.entry_as_response(entry, **response_kwargs)
-        elif entry is None:
-            return None
+        elif isinstance(entry, OPDSMessage):
+            return ProblemDetail(
```
jonathangreen, line 593: It looks like this new case isn't covered by a test, and it seems like the original code doesn't return a problem detail in this case, but converts the `OPDSMessage` into an error response?

Can you write up a test for this and verify that we aren't changing the behavior of the single item loans feed here?

core/feed_protocol/opds.py
```diff
@@ -11,11 +13,17 @@
 from core.util.opds_writer import OPDSMessage
 
 
-def get_serializer(content_type: Optional[str]) -> OPDS1Serializer | OPDS2Serializer:
-    if content_type and "json" in content_type:
-        return OPDS2Serializer()
-    else:
-        return OPDS1Serializer()
+def get_serializer(
```
jonathangreen, line 16: This is closer, but I think we need to write some tests around this function. 

I did a little bit of testing, and I don't think iterating over a `MIMEAccept` puts the results in priority order. It uses the order that the Accept header was sent in. It also returns a `Tuple[str, int]` mime / quality. 

I think we probably want to use the [best_match method](https://werkzeug.palletsprojects.com/en/2.0.x/datastructures/#werkzeug.datastructures.Accept.best_match) of `MIMEAccept`, which does the kind of sorting / matching we want here.

It would be nice to see some tests that test a few cases here: 
- No mime passed in
- A single simple mime type
- A couple more complicated Accept headers like: 
  - `Accept: application/atom+xml;profile=opds-catalog;kind=acquisition;q=0.08, application/opds+json;q=0.9`
  - `Accept: application/atom+xml;profile=opds-catalog;kind=acquisition, application/opds+json;q=0.9`

RishiDiwanTT: Ok, I didn't quite grasp the q-value sets the priority and not the order of appearance. I'll fix this.

core/feed_protocol/acquisition.py
jonathangreen: I think either: 
- we need to handle this at the serialization step, so that we get the same XML response we used to get in the case of an OPDS 1 feed
- Or check with the mobile apps and see how returning a problem detail here instead of a feed will be handled

Since our mobile apps are the primary consumer of this feed still, we need to make sure they are in sync with any changes that get made to the responses.

RishiDiwanTT: We can return to the XML format easily, will that be an acceptable behaviour when the client is expecting JSON though?
Re: ProblemDetail: The same method also returns problem details [in different circumstances](https://github.com/ThePalaceProject/circulation/pull/1340/files/8c4941b94515b085a85b482a88d366642315b588..508b81a66336ff8bd846c90401df0ce490ba6391#diff-915f371a570738a5b9970eca750c31619010b101b4ca87634fcbcaf01649c078R561) so a problem detail should be already handled by the client side.

core/feed_protocol/opds.py
jonathangreen: Can you add a test case for these Accept header? 
- `Accept: application/atom+xml;profile=opds-catalog;kind=acquisition;q=0.08, application/opds+json;q=0.9`
  - I'd expect an `application/opds+json` response in that case, but I believe the current implementation will response with `application/atom+xml`. 
- `Accept: application/atom+xml;profile=opds-catalog;kind=acquisition`
  - I'd expect an `application/atom+xml` response in that case, but I believe we will serve `application/opds+json`

RishiDiwanTT: Well... that was embarrassing. Seems like the MIMEAccept datatypes sorting is based FIRST on specificity THEN on q-value.
Changed the method to always sort on q-value.

jonathangreen: Is there a reason you aren't using the best_match method I mentioned above here? It would save a lot of sorting etc and lets you set a default.

RishiDiwanTT: The datatype could be a List, this is only used in a test case for now, but was done to future proof the method in case we needed to do generate feeds offline.

jonathangreen: @RishiDiwanTT I believe that taking the specificity into account is part of the usual Accept header content negotiation dance. It would be nice to be able to lean on the complex matching provided by Flask instead of trying to reimplement it ourselves.

For tests, can we just directly pass in a `MimeAccept` class, or use the (admittedly annoying) flask machinery to fake a request for testing so we can directly consume `MimeAccept`.

RishiDiwanTT: Updated. This caused one test case to change because specificity magic switched around the best match.

jonathangreen (commented):
I'm still concerned here about the change in behavior of the `OPDSMessage` class as part of this PR. I don't want to hold up merging this, but I also don't want it to have a negative impact on the experience of our users consuming the OPDS 1 feed through the app.

core/feed_protocol/opds.py
```diff
@@ -14,17 +14,19 @@
...
-            return OPDS1Serializer()
+    serializers: Dict[str, Type[Any]] = {
+        "application/opds+json": OPDS2Serializer,
+        "application/atom+xml": OPDS1Serializer,
+    }
+    if mime_types:
+        match = mime_types.best_match(
+            serializers.keys(), default="application/atom+xml"
+        )
+        return serializers[match]()  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
RishiDiwanTT: I introduced an abstract base type to better handle the serializer protocol

core/feed_protocol/acquisition.py
RishiDiwanTT: Ok, I reverted to protocol specific data, OPDS1 will be XML and OPDS2 will be JSON.

jonathangreen (approved):
🚀 I think this is looking good @RishiDiwanTT

PR #1342, opened 2023-08-23
Bugfix for OPDS 2 token fulfillment (PP-360)

jonathangreen:
## Description

This is a temporary bug fix. The real fix is going to require a deeper intervention. I think that work will have to take place as part of PP-313. 

This rolls back the changes to get the token URL from the `settings_dict` of the integration configuration that came in as part of the work on https://github.com/ThePalaceProject/circulation/pull/1189. Instead it uses the old behavior of getting the token URL from a configuration setting. Since we didn't remove any old settings, this will still work.

## Motivation and Context

This is a quick fix, to restore token auth for NYU and Columbia. The current code isn't working because the token URL isn't part of the settings model. So this value never got migrated over, because we only migrate values that are part of the model.  Additionally the value would get written if / when a full import was done, but this value would get overwritten any time the settings form is submitted.

## How Has This Been Tested?

- Tested locally with the DB from Columbia. I saw the same error on fulfill that I was seeing in production before this fix, and I was able to fulfill the content after this fix went in.

## Checklist


tdilauro (approved):
Looks good to me.

RishiDiwanTT:
Could this issue be fixed by adding `TOKEN_AUTH` to the `OPDS2ImporterSettings` model?

jonathangreen:
We could add `TOKEN_AUTH` to the model, but I don't think its something we actually want to be configurable, since its supplied in the feed. I think a better long term fix here is to find another place to store this value. Maybe as some context on the integration outside of the settings.

PR #1344, opened 2023-08-29
Remove locally hosted LCP Collections 🔥 (PP-393)

jonathangreen:
## Description

Remove locally hosted LCP collections from the CM, since we are not using it anywhere. 

**NOTE:** This does **NOT** impact LCP content hosted by other vendors, but borrowed through the CM. Just CM hosted LCP collections, which are not being used anywhere currently.

## Motivation and Context

As part of PP-95, the LCP mirror integration needs some updating. Since we are not using this anywhere, and it does not appear to be working currently, we decided to remove it rather then fix it.

## How Has This Been Tested?

- CI tests

## Checklist


jonathangreen:
🚨 The migration on this one has gotten stale and will need to be synced up with main before merging. Just leaving this as a note to myself so I make sure that happens before merging. This still should be fine for code review.

RishiDiwanTT (approved):
Looks good :+1: 
Does this removal of code not affect the lcp passphrase that gets added to the OPDS feeds from in `api/opds.py` ?

jonathangreen:
> Does this removal of code not affect the lcp passphrase that gets added to the OPDS feeds in api/opds.py ?

It should not change the addition of that lcp passphrase, since that is used for LCP content that is hosted by other providers, as well as locally hosted LCP content. So we will still need that element in the feed to support other LCP providers like Palace Marketplace.

PR #1345, opened 2023-08-29
Remove the directory importer script 🔥 (PP-213)

jonathangreen:
## Description

Remove the directory importer script, since we are not using it anywhere, and have decided to host content like this elsewhere in the future.

## Motivation and Context

Our database migration for #1189 failed on the Columbia staging server because it had an external integration configured with “Manual intervention“ as the protocol. This collection was added via the `directory_import` script. This script appears to create collections that cause problems with the new integration settings system. 

Instead of fixing this issue, we've decided to remove the script.

## How Has This Been Tested?

Running tests.

## Checklist


tdilauro (approved):
Looks good! 🔥

PR #1347, opened 2023-09-01
PP-398 Maintain type coercions during collections settings mutation

RishiDiwanTT:
## Description
The Admin CollectionSettingsController was not maintain type coercion from the pydantic configuration models.
This change ensures the type coercion is saved to the database as well.
Along with this we ensure the default loan period method always casts the response to an int.

## Motivation and Context
Not coercing the types would result in always saving the values as strings, which is incorrect.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-398)

## How Has This Been Tested?
Updated the unit tests.
Manually tested the "ebook load duration" setting in ODL collections.

## Checklist

### DIFF
--- a/alembic/versions/20230905_2b672c6fb2b9_type_coerce_collection_settings.py
+++ b/alembic/versions/20230905_2b672c6fb2b9_type_coerce_collection_settings.py
@@ -6,7 +6,9 @@ Create Date: 2023-09-05 06:40:35.739869+00:00
 
 """
 import json
-from typing import Callable, Dict
+from typing import Any, Callable, Dict, Type
+
+from pydantic import parse_obj_as
 
 from alembic import op
 
@@ -22,9 +24,9 @@ def _bool(value):
 
 
 # All the settings types that have non-str types
-ALL_SETTING_TYPES: Dict[str, Callable] = {
-    "verify_certificate": _bool,
-    "default_reservation_period": _bool,
+ALL_SETTING_TYPES: Dict[str, Type[Any]] = {
+    "verify_certificate": bool,
+    "default_reservation_period": bool,
     "loan_limit": int,
     "hold_limit": int,
     "max_retry_count": int,
@@ -38,7 +40,7 @@ def _coerce_types(settings: dict) -> None:
     setting_type: Callable
     for setting_name, setting_type in ALL_SETTING_TYPES.items():
         if setting_name in settings:
-            settings[setting_name] = setting_type(settings[setting_name])
+            settings[setting_name] = parse_obj_as(setting_type, settings[setting_name])
 
 
 def upgrade() -> None:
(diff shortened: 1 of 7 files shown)


core/model/collection.py
```diff
@@ -325,10 +325,11 @@ def default_loan_period(self, library, medium=EditionConstants.BOOK_MEDIUM):
         that someone who borrows a non-open-access item from this
         collection has it for this number of days.
         """
-        return (
+        value = (
             self.default_loan_period_setting(library, medium)
             or self.STANDARD_DEFAULT_LOAN_PERIOD
         )
+        return int(value) if value is not None else None
```
jonathangreen, line 332: Do we need this here? Since we're correctly handing this in the DB, we shouldn't need to force the type here. If we're worried about values that didn't get correctly saved, maybe a migration to clean them up would be better?

jonathangreen (commented):
This generally looks good to me, just one comment about casting the `default_loan_period_setting` to `int`.

core/model/collection.py
RishiDiwanTT: Ok, there is a logistical problem here. To target all the settings available the best course of action would be to load every settings_dict into their pydantic model and writing the `.dict()` back into the database.
However, the settings models are all defined within the application codebase which would lead to importing the ORM models into the migration code, which is bad practice.
We can either
1. Refactor the models into their own files 
2. Target very specific settings values individually

RishiDiwanTT: I'm going to use strategy 2, because 1 looks like a refactor we don't need.

alembic/versions/20230905_2b672c6fb2b9_type_coerce_collection_settings.py
```diff
@@ -0,0 +1,77 @@
...
+    "default_loan_duration": int,
+}
+
+
+def _coerce_types(settings: dict) -> None:
+    """Coerce the types, in-place"""
+    setting_type: Callable
+    for setting_name, setting_type in ALL_SETTING_TYPES.items():
+        if setting_name in settings:
+            settings[setting_name] = setting_type(settings[setting_name])
```
jonathangreen, line 41: Lets use pydantics type cohesion here instead of the `_bool` function. They provide [parse_obj_as](https://docs.pydantic.dev/1.10/usage/models/#parsing-data-into-a-specified-type), which should do a more robust conversion.

Something like: 
```python
for setting_name, setting_type in ALL_SETTING_TYPES.items():
    if setting_name in settings:
        settings[setting_name] = parse_obj_as(settings_type, settings[setting_name])
```

jonathangreen (commented):
Looks good, just a comment about how we do the type coercion.

jonathangreen (approved):

PR #1349, opened 2023-09-04
PP-356 Scripts CI test for crontab

RishiDiwanTT:
## Description
Crontab syntax is tested by installing the `/etc/cron.d/circulation` crontab on the container.
Any errors will result in an error of the script.

## Motivation and Context
Some cron expressions are supported only in certain distributions, this should avoid the case where a documented syntax is not supported.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-356)

## How Has This Been Tested?
Locally tested the `test_scripts.sh` against a badly written crontab.
The CI was also run against the badly written crontab [here](https://github.com/ThePalaceProject/circulation/actions/runs/6071529771/job/16469795441).

## Checklist

### DIFF
--- a/docker/ci/check_service_status.sh
+++ b/docker/ci/check_service_status.sh
@@ -29,3 +29,17 @@ function check_service_status()
     echo "  OK"
   fi
 }
+
+function check_crontab() {
+  container="$1"
+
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
Looks good!

PR #1358, opened 2023-09-05
Remove collection content mirroring (PP-415)

jonathangreen:
## Description

We are not using the collection content mirroring functionality anywhere in our production CMs and the settings to support this add additional complexity to collection configuration screens.

## Motivation and Context

This was done as part of the work for PP-95.

## How Has This Been Tested?

- Running tests in CI

## Checklist


api/admin/routes.py
```diff
@@ -155,32 +155,6 @@ def work_classifications(identifier_type, identifier):
     )
 
 
-@library_route(
```
RishiDiwanTT, line 158: Do only mirrored collections use the urls and methods for cover images?

RishiDiwanTT (approved):
A single clarifying comment, otherwise looks fairly :fire: to me

api/admin/routes.py
jonathangreen: Yes, thats correct. The only time the ability to change the cover image is shown in the admin UI is when the collection has mirror integrations configured.

jonathangreen:
Rebased against main, resolved conflicts and added a commit to update this PR to reflect the changes that came in with https://github.com/ThePalaceProject/circulation/pull/1308. Planning to merge if all the tests are green when CI runs.

PR #1359, opened 2023-09-06
PP-356 Test run scripts during CI

RishiDiwanTT:
## Description
The test_scripts.sh script tests for correctly configured variables as well by running a single script.
`search_index_refesh` was chosen since it hits both the DB and OpenSearch.

## Motivation and Context
Sometimes a misconfigured environment creeps into the deployments, the CI should catch these misconfigurations.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-356)

## How Has This Been Tested?
Locally ran a misconfigured docker to see failures

## Checklist

### DIFF
--- a/docker/ci/check_service_status.sh
+++ b/docker/ci/check_service_status.sh
@@ -48,7 +48,7 @@ function run_script() {
   container="$1"
   script="$2"
 
-  $(docker exec "$container" /bin/bash -c "$script")
+  output=$(docker exec "$container" /bin/bash -c "$script")
   script_status=$?
   if [[ "$script_status" != 0 ]]; then
     echo "  FAIL: script run failed"
--- a/docker/ci/check_service_status.sh
+++ b/docker/ci/check_service_status.sh
@@ -43,3 +43,17 @@ function check_crontab() {
     echo "  OK"
   fi
 }
+
+function run_script() {
+  container="$1"
+  script="$2"
+
+  $(docker exec "$container" /bin/bash -c "$script")
+  script_status=$?
+  if [[ "$script_status" != 0 ]]; then
+    echo "  FAIL: script run failed"
+    exit 1
+  else
+    echo "  OK"
+  fi
+}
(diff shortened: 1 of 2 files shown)


RishiDiwanTT:
> @RishiDiwanTT looks like this new test is failing in the container.

It looks like the docker container is running in a way that is not the same as the prod deployment

```
docker run --rm --name test_container -d --platform linux/amd64 --network github_network_41098063dc4d42ff854e954f26d41f5d -e SIMPLIFIED_PRODUCTION_DATABASE="***postgres:5432/palace_circulation" test_image
```

In this way we are testing the CI container command as opposed to the Prod container command, it doesn't seem to be useful.

jonathangreen:
We are testing the container environment, and the environment is always going to be different in CI. To me it still seems useful to make sure that scripts are able to run in the container. 

It looks like the error message here is:
```
core.config.CannotLoadConfiguration: No search integration configured.
```

Which makes sense, because the search integration has to be configured in the UI... but we have a fresh initialized database here, so no search config is present. I like your idea of testing a script that needs both search and DB, but the way we configure search currently makes that hard.

Maybe a couple options to fix: 
- Test another script that doesn't explicitly depend on search
- Configure search in DB before testing

RishiDiwanTT:
> We are testing the container environment, and the environment is always going to be different in CI. To me it still seems useful to make sure that scripts are able to run in the container.
> 
> It looks like the error message here is:
> 
> ```
> core.config.CannotLoadConfiguration: No search integration configured.
> ```
> 
> Which makes sense, because the search integration has to be configured in the UI... but we have a fresh initialized database here, so no search config is present. I like your idea of testing a script that needs both search and DB, but the way we configure search currently makes that hard.
> 
> Maybe a couple options to fix:
> 
> * Test another script that doesn't explicitly depend on search
> * Configure search in DB before testing

I will test another script that only uses the DB variable, since the test of whether the app responds to a DB configured value is probably a unit test somewhere already.

jonathangreen:
Can we run this one without the `-h` flag? It shouldn't depend on search

jonathangreen (approved):
🎉

PR #1361, opened 2023-09-08
PP-411 Fix incorrectly formatted FCM notification

RishiDiwanTT:
## Description
Fixed the `fcm_admin` method call, this wasn't caught during testing because `fcm_admin` is mocked and cannot be realistically tested in the unit tests.

## Motivation and Context
The desired behaviour was not occurring  after the previous fix, investigations led to the the discovery of a badly formatted API call.

## How Has This Been Tested?
Manually tested the FCM notification syntax with a dev token.

## Checklist


tdilauro (approved):
LGTM 🛠️

PR #1366, opened 2023-09-11
Fix issue with 2b672c6fb2b9 migration (PP-428)

RishiDiwanTT:
## Description
Fixed the type for default_reservation_period
Fixed the primary key queries for the library integration configurations

## Motivation and Context
The migration did not have the query for both primary keys when created.
This would cause incorrect settings migrations.

## How Has This Been Tested?
Updated the unit test to with the additional use cases

## Checklist


jonathangreen (approved):
Added some additional filtering on the queries, additional logging and some addition test cases.

PR #1368, opened 2023-09-12
Add timeout to our weekly baseimage docker build

jonathangreen:
## Description

Add `timeout-minutes: 60` to our `build-base-image.yml` workflow.

## Motivation and Context

Our last baseimage build froze for some reason, and hit the github actions build timeout of 6 hours. Since our base image build usually only takes about 30 minutes, we can add the `timeout-minutes` setting so we fail faster if we run into this issue again.

## How Has This Been Tested?

- Running CI

## Checklist


tdilauro (approved):
🚀

PR #1378, opened 2023-09-13
PP-337 Quicksight dashboard embed URL generation

RishiDiwanTT:
## Description
Depends on a comma separated string of dashboard ARNs under `QUICKSIGHT_AUTHORIZED_ARNS` environment variable.
Exposes the URL `/admin/quicksight_embed/<dashboard-id>?libraryIds=1,2,3` and returns `dict(embedUrl=....)`.

Currently allows access till the `LIBRARIAN` admin role.

## Motivation and Context
Certain library Admin users need on-the-go and anonymous access to the quicksight dashboards for analytics.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-337)

## How Has This Been Tested?
Manually tested the link generation with the tpp-dev account.
New unit tests were written.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,10 @@ More details in the [FCM documentation](https://firebase.google.com/docs/admin/s
 ##### Quicksight Dashboards
 
 For generating quicksight dashboard links the following environment variable is required
-`QUICKSIGHT_AUTHORIZED_ARNS` - A comma separated list of `arn:aws:quicksight:...` format ARN strings
+`QUICKSIGHT_AUTHORIZED_ARNS` - A dictionary of the format `"<dashboard name>": ["arn:aws:quicksight:...",...]`
+where each quicksight dashboard gets treated with an arbitrary "name", and a list of "authorized arns".
+The first the "authorized arns" is always considered as the `InitialDashboardID` when creating an embed URL
+for the respective "dashboard name".
 
 #### Email
 
--- a/README.md
+++ b/README.md
@@ -239,7 +239,14 @@ export SIMPLIFIED_FCM_CREDENTIALS_FILE="/opt/credentials/fcm_credentials.json"
 The FCM credentials can be downloaded once a Google Service account has been created.
 More details in the [FCM documentation](https://firebase.google.com/docs/admin/setup#set-up-project-and-service-account)
 
+##### Quicksight Dashboards
+
+For generating quicksight dashboard links the following environment variable is required
+`QUICKSIGHT_AUTHORIZED_ARNS` - A comma separated list of `arn:aws:quicksight:...` format ARN strings
+
 #### Email
+
+### Email sending
 
 To use the features that require sending emails, for example to reset the password for logged-out users, you will need
 to have a working SMTP server and set some environment variables:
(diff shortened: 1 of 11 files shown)


api/admin/routes.py
```diff
@@ -83,7 +87,11 @@ def returns_json_or_response_or_problem_detail(f):
 
     @wraps(f)
     def decorated(*args, **kwargs):
-        v = f(*args, **kwargs)
+        try:
+            v = f(*args, **kwargs)
+        except ProblemError as ex:
```
RishiDiwanTT, line 92: Added this in so we can globally manage raising ProblemErrors

jonathangreen:
This one looks okay to me, but I'd like @dbernstein to take a look since hes done the rest of the quicksight work.

README.md
```diff
@@ -180,6 +180,11 @@ export SIMPLIFIED_FCM_CREDENTIALS_FILE="/opt/credentials/fcm_credentials.json"
 The FCM credentials can be downloaded once a Google Service account has been created.
 More details in the [FCM documentation](https://firebase.google.com/docs/admin/setup#set-up-project-and-service-account)
 
+##### Quicksight Dashboards
+
+For generating quicksight dashboard links the following environment variable is required
+`QUICKSIGHT_AUTHORIZED_ARNS` - A comma separated list of `arn:aws:quicksight:...` format ARN strings
```
dbernstein, line 186: Since we'll likely have multiple dashboards that can be displayed within the CM, I think we'll need to pass in a dictionary where the keys are logical dashboard names such as "library" or "consortium" or "all_libraries" and the values are ARNs that link to environment specifi dashboards.   To give ourselves flexibility we'll use the logical name (ie the key) in the Circulation manager API calls below.   Does that make sense?

core/config.py
```diff
@@ -50,6 +50,10 @@ class Configuration(ConfigurationConstants):
     OD_FULFILLMENT_CLIENT_KEY_SUFFIX = "OVERDRIVE_FULFILLMENT_CLIENT_KEY"
     OD_FULFILLMENT_CLIENT_SECRET_SUFFIX = "OVERDRIVE_FULFILLMENT_CLIENT_SECRET"
 
+    # Quicksight
+    # Comma separated aws arns
+    QUICKSIGHT_AUTHORIZED_ARNS_KEY = "QUICKSIGHT_AUTHORIZED_ARNS"
```
dbernstein, line 55: See comment above re dictionary.

api/admin/routes.py
```diff
@@ -340,6 +348,18 @@ def stats():
     return statistics_response.api_dict()
 
 
+@app.route("/admin/quicksight_embed/<dashboard_id>")
```
dbernstein, line 351: I believe we'll want to use the logical name here instead of the dashboard id.   That way the Admin UI can use hard-coded logical names which can be completely decoupled from the dashboard ARN.  (Per my comments above on the README).

dbernstein (changes requested):
As usual, excellent work @RishiDiwanTT .   See my suggestion re use of a dictionary.   I'm curious what your thoughts are on that.

README.md
RishiDiwanTT: I changed this to a Dict format like:
```
{ "name": [<arn list>]}
```
This should be json dumped from ansible into the same env var. 
I also added an `/admin/quicksight_embed/names` API so the dashboard names are not hard coded anywhere (frontend or backend).
This brings up an issue with authorisation though, all dashboard names will be visible to all dashboard users regardless of whether there is any intersecting library between their authorisation and the dashboards available libraries.
In which case the admin API will throw authorisation errors when trying to create an URL without any libraries visible to the user.

PR #1379, opened 2023-09-13
Remove the externalintegrationslinks table

jonathangreen:
## Description

This PR removes the now unused `externalintegrationslinks` table.

## Motivation and Context

As of #1377 the `externalintegrationslinks` table is no longer being used. It is kept around in that PR in case we need to roll back those changes. Once those changes are successfully running everywhere, we should merge this PR to remove the unused table.

## How Has This Been Tested?

- Running tests in CI

## Checklist


jonathangreen:
Now that [v10.0.0](https://github.com/ThePalaceProject/circulation/releases/tag/v10.0.0) is out, I think we are good to merge this.

RishiDiwanTT (approved):
Looks good :+1:

PR #1383, opened 2023-09-15
Add patron "Basic Token" authentication feature switch. (PP-431)

tdilauro:
## Description

Enables/disables patron "Basic Token" authentication through setting the `SIMPLIFIED_ENABLE_BASIC_TOKEN_AUTH` environment variable to any (case-insensitive) value of "true"/"yes"/"on"/"1" or "false"/"no"/"off"/"0", respectively. If the value is the empty string or the variable is not present in the environment, it is disabled by default.

## Motivation and Context

The need to selectively enable and disable the patron "basic token" authentication capability has had us adding and removing the code via commits for the past few cycles. This has meant that we were sometimes deploying to production code different than what we had tested in our development and staging environments. This change will allow us to deploy the same code and selectively turn this feature on and off via configuration.

[Jira [PP-431](https://ebce-lyrasis.atlassian.net/browse/PP-431)]

## How Has This Been Tested?

- New tests added for new functionality.
- CA tests passed on the associated feature branch.

## Checklist

[PP-431]: https://ebce-lyrasis.atlassian.net/browse/PP-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -11,7 +11,7 @@ from sqlalchemy.exc import ArgumentError
 # from this module, alongside CannotLoadConfiguration.
 from core.exceptions import IntegrationException
 
-from .util import LanguageCodes
+from .util import LanguageCodes, ansible_boolean
 from .util.datetime_helpers import to_utc, utc_now
 
 
@@ -39,6 +39,10 @@ class Configuration(ConfigurationConstants):
     # Environment variables that contain URLs to the database
     DATABASE_TEST_ENVIRONMENT_VARIABLE = "SIMPLIFIED_TEST_DATABASE"
     DATABASE_PRODUCTION_ENVIRONMENT_VARIABLE = "SIMPLIFIED_PRODUCTION_DATABASE"
+
+    # TODO: We can remove this variable once basic token authentication is fully deployed.
+    # Patron token authentication enabled switch.
+    BASIC_TOKEN_AUTH_ENABLED_ENVVAR = "SIMPLIFIED_ENABLE_BASIC_TOKEN_AUTH"
 
     # Environment variables for Firebase Cloud Messaging (FCM) service account key
     FCM_CREDENTIALS_FILE_ENVIRONMENT_VARIABLE = "SIMPLIFIED_FCM_CREDENTIALS_FILE"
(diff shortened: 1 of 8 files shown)


core/config.py
```diff
@@ -207,6 +211,26 @@ def database_url(cls):
...
+    @classmethod
+    def basic_token_auth_is_enabled(cls) -> bool:
+        """Is basic token authentication enabled?
+
+        Return False, if the variable is unset or is an empty string.
+        Raises CannotLoadConfiguration, if the setting is invalid.
+        :raise CannotLoadConfiguration: If the setting contains an unsupported value.
+        """
+        try:
+            return ansible_boolean(
```
jonathangreen, line 224: Instead of defining a function to do this, I wonder if we would be better off leaning on Pydantic to do the parsing. I think we could accomplish the same thing with: 

```python
    @classmethod
    def basic_token_auth_is_enabled(cls) -> bool:
        """Is basic token authentication enabled?
        Return False, if the variable is unset or is an empty string.
        Raises CannotLoadConfiguration, if the setting is invalid.
        :raise CannotLoadConfiguration: If the setting contains an unsupported value.
        """
        try:
            return parse_obj_as(bool, os.environ.get(cls.BASIC_TOKEN_AUTH_ENABLED_ENVVAR))
        except (ValidationError) as e:
            raise CannotLoadConfiguration(
                f"Invalid value for {cls.BASIC_TOKEN_AUTH_ENABLED_ENVVAR} environment variable."
            ) from e
```

Which would parse the env var as a pydantic boolean https://docs.pydantic.dev/latest/usage/types/booleans/
> a str which when converted to lower case is one of '0', 'off', 'f', 'false', 'n', 'no', '1', 'on', 't', 'true', 'y', 'yes'

jonathangreen (approved):
Looks good. Just one comment about being able to lean on pydantic to do our parsing instead of doing it ourselves.

core/config.py
jonathangreen: Made a slight edit there to the code in my comment because I realized I wasn't handling the unset case in my first reply.

tdilauro: @jonathangreen That's a clever and interesting idea. It works right now because `False` is the default, but it might cause some confusion if we were to decide at some point switch to a default of True. In that case, making a naive change the default and subsequently setting the environment variable to a falsy value would result in a value of True. It might lead to some programming errors down the line.

I also thought it would be helpful to have a function feature flags going forward, since I don't think this is the last time we'll see the need. And it has tests, so it's pretty easy to reason about how it will behave.

Another option would be to replace part of the logic in the current `ansible_boolean` with the `pydantic.parse_obj_as` approach above.

What do you think?

jonathangreen: I'm okay with whatever you choose, including just leaving it as is.

tdilauro: Okay. Thanks!

PR #1390, opened 2023-09-18
Accept POST requests to the /patrons/me/token endpoint (PP-464)

jonathangreen:
## Description

Accept POST requests to the /patrons/me/token endpoint.

## Motivation and Context

Discussion on Notion here: https://www.notion.so/Create-plan-for-token-based-authentication-61bbd66923b94b38bccfa838546b6ad8?d=7ff681b5d2134f8682cf68f9e991669a&pvs=4#4ef3b7831e4f482daa4f380bb716d362

Looks like this was accidentally made to accept GET requests, when it should be accepting POST requests.

## Checklist

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -395,7 +395,7 @@ def delete_patron_devices():
     return app.manager.patron_devices.delete_patron_device()
 
 
-@library_dir_route("/patrons/me/token", methods=["GET"])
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Looks good. Doesn't appear that there is much in the way of tests for this route.

PR #1391, opened 2023-09-19
PP-463 Empty author tag fix

RishiDiwanTT:
## Description
Author tags could wind up empty in an OPDS feed if no name was present, this protects against such an eventuality.

## Motivation and Context
Minotaur has entries that demonstrate the problem, causing the Android App to crash
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-463)

## How Has This Been Tested?
Unit tests have been updated for this change

## Checklist

### DIFF
--- a/core/feed/annotator/base.py
+++ b/core/feed/annotator/base.py
@@ -95,7 +95,7 @@ class ToFeedEntry:
         name = contributor.display_name or contributor.sort_name
         name_key = name and name.lower()
         if not name_key or name_key in state[marc_role]:
-            # Either there is not valid name present or
+            # Either there is no valid name present or
             # we've already credited this person with this
             # MARC role. Returning a tag would be redundant.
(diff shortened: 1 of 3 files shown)


core/feed/annotator/base.py
```diff
@@ -93,8 +93,8 @@ def contributor(
             return None
 
         name = contributor.display_name or contributor.sort_name
-        name_key = name.lower()
-        if name_key in state[marc_role]:
+        name_key = name and name.lower()
+        if not name_key or name_key in state[marc_role]:
             # We've already credited this person with this
```
jonathangreen, line 98: Is this comment still accurate with this change? Seems like it needs an extra note about empty author.

jonathangreen (approved):
Looks good. Just have a minor question about a comment in the code.

PR #1392, opened 2023-09-19
PP-466 Fix notification script failures

RishiDiwanTT:
## Description
Ensured the optional attributes of a patron are vetted before being used in a notification

## Motivation and Context
Some attributes of a user are optional (NoneTypes), which are incompatible with the FCM notification format.
So we ensure only strings are used in the notifications.

## How Has This Been Tested?
Unit tests have been updated to test the situation

## Checklist


jonathangreen (approved):
Looks good.

PR #1396, opened 2023-09-20
Some patron authenticators delegate their patron lookup. (PP-461) 

tdilauro:
## Description

Adds method to patron authentication providers to indicate the (possibly different) authentication provider that handles its patron lookups.

## Motivation and Context

Patron lookups for reseting Adobe IDs were failing with basic token authentication enabled because `BasicTokenAuthenticationProvider` needs to delegate patron lookups to its associated `BasicAuthenticationProvider`.

[Jira [PP-461](https://ebce-lyrasis.atlassian.net/browse/PP-461)]

## How Has This Been Tested?

- Manual testing of the failing code path in local development environment.
- Added testing for the new functionality.
- CI build passing for associate branch.

## Checklist

[PP-461]: https://ebce-lyrasis.atlassian.net/browse/PP-461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/admin/controller/patron.py
+++ b/api/admin/controller/patron.py
@@ -35,13 +35,14 @@ class PatronController(CirculationManagerController, AdminPermissionsControllerM
 
         patron_data = PatronData(authorization_identifier=identifier)
         complete_patron_data = None
+        patron_lookup_providers = list(authenticator.unique_patron_lookup_providers)
 
-        if not authenticator.providers:
+        if not patron_lookup_providers:
             return NO_SUCH_PATRON.detailed(
                 _("This library has no authentication providers, so it has no patrons.")
             )
 
-        for provider in authenticator.providers:
+        for provider in patron_lookup_providers:
             complete_patron_data = provider.remote_patron_lookup(patron_data)
             if complete_patron_data:
                 return complete_patron_data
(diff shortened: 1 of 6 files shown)


jonathangreen (approved):
Looks good to me 🚀

jonathangreen:
This fix works, and makes sense given the structure we have here, and it 100% makes sense to merge as is.

That said, I do wonder if it makes any sense to have `BasicTokenAuthenticationProvider` separated from `BasicAuthenticationProvider`. They are intertwined anyway, just putting the Token functionality into `BasicAuthenticationProvider` rather then delegating parts of the functionality seems like a more reasonable long term structure to me.

tdilauro:
I'd like to see this change, too. Some of what I did was just to make it more clear what is happening and delineate at least a light distinction between authentication and lookup. I'd like the token auth capability to be a lighter layer atop our "traditional" patron auth providers, so that it's not something that we have to implement for every kind of provider in the CM.

jonathangreen:
Would you be up for making a ticket to handle this further refactoring here @tdilauro?

PR #1397, opened 2023-09-20
Fully type hint the OPDS Importer and OPDS2 Importer classes (PP-313)

jonathangreen:
## Description

Fully type hint the `core/opds_import.py` and `core/opds2_import.py` files.

## Motivation and Context

As part of the work on PP-313 I'm doing some refactoring in the `OPDSImporter` classes. Its so much easier to refactor classes that have type hints, so I did this work to add type hints to these classes before starting the bulk of the refactoring work. 

Jira: PP-313

## How Has This Been Tested?

Running tests in CI

## Checklist


core/model/identifier.py
```diff
@@ -561,6 +561,26 @@ def _parse_urn(
...
+        must_support_license_pools: bool = False,
+    ) -> tuple[Identifier, bool]:
+        ...
+
+    @classmethod
+    @overload
+    def parse_urn(
+        cls,
+        _db: Session,
+        identifier_string: str | None,
```
RishiDiwanTT, line 579: Should this be just the `str` type, since we've covered the `None` case?

RishiDiwanTT (approved):
Some very minor comments. Looks good otherwise :+1: .

core/model/identifier.py
jonathangreen: @RishiDiwanTT I'm a bit confused by this comment. Do you mean the opposite here? 
> Should this just be `None` since we've covered the `str` case?

We need a type hint for the type union case, that will be used when mypy doesn't know if the type is `None` or `str` statically.

api/odl2.py
```diff
@@ -251,6 +258,7 @@ def _extract_publication_metadata(self, feed, publication, data_source_name):
                     if not medium:
                         medium = Edition.medium_from_media_type(license_format)
 
+                    drm_schemes: List[str | None]
```
tdilauro, line 261: Do any of the calls in the following if/else block return a list that potentially includes `None`? Or does allowing `None1 here come in only for the `for` on l. 283?

tests/core/test_opds_import.py
```diff
@@ -1628,7 +1631,7 @@ def _is_open_access_link(self, url, type):
         result = good_link_importer.assert_importable_content(
             "feed", "url", max_get_attempts=5
         )
-        assert "this is a book" == result
+        assert True == result
```
tdilauro, line 1634: `result is True`?

tdilauro (approved):
Looks good. ✅

Just a question and a minor comment.

core/model/identifier.py
RishiDiwanTT: Yes, I meant the opposite.

PR #1402, opened 2023-09-22
Fix failure to set unlimited_access property on non-existent pool.

dbernstein:
## Description
Simply checks if the pool is not None before trying to set a property on it. 
Improves logging to show detailed failure info in logs during OPDS feed imports so that future issues will be easier to troubleshoot in the future.

One thing I'm not sure about:  does it make sense for the pool to be returned as a None from `super().update_work_for_e
dition`

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-192

## How Has This Been Tested?
I reran San Rafael BiblioBoard  collection locally and confirmed that all 139 failures are no longer occurring.

```
./bin/opds_for_distributors_import_monitor --collection "San Rafael BiblioBoard" --force
...
99d7-25f0535356c6: rel=alternate", "timestamp": "2023-09-22T17:57:25.570186+00:00"}
{"host": "8e97e3d1a6b0", "app": "simplified", "name": "OPDS for Distributors Import Monitor", "level": "INFO", "filename": "monitor.py", "message": "[San Rafael BiblioBoard] Items imported: 30379. Failures: 0.", "timestamp": "2023-09-22T17:57:25.652138+00:00"}
{"host": "8e97e3d1a6b0", "app": "simplified", "name": "OPDS for Distributors Import Monitor", "level": "INFO", "filename": "monitor.py", "message": "[San Rafael BiblioBoard] Ran OPDS for Distributors Import Monitor monitor in 3664.29 sec.", "timestamp": "2023-09-22T17:57:25.657573+00:00"}
```

## Checklist

- [/] I have updated the documentation accordingly.
- [] All new and existing tests passed.

### DIFF
--- a/api/opds_for_distributors.py
+++ b/api/opds_for_distributors.py
@@ -357,7 +357,10 @@ class OPDSForDistributorsImporter(OPDSImporter):
         pool, work = super().update_work_for_edition(
             *args, is_open_access=False, **kwargs
         )
-        pool.unlimited_access = True
+
+        if pool:
+            pool.unlimited_access = True
+
         return pool, work
 
     @classmethod
--- a/core/opds_import.py
+++ b/core/opds_import.py
@@ -524,6 +524,12 @@ class OPDSImporter(CirculationConfigurationMixin):
                 if work:
                     works[key] = work
             except Exception as e:
+                logging.warning(
+                    f"Non-fatal exception: Failed to import item - import will continue: "
+                    f"identifier={key}; collection={self.collection.name}; "
+                    f"data_source={self.data_source}; exception={e}",
+                    stack_info=True,
+                )
                 identifier, ignore = Identifier.parse_urn(self._db, key)
                 data_source = self.data_source
                 failure = CoverageFailure(


api/opds_for_distributors.py
```diff
@@ -357,7 +357,10 @@ def update_work_for_edition(self, *args, **kwargs):
         pool, work = super().update_work_for_edition(
             *args, is_open_access=False, **kwargs
         )
-        pool.unlimited_access = True
+
+        if pool:
```
dbernstein, line 361: One thing I'm not sure about: does it make sense for the pool to be returned as a None from super().update_work_for_edition.  The `update_work_for_edition` seems to allow for it  and downstream code seems to expect it, but I'm wondering if makes sense in this context.

jonathangreen: It does seem like the super class expects to return None in some cases. So I think this is an okay fix.

jonathangreen (approved):
Looks good to me.

PR #1404, opened 2023-09-22
Remove unused integration test code 🔥

jonathangreen:
## Description

We are not running the code in the integration tests folder. Its not been maintained since we forked from Library Simplified.

## Motivation and Context

Nice to clean up some more unused code. Since we've not run this since we forked the code, I'm not even convinced these files would still run given the changes in the wider codebase.

## How Has This Been Tested?

- Nope

## Checklist


tdilauro (approved):
🔥

PR #1411, opened 2023-09-26
PP-505 Switched the _cluster API body to the AWS supported flat format

RishiDiwanTT:
## Description
AWS does not support the nested API format.
Reference: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-operations.html#version_api_notes

## Motivation and Context
Minotaur could not connect to the cluster since the opening API request was failing.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-505)

## How Has This Been Tested?
Manually tested that the format works on an OpenSearch docker container, could not test with an actual Managed OpenSearch deployment due to "restrictive policies" on publicly accessible opensearch domains.

## Checklist


jonathangreen (approved):
🚀

PR #1419, opened 2023-10-02
PP-482 Notifications logs

RishiDiwanTT:
## Description
Added some descriptive logs to the Push Notification methods

## Motivation and Context
The production logs make it difficult to discern what the scripts are doing, additional logging has been added to alleviate the issue.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-482)

## How Has This Been Tested?
Manually ran the scripts for see the effect of the logs.

## Checklist


core/util/notifications.py
```diff
@@ -27,6 +28,8 @@ class PushNotifications:
 
     VALID_TOKEN_TYPES = [DeviceTokenTypes.FCM_ANDROID, DeviceTokenTypes.FCM_IOS]
 
+    log = logging.getLogger("PushNotifications")
```
jonathangreen, line 31: Minor: I've been trying to standardize our logger names, so that its a bit easier to turn on and off debug logging for particular classes. 

I'd suggest making this:
```python
@classmethod
def logger(cls) -> logging.Logger:
    return logging.getLogger(f"{cls.__module__}.{cls.__name__}")
```

Then update the calls to be `cls.logger()` so the logger ends up being named `core.util.PushNotifications`.

jonathangreen (approved):
Logging looks good, just a comment on the logger name.

core/util/notifications.py
RishiDiwanTT: Added a standard LoggerMixin to be used everywhere

RishiDiwanTT: The log ends up being `"name": "core.util.notifications.PushNotifications"`, should there be rules for truncation of the module name?

jonathangreen: That mixin looks great @RishiDiwanTT. The logger name looks good to me, I just mixed it up in my last message.

RishiDiwanTT: I meant for longer modules, for example: the new annotators may end up logging like `core.feed.annotators.loan_and_hold.LoanAndHoldLibraryAnnotator` in which case should we have a max module length, so it truncates to `core.feed.annota...LoanAndHoldLibraryAnnotator`. This is bridge hasn't been crossed yet, so we can defer this decision.

PR #1420, opened 2023-10-02
Fix missing required argument for update_lane_size script (PP-524)

jonathangreen:
## Description

Add default  of `None` to `_db` arg for the `update_lane_size` script.

## Motivation and Context

Working on another issue I noticed that running the update_lane_size script fails with:
```
  File "bin/update_lane_size", line 11, in <module>
    UpdateLaneSizeScript().run()
TypeError: __init__() missing 1 required positional argument: '_db'
```

This appears to have come in as part of #1268.

## How Has This Been Tested?

Ran script locally.

## Checklist

### DIFF
--- a/core/scripts.py
+++ b/core/scripts.py
@@ -2645,7 +2645,7 @@ class ListCollectionMetadataIdentifiersScript(CollectionInputScript):
 
 
 class UpdateLaneSizeScript(LaneSweeperScript):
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
🎸

Should be good once one of the API test checks re-runs.

PR #1426, opened 2023-10-03
Use LoggerMixin

jonathangreen:
## Description

This adds caching to the LoggerMixin class and add a property to make it easier to access from class instances and uses the LoggerMixin to replace a number of places we were calling logging.getLogger.

## Motivation and Context

Reduce the number of duplicate calls to `logging.getLogger`, since we have a mixin class for it.

## How Has This Been Tested?

Running tests.

## Checklist

### DIFF
--- a/api/admin/config.py
+++ b/api/admin/config.py
@@ -1,4 +1,3 @@
-import logging
 import os
 from enum import Enum
 from typing import Optional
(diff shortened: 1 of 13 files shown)


RishiDiwanTT (approved):
Looks excellent :+1:

PR #1427, opened 2023-10-03
PP-139 Facet ordering with last_update enabled for libraries

RishiDiwanTT:
## Description
Added a migration to retrospectively enable it for all libraries.
The admin panel will now show the additional ordering facet type.
image

## Motivation and Context
Aspen OPDS2 feeds need to be sorted by "update time", we already have this data locally within the `licensepool.last_checked` and the `work.last_update_time` attributes. The `Work.last_update_time` attribute gets indexed in the search indexed which enables the facet ordering.

## How Has This Been Tested?
Manually ran the migrations, and tested ordering.
Unit tests have been written for the migration.

## Checklist


jonathangreen (changes requested):
For this one, I don't think we need to add a data migration... 

We don't serialize the defaults out to the database, in the [dict()](https://github.com/ThePalaceProject/circulation/blob/main/core/integration/settings.py#L334-L338) method of the settings class. Specifically for occasions like this, where we want to update defaults, so we don't need to use a migration.

That way if someone has changed the defaults, they keep their settings, but if they are using the defaults they get the new defaults. 

The question arises, what happens if someone changes the defaults, and removes `ORDER_LAST_UPDATE`, but that question is the same with this PR. After the migration and site would be free to do that, and it might break the Aspin indexing. 

I think adding this as a default is a good first step here, without a migration, then we can continue to iterate on this.

RishiDiwanTT:
> For this one, I don't think we need to add a data migration...
> 
> We don't serialize the defaults out to the database, in the [dict()](https://github.com/ThePalaceProject/circulation/blob/main/core/integration/settings.py#L334-L338) method of the settings class. Specifically for occasions like this, where we want to update defaults, so we don't need to use a migration.
> 
> That way if someone has changed the defaults, they keep their settings, but if they are using the defaults they get the new defaults.
> 
> The question arises, what happens if someone changes the defaults, and removes `ORDER_LAST_UPDATE`, but that question is the same with this PR. After the migration and site would be free to do that, and it might break the Aspin indexing.
> 
> I think adding this as a default is a good first step here, without a migration, then we can continue to iterate on this.

Since overriding the defaults would break the entire premise of the PR I have discarded the given approach for a less elegant but more robust solution.
The ORDER_LAST_UPDATE facet is manually added during the `Facets.available_facets` method, thereby skipping the Admin UI entirely, making it impossible for a user to remove the new sorting.
The new sort order will still show up in the feed and hence the Mobile Apps.

jonathangreen:
I've been thinking about this one a bit, and I think the solution here is that instead of adding the facets to the feed, we should just be directing Aspin to use the crawlable feed, which I believe has all the facets enabled by default. That would let them sort by last_updated, without us having to add these facets to the feed that our apps consume.

The biggest change we would need is that the crawlable feed would have to respond with OPDS2. Does that make sense to you @RishiDiwanTT? 

Can you do some testing and make sure the crawlable feed gives us all the facets that aspin currently uses?

RishiDiwanTT:
Closed this PR in favour of the crawlable endpoint

PR #1428, opened 2023-10-03
OPDS Importer Base Classes (PP-500)

jonathangreen:
## Description

Refactor the OPDS importer class hierarchy so that we have some base classes, that the specific importers implement. There are two new base classes: `BaseOPDSImporter` and `BaseODLImporter`. These classes contain all of the shared functionality across `OPDSImporter`, `OPDS2Importer`, `ODLImporter`, `ODL2Importer` and `OPDSForDistributorsImporter`, so its a little easier to see what it overridden and when looking at the classes.

This also tried to make the tests for ODL and ODL2 a bit easier to understand by not having them inherit from each other and instead use a parameterized fixture that picks the correct API for the tests.

## Motivation and Context

This refactor is setting up for PP-502 to replace external integrations with configurationsettings in the collection integrations.

## How Has This Been Tested?

- Did some light local testing 
  - running the importers to make sure they work
  - setting up collections in admin UI
- Running unit tests

## Checklist

### DIFF
--- a/core/opds2_import.py
+++ b/core/opds2_import.py
@@ -193,8 +193,6 @@ class OPDS2Importer(IgnoredIdentifierImporterMixin, BaseOPDSImporter):
             If there is no DataSource with this name, one will be created.
             NOTE: If `collection` is provided, its .data_source will take precedence over any value provided here.
             This is only for use when you are importing OPDS metadata without any particular Collection in mind.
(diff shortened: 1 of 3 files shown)


RishiDiwanTT (approved):
Very minor comments, looks good :+1:

PR #1434, opened 2023-10-04
Update webpub manifest parser to the latest.

jonathangreen:
## Description

Update our webpub-manifest-parser dependency.

Dependabot can't update this because it requires updates to three dependencies in sync: 
- multipledispatch
- palace-webpub-manifest-parser
- pytz

## Motivation and Context

I'm trying to get https://github.com/ThePalaceProject/circulation/pull/1431 to work, which requires us to update flask-babel, which requires some transitive dependency updates that mean we need to update webpub-manifest-parser.

## How Has This Been Tested?

- Running tests in CI

## Checklist

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -24,20 +24,21 @@ tz = ["python-dateutil"]
 
 [[package]]
 name = "attrs"
-version = "21.4.0"
+version = "23.1.0"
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
👍🏽

PR #1437, opened 2023-10-04
Refactor CirculationAPI to remove unused parameters

jonathangreen:
## Description

Remove the unused `part` and `fulfill_part_url` parameters from the fulfill call in CirculationAPI. Instead of passing `internal_format` into each API, let the APIs opt into this with a `CirculationInternalFormatsMixin` mixin class. 

## Motivation and Context

Removing `internal_format` makes sense because only a couple APIs actually use it, the others ignore it. The forthcoming OPDS and OPDS2 APIs will also ignore this, so it made sense to refactor it out.

## How Has This Been Tested?

Running tests.

## Checklist


api/bibliotheca.py
```diff
@@ -138,19 +138,6 @@ class BibliothecaAPI(
...
-    adobe_drm = DeliveryMechanism.ADOBE_DRM
-    findaway_drm = DeliveryMechanism.FINDAWAY_DRM
-    delivery_mechanism_to_internal_format = {
-        (Representation.EPUB_MEDIA_TYPE, adobe_drm): "ePub",
-        (Representation.PDF_MEDIA_TYPE, adobe_drm): "PDF",
-        (None, findaway_drm): "MP3",
-    }
-    internal_format_to_delivery_mechanism = {
-        v: k for k, v in list(delivery_mechanism_to_internal_format.items())
-    }
```
RishiDiwanTT, line 152: This seemed to still have been in use within bibliotheca

api/opds_for_distributors.py
```diff
@@ -300,57 +284,65 @@ def fulfill(
...
+        # bearer token.
+        _db = Session.object_session(licensepool)
+        credential = self._get_token(_db)
 
-        # We couldn't find an acquisition link for this book.
-        raise CannotFulfill()
+        # Build a application/vnd.librarysimplified.bearer-token
+        # document using information from the credential.
+        now = utc_now()
+        assert credential.expires is not None
```
RishiDiwanTT, line 327: Does an AssertionError in production lead to an INTERNAL_SERVER_ERROR problem detail?

RishiDiwanTT (commented):
Some minor comments, I mostly am confused about the removal of internal formats from the Bibliotheca API.

RishiDiwanTT (approved):
All good :+1:

PR #1438, opened 2023-10-05
PP-414 Deleted all deprecated Acquisition feed and annotator classes :axe: 

RishiDiwanTT:
## Description
This includes any of the caching scripts that used the feed directly.

## Motivation and Context
The new feed and annotator implementations are in use currently, the old implementations were deprecated and only used in caching scripts which are now redundant due the CDN implementations.

## How Has This Been Tested?
All unit tests that referenced the old classes were deleted. The rest all pass.
Manually tested some of the feed endpoints.

## Checklist


jonathangreen:
@RishiDiwanTT I think we need a migration here as well to drop the unused cache DB columns right?

RishiDiwanTT:
> @RishiDiwanTT I think we need a migration here as well to drop the unused cache DB columns right?

@jonathangreen I was thinking of that as a follow up, since that should also include deleting all the lane caching scripts and the CachedFeed table entirely. Along with all the code that references the caches.
Unless they still have some uses that I'm not seeing.

jonathangreen (approved):
This looks good to me 🔥 

The only comment I have is that we are losing somewhat significant test coverage in two files, probably from removing some of the tests in this PR:
https://app.codecov.io/gh/ThePalaceProject/circulation/pull/1438/indirect-changes

[core/util/opds_writer.py](https://app.codecov.io/gh/ThePalaceProject/circulation/pull/1438/blob/core/util/opds_writer.py) -6.80%
[core/feed/opds.py](https://app.codecov.io/gh/ThePalaceProject/circulation/pull/1438/blob/core/feed/opds.py) -4.08%

It might be a good idea to see if there are any tests we can move over (or write) to preserve coverage in these files.

jonathangreen:
> > @RishiDiwanTT I think we need a migration here as well to drop the unused cache DB columns right?
> 
> @jonathangreen I was thinking of that as a follow up, since that should also include deleting all the lane caching scripts and the CachedFeed table entirely. Along with all the code that references the caches. Unless they still have some uses that I'm not seeing. I thought having a large PR with a lot of deletes tends to be confusing for review.

I created a follow up ticket in JIRA to track this work PP-554, can you take a look and make sure it captures what you were thinking in this comment?

RishiDiwanTT:
> This looks good to me 🔥
> 
> The only comment I have is that we are losing somewhat significant test coverage in two files, probably from removing some of the tests in this PR: https://app.codecov.io/gh/ThePalaceProject/circulation/pull/1438/indirect-changes
> 
> [core/util/opds_writer.py](https://app.codecov.io/gh/ThePalaceProject/circulation/pull/1438/blob/core/util/opds_writer.py) -6.80% [core/feed/opds.py](https://app.codecov.io/gh/ThePalaceProject/circulation/pull/1438/blob/core/feed/opds.py) -4.08%
> 
> It might be a good idea to see if there are any tests we can move over (or write) to preserve coverage in these files.

These ended up being mostly AtomFeed methods that were no longer in use due to the refactor. I deleted the code that was not in use anymore.

PR #1439, opened 2023-10-06
Fully type hint `core/util/xmlparser.py` (PP-500)

jonathangreen:
## Description

Fully type hint `core/util/xmlparser.py`. Went down a bit of a rabbit hole 🐰 trying to type annotate the circulation API code and ended up type annotating the `core/util/xmlparser.py` class.

## Motivation and Context

As part of PP-500. I'm adding some type annotations to the `CirculationAPI` classes. Since both axis and bibliotecha use `xmlparser.py`, it seemed like we needed to annotate it. Turned out to be more work then I expected 😅 

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 import html
 import json
 import logging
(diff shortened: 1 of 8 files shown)


RishiDiwanTT (approved):
Overall look good :+1:

PR #1440, opened 2023-10-06
Fix flaky quicksight test

jonathangreen:
## Description

Because the libraries we were getting back from the DB were not ordered, we could not rely on the order of the libraries in the tests, which was causing test failures. 

It seems desirable to have a stable sort of the libraries we pass to quicksight anyway, so I added order to the query that generates our library list. I also modified the query slightly, so it just returns the library name, instead of the whole ORM Library object, since we are only using the name.

## Motivation and Context

Test failures in CI:
https://github.com/ThePalaceProject/circulation/actions/runs/6431978534/job/17465903012#step:8:410

## How Has This Been Tested?

- Running tests.

## Checklist

### DIFF
--- a/api/admin/controller/quicksight.py
+++ b/api/admin/controller/quicksight.py
@@ -3,6 +3,7 @@ from typing import Dict
 
 import boto3
 import flask
+from sqlalchemy import select
 
 from api.admin.model.quicksight import (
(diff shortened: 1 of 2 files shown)


RishiDiwanTT (approved):
Looks good :+1:

PR #1443, opened 2023-10-07
Use codecov flags (PP-499)

jonathangreen:
## Description

Try running codecov with [flags](https://docs.codecov.com/docs/flags) enabled. 

## Motivation and Context

This should hopefully stop codecov from failing our CI as much because it won't think that coverage has dropped until all the CI reports have been submitted. 

Jira: PP-499

## How Has This Been Tested?

- Running codecov on this PR

## Checklist


RishiDiwanTT (approved):
:+1:

PR #1444, opened 2023-10-09
Force content encoding during OPDS serialization

RishiDiwanTT:
## Description
By forcing the encoding we ensure that unicode characters, which exist in large number due to the varying languages of the books available, are encoded correctly.

## Motivation and Context
The Gorgon feeds are failing due to the summaries of several books having unicode characters and failing to serialize correctly.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-520) 

## How Has This Been Tested?
Manually tested the feeds with multi lingual feeds.

## Checklist


core/feed/serializer/opds.py
```diff
@@ -381,7 +381,7 @@ def _serialize_data_entry(self, entry: DataEntry) -> etree._Element:
 
     @classmethod
     def to_string(cls, element: etree._Element) -> bytes:
```
jonathangreen, line 383: It seems really weird here to have a method called `to_string` returning `bytes`. It seems like we should probably either return a string, or rename the method to `to_bytes`.

jonathangreen (approved):
Fix looks good. I have a comment there about function naming, but that situation existed before this PR, so feel free to just merge as is if you want.

PR #1445, opened 2023-10-09
PP-309 Switched the marc exporter to upload files based on size

RishiDiwanTT:
## Description
Earlier this was based on number of works per file, this would sometimes fail to be the minimum required size for a multipart upload to s3 [[ref](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html)].
Switching to content size based uploads ensures we don't have this problem.

## Motivation and Context
The MARC Record Generator for some libraries in California is not showing all of the categories. The MARC generator displays categories for the swimlanes; however, some libraries are missing categories.
This was because the part-files were sometimes below the 5MB threshold for s3.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-309)

## How Has This Been Tested?
Wrote a new test for this.
Manually exported marc files, ensuring the size constraints are followed.

## Checklist


core/marc.py
```diff
@@ -710,14 +708,15 @@ def records(
                         work, annotator, force_refresh, self.integration
                     )
                     if record:
-                        this_batch.write(record.as_marc())
-                this_batch_size += pagination.this_page_size
-                if this_batch_size >= upload_batch_size:
+                        record_bytes = record.as_marc()
+                        this_batch_bytes += len(record_bytes)
+                        this_batch.write(record_bytes)
+                if this_batch_bytes >= self.MINIMUM_UPLOAD_BATCH_SIZE_BYTES:
```
jonathangreen, line 714: Minor: It might clearer to just directly use the length of the `BytesIO` object rather then track it separately in `this_batch_bytes`. Something like:

```python
if this_batch.getbuffer().nbytes >= self.MINIMUM_UPLOAD_BATCH_SIZE_BYTES:
```

jonathangreen (approved):
Looks good!

PR #1449, opened 2023-10-10
PP-536 String coercion for fcm messages

RishiDiwanTT:
## Description
FCM only allows string type message data

Added a test to ensure we do not send bad data again

## Motivation and Context
Sending Loan notifications sometimes errors out due to integer type data
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-536)

## Checklist


jonathangreen (approved):
Looks good

PR #1451, opened 2023-10-11
Fix some issues found during testing of push notification scripts (PP-570)

jonathangreen:
## Description

I saw a couple of issues while testing the push notification scripts:
- We were using authorization_identifier for the log messages in some places and external_identifier in others, making it a bit hard to correlate log messages. I standardized on authorization_identifier.
- We were getting `UnregisteredError` exceptions coming back from firebase from some of the tokens on minotaur. This is probably expected, it looks like tokens are automatically unregistered when apps are uninstalled. I think the correct way to deal with this is to just delete the token.
- I added a helper to send the notifications that deals with the exceptions and makes sure our data is strings before sending the push notifications. 

Questions: 
@RishiDiwanTT were we using `send_all` instead of `send` because we were hitting some API limit? If so this code may have to change, I'm assuming here we can just use `send` instead, so its easier to deal with the exceptions on a message by message basis.

## Motivation and Context

Dealing with testing done for PP-480 and PP-481.

## How Has This Been Tested?

Running CI tests.

## Checklist


RishiDiwanTT (approved):
Looks good :+1: 
One optional comment.

PR #1452, opened 2023-10-11
Disable codecov per flag status

jonathangreen:
## Description

See if we can disable per-flag status updates, since they are causing PRs to fail.

## Motivation and Context

Trying to get a reasonable codecov config.

## How Has This Been Tested?

Running codecov in CI

## Checklist


dbernstein (approved):
👍

PR #1454, opened 2023-10-11
Fix logging for monitors (PP-570)

jonathangreen:
## Description

This makes sure that we initialize logging before running monitors.

## Motivation and Context

While testing the notification scripts today, I noticed that I wasn't seeing any logging from monitors.

## How Has This Been Tested?

- Tests in CI

## Checklist


RishiDiwanTT (approved):
Looks good :+1:

PR #1455, opened 2023-10-11
Add PALACE_REPORTING_NAME env variable to time report subject

dbernstein:
…ime tracking report email.
## Description
Pulls the PALACE_CM_NAME env variable into the subject of the audio time tracking report email.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-539

## Checklist

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -61,8 +61,8 @@ class Configuration(ConfigurationConstants):
     # Environment variable for temporary reporting email
     REPORTING_EMAIL_ENVIRONMENT_VARIABLE = "SIMPLIFIED_REPORTING_EMAIL"
 
-    # Environment variable for collection manager name
-    CM_NAME_ENVIRONMENT_VARIABLE = "PALACE_CM_NAME"
+    # Environment variable for used to distinguish one CM environment from another in reports
+    REPORTING_NAME_ENVIRONMENT_VARIABLE = "PALACE_REPORTING_NAME"
 
     # ConfigurationSetting key for the base url of the app.
     BASE_URL_KEY = "base_url"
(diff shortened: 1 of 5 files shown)


core/config.py
```diff
@@ -61,6 +61,9 @@ class Configuration(ConfigurationConstants):
     # Environment variable for temporary reporting email
     REPORTING_EMAIL_ENVIRONMENT_VARIABLE = "SIMPLIFIED_REPORTING_EMAIL"
 
+    # Environment variable for collection manager name
+    CM_NAME_ENVIRONMENT_VARIABLE = "PALACE_CM_NAME"
```
jonathangreen, line 65: Minor comment on the name of this env variable. We might want to change this to either: 
- PALACE_NAME
- PALACE_REPORTING_NAME

I'm not sure having CM in the name adds much specificity to the variable.

core/jobs/playtime_entries.py
```diff
@@ -127,9 +127,10 @@ def do_run(self):
         formatted_start_date = start.strftime(self.REPORT_DATE_FORMAT)
         formatted_until_date = until.strftime(self.REPORT_DATE_FORMAT)
         report_date_label = f"{formatted_start_date} - {formatted_until_date}"
-        email_subject = (
-            f"Playtime Summaries {formatted_start_date} - {formatted_until_date}"
-        )
+
+        cm_name = os.environ.get(Configuration.CM_NAME_ENVIRONMENT_VARIABLE)
+
+        email_subject = f"{cm_name}: Playtime Summaries {formatted_start_date} - {formatted_until_date}"
```
jonathangreen, line 133: If that environment variable isn't set, `os.environ.get` will return `None`. We might want to gracefully handle that case. As this is written the subject will be:
```
None: Playtime Summaries ...
```

jonathangreen (commented):
This looks good, but I'd like to see a couple things addressed before its merged:
- Minor comment about env var naming. Feel free to ignore if you disagree. 
- Need to handle the case where the variable isn't set. 
- Should add the new env variable to the readme.

jonathangreen (approved):
Looks good!

PR #1462, opened 2023-10-13
Upgrade to Ubuntu 22.04 and Python 3.10 (PP-506)

jonathangreen:
## Description

Update the base image that we use to Ubuntu 22.04. This will let us deploy using Python 3.10. Opening the way for us to drop support for Python 3.8 and 3.9.

## Motivation and Context

Right now the CM runs on Python 3.8. We should upgrade to a newer Python version, so that we can eventually drop Python 3.8. Python 3.8 will be unsupported by 14 Oct 2024, but I’d like to move off it it long before then.

## How Has This Been Tested?

- Tests in CI
- Built image and ran it locally, everything seemed okay

## Checklist

### DIFF
--- a/docker/ci/test_webapp.sh
+++ b/docker/ci/test_webapp.sh
@@ -17,7 +17,7 @@ check_service_status "$container" /etc/service/nginx
 check_service_status "$container" /etc/service/uwsgi
 
 # Wait for UWSGI to be ready to accept connections.
-timeout 120s grep -q 'WSGI app .* ready in [0-9]* seconds' <(docker logs "$container" -f 2>&1)
+timeout 240s grep -q 'WSGI app .* ready in [0-9]* seconds' <(docker logs "$container" -f 2>&1)
 
 # Make sure the web server is running.
 healthcheck=$(docker exec "$container" curl --write-out "%{http_code}" --silent --output /dev/null http://localhost/healthcheck.html)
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good!

docker/ci/test_webapp.sh
```diff
@@ -17,7 +17,7 @@ check_service_status "$container" /etc/service/nginx
 check_service_status "$container" /etc/service/uwsgi
 
 # Wait for UWSGI to be ready to accept connections.
-timeout 120s grep -q 'WSGI app .* ready in [0-9]* seconds' <(docker logs "$container" -f 2>&1)
+timeout 240s grep -q 'WSGI app .* ready in [0-9]* seconds' <(docker logs "$container" -f 2>&1)
```
tdilauro, line 20: Also, wow! Is it really taking four minutes to spin up?

jonathangreen: No, looking at the logs where it failed, it seems like it was taking about 130s, just beyond the previous timeout. But I figured I'd give it a bit of a safety factor and just double what we had before.

jonathangreen:
I'm holding off on merging this until I cut the new staging release, so we have a bit of time to test this out on minotaur before it hits staging.

PR #1464, opened 2023-10-16
PP-483 Added last notified date to loan and hold models

RishiDiwanTT:
## Description
This allows us to run the notification scripts multiple times a day without bombarding the same patron multiple times with the same content.
The notifications code updates the timestamp every time a notification is sent out, and the script queries incorporate the new column.

## Motivation and Context
Currently, the push notification scripts only run once per day. 
This is because additional runs would send out duplicate notifications.
We need to add some way of disallowing the notification from going out twice.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-483)

## How Has This Been Tested?
Updated the unit tests to cover the new cases.

## Checklist


core/util/notifications.py
```diff
@@ -135,6 +136,7 @@ def send_loan_expiry_message(
             f"Patron {loan.patron.authorization_identifier} has {len(tokens)} device tokens. "
             f"Sending loan expiry notification(s)."
         )
+        loan.patron_last_notified = utc_now().date()
```
jonathangreen, line 139: If firebase gives us a transient exception in `send_messages`, this might cause the notification to be dropped, instead of retried on the next script run. Since we set the notification date before we know that we sent a message.

core/util/notifications.py
```diff
@@ -204,6 +206,7 @@ def send_holds_notifications(cls, holds: list[Hold]) -> list[str]:
             if hold.patron.authorization_identifier:
                 data["authorization_identifier"] = hold.patron.authorization_identifier
 
+            hold.patron_last_notified = utc_now().date()
```
jonathangreen, line 209: Same comment about when we set the notified date as above.

jonathangreen (commented):
A couple minor comments on this one:
1. Maybe we should set the notified field after we've successfully sent a notification, so that we gracefully retry in the case of a transient failure from firebase.
2. Since the ticket is to run the scripts more often, we should probably update the cron schedule in this PR as well.

core/util/notifications.py
RishiDiwanTT: I've updated both instances of the code to only update the `patron_last_notified` if at least one notification was sent successfully.

jonathangreen (approved):

PR #1469, opened 2023-10-18
PP-598 Removed the google analytics provider

RishiDiwanTT:
## Description
Removed the provider along with its test cases

Changed the analytics service test cases to use the s3 provider instead

## Motivation and Context
The code didn't work anymore and was deprecated.

## How Has This Been Tested?
All unit tests run post removal.

## Checklist


jonathangreen (approved):
This looks good. 

The only question / concern I have is: what happens when this code runs against a DB that already has a GA analytics provider configured? Will those settings still show in the admin UI? Will analytics fail to load since the class doesn't exist?

jonathangreen (approved):
Looks good

PR #1470, opened 2023-10-18
Remove circulation api post processor (PP-501)

jonathangreen:
## Description

As a follow up to #1442, this removes the `CirculationFulfillmentPostProcessor` class. All the logic has been moved into the OPDSAPI and OPDS2API classes.

## Motivation and Context

`CirculationFulfillmentPostProcessor` were only used to work around OPDS importers not having an API class. Now that we have a proper API class there, there is no reason to keep them around.

## How Has This Been Tested?

Running tests.

## Checklist


tests/api/test_axis.py
```diff
@@ -798,9 +799,7 @@ def test_integration_settings_url(
             api = MockAxis360API(axis360.db.session, axis360.collection)
             assert api.base_url == expected
         else:
-            pytest.raises(
-                ProblemError, MockAxis360API, axis360.db.session, axis360.collection
-            )
+            pytest.raises(ProblemError, Axis360Settings, **settings)
```
tdilauro, line 802: #1472 changes here, as well, I think.

tdilauro (approved):
This looks good!

There are a couple of places where it looks like there's a conflict with #1472, but they should be straightforward to resolve.

PR #1472, opened 2023-10-19
Add some helper methods to load and save config settings

jonathangreen:
## Description

While I was working on PP-501, I noticed we are loading / updating config settings in a number of places. This centralizes that in a couple functions, and adds classmethods for them to `HasIntegrationConfiguration`.

## Motivation and Context

Easier to work with integration settings.

## How Has This Been Tested?

- Running tests in CI

## Checklist

### DIFF
--- a/tests/api/admin/controller/test_library.py
+++ b/tests/api/admin/controller/test_library.py
@@ -210,7 +210,7 @@ class TestLibrarySettings:
                 FacetConstants.ORDER_TITLE,
                 FacetConstants.ORDER_AUTHOR,
             ] == settings_dict.get("facets_enabled_order")
-            assert ["French"] == settings_dict.get("large_collection_languages")
+            assert ["fre"] == settings_dict.get("large_collection_languages")
 
     def test_libraries_post_errors(self, settings_ctrl_fixture):
         with settings_ctrl_fixture.request_context_with_admin("/", method="POST"):
--- a/tests/api/test_authenticator.py
+++ b/tests/api/test_authenticator.py
@@ -682,9 +682,6 @@ class TestLibraryAuthenticator:
         # propagated.
         # Create an integration destined to raise CannotLoadConfiguration..
         library = db.default_library()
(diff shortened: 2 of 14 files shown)


tests/api/test_authenticator.py
```diff
@@ -682,24 +682,16 @@ def test_configuration_exception_during_from_config_stored(
...
         # The LibraryAuthenticator exists but has no AuthenticationProviders.
         assert auth.basic_auth_provider is None
 
-        # Both integrations have left their trace in
-        # initialization_exceptions.
-        not_configured = auth.initialization_exceptions[(misconfigured.id, library.id)]
-        assert isinstance(not_configured, CannotLoadConfiguration)
-        assert "Could not instantiate MilleniumPatronAPI" in str(not_configured)
-
+        # The integration has left it trace in initialization_exceptions.
```
tdilauro, line 692: Minor: typo -> "...left [its] trace..."

tests/api/admin/controller/test_library.py
```diff
@@ -210,7 +210,7 @@ def test_libraries_get_with_multiple_libraries(
                 FacetConstants.ORDER_TITLE,
                 FacetConstants.ORDER_AUTHOR,
             ] == settings_dict.get("facets_enabled_order")
-            assert ["French"] == settings_dict.get("large_collection_languages")
+            assert ["fre"] == settings_dict.get("large_collection_languages")
```
tdilauro, line 213: It's not obvious why this has changed.

tdilauro (approved):
Looks good to me!

tests/api/admin/controller/test_library.py
jonathangreen: It changed because the call to update settings on line 176 now validates the settings before saving, which normalizes the language codes. That normalization is tested elsewhere, so I think its fine to just update the test here.

tdilauro: I think it's fine, too. I just didn't understand it.

PR #1476, opened 2023-10-20
PP-554 Remove cachedfeeds

RishiDiwanTT:
## Description
Removed the CachedFeed table and the Work opds entry caching columns.
Mostly this is just the removal of any reference to the tables and deleting any scripts used to generate them.
Additionally, there is one logical change in how the `Lane.max_cache_age` is treated.
Originally this was used to maintain the CachedFeed entries for a specified amount of time, but now since we've outsourced the caching further down the request-response pipeline to a CDN, we must also make use of the `Lane.max_cache_age` values further down the line.
For this reason we now return the `Lane.max_cache_age` in the `Cache-Control` headers so that the CDNs may cache the data accordingly.

## Motivation and Context
Since the refactored Feeds and Annotators do not make use of the caching mechanisms and the use of CDNs have been introduced we no longer need the CacheFeed table and supporting functionalities.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-554)

## How Has This Been Tested?
All unit tests are passing.
Manually hit several different feed pages.

## Checklist


core/model/cachedfeed.py
```diff
@@ -1,400 +1,16 @@
 # CachedFeed, WillNotGenerateExpensiveFeed
```
jonathangreen, line 1: Minor: This comment can probably be dropped.

core/model/cachedfeed.py
```diff
@@ -1,400 +1,16 @@
...
-    "ix_cachedfeeds_library_id_lane_id_type_facets_pagination",
-    CachedFeed.library_id,
-    CachedFeed.lane_id,
-    CachedFeed.type,
-    CachedFeed.facets,
-    CachedFeed.pagination,
-)
-
 
 class WillNotGenerateExpensiveFeed(Exception):
```
jonathangreen, line 15: This exception now (maybe was before?) unused and can go.

api/config.py
```diff
@@ -282,14 +280,3 @@ def cipher(cls, key: bytes) -> PKCS1OAEP_Cipher:
...
-
-# We changed Configuration.DEFAULT_OPDS_FORMAT, but the Configuration
-# class from core still has the old value. Change that one to match,
-# so that core code that checks this constant will get the right
-# value.
-#
-# TODO: We should come up with a better solution for this, probably
-# involving a registry of Configuration objects that returns the
-# appropriate one in any situation. This is a source of subtle bugs.
-CoreConfiguration.DEFAULT_OPDS_FORMAT = Configuration.DEFAULT_OPDS_FORMAT
```
jonathangreen, line 295: 🎉 for this going away

api/controller.py
```diff
@@ -249,10 +248,7 @@ def load_facets_from_request(self, *args, **kwargs):
             ):
                 return NO_SUCH_LANE.detailed(_("Lane does not exist"))
 
-        if (
-            isinstance(facets, BaseFacets)
-            and getattr(facets, "max_cache_age", None) is not None
-        ):
+        if isinstance(facets, BaseFacets):
             # A faceting object was loaded, and it tried to do something nonstandard
```
jonathangreen, line 252: This comment seems out of date, now that we aren't checking `max_cache_age` here. 

Can this whole `if` block just go away? It seems that it was mainly concerned with facet caching, which we aren't doing locally anymore.

core/lane.py
```diff
@@ -325,10 +305,9 @@ def load_entrypoint(cls, name, valid_entrypoints, default=None):
     @classmethod
     def load_max_cache_age(cls, value):
         """Convert a value for the MAX_CACHE_AGE_NAME parameter to a value
```
jonathangreen, line 307: Minor: This comment trails off...

core/lane.py
```diff
@@ -1413,7 +1386,7 @@ class WorkList:
     def visible(self) -> bool:
         return True
 
-    def max_cache_age(self, type):
+    def max_cache_age(self):
         """Determine how long a feed for this WorkList should be cached
         internally.
```
jonathangreen, line 1391: Since we don't do internal caching, this comment is out of date.

core/model/__init__.py
```diff
@@ -515,11 +515,7 @@ def _bulk_operation(self):
     SAMLFederation,
 )
 from core.model.admin import Admin, AdminRole
-from core.model.cachedfeed import (
-    CachedFeed,
-    CachedMARCFile,
-    WillNotGenerateExpensiveFeed,
-)
+from core.model.cachedfeed import CachedMARCFile, WillNotGenerateExpensiveFeed
```
jonathangreen, line 518: When we remove `WillNotGenerateExpensiveFeed`, it should also be removed from this import.

core/model/work.py
```diff
@@ -59,12 +59,7 @@
 
 # Import related models when doing type checking
 if TYPE_CHECKING:
-    from core.model import (  # noqa: autoflake
-        CachedFeed,
-        CustomListEntry,
-        Library,
-        LicensePool,
-    )
+    from core.model import CustomListEntry, Library, LicensePool  # noqa: autoflake
```
jonathangreen, line 62: Minor: its unrelated to this PR, but I noticed it because of the changes on this line. We can remove the `# noqa: autoflake` comment here.

alembic/versions/20231020_7fceb9488bc6_drop_cachedfeeds.py
```diff
@@ -0,0 +1,92 @@
...
+    op.drop_index("ix_cachedfeeds_lane_id", table_name="cachedfeeds")
+    op.drop_index("ix_cachedfeeds_library_id", table_name="cachedfeeds")
+    op.drop_index(
+        "ix_cachedfeeds_library_id_lane_id_type_facets_pagination",
+        table_name="cachedfeeds",
+    )
+    op.drop_index("ix_cachedfeeds_timestamp", table_name="cachedfeeds")
+    op.drop_index("ix_cachedfeeds_work_id", table_name="cachedfeeds")
+    op.drop_table("cachedfeeds")
+    op.drop_column("works", "simple_opds_entry")
```
jonathangreen, line 31: We should be able to remove the `simple_opds_entry` field from the editions table as well. https://github.com/ThePalaceProject/circulation/blob/cd8ef179b0ec277151816522ebef7b34d3f2e403/core/model/edition.py#L141 It seems unused after these changes (possibly before as well?)

jonathangreen (approved):
This looks great @RishiDiwanTT 🔥 

Just a few small comments on the PR. I think this file can be removed as well: `core/model/files/view_definitions.sql` since it references `simple_opds_entry` and doesn't appear to be used.
https://github.com/ThePalaceProject/circulation/blob/de52bdaec1b330c45701476df73cd5e83a064af9/core/model/files/view_definitions.sql#L1-L40

I found myself wondering as I went through the code review if all of the `max_age` stuff here makes sense anymore, or if it should come out as well. Any thoughts on that? I think this PR can go in as is and we can handle that in a future PR if it needs to come out.

alembic/versions/20231020_7fceb9488bc6_drop_cachedfeeds.py
RishiDiwanTT: Yes, I think before as well. Removing.

api/controller.py
RishiDiwanTT: Yes, facets.max_cache_age is no longer used, will have to remove all occurences.

jonathangreen (approved):

jonathangreen:
There is a minor conflict that needs to be resolved, but looks good!

PR #1477, opened 2023-10-24
PP-564 Import duration from palace marketplace

RishiDiwanTT:
## Description
In case a "duration" is available in the metadata of any publication of an OPDS2/ODL2 feed, we import it into the Edition.
The OPDS1.2/2.0 serializers export this value in the feeds as well.

## Motivation and Context
We now get audiobook duration metadata in the feeds from the Palace Marketplace. We want to include this metadata in the information we provide for audiobooks in the mobile apps, so we need to add this information to the feeds we provide to the apps.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-564)

## How Has This Been Tested?
Manually tested the import and export with duration values.
Updated the unit tests.

## Checklist


core/feed/serializer/opds.py
```diff
@@ -153,6 +153,10 @@ def serialize_work_entry(self, feed_entry: WorkEntryData) -> etree._Element:
                     feed_entry.subtitle.text,
                 )
             )
+        if feed_entry.duration is not None:
+            entry.append(
+                OPDSFeed.E(f"{{{OPDSFeed.ATOM_NS}}}duration", str(feed_entry.duration))
```
jonathangreen, line 158: I think we should align this with the XML OPDS 1.2 feed from DeMarque. Which would mean that we should put it into the DCTERMS_NS, rather then atom. 

Here is how is how it looks when I pull the DeMarque XML feed: 
`<dcterms:duration>57894</dcterms:duration>`

core/model/edition.py
```diff
@@ -126,6 +136,8 @@ class Edition(Base, EditionConstants):
 
     medium = Column(MEDIUM_ENUM, index=True)
 
+    duration = Column(Float, nullable=True)
```
jonathangreen, line 139: I think it would be helpful to add a small comment here with some details of what this number represents. We can probably link out to https://github.com/readium/webpub-manifest/tree/master/contexts/default#duration-and-number-of-pages. It looks like its the float number of seconds in an audiobook.

core/feed/annotator/base.py
```diff
@@ -304,6 +304,9 @@ def annotate_work_entry(
         if edition.series:
             computed.series = self.series(edition.series, edition.series_position)
 
+        if edition.duration is not None:
```
jonathangreen, line 307: Codecov is saying these lines aren't covered. Is it just being broken, or are the new tests bypassing this line somehow?

jonathangreen (commented):
This looks good Rishi! Just a couple small comments there.

core/feed/annotator/base.py
RishiDiwanTT: Didn't create a test of a serialisation without duration, will create one for completeness

jonathangreen (approved):
I think the migration will need updating, since another migration merged in since this PR was created. Otherwise this is good to go!

PR #1481, opened 2023-10-25
Type check collection.py (PP-503)

jonathangreen:
## Description

Add type checking information to `collection.py` before refactoring.

## Motivation and Context

First step for PP-503.

## How Has This Been Tested?

Running tests locally and in CI.

## Checklist

- [X] I have updated the documentation accordingly.
- [X] All new and existing tests passed.

PR #1482, opened 2023-10-25
Remove unused functions from collection model (PP-503)

jonathangreen:
## Description

This is a follow up to #1481. Remove a number of functions that are not used on collections. It looks like they existed to support the metadata wrangler at some point.

## Motivation and Context

Cleaning up the collection model before refactoring it.

## How Has This Been Tested?

Running CI.

## Checklist


RishiDiwanTT (approved):
Looks good :+1:

PR #1483, opened 2023-10-26
No retries for web application based requests

RishiDiwanTT:
Remove retries for any request that are within the context of the web application.
This is to ensure we do not hang web requests due to issues with third party APIs.
Also add a log for request times.
## Motivation and Context
Having retry logic for outgoing requests during an incoming web request is a hindrance, and should be removed.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-586)

## How Has This Been Tested?
Manually tested the retry failures within the /loans requests, and script based retries with the axis_monitor.
Added a unit test for the web application reduced retries.

## Checklist

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -28,6 +28,7 @@ from core.model import (
 from core.service.container import Services, container_instance
 from core.util import LanguageCodes
 from core.util.cache import CachedData
+from core.util.http import HTTP
 from scripts import InstanceInitializationScript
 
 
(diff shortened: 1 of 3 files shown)


jonathangreen:
This looks good @RishiDiwanTT. I know you mentioned this in planning for this ticket, and I disagreed, but I've come around. Can you also reduce the timeout on requests in the web context to 5s. We can use the data gathered by the added logging to reduce it further if we need.

jonathangreen (approved):

PR #1492, opened 2023-11-01
PP-637 Duration information for Overdrive

RishiDiwanTT:
## Description
Added duration extraction during the overdrive format sweep.
This is done via the `book_info_to_metadata` method. 
The changes to the test file `audiobook.json` are primarily the file formatting and adding the `duration` attribute to the `formats`.

## Motivation and Context
Duration information is available from some feed publishers and should be recorded within the CM.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-637)

## How Has This Been Tested?
Manually tested the `overdrive_format_sweep` script, and observed durations added to the DB.
Updated the unit tests.

## Checklist


tests/api/files/overdrive/audiobook.json
```diff
@@ -1 +1,177 @@
...
+{
```
jonathangreen, line 1: Maybe we should we leave this fixture stored without new-lines, since that is what gets served to us from the overdrive API?

jonathangreen (approved):
Minor comment about the test fixture, but this looks great!

PR #1494, opened 2023-11-01
Remove external integrations from collection integrations (PP-503)

jonathangreen:
## Description

This PR changes the data model for our collections in a number of ways with the goal of removing all reliance on external integrations from our collections.

All the database columns and tables "removed" by this PR have just been renamed on the Python side, but left in place for now. Once this goes into a release https://github.com/ThePalaceProject/circulation/pull/1498 can be merged to remove the now unused tables and columns.

The major changes: 
- The `external_integration_id` column has been removed.
- Collections no longer have a `name` column. 
  - Previously both integrations and collections had names, this meant that they could get out of sync. Now collection names are driven by the name of their integration. A migration handles making sure that the existing integrations that have gotten out of sync are updated to have their collections name.
- The `collections_libraries` table has been removed. 
  - Collections used to be able to be related to libraries in two ways: through the `collections_libraries` table, and via their `integration_library_configuration`.  This situation was confusing, and meant that these could get out of sync. 
- The collection table no longer uses its `external_account_id` column. This was duplicated between the table and the `settings` before, it now lives only in the `settings`. As with the others the column is still present, but is renamed and unused and will be dropped once this is rolled out everywhere.
- The `settings` column get a [gin](https://www.postgresql.org/docs/current/gin-intro.html) index, to make it easier to query the jsonb column.
- Some of the properties on the collection class have been moved out to the specific license provider API that used them. These were on the collection model because previously some providers didn't have their own API classes.
   - There are still some of these that could be moved into a more suitable location, but that can happen in a follow up PR.
- The collection controller and collection self test controller have been updated to use the integration settings controller base classes. This lets us keep all the logic for updating these settings isolated there. The biggest consequence of this change is that we now pass around the integration ID rather then the collection ID when referencing collections and collection self tests. 
  - This is transparent to the admin UI, since it just sends back the IDs given to it when it queries for the collection settings.
  - This makes the collection controllers act like the other controllers that modify integrations, rather then using different IDs in the collection controller.
- The collection self test controller removes the special case that used to exist for OPDS importer collections, since they now have their own API class.
  - This PR removes the self test functionality for the OPDS 1.x importer class, since we don't have any OPDS 1.x collections in our production CMs. [Here is a ticket](https://ebce-lyrasis.atlassian.net/browse/PP-674) to add self tests for the collection classes missing them (OPDS 2, OPDS 2 + ODL), but this is the situation we've been in for awhile, so no impact from this PR. This PR will make it easier to add these self tests by just implementing the self test interface in the api classes. 
  - There is also a DB migration to apply these database changes and an associated migration test.
- All the major files touched by this PR are now fully type hinted, and added to the strict mypy group in `pyproject.toml`. 

## Motivation and Context

Remove external integrations from collections PP-503.

## How Has This Been Tested?

- Light local testing with the CM admin UI:
  - Setting up collections
  - Running migrations
  - Associating and disassociating libraries
- CI tests

## Checklist


api/admin/controller/patron_auth_service_self_tests.py
```diff
@@ -2,116 +2,54 @@
...
-    IntegrationConfiguration,
-    IntegrationLibraryConfiguration,
-)
+from core.model.integration import IntegrationConfiguration
 from core.util.problem_detail import ProblemDetail, ProblemError
 
 
-class PatronAuthServiceSelfTestsController:
+class PatronAuthServiceSelfTestsController(
+    IntegrationSelfTestsController[AuthenticationProviderType]
```
jonathangreen, line 18: Much of what used to exist in this class has been moved to `IntegrationSelfTestsController`, so it can be shared with `CollectionSelfTestsController`.

api/admin/problem_details.py
```diff
@@ -168,15 +168,6 @@
     detail=_("You must identify the analytics service by its name."),
 )
 
-COLLECTION_NAME_ALREADY_IN_USE = pd(
```
jonathangreen, line 171: This has been replaced with the more generic `INTEGRATION_NAME_ALREADY_IN_USE` problem detail, since collections take their names from integrations now.

api/controller.py
```diff
@@ -650,13 +652,27 @@ def load_licensepools(self, library, identifier_type, identifier):
...
         pools = (
-            _db.query(LicensePool)
-            .join(LicensePool.collection)
-            .join(LicensePool.identifier)
-            .join(Collection.libraries)
-            .filter(Identifier.type == identifier_type)
-            .filter(Identifier.identifier == identifier)
-            .filter(Library.id == library.id)
+            _db.scalars(
+                select(LicensePool)
```
jonathangreen, line 656: Took the opportunity where I could to replace `query` with sqlalchemy 2.x style `select`, to make that move easier when we get to it.

api/integration/registry/license_providers.py
```diff
@@ -6,13 +6,10 @@
...
 if TYPE_CHECKING:
-    from api.circulation import BaseCirculationAPI  # noqa: autoflake
-    from core.integration.settings import BaseSettings  # noqa: autoflake
+    from api.circulation import CirculationApiType  # noqa: autoflake
 
 
-class LicenseProvidersRegistry(
-    IntegrationRegistry["BaseCirculationAPI[BaseSettings, BaseSettings]"]
-):
+class LicenseProvidersRegistry(IntegrationRegistry["CirculationApiType"]):
```
jonathangreen, line 12: Unfortunately future annotations doesn't cover the parameters for generic classes, so we still need to use a string annotation here.

core/model/__init__.py
```diff
@@ -54,6 +54,13 @@ def pg_advisory_lock(
         connection.execute(text(f"SELECT pg_advisory_lock({lock_id});"))
         try:
             yield
+        except IntegrityError:
+            # If there was an IntegrityError, and we are in a transaction,
+            # we need to roll it back before we are able to release the lock.
+            transaction = connection.get_transaction()
+            if transaction is not None:
+                transaction.rollback()
+            raise
```
jonathangreen, line 63: This came up while testing this PR, but it uncovered a bug that existed in our `pg_advisory_lock` function. If something within the lock context manager causes an `IntegrityError` the connection won't accept any other statements until the transaction is rolled back. This means that the `SELECT pg_advisory_unlock` in the `finally` statement will raise an exception, and the lock won't get released. 

In this case we catch the error, and rollback the transaction here, before continuing to raise the error. That way we are able to successfully release the lock.

core/model/collection.py
```diff
@@ -51,34 +48,38 @@ class Collection(Base, HasSessionCache):
...
 
-    # How does the provider of this collection distinguish it from
-    # other collections it provides? On the other side this is usually
-    # called a "library ID".
-    external_account_id = Column(Unicode, nullable=True)
+    # TODO: This should no longer be used. And will be removed in the next release.
+    #  Collections store their configurations in integration configurations now.
+    #  This is only left here in case there needs to be a rollback to the current
+    #  release.
+    _external_integration_id_deprecated = Column("external_integration_id", Integer)
```
jonathangreen, line 69: This keeps the column, but removes the constraints on it. This means we should be able to roll back if we need to, but otherwise this column should be removed soon after as it will begin to get stale.

core/model/collection.py
```diff
@@ -51,34 +48,38 @@ class Collection(Base, HasSessionCache):
...
-    _external_integration: ExternalIntegration
-
+    # integration configuration.
     integration_configuration_id = Column(
         Integer,
-        ForeignKey("integration_configurations.id", ondelete="SET NULL"),
+        ForeignKey("integration_configurations.id"),
         unique=True,
         index=True,
+        nullable=False,
```
jonathangreen, line 82: We now require each collection to always have an integration configuration.

core/model/collection.py
```diff
@@ -241,35 +271,17 @@ def by_protocol(cls, _db: Session, protocol: str | None) -> Query[Collection]:
...
+        """What is the name of this collection?"""
+        if self.integration_configuration is None:
+            raise ValueError("Collection has no integration configuration.")
+        name = self.integration_configuration.name
+        if not name:
+            raise ValueError("Collection has no name.")
+        return name
 
-    @hybrid_property
+    @property
```
jonathangreen, line 284: This never made sense as a `hybrid_property` since this for loop would not work in the class method case, so I switch it to a `property`.

core/model/integration.py
```diff
@@ -58,45 +78,44 @@ class IntegrationConfiguration(Base):
...
-    def for_library(
-        self, library_id: int, create: Literal[True]
-    ) -> IntegrationLibraryConfiguration:
-        ...
+    # https://docs.sqlalchemy.org/en/14/orm/extensions/associationproxy.html#simplifying-association-objects
+    libraries: Mapped[List[Library]] = association_proxy(
+        "library_configurations",
+        "library",
+        creator=lambda library: IntegrationLibraryConfiguration(library=library),
+    )
```
jonathangreen, line 94: This makes it easy to access the libraries for this integration, if you don't need to access the intermediate settings object.

core/model/integration.py
```diff
@@ -112,6 +131,10 @@ class IntegrationLibraryConfiguration(Base):
     It stores the configuration settings for each external integration in
     a single json row in the database. These settings are then serialized
     using Pydantic to a python object.
+
+    This is a many-to-many relationship between IntegrationConfiguration and
+    Library. Implementing the Association Object pattern:
+    https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#association-object
     """
```
jonathangreen, line 138: This has always been the case, I just wanted to clarify and formalize how this table was being used.

core/model/listeners.py
```diff
@@ -108,8 +108,6 @@ def _site_configuration_has_changed(_db, cooldown=1):
 # catch most that slip through the cracks.
 @event.listens_for(Collection.children, "append")
 @event.listens_for(Collection.children, "remove")
-@event.listens_for(Collection.libraries, "append")
-@event.listens_for(Collection.libraries, "remove")
```
jonathangreen, line 112: This is now handled by calling the `site_configuration_changed` function from the controller explicitly instead of using this listener.

tests/api/admin/controller/test_collection_settings.py
```diff
@@ -1,58 +0,0 @@
...
-from unittest.mock import PropertyMock, create_autospec, patch
-
-from api.admin.controller.collection_settings import CollectionSettingsController
-from api.controller import CirculationManager
-from core.integration.goals import Goals
-from core.integration.registry import IntegrationRegistry
-from tests.fixtures.database import DatabaseTransactionFixture
-
-
-class TestCollectionSettingsController:
```
jonathangreen, line 11: We had two files to test the collection settings `test_collection_settings.py` and `test_collections.py`. I put all the tests into one file `test_collections.py` which matches the filename of the controller.

core/model/library.py
```diff
@@ -189,8 +192,25 @@ class Library(Base, HasSessionCache):
...
+        _db = Session.object_session(self)
+        return _db.scalars(
+            select(Collection)
+            .join(IntegrationConfiguration)
+            .join(IntegrationLibraryConfiguration)
+            .where(
+                IntegrationConfiguration.goal == Goals.LICENSE_GOAL,
+                IntegrationLibraryConfiguration.library_id == self.id,
+            )
+        ).all()
```
jonathangreen, line 213: Turning this into a read only property caused a lot of little changes in tests, not much in our production code though. Instead of doing:
`library.collections.append(collection)` we end up doing `collection.libraries.append(library)`

RishiDiwanTT (commented):
This took a while :sweat_smile: 
I mostly had comments about the migration, most of the rest seemed A-OK :ok_hand:

api/admin/controller/settings.py
```diff
@@ -307,14 +304,16 @@ def _set_configuration_library(
...
-        config = None
 
         # Validate first
         validated_data = protocol_class.library_settings_class()(**info_copy)
+
         # Attach the configuration
-        config = configuration.for_library(cast(int, library.id), create=True)
-        config.settings_dict = validated_data.dict()
-        return config
+        library_configuration = IntegrationLibraryConfiguration(
```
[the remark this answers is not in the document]
jonathangreen: I updated `for_library` to only return existing configurations, instead of also creating them. So it broke these lines unless you create the `IntegrationLibraryConfiguration` before.

api/admin/controller/settings.py
```diff
@@ -407,45 +406,14 @@ def _get_prior_test_results(self, item, protocol_class=None, *extra_args):
...
-                    protocol_class = OPDSImportMonitor
-                    extra_args = (OPDSImporter,)
-
-                if issubclass(protocol_class, BaseHasSelfTests):
-                    self_test_results = protocol_class.prior_test_results(
-                        self._db, protocol_class, self._db, item, *extra_args
-                    )
-
-            elif self.type == "search service":
+            if self.type == "search service":
```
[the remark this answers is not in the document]
jonathangreen: So I removed them here, since they were no longer getting tested, and it was causing out code coverage to go down, but these are only removed here since we are no longer using the settings controller for collection tests. The collection tests run through `IntegrationSelfTestsController`, which still handles them.

api/odl.py
```diff
@@ -78,7 +78,7 @@ class ODLAPIConstants:
 
 
 class ODLSettings(OPDSImporterSettings):
-    external_account_id: Optional[HttpUrl] = FormField(
+    external_account_id: HttpUrl = FormField(
```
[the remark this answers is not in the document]
jonathangreen: Hmm that's a good point... I maybe I'll change this one to `AnyHttpUrl`, so its a bit less strict.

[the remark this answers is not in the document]
jonathangreen: Updated this one.

alembic/versions/20231101_2d72d6876c52_remove_collection_external_integration.py
```diff
@@ -0,0 +1,273 @@
...
+Create Date: 2023-11-01 22:42:06.754873+00:00
+
+"""
+import logging
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+from api.integration.registry.license_providers import LicenseProvidersRegistry
```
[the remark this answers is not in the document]
jonathangreen: Hmmm this is sticky isn't it... I don't think this will directly import the ORM models, since we don't do most of the imports until we create a registry, but creating a registry will import the models for sure. I'm not sure how having the models imported late like that will impact things... I'll make a couple tests.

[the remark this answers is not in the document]
jonathangreen: We imported the `LicenseProvidersRegistry` in a previous migration here:
https://github.com/ThePalaceProject/circulation/blob/4763d83386dde174a0a4ea7e9fa0cc66a394c7ae/alembic/versions/20230531_0af587ff8595_migrate_license_integrations_to_.py#L12

I think we've added columns since then, and it doesn't seem like its caused issues with the migration. I might be misunderstanding how this would break though, does that make sense?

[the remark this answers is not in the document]
RishiDiwanTT: Yes, I tested this out by adding a new column in the Collections model, it did not fail due to a schema mismatch since we're not actively using the ORM (the models do get imported though).

alembic/versions/20231101_2d72d6876c52_remove_collection_external_integration.py
```diff
@@ -0,0 +1,273 @@
...
+
+    # We have moved the setting for the TOKEN_AUTH integration from an external integration
+    # to a new JSONB column on the integration_configurations table (context). We need to move
+    # the data into the new column as part of this migration.
+    # The context column is not nullable, so we need to set a default value for the existing
+    # rows. We will use an empty JSON object. We create the column as nullable, set the default
+    # value, then make it non-nullable.
+    op.add_column(
+        "integration_configurations",
+        sa.Column("context", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
```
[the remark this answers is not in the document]
RishiDiwanTT: Ah ok, yes `default` is an application level default. Using `server_default` will work, I tested it out.

api/odl2.py
```diff
@@ -37,7 +37,7 @@
     from core.model.patron import Hold, Loan, Patron
 
 
-class ODL2Settings(OPDS2ImporterSettings, ODLSettings):
```
RishiDiwanTT, line 40: I assume the side effect of this making username and password mandatory was on purpose

alembic/versions/20231101_2d72d6876c52_remove_collection_external_integration.py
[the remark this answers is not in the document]
jonathangreen: I think this changes the DB level schema though, right?

So migrated tables would have a different schema definition then tables that weren't migrated. We could mitigate this by changing the column definition to use `server_default`, but since none of our other tables with defaults use `server_default`, I'd rather leave it as `default` for consistency.

api/odl2.py
jonathangreen: It is intentional, I think these were changed to non-required by accident in https://github.com/ThePalaceProject/circulation/pull/1189. If you checkout https://github.com/ThePalaceProject/circulation/commit/299003ea0f3a6322dbcb6068219f4c0d27cd5612 and take a look at the config, the configuration form in this PR matches what I see there.

PR #1495, opened 2023-11-01
Adds new type of dialect for SIP2 servers that require spaces instead…

dbernstein:
… of zeros in the time zone segment of the date time values sent via the SIP2 protocol.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-546

## How Has This Been Tested?

Manually test in addition to the unit tests.

## Checklist

### DIFF
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -120,7 +120,7 @@ class SIP2Settings(BasicAuthProviderSettings):
             options={
                 Sip2Dialect.GENERIC_ILS: "Generic ILS",
                 Sip2Dialect.AG_VERSO: "Auto-Graphics VERSO",
-                Sip2Dialect.TZ_SPACES: "TZ Spaces",
+                Sip2Dialect.FOLIO: "Folio",
             },
             required=True,
         ),
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -120,6 +120,7 @@ class SIP2Settings(BasicAuthProviderSettings):
             options={
                 Sip2Dialect.GENERIC_ILS: "Generic ILS",
                 Sip2Dialect.AG_VERSO: "Auto-Graphics VERSO",
+                Sip2Dialect.TZ_SPACES: "TZ Spaces",
             },
             required=True,
         ),
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -345,7 +345,11 @@ class SIPClient(Constants):
             # We're implicitly logged in.
             self.must_log_in = False
         self.login_password = login_password
-        self.dialect_config = dialect.config
+
+        if isinstance(dialect, str):
+            self.dialect_config = Dialect(dialect).config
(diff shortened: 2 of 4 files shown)


api/sip/client.py
```diff
@@ -345,7 +345,11 @@ def __init__(
             # We're implicitly logged in.
             self.must_log_in = False
         self.login_password = login_password
-        self.dialect_config = dialect.config
+
+        if isinstance(dialect, str):
+            self.dialect_config = Dialect(dialect).config
```
dbernstein, line 350: This fix here also addresses a problem with the AG Verso dialect.  Without this conversion from a string value to the enum an exception will be thrown when trying to access the non-existent property.

api/sip/__init__.py
```diff
@@ -120,6 +120,7 @@ class SIP2Settings(BasicAuthProviderSettings):
             options={
                 Sip2Dialect.GENERIC_ILS: "Generic ILS",
                 Sip2Dialect.AG_VERSO: "Auto-Graphics VERSO",
+                Sip2Dialect.TZ_SPACES: "TZ Spaces",
```
jonathangreen, line 123: I know there was some discussion of this in the ticket, I think this would be better named `Folio` then `TZ Spaces`. I pinged Carissa and Courtney on the ticket for some feedback from their perspective.

api/sip/client.py
jonathangreen: This is a good catch @dbernstein.

I spent a bit of time looking into it, and this is actually a more pervasive issue then just in this integration. Instead of adding the code to deal with the dialect coming in as a string here, I'd rather deal with the up-steam cause so that we get the correct enum type passed in here.

I pushed a new PR here that should do that: https://github.com/ThePalaceProject/circulation/pull/1500

api/sip/client.py
```diff
@@ -842,8 +847,10 @@ def parse_patron_status(cls, status_string):
 
     def now(self):
         """Return the current time, formatted as SIP expects it."""
+        tz_spaces = self.dialect_config.tz_spaces
         now = utc_now()
-        return datetime.datetime.strftime(now, "%Y%m%d0000%H%M%S")
+        zzzz = " " * 4 if tz_spaces else "0" * 4
+        return datetime.datetime.strftime(now, f"%Y%m%d{zzzz}%H%M%S")
```
jonathangreen, line 853: Can we add a small test for this that makes sure `now()` formats the date as we would expect based on the dialect selected?

jonathangreen (commented):
This looks good @dbernstein! Just a couple comments.

api/sip/client.py
dbernstein: 👍

dbernstein:
@jonathangreen : I believe I've addressed your change requests.  I believe this one can be merged.

jonathangreen (approved):
Looks good!

PR #1499, opened 2023-11-03
PP-659 Fix multiple loan notifications

RishiDiwanTT:
Added a db commit to the loan notifications loop
Without this commit the patron_last_notified changes are lost

## Checklist


jonathangreen (approved):
Looks good

PR #1500, opened 2023-11-03
Update how we load settings from database

jonathangreen:
## Description

Load our settings from the database with full validation, rather then short-cutting that using the `construct()` method.

## Motivation and Context

@dbernstein noted in https://github.com/ThePalaceProject/circulation/pull/1495 that we had a problem in our sip2 integration where settings were not being loaded with the proper enum types from the database. This turned out to be a more pervasive problem then just for that integration, and necessitated this fix.

## How Has This Been Tested?

Tested with the SIP2 code where this problem was first noted and ran unit tests.

## Checklist

### DIFF
--- a/core/integration/base.py
+++ b/core/integration/base.py
@@ -26,9 +26,9 @@ def integration_settings_load(
     """
     Load the settings object for an integration from the database.
 
-    These settings ARE NOT validated when loaded from the database. It is assumed that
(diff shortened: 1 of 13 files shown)


dbernstein (approved):
@jonathangreen :  looks good. I don't fully understand the intricacies of the test code changes, but I trust that those are required by the change to base.py.

PR #1501, opened 2023-11-06
Type hint base64 & worker pools

jonathangreen:
## Description

Type hint the base64 and worker pool utility functions.

## Motivation and Context

Slowly adding type hints to frequently used functions.

## How Has This Been Tested?

Running unit tests.

## Checklist


tests/core/util/test_string_helpers.py
```diff
@@ -1,76 +1,8 @@
...
-
-        # Run the same tests against two different encodings that can
-        # handle Hebrew characters.
-        self._test_encoder(string, UnicodeAwareBase64("utf8"))
-        self._test_encoder(string, UnicodeAwareBase64("iso-8859-8"))
-
-        # If UnicodeAwareBase64 is given a string it can't encode in
-        # its chosen encoding, an exception is the result.
-        shift_jis = UnicodeAwareBase64("shift-jis")
-        pytest.raises(UnicodeEncodeError, shift_jis.b64encode, string)
```
tdilauro, line 23: Do we no longer need to test these encodings / exception or are they no longer applicable for some reason?

pyproject.toml
```diff
@@ -100,6 +100,8 @@ module = [
     "core.util.cache",
     "core.util.notifications",
     "core.util.problem_detail",
+    "core.util.string_helpers",
```
tdilauro, line 103: Should be able to add "core.util.base64" here, as well.

tdilauro (commented):
Looks good! Just a few small questions/comments.

tests/core/util/test_base64.py
```diff
@@ -0,0 +1,87 @@
...
+        "b64decode",
+        "standard_b64encode",
+        "standard_b64decode",
+        "urlsafe_b64encode",
+        "urlsafe_b64decode",
+        "encodebytes",
+        "decodebytes",
+    ],
+)
+def test_default_is_base64(func):
```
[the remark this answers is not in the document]
jonathangreen: I updated the test name to `test_base64_wraps_stdlib`

tests/core/util/test_string_helpers.py
jonathangreen: I don't think this test was ever very meaningful, since we only ever created a `utf8` `UnicodeAwareBase64` object. We don't have a wrapper class now, just some wrapped functions that all use `utf8`, so it didn't seem meaningful to test other encoding that we don't use.

We could have tests that test `_wrap_func_bytes_string` with other encodings, but I don't know how helpful these tests are, since we're not using these encodings. If you think we need these tests I can add them here.

jonathangreen:
@tdilauro I think I addressed all the feedback here if you want to give this another look.

tdilauro (approved):
Looks good! 🚀

PR #1502, opened 2023-11-06
Enforce SIP2 timeout across calls to read on the socket (PP-659)

jonathangreen:
## Description

Enforce a timeout across multiple recv calls and fix a potential issue with our send call.

## Motivation and Context

@RishiDiwanTT here is my suggestion about how we might handle PP-659. I'm not sure it will fix the issue, but this seems like a potential way we could be taking a long time for SIP connections. Every time we call `recv` the timeout is reset, so we might be just spending a lot of time waiting in this loop for data.

I also noticed that we call [`send` ](https://docs.python.org/3/library/socket.html#socket.socket.send) to send the SIP data, which calls out in the documentation:
> Applications are responsible for checking that all data has been sent; if only some of the data was transmitted, the application needs to attempt delivery of the remaining data. 

We don't do that checking. So its possible that sometimes we are not sending a complete SIP message, which might be slowing things down with retries etc. I think we want to be calling [`sendall`](https://docs.python.org/3/library/socket.html#socket.socket.sendall) 
> Unlike [send()](https://docs.python.org/3/library/socket.html#socket.socket.send), this method continues to send data from bytes until either all data has been sent or an error occurs.

That might sort out some issues we are seeing here as well.

## How Has This Been Tested?

Running locally + tests in CI.

## Checklist

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -39,6 +39,7 @@ import certifi
 
 from api.sip.dialect import Dialect
 from core.util.datetime_helpers import utc_now
+from core.util.log import LoggerMixin
 
 # SIP2 defines a large number of fields which are used in request and
 # response messages. This library focuses on defining the response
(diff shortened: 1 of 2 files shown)


RishiDiwanTT (approved):
Looks good, maybe we should also add a log for the total time taken per `sendall` and `recv` in case the problem persists

PR #1504, opened 2023-11-08
PP-685 Install the tzdata package 

RishiDiwanTT:
## Description
The tzdata package needs to be installed in the docker containers in order for the TZ env var to be respected.
This change installs the package in the base container.

## Motivation and Context
The crontab entries are being executed at the associated UTC, rather than US/Eastern, time. This has them running 4-5 hours earlier than expected, depending on daylight savings time.
[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-685)

## How Has This Been Tested?
The built script container had the right time and timezone when using the `date` command.
The `/var/log/cron.log` also displayed the EST times.

## Checklist

### DIFF
--- a/docker/Dockerfile.baseimage
+++ b/docker/Dockerfile.baseimage
@@ -36,6 +36,7 @@ RUN install_clean \
       # needed for xmlsec
       libxmlsec1-dev \
       libxmlsec1-openssl \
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
🚀

PR #1505, opened 2023-11-08
Bugfix make sure patron profile doesn't authenticate twice (PP-701)

jonathangreen:
## Description

Make sure requests to `patrons/me` only call out to the remote authentication service once.

https://github.com/ThePalaceProject/circulation/blob/7c2d0eeba9c3e0aee1589be41233ed868b39f402/api/routes.py#L321-L327

In `routes.py` the patron profile controller has the `requires_auth` decorator, which does the ILS auth, and puts the patron on the request. If there is no patron, then an error is returned and the route function is never called. This is what was causing the first auth request. 

Inside `app.manager.profiles.protocol()` we were making another call to `authenticated_patron_from_request` which was causing a second authentication request. Instead of doing this, we just use the patron object returned by the first call from the request.

## Motivation and Context

While testing the fix for PP-659, I noticed that requests for the `patrons/me` endpoint authenticate twice, which isn’t ideal. It increases load on our partner ILS, and its the slowest action we take, so this was making loading the patron profile document slower then it needed to be. 

## How Has This Been Tested?

Tested locally with a authentication provider to make sure that I no longer see duplicated requests to the remote auth service.

I don't think we need an additional test here since this test already makes sure that the `requires_auth` decorator is called and the patron is set on for calls to `patrons/me` https://github.com/ThePalaceProject/circulation/blob/7c2d0eeba9c3e0aee1589be41233ed868b39f402/tests/api/test_routes.py#L116-L129

## Checklist

### DIFF
--- a/tests/api/test_controller_profile.py
+++ b/tests/api/test_controller_profile.py
@@ -98,11 +98,11 @@ class TestProfileController:
             assert request_patron.synchronize_annotations is True
 
             # The other patron is unaffected.
-            assert profile_fixture.other_patron.synchronize_annotations is False
+            assert profile_fixture.other_patron.synchronize_annotations is False  # type: ignore[unreachable]
 
         # Now we can create an annotation for the patron who enabled
         # annotation sync.
-        annotation = Annotation.get_one_or_create(
+        Annotation.get_one_or_create(  # type: ignore[unreachable]
             profile_fixture.db.session, patron=request_patron, identifier=identifier
         )
         assert 1 == len(request_patron.annotations)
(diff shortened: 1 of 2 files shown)


tests/api/test_controller_profile.py
```diff
@@ -115,11 +117,12 @@ def test_put(self, profile_fixture: ProfileFixture):
             content_type=ProfileController.MEDIA_TYPE,
             data=json.dumps(payload),
         ):
-            response = profile_fixture.manager.profiles.protocol()
+            profile_fixture.controller.authenticated_patron_from_request()
+            profile_fixture.manager.profiles.protocol()
 
             # ...the annotation goes away.
             profile_fixture.db.session.commit()
-            assert False == request_patron.synchronize_annotations
```
RishiDiwanTT, line 122: Super minor: All other tests were changed to use the "is" keyword.

RishiDiwanTT (approved):
Looks good :+1:

PR #1508, opened 2023-11-09
Instead of raising an exception, we should return a problem detail (PP-620)

jonathangreen:
## Description

When there is an error or a timeout contacting a SIP2 server, we are currently raising a `RemoteIntegrationException`. This isn't being handled further up the stack, so its causing an unhandled exception in our webserver. This is causing us to see spikes of 500 errors in the logs. 

I think the correct thing to do here is to return a `ProblemDetail` with the status code 401 and an appropriate message as well as logging the detailed exception, so its there if we need to investigate why the SIP2 connection failed. That way we won't see this as a server issue in our monitoring and the apps should display the message from the `ProblemDetail`.

## Motivation and Context

Investigating alerts in our server monitoring.

## How Has This Been Tested?

Running unit tests.

## Checklist


RishiDiwanTT (approved):
Looks good :+1:

PR #1516, opened 2023-11-16
Add ISBN, when available, to playback time reports. (PP-572)

tdilauro:
## Description

- Adds an ISBN column to the audiobook playbook time reports.
- Attempts to find the best equivalent ISBN identifier, if the book identifier is not an ISBN.

## Motivation and Context

[Jira [PP-572](https://ebce-lyrasis.atlassian.net/browse/PP-572)]

## How Has This Been Tested?

- New tests for report output and ISBN identifier matching.
- Manual testing in local development environment, including verify report delivery.
- CI tests for the associated branch passed.

## Checklist

[PP-572]: https://ebce-lyrasis.atlassian.net/browse/PP-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/jobs/playtime_entries.py
+++ b/core/jobs/playtime_entries.py
@@ -16,7 +16,7 @@ from sqlalchemy.sql.functions import sum
 from core.config import Configuration
 from core.model import get_one
 from core.model.edition import Edition
-from core.model.identifier import Identifier
+from core.model.identifier import Identifier, RecursiveEquivalencyCache
 from core.model.time_tracking import PlaytimeEntry, PlaytimeSummary
 from core.util.datetime_helpers import previous_months, utc_now
 from core.util.email import EmailManager
@@ -257,11 +257,11 @@ class PlaytimeEntriesEmailReportsScript(Script):
         if _is_usable_isbn_identifier(identifier):
             return cast(str, identifier.identifier)
 
-        # If our identifer is not an ISBN itself, we'll use our Recursive Equivalency
-        # mechanism to find the next best one, if available.
+        # If our identifier is not an ISBN itself, we'll use our Recursive Equivalency
+        # mechanism to find the next best one that is, if available.
         db = Session.object_session(identifier)
-        eq_subquery = Identifier.recursively_equivalent_identifier_ids_query(
-            identifier.id
+        eq_subquery = db.query(RecursiveEquivalencyCache.identifier_id).filter(
+            RecursiveEquivalencyCache.parent_identifier_id == identifier.id
         )
         equivalent_identifiers = (
             db.query(Identifier)
--- a/core/jobs/playtime_entries.py
+++ b/core/jobs/playtime_entries.py
@@ -254,7 +254,7 @@ class PlaytimeEntriesEmailReportsScript(Script):
         if identifier is None:
             return default_value
 
-        if _is_usable_isbn_identifier(identifier):
+        if identifier.type == Identifier.ISBN:
             return cast(str, identifier.identifier)
 
         # If our identifier is not an ISBN itself, we'll use our Recursive Equivalency
@@ -270,24 +270,7 @@ class PlaytimeEntriesEmailReportsScript(Script):
         )
 
         isbn = next(
-            map(
-                lambda id_: id_.identifier,
-                filter(
-                    lambda id_: _is_usable_isbn_identifier(id_), equivalent_identifiers
-                ),
-            ),
+            map(lambda id_: id_.identifier, equivalent_identifiers),
             None,
         )
         return isbn or default_value
-
-
-def _is_usable_isbn_identifier(id_: Identifier) -> bool:
-    """Is this a useful ISBN identifier?
-
-    :param identifier: The identifier to check.
-    """
-    return (
-        id_.type == Identifier.ISBN
-        and id_.identifier is not None
-        and id_.identifier != ""
-    )
--- a/core/jobs/playtime_entries.py
+++ b/core/jobs/playtime_entries.py
@@ -176,9 +176,15 @@ class PlaytimeEntriesEmailReportsScript(Script):
                 edition: Optional[Edition] = None
                 identifier: Optional[Identifier] = None
                 if identifier_id:
-                    identifier = get_one(self._db, Identifier, id=identifier_id)
                     edition = get_one(
                         self._db, Edition, primary_identifier_id=identifier_id
+                    )
+                    # Use the identifier from the edition where available.
+                    # Otherwise, we'll have to look it up.
+                    identifier = (
+                        edition.primary_identifier
+                        if edition
+                        else get_one(self._db, Identifier, id=identifier_id)
                     )
                 isbn = self._isbn_for_identifier(identifier)
                 title = edition and edition.title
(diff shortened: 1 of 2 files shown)


core/jobs/playtime_entries.py
```diff
@@ -218,3 +231,52 @@ def _fetch_report_records(self, start: datetime, until: datetime) -> Query:
                 PlaytimeSummary.identifier_id,
             )
         )
+
+    @staticmethod
+    def _isbn_for_identifier(
+        identifier: Optional[Identifier],
+        /,
```
RishiDiwanTT, line 238: Minor: The positional and name only arguments seem unneeded, but don't cause any hindrance either.

RishiDiwanTT (commented):
Some minor feedback.

core/jobs/playtime_entries.py
```diff
@@ -164,15 +173,19 @@ def do_run(self):
                 identifier_id,
                 total,
             ) in self._fetch_report_records(start=start, until=until):
-                edition = None
+                edition: Optional[Edition] = None
+                identifier: Optional[Identifier] = None
                 if identifier_id:
+                    identifier = get_one(self._db, Identifier, id=identifier_id)
```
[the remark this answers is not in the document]
tdilauro: That is true only if there is an edition associated with the identifier. But I could optimize it avoid the extra query where possible.

core/jobs/playtime_entries.py
```diff
@@ -218,3 +231,52 @@ def _fetch_report_records(self, start: datetime, until: datetime) -> Query:
...
+        """Find the strongest ISBN match for the given identifier.
+
+        :param identifier: The identifier to match.
+        :param default_value: The default value to return if the identifier is missing or a match is not found.
+        :param default_match_strength: The default strength for an equivalent, if none is present.
+        """
+        if identifier is None:
+            return default_value
+
+        isbn = (
```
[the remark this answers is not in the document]
tdilauro: Yeah, that makes sense. Although, I'm thinking that this should probably switch to using the `RecursiveEquivalencyCache` to expose the full pool of equivalent identifiers, regardless of which direction the equivalency goes.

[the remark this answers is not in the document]
tdilauro: @RishiDiwanTT I changed the code and associated tests to use Recursive Equivalency here. Would appreciate special attention to this, as I'm not 100% sure of myself with this mechanism.

core/jobs/playtime_entries.py
tdilauro: I've been thinking about using these more to help ensure that the methods are called as expected. Would be interested in your thoughts -- either here or elsewhere -- on doing this.

core/jobs/playtime_entries.py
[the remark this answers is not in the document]
tdilauro: I added the optimization I mentioned above.

tdilauro:
This is ready for another look, @RishiDiwanTT.

core/jobs/playtime_entries.py
```diff
@@ -218,3 +238,56 @@ def _fetch_report_records(self, start: datetime, until: datetime) -> Query:
...
+        if identifier is None:
+            return default_value
+
+        if _is_usable_isbn_identifier(identifier):
+            return cast(str, identifier.identifier)
+
+        # If our identifer is not an ISBN itself, we'll use our Recursive Equivalency
+        # mechanism to find the next best one, if available.
+        db = Session.object_session(identifier)
+        eq_subquery = Identifier.recursively_equivalent_identifier_ids_query(
```
RishiDiwanTT, line 263: This method does not use the RecursiveEquivalencyCache table, but instead uses the on-the-go sql function that is deprecated at the moment.

RishiDiwanTT: The recursive cache table is simpler in usage (and less demanding).
`_db.query(RecursiveEquivalencyCache).filter(parent_id=identifier.id)`

core/jobs/playtime_entries.py
RishiDiwanTT: That's fine, was just curious for the use of the style :+1:

RishiDiwanTT (commented):
Only one comment about the recursive equivalency cache

core/jobs/playtime_entries.py
tdilauro: Okay. I switched to the cache. It wasn't much simpler to use (unless I'm doing it wrong, which is quite possible). In a way, it makes the tests less convenient because of the need to refresh the cache. But hopefully the performance will be better.

tdilauro:
@RishiDiwanTT Third time's a charm? 🤷🏽‍♂️

tests/core/jobs/test_playtime_entries.py
```diff
@@ -229,6 +230,28 @@ def test_do_run(self, db: DatabaseTransactionFixture):
...
+            ),
+            Equivalency(
+                input_id=isbn_ids["i1"].id, output_id=isbn_ids["i2"].id, strength=1
+            ),
+        ]
+        strongest_isbn = isbn_ids["i2"].identifier
+        no_isbn = ""
+
+        # We're using the RecursiveEquivalencyCache, so must refresh it.
+        EquivalentIdentifiersCoverageProvider(db.session).run()
```
RishiDiwanTT, line 253: While this is the pristine method of creating the Recursive cache, for the purpose of the test we can just create RecursiveEquivalencyCache entries directly, since the coverage provider is not the target of the test.
Eg.
```
create(db, RecursiveEquivalencyCache, parent_id=1, identifier_id=2)
create(db, RecursiveEquivalencyCache, parent_id=1, identifier_id=3)
```

RishiDiwanTT (commented):
Last comments, I promise :crossed_fingers:

tests/core/jobs/test_playtime_entries.py
tdilauro: I have mixed feelings about this. On the one hand, I don't want to conflate the testing of different parts of the system too much; on the other hand, I don't want to get bogged down in the details of how the RecursiveEquivalencyCache optimization is set up in every test that uses it (I expect it to get more adoption).

core/jobs/playtime_entries.py
```diff
@@ -218,3 +238,56 @@ def _fetch_report_records(self, start: datetime, until: datetime) -> Query:
...
+            db.query(Identifier)
+            .filter(Identifier.id.in_(eq_subquery))
+            .filter(Identifier.type == Identifier.ISBN)
+        )
+
+        isbn = next(
+            map(
+                lambda id_: id_.identifier,
+                filter(
+                    lambda id_: _is_usable_isbn_identifier(id_), equivalent_identifiers
```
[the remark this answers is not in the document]
tdilauro: I guess not. I meant to handle an edge case, since it seems possible to have an empty or null value for the identifier. It does seem unlikely, so I'm comfortable removing the checks. I think I'll just remove the check everywhere and count on the `identifier` string always being available for ISBN identifiers.

tdilauro:
> Last comments, I promise 🤞

Hmmm. Crossing your fingers doesn't inspire confidence. 🤣

RishiDiwanTT (approved):
Looks good :+1: 

Eventually for the Recursive Cache a fixture can be written when adoption picks up

PR #1524, opened 2023-11-22
PP-745 Removed patron synchronization checks for annotations

RishiDiwanTT:
## Description
Now the CM will record annotations regardless of patron settings
The checks were removed from the controller, and the patron data model.

There is a side-effect of the patron disabling their sync which deletes all previous annotations. Should that be removed as well?

## Motivation and Context
The Apps should be able to sync their data regardless of Patron settings.

## How Has This Been Tested?
Unit tests were updated to reflect the changes.

## Checklist

### DIFF
--- a/api/annotations.py
+++ b/api/annotations.py
@@ -149,9 +149,6 @@ class AnnotationWriter:
 class AnnotationParser:
     @classmethod
     def parse(cls, _db, data, patron):
-        if patron.synchronize_annotations != True:
-            return PATRON_NOT_OPTED_IN_TO_ANNOTATION_SYNC
-
         try:
             data = json.loads(data)
             if "id" in data and data["id"] is None:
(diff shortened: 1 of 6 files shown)


api/annotations.py
```diff
@@ -149,9 +149,6 @@ def detail(cls, annotation, with_context=True):
 class AnnotationParser:
     @classmethod
     def parse(cls, _db, data, patron):
-        if patron.synchronize_annotations != True:
-            return PATRON_NOT_OPTED_IN_TO_ANNOTATION_SYNC
```
jonathangreen, line 153: We should be able to remove the `PATRON_NOT_OPTED_IN_TO_ANNOTATION_SYNC` ProblemDetail as well, since nothing else will be using it.

jonathangreen (approved):
Looks good. One minor comment.

PR #1526, opened 2023-11-22
Make sure default audience can be selected (PP-752)

jonathangreen:
## Description

We were unable to select "Default Audience" in the OPDS configuration form prior to this change. It would always just report that a required value was missing.

## Motivation and Context

Fix issue submitting the OPDS configuration form.

## How Has This Been Tested?

Local testing in CM admin interface.

## Checklist


tdilauro (approved):
Looks good!

PR #1530, opened 2023-11-27
PP-750 Time tracking flag

RishiDiwanTT:
## Description
Added the should_track_playtime attribute to licensepools

The migrations and migration tests have been added as well
Any 'OPDS for Distributor' audiobooks should be enabled by default
Feed items should contain the playtime links only if the pools allow for it
Importing any Biblioboard feed will result in time tracked audiobooks

For OPDS2 feeds the timeTracking flag could only be imported by customising the OPDS2Feed metadata definition.

## Motivation and Context
We need to add code to the CM that enables time tracking when we see these flags in the OPDS feed. One Audio is going to add this, and we should make sure that these books get time tracked. 

[JIRA](https://ebce-lyrasis.atlassian.net/browse/PP-750)

## How Has This Been Tested?
Updated unit tests to support the flags.
Could not manually test the feeds, as there are no feeds with this flag at the moment.

## Checklist

### DIFF
--- a/core/model/licensing.py
+++ b/core/model/licensing.py
@@ -270,7 +270,7 @@ class LicensePool(Base):
     licenses_available: int = Column(Integer, default=0, index=True)
     licenses_reserved: int = Column(Integer, default=0)
     patrons_in_hold_queue = Column(Integer, default=0)
-    should_track_playtime = Column(Boolean, default=False)
+    should_track_playtime = Column(Boolean, default=False, nullable=False)
 
     # This lets us cache the work of figuring out the best open access
     # link for this LicensePool.
--- a/tests/migration/conftest.py
+++ b/tests/migration/conftest.py
@@ -4,17 +4,7 @@ import json
 import random
 import string
 from pathlib import Path
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    Generator,
-    Optional,
-    Protocol,
-    Tuple,
-    Union,
-    cast,
-)
+from typing import TYPE_CHECKING, Any, Dict, Generator, Optional, Protocol, Union, cast
 
 import pytest
 import pytest_alembic
@@ -98,22 +88,6 @@ def random_name() -> RandomName:
         return "".join(random.choices(string.ascii_lowercase, k=length))
 
     return fixture
-
-
-def flexible_insert_statement(
-    table_name: str, **values: str | int | None
-) -> Tuple[str, Tuple[Any, ...]]:
-    """Sometimes after a few migrations the tables go out of sync for testing.
-    A fixture may want to insert a column that doesn't yet exist.
-    This function allows any fixture to provide only the required number of values
-    it wants to insert, the method generates an INSERT statement with only the provided
-    column names and corresponding VALUES"""
-    keys = tuple(values.keys())
-    sses = ",".join(["%s" for _ in range(len(keys))])
-    return (
-        f"INSERT INTO {table_name} ({','.join(keys)}) VALUES ({sses}) returning id",
-        tuple(values.values()),
-    )
 
 
 class CreateLibrary(Protocol):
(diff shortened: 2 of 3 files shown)


alembic/versions/20231124_6af9160a578e_licensepools_time_tracking_flag.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = "6af9160a578e"
+down_revision = "2d72d6876c52"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "licensepools", sa.Column("should_track_playtime", sa.Boolean(), nullable=True)
```
jonathangreen, line 21: Since `null` means the same as `false`, it might be more clear to make this non-nullable, but with a default of `false`, so we don't have to worry about null entries in the future?

core/opds2_import.py
```diff
@@ -764,6 +823,8 @@ def _extract_publication_metadata(
         )
         # Audiobook duration
         duration = publication.metadata.duration
+        # Not all parsers support time_tracking
+        time_tracking = getattr(publication.metadata, "time_tracking", False)
```
jonathangreen, line 827: Since playtime tracking is really only meaningful for audiobooks, it might be a good idea to make sure that if the flag is True, we are working with an audiobook. If the item isn't an audiobook, maybe we log a warning and ignore the flag.

core/opds_import.py
```diff
@@ -818,6 +819,9 @@ def extract_feed_data(
                     combined_circ["data_source"] = self.data_source_name
 
                 combined_circ["primary_identifier"] = identifier_obj
+                combined_circ["should_track_playtime"] = xml_data_dict.get(
+                    "should_track_playtime", False
+                )
```
jonathangreen, line 824: Same comment above about a `should_track_playtime` entry on a non-audiobook.

core/opds2_import.py
```diff
@@ -135,6 +147,53 @@ def parse_manifest(
         return result
 
 
+class PalaceOPDS2PresentationMetadata(opds2_ast.PresentationMetadata):  # type: ignore[misc]
```
jonathangreen, line 150: Minor: I wonder if this code would be better suited to live in the `webpub-manifest-parser` codebase, instead of extending it here. Even if it continues to be a separate class.

jonathangreen (commented):
This looks good! Just a couple comments. 

Also before we can merge this, need to sort out what is causing the migration test to fail on this PR.

core/opds2_import.py
RishiDiwanTT: I had thought the same thing when I realised what needed to be extended, however, since the "timeTracking' flag is Palace specific I thought it should stay within the CM codebase.
This also allows for easier future extensions in the protocol. 
The changes could go in either place though :shrug:

core/model/licensing.py
```diff
@@ -270,6 +270,7 @@ class LicensePool(Base):
     licenses_available: int = Column(Integer, default=0, index=True)
     licenses_reserved: int = Column(Integer, default=0)
     patrons_in_hold_queue = Column(Integer, default=0)
+    should_track_playtime = Column(Boolean, default=False)
```
jonathangreen, line 273: Need to update this to be `nullable=False` as well.

tests/migration/conftest.py
```diff
@@ -90,6 +100,22 @@ def fixture(length: Optional[int] = None) -> str:
     return fixture
 
 
+def flexible_insert_statement(
```
jonathangreen, line 103: Is this actually being used anywhere? If not can we remove it?

core/model/licensing.py
RishiDiwanTT: No, I just liked my code :(

alembic/versions/20231124_1c14468b74ce_bibliotheca_licensepools_time_tracking_.py
```diff
@@ -0,0 +1,57 @@
+"""Bibliotheca licensepools time tracking flag
```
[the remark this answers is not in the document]
RishiDiwanTT: Changed the file name as well

PR #1548, opened 2023-12-02
Change the generate_quicksight_url command params to library_uuids fr…

dbernstein:
…om library_ids

since library ids are not exposed in the API (and it is probably better that they aren't.

## Checklist


api/admin/model/quicksight.py
```diff
@@ -6,12 +6,12 @@
 
 
 class QuicksightGenerateUrlRequest(CustomBaseModel):
-    library_ids: List[int] = Field(
+    library_uuids: List[str] = Field(
```
jonathangreen, line 9: Instead of a string, you can probably make this `uuid.UUID` (see: https://docs.pydantic.dev/1.10/usage/types/#standard-library-types ), which should give us some validation on the UUID.

jonathangreen (approved):
Looks good to me! One minor comment about making the request validation a bit more strict for UUIDs.

dbernstein:
@jonathangreen :  this one is ready to go.  I accidentally squashed my updates to the PR, but they you can see the main updates here: https://github.com/ThePalaceProject/circulation/pull/1548/files#diff-33fd2e2c8b0e67d9c036998775a3978176289c680aa148c4ef09c41f84fb4606R14.

api/admin/model/quicksight.py
```diff
@@ -1,18 +1,21 @@
...
     )
 
-    @validator("library_ids", pre=True)
-    def parse_library_ids(cls, value):
-        return str_comma_list_validator(value)
+    @validator("library_uuids", pre=True)
+    def parse_library_uuids(cls, value) -> List[UUID]:
+        uuid_list = str_comma_list_validator(value)
+        # verify that all strings in the list are UUIDs
+        return list(map(UUID, uuid_list))
```
jonathangreen, line 18: This works as is, but you don't actually need to create `UUID` objects in this validator. Since its marked as `pre=True` it runs before Pydantic's usual data conversion. So if we return a `List[str]` Pydantic will handle turning it into a `List[UUID]`. 

Something like the validation function before this change.
```python
    @validator("library_uuids", pre=True)
    def parse_library_uuids(cls, value) -> List[str]:
        return str_comma_list_validator(value)
```

jonathangreen (approved):
Looks good @dbernstein. Made a comment about the validator function, but this one is approved, so feel free to merge as is, or make that change and merge.

api/admin/model/quicksight.py
dbernstein: Ah - okay.  I'm remove the unnecessary conversion and merge.

PR #1553, opened 2023-12-05
Hold notifiations now include a short title and a body with details (PP-753)

jonathangreen:
## Description

Use a short title, so the title text doesn't get cut off on iOS, and add a body with a longer description of the hold that is available. This will let users on iOS expand the notification to see the body text.

I also modified the tests a bit, as they were not testing what `title` and `body` args were given to the `Notification` class before. 

## Motivation and Context

Bug report from A1QA that the text of the hold notification is cutoff on iOS. See PP-753.

## How Has This Been Tested?

- Running unit tests

This one is hard to test fully. We will need to do a full end to end test in the development environment to make sure this one is working as expected.

## Checklist


dbernstein (approved):
Looks good.

PR #1554, opened 2023-12-05
Search integration configured as a service (PP-93)

jonathangreen:
## Description

This PR updates Opensearch to be configured as a service via environment variables, rather then in the database. This means that Opensearch is not necessary to run the CM, it won't start without the necessary environment variables set. Opensearch is required for the CM to function currently, but the code will start without it, so that it can be configured. Now that it is configured via environment variables, it needs to be supplied.

The new environment variable is: `PALACE_SEARCH_URL` and the environment variable to configure the URL used when running unit tests is updated to `PALACE_TEST_SEARCH_URL`.

Now that we require Opensearch to be running to start the CM instance, the github actions workflow is updated to use the `docker-compose.yml` to start the CM for testing. This makes sure all the services are started and that `docker-compose.yml` stays up to date.

## Motivation and Context

🚨 Note: This is ready for review, but it cannot be merged until:
- ~~Need update for admin UI~~ https://github.com/ThePalaceProject/circulation-admin/pull/100
- ~~Need update for playbook to deploy the new environment variable~~ https://github.com/ThePalaceProject/hosting-playbook/pull/560

This is part of PP-4, and will make it easier to configure the elasticsearch service without having to manually write settings to the database and removes the reliance on external integration settings.

## How Has This Been Tested?

- Manual testing in CM locally
- Running unit tests

## Checklist


tdilauro (approved):
Looks good to me!

A few comments related to parenthesizing of multiple context managers and a possible extraneous file.

I also note that self-test functionality and some testing for misconfigured search configuration has been removed; however, that appears to be intentional.

PR #1556, opened 2023-12-06
Display warning when MARC exporter has no collections configured (PP-59)

jonathangreen:
## Description

Adds a warning when viewing the MARC records page, when the MARC exporter is configured, but no collections have MARC export enabled.

## Motivation and Context

I was testing the new MARC export functionality this morning on Minotaur and noticed that the message "MARC files aren't ready to download yet" isn't really accurate in this situation, since nothing is configured to export.

## How Has This Been Tested?

- Running tests
- Local testing in admin UI

## Checklist

### DIFF
--- a/api/controller_marc.py
+++ b/api/controller_marc.py
@@ -140,7 +140,15 @@ class MARCRecordController:
         marc_files = self.get_files(session, library)
 
         if len(marc_files) == 0:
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🔧

PR #1557, opened 2023-12-06
Allow pyinstrument to profile our test runs

jonathangreen:
## Description

- Add a note to the README about how we can use `pyinstrument` to profile tests.
- Make sure that `freezetime` is configured to ignore `pyinstrument`, so it doesn't throw off the profiling.

## Motivation and Context

Added this because some changes I made in PP-93 slowed down test runs significantly, and I needed to do some profiling to figure out what was going on there. Figured it would be good to have some documentation about how to do this.

## How Has This Been Tested?

- Running tests locally

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -743,6 +743,22 @@ module under the hood to do the profiling.
 
 This profiler uses [PyInstrument](https://pyinstrument.readthedocs.io/en/latest/) to profile the code.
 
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
🏎️ 🏁

PR #1560, opened 2023-12-06
Fix 2d72d6876c52 migration failure with overdrive advantage collections

jonathangreen:
## Description

`20231101_2d72d6876c52_remove_collection_external_integration.py` failed to properly account for Overdrive advantage collections, so the migration would fail with the following error:

```
pydantic.error_wrappers.ValidationError: 3 validation errors for OverdriveSettings
overdrive_website_id
  field required (type=value_error.missing)
overdrive_client_key
  field required (type=value_error.missing)
overdrive_client_secret
  field required (type=value_error.missing)
```

Added a test for this case, and fixed the error.

## Motivation and Context

Fix for an issue that came up when upgrading our production CM instances. We can just directly fix the migration, since this won't have an impact if the migration succeeded, and if the migration failed, it will need this fix.

## How Has This Been Tested?

- Running tests in CI

## Checklist

### DIFF
--- a/alembic/versions/20231101_2d72d6876c52_remove_collection_external_integration.py
+++ b/alembic/versions/20231101_2d72d6876c52_remove_collection_external_integration.py
@@ -11,6 +11,7 @@ from sqlalchemy.dialects import postgresql
 
 from alembic import op
 from api.integration.registry.license_providers import LicenseProvidersRegistry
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good!

PR #1563, opened 2023-12-07
Fix for e06f965879ab MARC file deletion migration (PP-773)

jonathangreen:
## Description

`e06f965879ab` was hard to test, since it involves both the DB and S3. After the migration ran on Minotaur and Cerberus, there are a couple changes needed:

- The latest code puts the MARC file URL in `Representation.mirror_url` and `Representation.url` but older versions of the code only put the url in `Representation.url`. So older files were failing this migration because `mirror_url` is `NULL`. So its safe to just use URL for everything.
- Some records have NULL in `url` (as well as `mirror_url`). I'm not sure how this happened. But there is no way we can delete the file in this case, so it doesn't fail the migration, just outputs a log message.
- The URLs are stored URL escaped, and S3 keys are expected not to be escaped, so we have to call `unquote` on the keys.
- Don't fail the migration if we can't parse the key out of the URL. Just log what happened and move on.

## Motivation and Context

Troubleshooting failed migration on `Cerberus`.

## How Has This Been Tested?

- Tested with local DB copy, but no s3... so 🤞🏻
- Running unit tests

## Checklist

### DIFF
--- a/alembic/versions/20231206_e06f965879ab_marc_s3_file_cleanup.py
+++ b/alembic/versions/20231206_e06f965879ab_marc_s3_file_cleanup.py
@@ -6,7 +6,7 @@ Create Date: 2023-12-06 16:04:36.936466+00:00
 
 """
 from typing import Optional
-from urllib.parse import urlparse
+from urllib.parse import unquote, urlparse
 
 from alembic import op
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good!

PR #1564, opened 2023-12-08
Refactor mock webserver into a fixture

jonathangreen:
## Description

This makes two changes: 
- Turn `mock_web_server` into a fixture, since we use the same fixture function in a few places
- Make sure when we are calling `request_with_timeout` we set `timeout`, `backoff_factor`, so that the tests don't have to wait for retries. This was making tests slower then they need to be.

## Motivation and Context

I was troubleshooting what was causing slow tests in #1554. This was not the cause, but it was making out tests take longer then they needed to, so I thought I'd fix it while I noticed since it cuts down the time to run our core tests by about a minute for me.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,6 +114,7 @@ module = [
     "core.util.worker_pools",
     "core.util.xmlparser",
     "tests.fixtures.authenticator",
+    "tests.fixtures.webserver",
     "tests.migration.*",
 ]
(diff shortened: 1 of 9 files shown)


tdilauro (approved):
The changes look good to me! 🍾

PR #1565, opened 2023-12-11
Break api/controller.py into multiple files

jonathangreen:
## Description

`api/controller.py` was becoming a bit of a monster at 2352 lines. This PR breaks each controller there out into its own file in the `api/controller` directory. No code or logic changes included other then this rename.

I took the opportunity to rename the tests that were already broken out, into the same naming scheme.

## Motivation and Context

Make it a bit easier to work on the circulation manager controllers.

## How Has This Been Tested?

- Running tests

## Checklist


tdilauro (approved):
Looks good!

PR #1571, opened 2023-12-14
Fix flaky marc test

jonathangreen:
## Description

I saw a MARC test failing in https://github.com/ThePalaceProject/circulation/pull/1570 that looked like it was just the order of the test results. This makes that order into a set, so we know the results are there, but don't care about ordering as its not defined.

## Motivation and Context

Fix flaky test.

## How Has This Been Tested?

- Running unit tests.

## Checklist


dbernstein (approved):
Looks good.

PR #1573, opened 2023-12-14
Make it easier to use log_elapsed_time with LoggerMixin

jonathangreen:
## Description

Right now the `log_elapsed_time` decorator cannot use the logger defined by `LoggerMixin` because the logging function needs to be passed into the decorator, and it is not available when the function is decorated. Modify the decorator so it is easier to use in this case.

## Motivation and Context

I'd like to switch `CirculationManager` to inherent from `LoggerMixin` but still be able to use `log_elapsed_time`.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/api/circulation_manager.py
+++ b/api/circulation_manager.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 import urllib.parse
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


dbernstein (approved):
Nice.

PR #1580, opened 2023-12-20
Fix broken Bibliotheca self test.

dbernstein:
Resolves: https://ebce-lyrasis.atlassian.net/browse/SUP-90

## Description
Just a one liner fix that likely broke due to an upgraded dependency.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/SUP-90

## How Has This Been Tested?
Manually + all tests continue to pass.

## Checklist


jonathangreen (approved):
LGTM!

PR #1590, opened 2024-01-04
Fix overdrive advantage collection parent reporting (PP-806)

jonathangreen:
## Description

The changes in https://github.com/ThePalaceProject/circulation/pull/1494 make it so the ID that the admin UI works with for collections is the ID of the integration, not the ID of the collection. We had a bit of a mixup, where the ID of the collection was still being reported as the ID for the parent collection. 

This should fix that issue, so the admin UI reports the correct collection. 

## Motivation and Context

In the Collection Managers, previously configured OverDrive Advantage accounts are showing “None” for Parent or an incorrect parent account. Changing the Parent account to the correct account then saving does not actually save – the save appears to work after clicking the Submit button; however, after navigating away and back to the collection the Parent account reverts to None or to the incorrect parent account it had. 

See JIRA: https://ebce-lyrasis.atlassian.net/browse/PP-806

This probably points to a bit more refactoring we should do, so we actually store on the CM side the parents integration ID, instead of collection ID, but I didn't want to make that large a change as part of this bugfix. I'm going to make a separate ticket for that.

## How Has This Been Tested?

Tested with CT and CA DBs.

## Checklist

### DIFF
--- a/api/admin/controller/collection_settings.py
+++ b/api/admin/controller/collection_settings.py
@@ -43,8 +43,11 @@ class CollectionSettingsController(
         if service_info:
             # Add 'marked_for_deletion' to the service info
             service_info["marked_for_deletion"] = service.collection.marked_for_deletion
-
-            service_info["parent_id"] = service.collection.parent_id
+            service_info["parent_id"] = (
+                service.collection.parent.integration_configuration_id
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good!

PR #1591, opened 2024-01-04
Migration to clean up overdrive settings (PP-797)

jonathangreen:
## Description

This PR adds a migration that cleans up the configuration for the License Provider integrations.  

This fixes some of symptoms noted in PP-797, specifically when updating the parent collection to None, the child having all the settings filled in. This isn't the expected behavior and is being caused by this bug.

## Motivation and Context

After the couple rounds of license provider refactoring, we have the situation where the settings we have stored have some old values stored in them. Specifically the settings for Overdrive advantage collections contain all of their parent collections settings. 

This situation pollutes our logs with warnings like: 
```json
{"host": "3933b57f3c89", "name": "api.overdrive.OverdriveChildSettings", "level": "INFO", "filename": "settings.py", "message": "Unexpected extra argument 'overdrive_client_key' for model OverdriveChildSettings", "timestamp": "2023-12-07T16:48:13.904671+00:00"}
{"host": "3933b57f3c89", "name": "api.overdrive.OverdriveChildSettings", "level": "INFO", "filename": "settings.py", "message": "Unexpected extra argument 'overdrive_website_id' for model OverdriveChildSettings", "timestamp": "2023-12-07T16:48:13.904803+00:00"}
{"host": "3933b57f3c89", "name": "api.overdrive.OverdriveChildSettings", "level": "INFO", "filename": "settings.py", "message": "Unexpected extra argument 'overdrive_client_secret' for model OverdriveChildSettings", "timestamp": "2023-12-07T16:48:13.905076+00:00"}
```

This will cause some issues if the parents settings get updated. So this migration goes through the settings and only stores the settings that are expected on the model. The migration logs the settings before and after so we have a record if needed.

## How Has This Been Tested?

Tested this locally with the CT, CA and NJ CM instances.

## Checklist

### DIFF
--- a/alembic/versions/20240104_735bf6ced8b9_update_license_provider_settings.py
+++ b/alembic/versions/20240104_735bf6ced8b9_update_license_provider_settings.py
@@ -25,7 +25,7 @@ def upgrade() -> None:
     conn = op.get_bind()
 
     rows = conn.execute(
-        "SELECT ic.id as integration_id, ic.settings, ic.protocol, ic.goal, c.parent_id "
+        "SELECT ic.id as integration_id, ic.settings, ic.protocol, ic.goal, c.parent_id, ic.name "
         "FROM collections c JOIN integration_configurations ic ON c.integration_configuration_id = ic.id"
     ).all()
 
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
This looks good, though `codecov` is complaining about lines that do appear to be covered by tests.

PR #1597, opened 2024-01-05
CI check making sure DB migrations keep database in sync

jonathangreen:
## Description

Previously our tests could not check that when upgrading / downgrading the database, we ended up in the expected state because you need older versions of the code to do this operation. This automates the process and adds it to CI, so that we can be more confident that upgraded databases match the state of newly initialized databases.

## Motivation and Context

This ended up as a bit of a side quest when I noticed while working with production DBs for https://github.com/ThePalaceProject/circulation/pull/1590 and https://github.com/ThePalaceProject/circulation/pull/1591 that one of our down migrations wasn't matching. 

That migration is fixed in this PR, and when that commit is left out, this PR catches the error.

## How Has This Been Tested?

Running CI, and it caught a DB downgrade error in our codebase already.

## Checklist

### DIFF
--- a/docker/ci/test_migrations.sh
+++ b/docker/ci/test_migrations.sh
@@ -24,6 +24,37 @@ run_in_container()
   compose_cmd run --build --rm webapp /bin/bash -c "source env/bin/activate && $*"
 }
 
+cleanup() {
+  compose_cmd down
+  git checkout -q "${current_branch}"
+}
+
+run_migrations() {
+  ALEMBIC_CMD=$1
+  run_in_container "alembic ${ALEMBIC_CMD}"
+  exit_code=$?
+  if [[ $exit_code -ne 0 ]]; then
+    echo "ERROR: Running database migrations failed."
+    cleanup
+    exit $exit_code
+  fi
+  echo ""
+}
+
+check_db() {
+  DETAILED_ERROR=$1
+  run_in_container "alembic check"
+  exit_code=$?
+  if [[ $exit_code -eq 0 ]]; then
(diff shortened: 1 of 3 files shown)


docker/ci/test_migrations.sh
```diff
@@ -0,0 +1,135 @@
+#!/bin/bash
+
+# This script makes sure that our database migrations bring the database up to date
+# so that the resulting database is the same as if we had initialized a new instance.
+#
+# This is done by checking out the an older version of our codebase. The commit when
+# the first migration was added and initializing a new instance. Then we check out
```
tdilauro, line 7: Some extra words/punctuation here? Maybe:

> This is done by (1) checking out an older version of our codebase at the commit on which the first migration was added and then (2) initializing a new instance. Then we check out...

docker/ci/test_migrations.sh
```diff
@@ -0,0 +1,135 @@
...
+# the current version of our codebase and run the migrations. If the database is in
+# sync, then the migrations are up to date. If the database is out of sync, then
+# a new migration is required. We then repeat this process with our down migrations
+# to make sure that the down migrations stay in sync as well.
+#
+# This test is cannot be added to the normal migration test suite since it requires
```
tdilauro, line 13: Extra word:
> This test [is] cannot be added

tdilauro (approved):
Looks good! 🎸 

A couple minor suggestions for the script comments.

jonathangreen:
Thanks @tdilauro, I updated those comments!

PR #1611, opened 2024-01-16
Ensure that drm_licensor (FeedEntryType) is serialized properly.

dbernstein:
## Description

Prior to this change, the drm_licensor attribute (which is a FeedEntryType) of the AcquisitionLink wasn't being serialized properly because the clientToken field of the drm_licensor (which is a FeedEntryType) is not a string but a nested FeedEntryType.  

The fix ensures that the clientToke value is extracted getting the text "text" property of the FeedEntryType.
## Motivation and Context
 https://ebce-lyrasis.atlassian.net/browse/PP-830

## How Has This Been Tested?
Unit tests updated.  
Tested manually in my local instance.

## Checklist

### DIFF
--- a/core/feed/serializer/opds2.py
+++ b/core/feed/serializer/opds2.py
@@ -171,7 +171,9 @@ class OPDS2Serializer(SerializerInterface[dict[str, Any]]):
 
         if link.drm_licensor:
             props["licensor"] = {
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
Looks good!

PR #1627, opened 2024-01-24
Convert metadata integrations to integration settings (PP-497)

jonathangreen:
## Description

~~🚨 This PR is a follow up to: https://github.com/ThePalaceProject/circulation/pull/1630 and that PR should go in first.~~

Convert all our metadata services to use integration settings as part of the PP-4 epic. Following on the work done in https://github.com/ThePalaceProject/circulation/pull/1630 the self tests controller is combined with `MetadataServicesController`.

This includes a migration to update the metadata integrations to the new config system. 

There is some dead code leftover after this PR in the settings controller class, but I'd like to handle cleaning that up in a separate PR, because I think there will be a lot of changes come out of that cleanup. That dead code is why our test coverage went down here, but I think that is okay as long as we clean it up eventually.

## Motivation and Context

PP-4 epic to convert our configurations.

## How Has This Been Tested?

- Local testing with CM interface
- Testing in CI

## Checklist


tdilauro (approved):
This one looks good, too! 🥇

PR #1630, opened 2024-01-25
Refactor self test controllers (PP-497)

jonathangreen:
## Description

~~🚨 This PR is a follow up to: https://github.com/ThePalaceProject/circulation/pull/1632 and that PR should get merged in first.~~

Refactor our self test controllers, so they are shared with the settings controllers, since there is so much overlap between them. This work is a precursor to refactoring the metadata controller to use integration settings.

In this PR I collapse together these classes:
- CollectionSelfTestsController & CollectionSettingsController
- PatronAuthController & PatronAuthSelfTestsController

And I update `IntegrationSelfTestsController` to be a subclass of `IntegrationSettingsController`, since they share a lot of common functionality.

## Motivation and Context

- Reduce the duplication of code between settings controllers and self test controllers

## How Has This Been Tested?

Tested locally, and running tests in CI.

## Checklist


tdilauro (approved):
Looks good! 🎸

PR #1632, opened 2024-01-26
Refactor tests for several of the admin controllers (PP-497)

jonathangreen:
## Description

This PR does some refactoring of the tests for the following controllers:
- `CollectionSelfTestsController`
- `CollectionSettingsController`
- `MetadataServicesController`
- `MetadataServiceSelfTestsController`
- `PatronAuthController`
- `PatronAuthSelfTestsController`

The test refactoring updates the tests to use a minimal new `flask_app_fixture` that lets us get a flask context without the overhead of instantiating a while `CirculationManager` instance.

This speeds up the tests and makes them a bit easier to reason about. It also breaks these tests out from the deeply nested hierarchy of fixtures we are using for admin controller tests. I'm hoping we can use this fixture to do similar work to other controller tests as we have the opportunity.

## Motivation and Context

This test refactoring was done in advance of work to refactor the controllers themselves. Since the test diffs are huge, I figured it would be helpful to land them in a separate commit.

## How Has This Been Tested?

- Running tests in CI

## Checklist

### DIFF
--- a/tests/api/admin/controller/test_collection_self_tests.py
+++ b/tests/api/admin/controller/test_collection_self_tests.py
@@ -16,6 +16,7 @@ from core.selftest import HasSelfTestsIntegrationConfiguration
 from core.util.problem_detail import ProblemDetail, ProblemError
 from tests.api.mockapi.axis import MockAxis360API
 from tests.fixtures.database import DatabaseTransactionFixture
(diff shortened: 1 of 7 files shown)


dbernstein (approved):
🥇

PR #1635, opened 2024-01-29
Remove SettingsController from IndividualAdminSettingsController (PP-893)

jonathangreen:
## Description

Remove `SettingsController` usage from the `IndividualAdminSettingsController`, since it largely doesn't use it anyway. The biggest thing is used is validate formats for the email validation. Since the rest of our email validation is done by Pydantic now, switch this over to use Pydantic as well.

## Motivation and Context

Trying to get to the place where we can remove all the old settings code including `SettingsController`.

## How Has This Been Tested?

-Testing in Admin UI locally
-Running unit tests

## Checklist


tdilauro (approved):
Looks good! 🎸

PR #1636, opened 2024-01-29
Remove SettingsController from AnnouncementSettings controller

jonathangreen:
## Description

Remove SettingsController usage from the AnnouncementSettings, since it largely doesn't use it anyway.

## Motivation and Context

Trying to get to the place where we can remove all the old settings code including SettingsController.

## How Has This Been Tested?

Running locally.

## Checklist


tdilauro (approved):
🎺 📜

PR #1641, opened 2024-01-29
Don't pre-load our configuration settings table cache anymore (PP-893)

jonathangreen:
## Description

Now that we aren't using ConfigurationSettings anywhere. We don't need to populate the cache for them on configuration loads.

## Motivation and Context

Removing the remaining Config settings pieces.

## How Has This Been Tested?

Tested locally.

## Checklist


tdilauro (approved):
👍🏽

PR #1642, opened 2024-01-29
Remove first book auth API (PP-903) 🔥

jonathangreen:
## Description

- Remove the unused Firstbook Auth API. 
- Update the static files route so its only added if its needed.
  - Update the static files controller, since its only used in a very simple case now.

## Motivation and Context

- I was looking into what the static files sitewide setting was used for, and it was only used here. Looking at firstbook, it doesn't seem like we are using it anywhere.

## How Has This Been Tested?

- Tested locally
- Tested in debug config with local admin UI
- Running tests in CI

## Checklist


tdilauro (approved):
Looks good!

PR #1649, opened 2024-02-01
Relax pre-sharing validation on list entries with no associated work.

dbernstein:
## Description

This update simply skips validation on entries without associated works before sharing a list with a library.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-708

## Checklist

### DIFF
--- a/core/query/customlist.py
+++ b/core/query/customlist.py
@@ -43,6 +43,7 @@ class CustomListQueries(LoggerMixin):
         # All entries must be valid for the library
         library_collection_ids = [c.id for c in library.collections]
         entry: CustomListEntry
+        missing_work_id_count = 0
         for entry in customlist.entries:
             # It appears that many many lists have entries without works.
             # see https://ebce-lyrasis.atlassian.net/browse/PP-708 for the full story.
@@ -53,10 +54,7 @@ class CustomListQueries(LoggerMixin):
             # for enabling sharing to work again for many existing lists would be to relax the
             # validation when an entry does not have an associated work.
             if not entry.work:
-                log.warning(
-                    f"Skipping license pool validation for list entry: No work associated with custom list entry where "
-                    f"entry.id = {entry.id}"
-                )
+                missing_work_id_count += 1
                 continue
 
             valid_license = (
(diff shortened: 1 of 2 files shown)


core/query/customlist.py
```diff
@@ -44,6 +44,21 @@ def share_locally_with_library(
...
+            # It appears that many many lists have entries without works.
+            # see https://ebce-lyrasis.atlassian.net/browse/PP-708 for the full story.
+            # Because of this frequently occurring condition, lists are quietly not shared
+            # with the majority of libraries causing confusion for our users.  As it stands
+            # there is nothing that prevents lists with work-less entries that have already been
+            # shared from being unshared.  So for the time being the least intrusive intervention
+            # for enabling sharing to work again for many existing lists would be to relax the
+            # validation when an entry does not have an associated work.
+            if not entry.work:
+                log.warning(
```
tdilauro, line 56: This might produce a lot of not really helpful log messages. Another approach would be to count this type of error and then emit a warning with the count, if the count is greater than zero.

tdilauro (commented):
One small comment on the code. I think we should also add a test for this behavior, probably in `tests/api/admin/controller/test_custom_lists.py`.

dbernstein:
@tdilauro : thanks for the review.  I've made updates - should be good to go.

tests/api/admin/controller/test_custom_lists.py
```diff
@@ -986,6 +986,26 @@ def test_share_locally_with_invalid_entries(
         assert response["failures"] == 2
         assert response["successes"] == 0
 
+    def test_share_locally_with_entry_with_missing_work(
```
tdilauro, line 989: It would also be good to verify that the new warning is logged when the condition is met, but not logged when the condition is not met. We can use the pytest `capsys` fixture or similar to access this information.

dbernstein: Done - I used caplog :)

dbernstein:
@tdilauro : Can I merge this one?

PR #1650, opened 2024-02-02
Remove app initialization lock (PP-893)

jonathangreen:
## Description

Now that our site settings are converted, we should be able to remove the app initialization lock.

## Motivation and Context

This should speed up our container startup. I broke this one out into its own PR in case we have to roll it back. I did some testing, but there are a lot of cases to test here and its timing dependant. If there is a problem with this we will find out soon enough once its merged though.

## How Has This Been Tested?

- Tested combinations of: 
  - startup in container
  - startup in dev mode
  - new database
  - existing database

And everything seemed to work okay.

## Checklist

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -21,7 +21,7 @@ from api.util.profilers import (
 )
 from core.app_server import ErrorHandler
 from core.flask_sqlalchemy_session import flask_scoped_session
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🚀

PR #1651, opened 2024-02-02
Remove custom_patron_index and custom_patron_catalog (PP-893)

jonathangreen:
## Description

Remove this code as part of the cleanup of external integrations code. This code relies on having external integrations in the DB, but isn't configurable through the admin UI. 

That makes me think it was something custom NYPL had going on for their collections, or is a partially removed feature. Either way we don't have it configured, and we don't have a way to configure it, so it seems like a good candidate for 🔥 

## Motivation and Context

This is part of the work for the PP-4 epic.

## How Has This Been Tested?

- Running tests in CI

## Checklist


tdilauro (approved):
🔥

PR #1653, opened 2024-02-02
Modify app initialization to speed up loading (PP-893)

jonathangreen:
## Description

This PR updates the app initialization to make sure we are not repeating work:
- Get the collections for each of our libraries from the DB once
- Create API classes for these collections in a single place instead of creating a API class for every library, collection pair.
- Pass these API classes to where they are needed, since the API classes themselves are library agnostic.
- And all the test changes needed to cope with this change in initialization.

## Motivation and Context

As part of cleaning up the old configuration settings, I did some testing to see how our app initialization was doing without having to do so many queries to load configuration settings. This showed that we were still slow on big CM instances to do a settings reload. Doing a bit of profiling, the big thing that is left causing problems is creating a collection API for each library a collection is used in.

I think there is more that can be done here, but it will require more changes to our initialization, and this gave a pretty solid performance boost given the fairly small changes.

## How Has This Been Tested?

Tested locally, with DB data from the CT and CA CM instances. 

These changes make a settings reload on CT go from ~8s to ~2s and CA go from ~16s to ~2s on my machine. 

## Checklist

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -5,7 +5,7 @@ import logging
 import sys
 import time
 from abc import ABC, abstractmethod
-from collections.abc import Iterable
+from collections.abc import Iterable, Mapping
 from threading import Thread
 from types import TracebackType
 from typing import Any, Literal, TypeVar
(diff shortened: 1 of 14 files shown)


tdilauro (approved):
🏁

PR #1676, opened 2024-02-14
Remove Work.to_search_documents__DONOTUSE function 🔥 (PP-939)

jonathangreen:
## Description

Remove the `Work.to_search_documents__DONOTUSE` function, which was only used in tests. This appears to be a function that was replaced, but the old function was left behind to be used in tests. 

I thought about just moving the function out to the test suite, but its not a great test. So I re-wrote the test to be a bit more direct and readable. I also removed the `test_to_search_documents_performance` test which doesn't appear to really be testing anything anymore. 

## Motivation and Context

I was testing some ideas for PP-939 and one of them needed changes to the `to_search_documents` function. This caused the test the fail, and looking at the test, in order to update it, I would have had to update `to_search_documents__DONOTUSE` to add the new values. 

This all seemed a bit roundabout, thus removing this function, and updating the test.

## How Has This Been Tested?

- Running tests locally

## Checklist


tdilauro (approved):
Makes sense! 🧠

Looks like that `test_to_search_documents_performance` test never worked, since it used the same method for both `result` and `inapp`. Do we have a sense of how relatively performant the new method is?

PR #1677, opened 2024-02-15
Make sure expired tokens give correct status / problem detail (PP-970)

jonathangreen:
## Description

In https://github.com/ThePalaceProject/circulation/pull/1664 I made token validation a bit too strict, so expired tokens were failing validation, which meant they were then returning 400 status instead of 401. They should be returning a 401 and the token expired problem detail document.

Fixed this so the initial validation is much more lax, it just determines what type of token we are dealing with and added some tests cases for this.

## Motivation and Context

Android app was not refreshing the token when getting a 400 status, so token auth was then failing.

## How Has This Been Tested?

- Running tests locally

Once this is merged we'll roll out to minotaur and test that everything is happy in apps.

## Checklist

### DIFF
--- a/api/authentication/access_token.py
+++ b/api/authentication/access_token.py
@@ -166,19 +166,6 @@ class PatronJWEAccessTokenProvider(LoggerMixin):
         return TokenPatronInfo(**payload)
 
     @classmethod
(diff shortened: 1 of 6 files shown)


tdilauro (approved):
LGTM!

PR #1680, opened 2024-02-16
Suppress works on a per library basis (PP-939)

jonathangreen:
## Description

A prototype for review to suppress works on a per-library basis. 

This PR adds a new many to many relationship between the `works` and `libraries` table, that is used to capture which works are suppressed for what libraries. This data is added to the search index and is added to the default queries used to build lanes.

## Motivation and Context

This will allow libraries with shared collections to hide titles they do not want to appear in their local catalog. 

This would need to be done manually via a script for now. There is no way to do it via the admin interface. But if we are happy with the approach here, we can add admin UI support via a follow up PR.

## How Has This Been Tested?

Tested with CM running locally, and with the newly added tests.

## Checklist

### DIFF
--- a/core/scripts.py
+++ b/core/scripts.py
@@ -2668,6 +2668,96 @@ class LoanNotificationsScript(Script):
             self.notifications.send_loan_expiry_message(loan, delta.days, tokens)
 
 
+class SuppressWorkForLibraryScript(Script):
+    """Suppress works from a library by identifier"""
+
+    BY_DATABASE_ID = "Database ID"
+
+    @classmethod
+    def arg_parser(cls, _db: Session | None) -> argparse.ArgumentParser:  # type: ignore[override]
+        parser = argparse.ArgumentParser()
+        if _db is None:
+            raise ValueError("No database session provided.")
+        library_name_list = sorted(str(l.short_name) for l in _db.query(Library))
+        library_names = '"' + '", "'.join(library_name_list) + '"'
+        parser.add_argument(
+            "-l",
+            "--library",
+            help="Short name of the library. Libraries on this system: %s."
+            % library_names,
+            required=True,
+            metavar="SHORT_NAME",
+        )
+        parser.add_argument(
+            "-t",
+            "--identifier-type",
+            help="Identifier type (default: ISBN). "
+            f'To name identifiers by their database ID, use --identifier-type="{cls.BY_DATABASE_ID}".',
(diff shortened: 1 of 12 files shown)


core/scripts.py
```diff
@@ -2668,6 +2668,96 @@ def process_loan(self, loan: Loan):
...
+            help="Identifier type (default: ISBN). "
+            f'To name identifiers by their database ID, use --identifier-type="{cls.BY_DATABASE_ID}".',
+            default="ISBN",
+        )
+        parser.add_argument(
+            "-i",
+            "--identifier",
+            help="The identifier to suppress.",
+            required=True,
+        )
```
tdilauro, line 2703: Minor: Since both library short name and identifier are both required, could we use the following syntax?

```
command [-i id-type] library identifiers...
```

That would mean a some reworking below and in the tests, but might make our work a little easier when we have to suppress more than one title with same `id-type`.

tests/core/test_lane.py
```diff
@@ -4627,6 +4627,87 @@ def test_works_and_works_from_database(
...
+        data = fixture.populate_works()
+        lane_data = fixture.create_lanes(data)
+
+        decoy_library = db.library()
+        another_library = db.library()
+
+        db.default_collection().libraries += [decoy_library, another_library]
+
+        # Add a couple suppressed works, to make sure they don't show up in the results.
+        suppressed_work1 = db.work(
```
tdilauro, line 4648: Minor: Suggest naming this to `globally_suppressed_work`, `licensepool_suppressed_work` or similar to make this a little easier to follow below.

tdilauro (approved):
Looks good!

A couple of minor comments and suggestions.

core/scripts.py
jonathangreen: I'd like to leave this to a follow up. I did the easiest thing here, and just re-implemented some bits of the `IdentifierScript` and `LibraryScript` types, since this script needs both. I think a better solution would be to update those base classes to be able to cooperate or just be mixin classes, but I was feeling like I had already brought in too much tangential refactoring already for this ticket, so I just left it.

I'll make a follow up ticket for that, and once thats done this can just be an IdentiferScript that can take multiple identifiers.

tests/core/test_lane.py
jonathangreen: Renamed

PR #1688, opened 2024-02-21
Fixes lane saving bug

dbernstein:
## Description
This PR resolves the bug that was manifesting as a failure to save lanes.

I'm not entirely why not passing the search_engine argument without a key causes the following error:
```
{"host": "eb21d6c68737", "name": "core.app_server.ErrorHandler", "level": "ERROR", "filename": "app_server.py", "message": "Exception in web app: Lane.update_size() got multiple values for argument 'search_engine'", "timestamp": "2024-02-20T23:43:55.588382+00:00", "traceback": "Traceback (most recent call last):\n  File \"/var/www/circulation/env/lib/python3.10/site-packages/flask/app.py\", line 870, in full_dispatch_request\n    rv = self.dispatch_request()\n  File \"/var/www/circulation/env/lib/python3.10/site-packages/flask/app.py\", line 855, in dispatch_request\n    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]\n  File \"/var/www/circulation/api/routes.py\", line 115, in decorated\n    return f(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/routes.py\", line 93, in decorated\n    v = f(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/routes.py\", line 64, in decorated\n    return f(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/routes.py\", line 82, in decorated\n    return f(*args, **kwargs)\n  File \"/var/www/circulation/api/admin/routes.py\", line 658, in lanes\n    return app.manager.admin_lanes_controller.lanes()\n  File \"/var/www/circulation/api/admin/controller/lanes.py\", line 163, in lanes\n    lane.update_size(self._db, self.search_engine)\n  File \"src/dependency_injector/_cwiring.pyx\", line 28, in dependency_injector._cwiring._get_sync_patched._patched\nTypeError: Lane.update_size() got multiple values for argument 'search_engine'"}

``` 
When I debugged it, there was only one value being passed.  I'm guessing it has something to do with the @Inject tag trying to inject the available search_engine in addition to the explicit `self.search_engine`.  

I noticed that removing the self.seach_engine argument altogether also solved the problem which makes sense because it is injecting the available search_engine in the context. Presumably the injected search_engine instance is the same instance being passed in explicitly.   The unit tests pass because the search_engine is explicitly using the keyword.  I'm not seeing it being passed implicitly via injection anywhere else in the code. 

For the sake of completeness, I also went ahead and updated other references to `Lane.update_size` where the search_engine wasn't being passed with the keyword.

@jonathangreen :  do you have any insight into what we should be doing here? Should rely on the injection or is it better to be explicit?

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-986

## How Has This Been Tested?
Manually tested on a local instance.  Unit tests still pass.

## Checklist

### DIFF
--- a/api/admin/controller/custom_lists.py
+++ b/api/admin/controller/custom_lists.py
@@ -290,7 +290,7 @@ class CustomListsController(
 
             # If this list was used to populate any lanes, those lanes need to have their counts updated.
             for lane in Lane.affected_by_customlist(list):
-                lane.update_size(self._db, self.search_engine)
+                lane.update_size(self._db, search_engine=self.search_engine)
 
         new_collections = []
         for collection_id in collections:
@@ -399,7 +399,7 @@ class CustomListsController(
             # Update the size for any lanes affected by this
             # CustomList which _weren't_ deleted.
             for lane in surviving_lanes:
-                lane.update_size(self._db, self.search_engine)
+                lane.update_size(self._db, search_engine=self.search_engine)
             return Response(str(_("Deleted")), 200)
 
         return None
--- a/api/admin/controller/lanes.py
+++ b/api/admin/controller/lanes.py
@@ -160,7 +160,7 @@ class LanesController(CirculationManagerController, AdminPermissionsControllerMi
             for list in lane.customlists:
                 if list.id not in custom_list_ids:
                     lane.customlists.remove(list)
-            lane.update_size(self._db, self.search_engine)
+            lane.update_size(self._db, search_engine=self.search_engine)
 
             if is_new:
                 return Response(str(lane.id), 201)
--- a/core/lane.py
+++ b/core/lane.py
@@ -2929,10 +2929,7 @@ class Lane(Base, DatabaseBackedWorkList, HierarchyWorkList):
             return True
         return False
 
-    @inject
-    def update_size(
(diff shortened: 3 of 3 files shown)


jonathangreen:
@dbernstein the reason its happening, as you suspected is the `@inject` decorator. It always injects args using the keyword form. So the function is being called with both a positional and keyword arg for the `search_engine` parameter. 

I think generally explicit is better then implicit for arguments, so I'd be happy to drop the `@inject` decorator here. I've been trying to use it very sparingly. I think I added it here because the `search_engine` parameter used to be optional, and I was having trouble tracing down all the possible lane calls, so I thought this was safer, but didn't think of the possible kwargs issues.

jonathangreen (approved):
Looks good as is, but if the parameter is getting set everywhere already, it probably makes sense to drop `@inject` decorator and the default for the arg so its more explicit:

```python
def update_size(
    self, _db, search_engine: ExternalSearchIndex
):
```

jonathangreen:
This one should probably go out as a hotfix once its merged. So we might want to wait on the production rollout until its merged.

PR #1689, opened 2024-02-21
Better handle RDS restarts (PP-991)

jonathangreen:
## Description

This adds the `pool_pre_ping` arg to our DB connections. This should let us deal more gracefully with RDS restarts that happen automatically during maintenance windows.

See SQL alchemy docs here:
https://docs.sqlalchemy.org/en/14/core/pooling.html#dealing-with-disconnects

This also updates the `ErrorHandler` function to:
- Also handle `ProblemError` exceptions, logging them as a warning, and returning the problem detail doc. We should handle these at another level of the stack, but its good to handle them correctly.
- Log `OperationalError` as warnings. Since `pool_pre_ping` should reduce the number we see, but we will still get some happening when a restart happens in the middle of a transaction. `pool_pre_ping` will just make sure the rest of the connection pool doesn't experience the error because its holding on to a old connection.
- Remove the `debug` functionality from `ErrorHandler` since its correctly flagged as a possible security issue. I think its unexpected to possibly leak stack traces if we were to set debug logging on a production instance. This is better handled by having the stack traces go to log, which they were already.
- I believe that `pool_pre_ping` removes the need to kill the entire process when we have a database error. Since the next connection should be refreshed from the connection pool.

## Motivation and Context

Looking at the top unhandled exceptions in our app in PP-991. I'm seeing this as the second most common, after the exception that is sorted out in https://github.com/ThePalaceProject/circulation/pull/1688.

## Checklist

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -111,7 +111,7 @@ def initialize_application() -> PalaceFlask:
         container.init_resources()
 
         # Initialize the applications error handler.
-        error_handler = ErrorHandler(app, container.config.logging.level())
+        error_handler = ErrorHandler(app)
         app.register_error_handler(Exception, error_handler.handle)
 
         # Initialize the circulation manager
--- a/core/app_server.py
+++ b/core/app_server.py
@@ -2,8 +2,6 @@
 from __future__ import annotations
 
 import gzip
-import sys
-import traceback
 from functools import wraps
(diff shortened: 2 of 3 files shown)


tdilauro (approved):
Looks good! 🎸

PR #1694, opened 2024-02-23
Refactor circulation exceptions (PP-991)

jonathangreen:
## Description

This refactors how we are handling `CirculationExceptions`, so they all inherit from `BaseProblemDetailException` and are able to be turned into a ProblemDetail. This lets us simplify the exception handlers in the Loans controller to catch any `CirculationException` that the Circulation APIs might throw, and return the appropriate ProblemDetail, instead of continuing to grow the huge list of different Exceptions and actions we were handling.

## Motivation and Context

Reviewing the list of unhandled exceptions we are seeing in production, multiple in the top 10 were CirculationExceptions that were not in the `Except` block, or were handled for some actions like `borrow` but not others like `hold`. This PR makes the handling we have here more consistent. 

Having a base class `BaseProblemDetailException` also makes it easier to know if we are able to turn an exception into a ProblemDetail or not. 

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -30,7 +30,21 @@ from api.circulation import (
     LoanInfo,
     PatronActivityCirculationAPI,
 )
-from api.circulation_exceptions import *
+from api.circulation_exceptions import (
+    AlreadyCheckedOut,
+    AlreadyOnHold,
+    CannotHold,
+    CannotLoan,
+    CannotReleaseHold,
+    CurrentlyAvailable,
+    NoAvailableCopies,
+    NoLicenses,
+    NotCheckedOut,
+    NotOnHold,
+    PatronHoldLimitReached,
+    PatronLoanLimitReached,
+    RemoteInitiatedServerError,
+)
 from api.selftest import HasCollectionSelfTests, SelfTestResult
 from api.web_publication_manifest import FindawayManifest, SpineItem
 from core.analytics import Analytics
@@ -79,6 +93,7 @@ from core.service.container import Services
 from core.util import base64
 from core.util.datetime_helpers import datetime_utc, strptime_utc, to_utc, utc_now
 from core.util.http import HTTP
+from core.util.problem_detail import BaseProblemDetailException
 from core.util.xmlparser import XMLParser, XMLProcessor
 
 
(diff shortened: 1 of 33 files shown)


api/bibliotheca.py
```diff
@@ -894,23 +909,7 @@ def date_from_subtag(self, tag, key, required=True):
         return self.parse_date(value)
 
 
-class BibliothecaException(Exception):
-    pass
-
-
-class WorkflowException(BibliothecaException):
```
jonathangreen, line 901: This wasn't being used anywhere.

api/bibliotheca.py
```diff
@@ -894,23 +909,7 @@ def date_from_subtag(self, tag, key, required=True):
         return self.parse_date(value)
 
 
-class BibliothecaException(Exception):
```
jonathangreen, line 897: Replaced this exception with `RemoteInitiatedServerError`, so that we get problem details from all the exceptions thrown by the Bibliotecha API.

api/enki.py
```diff
@@ -416,7 +424,7 @@ def checkout(
                     "User validation against Enki server with %s / %s was unsuccessful."
                     % (patron.authorization_identifier, pin)
                 )
-                raise AuthorizationFailedException()
```
jonathangreen, line 419: This was the only place that we used `AuthorizationFailedException`, everywhere else used `PatronAuthorizationFailedException` and it had the same problem detail, so I removed  `AuthorizationFailedException`.

api/problem_details.py
```diff
@@ -57,7 +57,7 @@
 
 NO_ACCEPTABLE_FORMAT = pd(
```
jonathangreen, line 58: The status codes here were changed because the only place that used these `ProblemDetails` was the loans controller exception handler, where they were called like this:
```
CANNOT_FULFILL.with_debug(str(e), status_code=e.status_code)
```

So they always had the status code of the exception, rather then the status code associated with the `ProblemDetail`. This let me drop the status code on the exceptions and just use the one on the ProblemDetail.

dbernstein (approved):
Looks good.

PR #1695, opened 2024-02-23
Refactor exceptions to have a common base class (PP-991)

jonathangreen:
## Description

Rename `BaseError` to be `BasePalaceException`. Refactor this class to remove the inner exception parameter, since Python 3 has exception chaining. Then refactor all the exceptions defined in our codebase to all have this as a base class, so its easier to catch all of our exceptions if we need to.

## Motivation and Context

Follow up to the work in https://github.com/ThePalaceProject/circulation/pull/1694 which gave all our exceptions that return problem details a common base class.

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/core/exceptions.py
+++ b/core/exceptions.py
@@ -6,10 +6,11 @@ class BasePalaceException(Exception):
 
         :param message: String containing description of the error occurred
         """
+        self.message = message
         super().__init__(message)
 
 
-class IntegrationException(Exception):
+class IntegrationException(BasePalaceException):
     """An exception that happens when the site's connection to a
     third-party service is broken.
 
--- a/core/exceptions.py
+++ b/core/exceptions.py
@@ -6,8 +6,8 @@ class BasePalaceException(Exception):
 
         :param message: String containing description of the error occurred
         """
-        self.message = message
         super().__init__(message)
+        self.message = message
 
 
 class IntegrationException(BasePalaceException):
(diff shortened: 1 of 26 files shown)


core/exceptions.py
```diff
@@ -1,50 +1,16 @@
-class BaseError(Exception):
-    """Base class for all errors"""
+class BasePalaceException(Exception):
+    """Base class for all Exceptions in the Palace manager."""
 
-    def __init__(
-        self, message: str | None = None, inner_exception: Exception | None = None
-    ):
+    def __init__(self, message: str | None = None):
         """Initializes a new instance of BaseError class
```
dbernstein, line 5: BaseError -> BasePalaceException

core/exceptions.py
```diff
@@ -1,50 +1,16 @@
...
 
         :param message: String containing description of the error occurred
-        :param inner_exception: (Optional) Inner exception
```
dbernstein, line 10: error -> exception that 

And the inner_exception has gone away yes?

jonathangreen: Yes, `inner_exception` was removed. Its leftover from Python 2 days when there was no built in way to chain exceptions. Now that we are on Python 3 I replaced it with PEP3134 exception chaining in the couple places where `inner_exception` was previously used.

dbernstein (approved):

PR #1700, opened 2024-02-27
Raise RemoteInitiatedServerError on expired session key (PP-996)

jonathangreen:
## Description

When doing a Findaway fulfillment from B&T, we now check if `FNDSessionKey` is set to `Expired` and raise a `RemoteInitiatedServerError` if that is the case.

The `debug_info` included with this problem_detail gives the response details they want when troubleshooting these messages, so its easier to resolve them in the future. 

JIRA: PP-996

## Motivation and Context

This appears to be a re-occurring problem, and B&T isn't putting a systematic fix in place, instead they are fixing individual accounts when we make support requests. So we want to make it easier to get this information, and to track in our logs when this is happening. 

## How Has This Been Tested?

- Ran unit tests locally
- Tested locally with prod CM DB where this error was occurring

## Checklist

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -1773,12 +1773,12 @@ class Axis360FulfillmentInfoResponseParser(
         checkoutId = k("FNDTransactionID", parsed)
 
         if sessionKey == "Expired":
-            if license_pool.identifier is None:
-                identifier = f"LicensePool {license_pool.id}"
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 🎸🤘🏽

PR #1714, opened 2024-03-08
Bugfix for hold notifications happening for expired holds (PP-1048)

jonathangreen:
## Description

This PR makes two changes:
- Makes sure we only send hold notifications for holds that are not expired
- Add the library name to the hold notification text

For collections not coming in through ODL where we control the holds queues, we don't clean up old holds until they are very old. So we just keep notifying patrons 😓. This will make sure we are only notifying for valid holds.

## Motivation and Context

In PP-997 Courtney is getting holds for an item where the hold has expired and its not clear which library the hold is coming from.

PP-1048

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/core/jobs/holds_notification.py
+++ b/core/jobs/holds_notification.py
@@ -39,11 +39,13 @@ class HoldsNotificationMonitor(SweepMonitor):
         return qu
 
     def item_query(self) -> Query:
+        now = utc_now()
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
Looks good! 🎉

PR #1721, opened 2024-03-11
Add library name to loan notifications (PP-1011)

jonathangreen:
## Description

Currently, holds and loan expiry notifications from the Palace app do not contain the library name. While many users likely have only one library added in the Palace app, for any user having more than 1 library added to the app, the library name not included in the Hold Ready notification can be confusing when trying to borrow a book that has become available. 

## Motivation and Context

See: PP-1011. Related to the changes in https://github.com/ThePalaceProject/circulation/pull/1714.

## How Has This Been Tested?

Running unit tests.

## Checklist


tdilauro (approved):
🏁

PR #1722, opened 2024-03-12
Separate docker builds to own workflow (PP-810)

jonathangreen:
## Description

Updates CI so that:
- Docker builds and unit testing are separate workflows
- Unit tests run on PRs and on push to `main`
- Docker builds run on `pushes`
- Unit tests get run inside of the built docker container before the container gets pushed
- A couple small test updates to address issues with the tests found running in the container
- Adds some caching, so we don't have to compile the lxml wheel every time

## Motivation and Context

As part of PP-810, we need to update how we install lxml to address a segfault. This segfault was happening when running the unit tests, and is resolved be making sure lxml gets built from source. However since we don't run the tests inside out our containers, its hard to make sure the issue is resolved there as well.

I've been using these changes to make sure things are properly compiled in the docker build in https://github.com/ThePalaceProject/circulation/pull/1602 and they have been helpful there. Wanted to merge them as a separate PR though, since these changes are directly related to https://github.com/ThePalaceProject/circulation/pull/1602.

## How Has This Been Tested?

- Running workflows

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,10 +7,6 @@
 # Readmes
 **/*.md
 
-# Tests
-tests/
-integration_tests/
-
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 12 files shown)


tdilauro (approved):
This looks good.

One thing: I do find it handy to get linting and mypy checks for branches/pushes. Could we keep that in?

PR #1723, opened 2024-03-12
Cache poetry install

jonathangreen:
## Description

Cache our poetry install, so we don't have to download it every time and bump up the poetry version we install, to match the version being used by dependabot.

## Motivation and Context

Makes poetry install slightly faster. Also should eliminate the build failures we get every now and then, when poetry fails to install.

## How Has This Been Tested?

- Running the actions

## Checklist


tdilauro (approved):
🏎️

PR #1734, opened 2024-03-14
Feature/pp 1027 implement backend report functionality

dbernstein:
## Description
This PR adds an AsyncTask database structure with associated facilities for pushing tasks onto a logical queue, pulling them for processing, and removing them when complete (by flagging them as complete as rather than deleting them).

Additionally it provides support for requesting for an inventory report to be generated asynchronously for a given library and user.

Finally it includes a script that runs every 15 minutes for pulling inventory report tasks off the queue, generating a CSV report for "Palace Marketplace", "BiblioBoard", "Unlimited Listens", and "Palace Bookshelf" titles, and then emailing them together to the requestor's email address.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1027

## Checklist


dbernstein:
@jonathangreen and @tdilauro :  for the life of me I can't figure out why the migration is failing.  I tried reworking the migration to simply add a single table with a single primary key and no other fields,  but it doesn't seem to work.  The table is being detected as removed after that upgrade operation. Strangely it works when I run against a copy of minotaur.   Do either of you guys have any suggestions how to sort this out?

jonathangreen:
@dbernstein I believe its because your new tables are not being imported in `core/model/__init__.py`, so the migration script does not know they exist.

https://github.com/ThePalaceProject/circulation/blob/24900ec757e6aa171281525f57f8ce08b877ab94/core/model/__init__.py#L513-L516

core/scripts.py
```diff
@@ -2758,6 +2766,166 @@ def suppress_work(self, library: Library, identifier: Identifier) -> None:
...
+                text(self.inventory_report_query()),
+                {"data_source_name": data_source_name, "library_id": library_id},
+            )
+            writer.writerow(rows.keys())
+            writer.writerows(rows)
+        return temp.name
+
+    def inventory_report_query(self) -> str:
+        return """
+            select
```
dbernstein, line 2858: This SQL statement is big.  Could it be written using the sqlalchemy api?  Surely.  Would it be more or less readable?  I would argue that it would be less readable.  Also building the query by directly talking to the database allowed me to iterate quickly in psql rather than trying to do it iteratively in python.  So IMHO making modifications to it will be relatively easy because they can be pasted into a `psql` terminal using production data.

core/scripts.py
```diff
@@ -2758,6 +2766,166 @@ def suppress_work(self, library: Library, identifier: Identifier) -> None:
         self._db.commit()
 
 
+class GenerateInventoryReports(Script):
+    """Generate inventory reports from queued report tasks"""
+
+    DATA_SOURCES = [
```
jonathangreen, line 2772: Since datasource is just a free form text input in the OPDS importers, I don't think hard coding the data sources here like this will accomplish what we are hoping for. Maybe this needs to be configurable, or you need to select the collection when you are generating the report. Otherwise we will be relying on the CMs to be setup with these exact strings set as the datasource.

core/scripts.py
jonathangreen: FYI: one of the things that I don't like about sprinkling text SQL queries into our code base, is that it makes it hard when refactoring our SQLAlchemy models because the IDE can no longer tell you where particular attributes are used. 

Your point is valid here though, and I'm willing to defer to your judgment that this is the best way to approach this query.

api/admin/controller/report.py
```diff
@@ -0,0 +1,42 @@
+import json
```
jonathangreen, line 1: Minor: I've been trying to slowly move our code towards being able to run with mypy in strict mode. I think this code will pass those tests with almost no changes, so it would be nice to add this new file to `pyproject.toml` in the `[[tool.mypy.overrides]]` section that enables strict mode.

core/model/asynctask.py
```diff
@@ -0,0 +1,115 @@
+# Async
```
jonathangreen, line 1: Minor: I've been trying to slowly move our code towards being able to run with mypy in strict mode. I think this code will pass those tests with almost no changes, so it would be nice to add this new file to `pyproject.toml` in the `[[tool.mypy.overrides]]` section that enables strict mode.

core/model/asynctask.py
```diff
@@ -0,0 +1,115 @@
...
+
+class AsyncTaskStatus(str, Enum):
+    READY = "READY"
+    PROCESSING = "PROCESSING"
+    SUCCESS = "SUCCESS"
+    FAILURE = "FAILURE"
+
+
+class AsyncTaskType(Enum):
+    INVENTORY_REPORT = "INVENTORY_REPORT"
```
jonathangreen, line 26: Minor: Same comment about `auto()` as above.

core/model/asynctask.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class AsyncTask(Base):
+    """An asynchronous task."""
+
+    __tablename__ = "asynctasks"
```
jonathangreen, line 32: Do we need a job of some sort to clear out old entries in this table? Maybe tasks that has finished more then a month ago or something like that. It seems like it might be a bit of a problem to have this table continue to grow forever.

core/model/asynctask.py
```diff
@@ -0,0 +1,115 @@
...
+    task_type = Column(SqlAlchemyEnum(AsyncTaskType), index=True, nullable=False)
+    status = Column(SqlAlchemyEnum(AsyncTaskStatus), index=True, nullable=False)
+    processing_start_time = Column(DateTime(timezone=True), nullable=True)
+    processing_end_time = Column(DateTime(timezone=True), nullable=True)
+    status_details = Column(String, nullable=True)
+    data: dict[str, Any] = Column(MutableDict.as_mutable(JSONB), default={})
+
+    def complete(self):
+        if self.status != AsyncTaskStatus.PROCESSING:
+            raise Exception(
```
jonathangreen, line 46: It looks like we're catching `ProblemDetailException`s in the controller, so perhaps we should be throwing a `ProblemDetailException` here. Even if we don't need them to be `ProblemDetailException` it would be nice to have these live in the `BasePalaceException` hierarchy.

jonathangreen (commented):
This generally looks good @dbernstein! Added some comments to review. 

As far as functionality, my biggest concern is around using a hard coded list of datasources, since we can't rely on these always being the same. Might need a bit of discussion on the ticket about how we are finding the collections to include in this report.

api/admin/controller/report.py
```diff
@@ -0,0 +1,42 @@
...
+from flask import Response
+
+from api.controller.circulation_manager import CirculationManagerController
+from core.model import Library
+from core.model.admin import Admin
+from core.model.asynctask import AsyncTaskType, InventoryReportTaskData, queue_task
+from core.util.problem_detail import ProblemDetail, ProblemDetailException
+
+
+class ReportController(CirculationManagerController):
```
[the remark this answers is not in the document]
dbernstein: The test in test_report seems to rely on the CirculationManagerController.  I left it in there.

core/model/asynctask.py
```diff
@@ -0,0 +1,115 @@
...
+        .filter(AsyncTask.task_type == task_type)
+        .filter(AsyncTask.status == AsyncTaskStatus.READY)
+        .filter(AsyncTask.data.cast(String) == json.dumps(data))
+        .first()
+    )
+
+    if t:
+        return t, False
+    else:
+        return create(
```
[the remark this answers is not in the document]
dbernstein: I have a solution for this.

api/admin/controller/report.py
dbernstein: Re the SQL query:  there is a test covering it, so if there is a problem with the model, the test will fail.

core/model/asynctask.py
dbernstein: I'll add it in.

api/admin/controller/report.py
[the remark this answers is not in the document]
jonathangreen:
```diff
diff --git a/api/admin/controller/__init__.py b/api/admin/controller/__init__.py
index ddf6fef3a..83eaaf4cb 100644
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -64,4 +64,4 @@ def setup_admin_controllers(manager: CirculationManager):
     manager.admin_announcement_service = AnnouncementSettings(manager._db)
     manager.admin_search_controller = AdminSearchController(manager)
     manager.admin_quicksight_controller = QuickSightController(manager)
-    manager.admin_report_controller = ReportController(manager)
+    manager.admin_report_controller = ReportController(manager._db)
diff --git a/api/admin/controller/report.py b/api/admin/controller/report.py
index ccb182681..97c9463fa 100644
--- a/api/admin/controller/report.py
+++ b/api/admin/controller/report.py
@@ -4,6 +4,7 @@ from http import HTTPStatus

 import flask
 from flask import Response
+from sqlalchemy.orm import Session

 from api.controller.circulation_manager import CirculationManagerController
 from core.model import Library
@@ -14,10 +15,14 @@ from core.model.deferredtask import (
     queue_task,
 )
 from core.problem_details import INTERNAL_SERVER_ERROR
+from core.util.log import LoggerMixin
 from core.util.problem_detail import ProblemDetail


-class ReportController(CirculationManagerController):
+class ReportController(LoggerMixin):
+    def __init__(self, db: Session):
+        self._db = db
+
     def generate_inventory_report(self) -> Response | ProblemDetail:
         library: Library = getattr(flask.request, "library")
         admin: Admin = getattr(flask.request, "admin")
```

I think that patch removes the requirement on `CirculationManagerController`. There is a lot of code in `CirculationManagerController` and its parent class `BaseCirculationManagerController` that would be nice to avoid pulling into this controller if we don't need.

dbernstein:
@jonathangreen :  thanks for the reviews on this one.  I redid the integration query,  removed the pg_advisory_lock and made a note about race conditions.

tests/core/test_scripts.py
```diff
@@ -2564,6 +2577,194 @@ def test_suppress_work(self, db: DatabaseTransactionFixture):
...
+            external_account_id="http://opds.com",
+            data_source="BiblioBoard",
+        )
+        collection = db.collection(
+            name="BiblioBoard Test Collection", settings=settings.dict()
+        )
+        collection.libraries = [library]
+
+        # Configure test data we expect will not be picked up.
+        no_inventory_report_settings = OPDSImporterSettings(
```
dbernstein, line 2595: I added this second collection/datasource  to ensure we are not picking up collections that should be out of scope.

jonathangreen (approved):
Looks good 🚀 

Might be nice to try the change suggested here https://github.com/ThePalaceProject/circulation/pull/1734#discussion_r1534234710 but this is good to merge either way. 

https://github.com/ThePalaceProject/circulation/pull/1739 is also approved and also has a DB migration, so whichever PR goes in first, the other one will need to resolve the conflicts with the migration ordering before merging.

PR #1738, opened 2024-03-19
Fix some files excluded from mypy (PP-1080)

jonathangreen:
## Description

Fix a couple files that were excluded from `mypy`. That we had a todo comment to go do.

## Motivation and Context

Was trying to find a bit of work to do this morning that wasn't too taxing on my 🧠, since I'm still feeling sick and my 🧠 🌫️ is real.

## How Has This Been Tested?

Running tests in CI.

## Checklist


tdilauro (approved):
🎉😷

PR #1739, opened 2024-03-19
Playback time tables are independent of related tables. (PP-1044)

tdilauro:
## Description

- Removes dependency on related identifier, collection, and library tables from the audio playback time entry (from clients) and summary tables.
- Removes constraints that require the associated keys not to be null.
- Captures the identifier urn, collection name, library name, isbn, and title at multiple points along the way.
- Adds tests to verify this functionality.

## Motivation and Context

Need to be able to remove collections, libraries, and identifiers without losing tracked playback time.

[Jira [PP-1044](https://ebce-lyrasis.atlassian.net/browse/PP-1044)]

## How Has This Been Tested?

- New tests to verify functionality.
- Performed tests in local dev environment.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/8351483916) pass for associated branch.

## Checklist

[PP-1044]: https://ebce-lyrasis.atlassian.net/browse/PP-1044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/alembic/versions/20240318_3e43ed59f256_playtime_tables_have_no_dependencies.py
+++ b/alembic/versions/20240318_3e43ed59f256_playtime_tables_have_no_dependencies.py
@@ -1,0 +1,287 @@
+"""Playtime tables have no dependencies.
+
+Revision ID: 3e43ed59f256
+Revises: 9d2dccb0d6ff
+Create Date: 2024-03-18 01:34:28.381129+00:00
+
+"""
+from functools import cache
+
+import sqlalchemy as sa
+from sqlalchemy.engine import Connection
+from sqlalchemy.orm.session import Session
+
+from alembic import op
+from core.model import Identifier, get_one
+from core.model.time_tracking import _isbn_for_identifier, _title_for_identifier
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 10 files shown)


jonathangreen (approved):
This looks good 👍🏻 

I'm still a bit confused if we need to be keep track of `identifier_id`, `collection_id` and `library_id`, rather then serializing them immediately, but I'll defer to your expertise here, I haven't delved fully into the playtime entry tracking.

jonathangreen:
FYI: Just approved https://github.com/ThePalaceProject/circulation/pull/1734 and it had a DB migration as well. So we'll need to coordinate the merging of these two PRs, since the DB migrations will conflict.

jonathangreen:
I've been reading through the playtime tracking code, I think its my first stroll through that part of the codebase. I see now what you mean about entries being merged into summaries.  This one looks good to me 🎸.

tdilauro:
I coordinated with @dbernstein on merging this. He will refit the migration on #1734 now that this one has been merged.

PR #1740, opened 2024-03-20
Output URNs for hrefs in findaway manfiests

jonathangreen:
## Description

Output manifests with href that look like: `urn:org.thepalaceproject:findaway:{fulfillmentId}:{item.part}:{item.sequence}` instead of `None` for findaway, so its easier for the mobile clients to use the readingOrder href to track things like bookmarks.

## Motivation and Context

See slack thread here: https://lyrasis.slack.com/archives/C03GKDAKU0H/p1710949080047149

## How Has This Been Tested?

Running tests in CI

## Checklist

### DIFF
--- a/api/web_publication_manifest.py
+++ b/api/web_publication_manifest.py
@@ -118,11 +118,11 @@ class FindawayManifest(AudiobookManifest):
         for item in spine_items:
             kwargs = {part_key: item.part, sequence_key: item.sequence}
             self.add_reading_order(
-                href=None,
+                href=f"urn:org.thepalaceproject:findaway:{item.part}:{item.sequence}",
                 title=item.title,
                 duration=item.duration,
                 type=item.media_type,
-                **kwargs
+                **kwargs,
             )
             total_duration += item.duration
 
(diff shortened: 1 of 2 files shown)


tests/api/test_bibliotheca.py
```diff
@@ -684,18 +685,23 @@ def test_findaway_license_to_webpub_manifest(
         assert 16.201 == first["duration"]
         assert "Track 1" == first["title"]
 
-        # There is no 'href' value for the readingOrder items because the
+        # The 'href' value for the readingOrder is a URN because the
```
tdilauro, line 688: Minor: This first paragraph could be taken to imply that the URN is used by the Findaway SDK. 

Possible rephrasing:
> The 'href' value for the `readingOrder` is a URN. We do not use it to retrieve files, since that must be done through the Findaway SDK.

api/web_publication_manifest.py
```diff
@@ -111,18 +111,21 @@ def __init__(
         # Add the SpineItems as reading order items. None of them will
         # have working 'href' fields -- it's just to give the client a
         # picture of the structure of the timeline.
+        # For the href we return a URN that includes the part and sequence
```
tdilauro, line 114: Minor: Might be worthwhile to remove the second sentence of the previous paragraph.

tdilauro (approved):
Looks good! Two minor comments on comments.

PR #1742, opened 2024-03-22
Ensure LibraryAnnotator passes its associated library when resolving active license pools.

dbernstein:
… resolving the active license pool for a work.

## Description

The issue appears to be that the library associated with the annotator in the context of the call to book linked via a lane (vs via search) - that execution passes through https://github.com/ThePalaceProject/circulation/blob/main/api/admin/controller/work_editor.py#L53 via (https://github.com/ThePalaceProject/circulation/blob/main/api/admin/routes.py#L152) is not being passed to the active license pool resolver resulting in potentially the wrong license pool being returned.

This update ensures that 1) LibraryAnnotators may not pass a library other than the one associated with the annotator to the active license pool resolver method and 2) a LibraryAnnotator will pass the annotator's library to the super class' resolver method even when the library is not explicitly passed as an argument to the subclass method.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-989

## Checklist

### DIFF
--- a/tests/api/feed/test_library_annotator.py
+++ b/tests/api/feed/test_library_annotator.py
@@ -1813,12 +1813,13 @@ class TestLibraryAnnotator:
         annotator_fixture: LibraryAnnotatorFixture,
         library_fixture: LibraryFixture,
     ):
-        # we would never want to
-        other_library = library_fixture.library("otherlibrary")
+        # we would never want to resolve an active license pool with a library other
+        # than the library that was passed to construct the LibraryAnnotator.
         annotator = LibraryAnnotator(None, None, annotator_fixture.db.default_library())
 
-        # This book has two delivery mechanisms
         work = annotator_fixture.db.work(with_license_pool=True)
+
+        other_library = library_fixture.library("otherlibrary")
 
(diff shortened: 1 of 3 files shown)


tests/api/feed/test_library_annotator.py
```diff
@@ -1806,3 +1807,23 @@ class MockAPI:
         assert (
             mech2.delivery_mechanism.content_type == link.indirect_acquisitions[0].type
         )
+
+    def test_library_conflict_when_resolving_active_license_pool(
+        self,
+        annotator_fixture: LibraryAnnotatorFixture,
+        library_fixture: LibraryFixture,
+    ):
+        # we would never want to
```
jonathangreen, line 1816: This comment seems incomplete?

jonathangreen (approved):
Looks good @dbernstein! 🎸 

Made one small note about a comment in a test that seems truncated that you might want to clean up before you merge this one.

tests/api/feed/test_library_annotator.py
dbernstein: yes - fill fix.

PR #1746, opened 2024-03-26
Update how we handle axis availability information 401 response (PP-1081)

jonathangreen:
## Description

This updates the logic we use to refresh the token when the axis API returns a 401 response. Looking at their API documentation, they don't always use a 401 response to mean that the token needs a refresh. 

When we are asking for availability information for patrons that haven't loaned content from a particular library before, the API returns a 401 response that looks like this:
```xml
<availabilityResponse xmlns="http://axis360api.baker-taylor.com/vendorAPI" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <status>
    <code>3122</code>
    <statusMessage>Patron information is not found.</statusMessage>
  </status>
</availabilityResponse>
```

Now we parse the message that comes back with the 401 response. If we are unable to parse the response, or if the response has a code in it that indicates we need to refresh the token then we attempt a token refresh.

## Motivation and Context

In PP-1081 the patron was seeing a book on their bookshelf that shouldn't have been there. This was happening because we were getting a failure when refreshing the users bookshelf from Axis which meant that we were not cleaning up the bookshelf entries since we didn't know the complete state of the bookshelf. 

Looking through our logs, it seems like this is happening fairly commonly. It looks like we saw the log message that this happened 4532 times in the last week. This could be causing these users to see their bookshelf as out of sync, so this will be good to get fixed.

## How Has This Been Tested?

- Running tests in CI
- Tested against the VT DB with the user noted in PP-1081

## Checklist

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -948,12 +948,9 @@ class StatusResponseParser(Axis360Parser[tuple[int, str]]):
     def process_one(
         self, tag: _Element, namespaces: dict[str, str] | None
     ) -> tuple[int, str] | None:
-        try:
-            status_code = self.int_of_subtag(tag, "axis:code", namespaces)
-            message = self.text_of_subtag(tag, "axis:statusMessage", namespaces)
-            return status_code, message
-        except ValueError:
-            return None
+        status_code = self.int_of_subtag(tag, "axis:code", namespaces)
+        message = self.text_of_subtag(tag, "axis:statusMessage", namespaces)
+        return status_code, message
 
     def process_first(
         self,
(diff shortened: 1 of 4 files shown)


api/axis.py
```diff
@@ -323,36 +323,34 @@ def request(
...
-        else:
-            return response
+        if response.status_code == 401 and not request_retried:
+            parsed = StatusResponseParser().process_first(response.content)
+            if parsed is None or parsed[0] in [1001, 1002, 1003]:
+                # The token is probably expired. Get a new token and try again.
+                # Axis 360's status codes mean:
+                #   1001: Invalid token
+                #   1002: Token expired
+                #   1003: Token is invalid for given library
```
tdilauro, line 341: Wondering if we should retry for `1003`, since it might mean retrying more than we need to. But it's probably 50/50 at worst, so maybe worth attempting anyway.

tdilauro (approved):
Looks good! 🛠️

Approved, but one small question below.

api/axis.py
jonathangreen: I wasn't sure on that one. I'll pull it out and we can add it back if needed.

PR #1747, opened 2024-03-27
Additional test for Axis token refresh

jonathangreen:
## Description

Adds one more test case for https://github.com/ThePalaceProject/circulation/pull/1746, the case where we get a expired token response. Also does a little bit of cleanup of the tests, since they were getting a little duplicative. 

## Motivation and Context

When I was working on https://github.com/ThePalaceProject/circulation/pull/1746 I hadn't actually been able to test with an expired token response. However today I was able to use a token I was testing with yesterday to generate one and verify that the API behavior matched the docs. Since I had the response, I figured it was also worth adding to the tests.

## How Has This Been Tested?

- Running tests in CI

## Checklist


tdilauro (approved):
Looks good! 🎸

PR #1760, opened 2024-04-03
Task queue using python celery + redis

jonathangreen:
## Description

Prototype task queue for CM using [Celery](https://docs.celeryq.dev/en/stable/getting-started/introduction.html) + [Redis](https://redis.io/). It implements a two tasks
 - delete a collection in the background when collection delete is selected in through the admin interface
 - populating customlists, both the scheduled maintenance task, and triggering a list to populate when changes are made in the admin UI.

~🚨 This is a work in progress, it can't be merged until more work is done:~
I believe all the pre-requisite work is now in place to deploy out to `tpp-dev`.
- [x] Celery workers running in scripts container
  - [x] Daemonized
  - [x] Logging
- [x] See if we are able to configure worker to preferentially choose a task from the `high` queue or if we need separate workers to prioritize like this.
- [x] Documentation added to README
  - [x] Redis
  - [x] New env vars
  - [x] Celery 
- [x] Tested multiple CM + single redis configuration
- [x] Hosting playbook updated
  - [x] Deploy redis
  - [x] New env vars set

This one is fairly large so it spans several jira tickets:
- PP-1142
- PP-1143
- PP-1147

## Motivation and Context

Some useful commands if you want to try running this locally:
```bash
# start redis
docker run --name redis -d redis -p 6379:6379
# setup env vars (assuming you have PG and OS containers already running and env vars set)
PALACE_CELERY_BROKER_URL="redis://localhost:6379"
# start CM
python app.py
# start celery worker
celery -A "core.celery.worker.app" worker --concurrency 1 --pool solo
```

Then you should be able to add a collection in the CM and delete it, and see the worker delete it in the background.

## How Has This Been Tested?

- Tested locally so far
- Tested with unit tests

## Checklist


core/celery/worker.py
```diff
@@ -0,0 +1,35 @@
...
+) -> None:
+    # Override the default Celery logger setup to use the logger configuration from the service container,
+    # this will likely need to be updated so that we respect some of the Celery specific configuration options.
+    ...
+
+
+services = container_instance()
+services.init_resources()
+import_celery_tasks()
+app = services.celery.app()
```
jonathangreen, line 35: This defines a global `app` instance, so that we can use it in the Celery worker script. That is the only place this file is used.

core/celery/worker.py
```diff
@@ -0,0 +1,35 @@
...
+from core.service.container import container_instance
+
+
+def import_celery_tasks() -> None:
+    tasks_path = Path(__file__).parent / "tasks"
+    for task_file in tasks_path.glob("*.py"):
+        if task_file.stem == "__init__":
+            continue
+        module = f"core.celery.tasks.{task_file.stem}"
+        importlib.import_module(module)
```
jonathangreen, line 20: This does some programmatic imports to make sure that all our tasks are imported when starting the worker processes. This assumes that all our tasks will live in the `core.celery.tasks` package.

core/service/celery/celery.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any
+
+from celery import Celery
+from kombu import Exchange, Queue
+
+
+def task_queue_config(cm_name: str) -> dict[str, Any]:
+    high_prefixed = f"{cm_name}:high"
+    default_prefixed = f"{cm_name}:default"
```
jonathangreen, line 9: Prefix our task queues with the `cm_name`, so that hopefully we are able to have a redis instance per-shard when we roll this out. This has only been lightly tested. It seems like it will work because Celery creates redis keys for each queue, but I need to do a lot more testing of this part.

core/service/celery/celery.py
```diff
@@ -0,0 +1,31 @@
...
+
+    return {
+        "task_queues": (
+            Queue(high_prefixed, Exchange(high_prefixed), routing_key="high"),
+            Queue(default_prefixed, Exchange(default_prefixed), routing_key="default"),
```
jonathangreen, line 14: For this prototype I've defined two queues, one high priority queue, and one low priority queue. The idea is to use the high priority queue for fast running tasks, and the default queue for longer running tasks, where starvation is less of a concern. As we roll this out we may need to adjust the queues we use to be finer grained then this, but this seemed reasonable for a prototype.

core/service/celery/celery.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def celery_factory(config: dict[str, Any]) -> Celery:
+    # Create a new Celery app
+    cm_name = config.get("cm_name")
+    assert isinstance(cm_name, str)
+    app = Celery(cm_name, task_cls="core.celery.task:Task")
+    app.conf.update(config)
+    app.conf.update(task_queue_config(cm_name))
+    app.set_default()
```
jonathangreen, line 29: This sets up the Celery app from the dependency injection container as the "default app", so that any web tasks interacting with the `shared_task` decorator know to use this `app` instance.

tests/fixtures/celery.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.fixture()
+def celery_fixture(
+    services_fixture: ServicesFixture,
+    mock_session_maker: MockSessionMaker,
+    celery_app: Celery,
+    celery_worker: WorkController,
+    celery_pydantic_config: CeleryConfiguration,
+) -> Generator[CeleryFixture, None, None]:
```
jonathangreen, line 85: This is the main fixture for `Celery` that can be imported to be able to do a complete end to end run of a task inside a test. The task will be run in a worker thread.

core/celery/task.py
```diff
@@ -0,0 +1,26 @@
...
+import celery
+from sqlalchemy.orm import Session, sessionmaker
+from sqlalchemy.pool import NullPool
+
+from core.model import SessionManager
+from core.service.container import Services, container_instance
+from core.util.log import LoggerMixin
+
+
+class Task(celery.Task, LoggerMixin):
```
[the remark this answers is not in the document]
jonathangreen: Moved into a comment in the code

core/celery/task.py
```diff
@@ -0,0 +1,26 @@
...
+    _session_maker = None
+
+    @property
+    def session_maker(self) -> sessionmaker[Session]:
+        if self._session_maker is None:
+            engine = SessionManager.engine(poolclass=NullPool)
```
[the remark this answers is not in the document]
jonathangreen: Moved into a comment in the code

core/service/celery/celery.py
jonathangreen: Moved into a comment in the code

core/service/celery/celery.py
jonathangreen: Moved into a comment in the code

tests/core/celery/tasks/test_collection_delete.py
```diff
@@ -0,0 +1,61 @@
...
+    assert db.session.execute(query).scalar_one_or_none() == collection
+
+    assert collection.id is not None
+    job = CollectionDeleteJob(mock_session_maker, collection.id)
+    job.run()
+    assert db.session.execute(query).scalar_one_or_none() is None
+    assert f"Deleting collection" in caplog.text
+
+
+def test_collection_delete_task(
```
dbernstein, line 53: Does the celery_fixture need to be here?  It doesn't look like it is referenced.

dbernstein:
@jonathangreen :  this is very slick.  The comments in the PR and in the code are very helpful.  I like how the handling of the task queue is completely transparent.   I'll be able to abandon the deferredtasks table for the inventory reports more or less immediately and with little effort.

tests/core/celery/tasks/test_collection_delete.py
jonathangreen: It needs to be here, because pulling in this fixture starts the worker thread the runs the task. Without the fixture reference the job will fail to run.

core/service/celery/celery.py
jonathangreen: Changed this to use a broker option instead.

tdilauro:
@jonathangreen I know that @dbernstein is reviewing this one, but I wanted to advocate for a couple extra bits of documentation:
- It would be useful to add some details on how we would manually kick off tasks that we are currently able to start by hopping onto a scripts server and running from the command line.
- Some guidance on (or pointers to same) on developing and debugging celery tasks.

jonathangreen:
> It would be useful to add some details on how we would manually kick off tasks that we are currently able to start by hopping onto a scripts server and running from the command line.

This facility doesn't exist, but its a good idea. I added a ticket for it here PP-1153, so we can work on adding that. I don't think having that is a blocker for this going in, given the couple tasks that this PR converts to use Celery but let me know if you disagree.

> Some guidance on (or pointers to same) on developing and debugging celery tasks.

I think this is something we are going to work out over time. I'm sure we'll know a lot more after moving all the scripts over to use celery. The most valuable source of information I've found is the celery documentation: https://docs.celeryq.dev/en/stable/index.html but I already have that linked in the README.

I did try to write some detailed inline comments about the Palace specific celery parts, with links out to the Celery documentation:
https://github.com/ThePalaceProject/circulation/blob/b2dda94106201da1d41f32d96e45e68485529747/core/celery/job.py#L11-L30
https://github.com/ThePalaceProject/circulation/blob/b2dda94106201da1d41f32d96e45e68485529747/core/celery/task.py#L13-L38

Is there anything more specific you would like to see here?

tdilauro:
No, I didn't have anything specific in mind, other than achieving the things mentioned in those bullet points. I was mainly thinking of any insights that you've gleaned as you've done this work, especially those that would help/save time for others (like me 😅) coming along and working on, trying to test, and debugging them. Also, I guess, anything gotchas that surprised you or weren't straightforward from the official docs. Thanks!

jonathangreen:
Despite my initial reluctance, I've been impressed with Celery so far. The docs are really good, and I haven't hit a lot of rough edges yet. The biggest issue I had was getting the worker to run on my mac, but the `celery -A "core.celery.worker.app" worker --concurrency 1 --pool solo` command I put in the docs gets a worker running locally. 

The workers don't automatically reload on code changes, that might be something nice to look at for development in the future. I saw a couple things in google about using the `watch` package to accomplish this, but I haven't tried it yet. I've just been restarting the service. 

That hasn't been too bad, since the `Job` class is easily testable. I've just been doing the conversion of the existing code via `Job` getting tests running, then doing a bit of testing with the worker. The two scripts I converted though were fairly simple ones. I'm sure we'll learn a lot converting one of the importers, since its a much more complicated operation.

dbernstein (approved):
@jonathangreen :  very impressive work.   Including an example of a script conversion to the celery approach (ie the custom lists update script) is very helpful.  It is encouraging to see that the process is fairly straightforward. 

The PR looks good.  I have no specific changes to request. 

However I am wondering about one thing:   What happens if the celery task queue processing slows down to the point where more tasks are going on than are coming off?  In the case of the scheduled custom list updates,  should the custom list update tasks not complete in the hour before the task scheduler runs again,  will we see duplicate tasks going on the queue? 

I don't see a problem with this as long as we can eventually vacate the queue.   The autoscaling of the celery workers that you've set up should take care of this.   But I'm guessing we'll need an alert to tell us if the task queues getting backed up.    Tracking the size of the queues and having some visibility into the composition (broken down by task type) might be useful.  You may have already called these wants/needs out - I may have overlooked it or forgotten - but I put them here in case you haven't.

jonathangreen:
> However I am wondering about one thing: What happens if the celery task queue processing slows down to the point where more tasks are going on than are coming off? In the case of the scheduled custom list updates, should the custom list update tasks not complete in the hour before the task scheduler runs again, will we see duplicate tasks going on the queue?

Yes we will see duplicate tasks going into the queue in this case. This will likely pose a problem for larger scripts, so as we convert those we will have to be thinking about this case. The celery docs have a suggestion for using locks for this purpose: https://docs.celeryq.dev/en/stable/tutorials/task-cookbook.html

The custom list updates process reasonably fast as is, so I thought it was a good first candidate to convert without having to immediately deal with this possibility.

> I don't see a problem with this as long as we can eventually vacate the queue. The autoscaling of the celery workers that you've set up should take care of this. But I'm guessing we'll need an alert to tell us if the task queues getting backed up. Tracking the size of the queues and having some visibility into the composition (broken down by task type) might be useful. You may have already called these wants/needs out - I may have overlooked it or forgotten - but I put them here in case you haven't.

Agreed we will need some visibility into the queues. I've got an epic going with additional work for task queues: PP-107. That is where I've been trying to capture the future work, since this PR is just the beginning of the work in getting things converted. I think PP-1150 in that epic will cover adding the monitoring we need. Any input / details on those additional tasks is appreciated though. I'm hoping to be able to pull PP-1150 into the next sprint to work on.

PR #1761, opened 2024-04-04
PP-1106 Inventory report updates

dbernstein:
## Description
This update includes several changes: 
* The genre, audience and data source columns have been added.
* The inventory report is now consolidated into one CSV with all relevant data sources included
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1106

## How Has This Been Tested?

Unit tests updated.
## Checklist

### DIFF
--- a/core/scripts.py
+++ b/core/scripts.py
@@ -2914,7 +2914,7 @@ class GenerateInventoryReports(Script):
                d.name data_source,
                ic.name collection_name,
                l.expires license_expiration,
-               DATE_PART('day', l.expires - CURRENT_DATE) days_remaining_on_license,
+               DATE_PART('day', l.expires - now()) days_remaining_on_license,
                l.checkouts_left remaining_loans,
                l.terms_concurrency allowed_concurrent_users,
                coalesce(lib_loans.active_loan_count, 0) library_active_loan_count,
@@ -2956,7 +2956,7 @@ class GenerateInventoryReports(Script):
                                                      checkouts_left,
                                                      expires,
                                                      terms_concurrency
-                                              from licenses) l on lp.id = l.license_pool_id
+                                              from licenses where status = 'available') l on lp.id = l.license_pool_id
         where lp.identifier_id = i.id and
               e.primary_identifier_id = i.id and
               e.id = w.presentation_edition_id and
@@ -3007,7 +3007,10 @@ class GenerateInventoryReports(Script):
                                        patrons p
                                   where p.id = h.patron_id and
                                         h.license_pool_id = lp.id and
-                                        p.library_id = :library_id
+                                        p.library_id = :library_id and
+                                        (h.end is null or
+                                        h.end > now() or
+                                        h.position > 0)
                                   group by p.library_id, lp.presentation_edition_id) lib_holds,
              identifiers i,
              (select ilc.parent_id,
--- a/core/scripts.py
+++ b/core/scripts.py
@@ -2863,11 +2863,8 @@ class GenerateInventoryReports(Script):
 
     def generate_report(self, data_source_name: str, library_id: int) -> str:
         """Generate a csv file and return the file path"""
-        with tempfile.NamedTemporaryFile(
-            "w",
-            delete=False,
-        ) as temp:
-            writer = csv.writer(temp, delimiter=",", encoding="utf-8")
+        with tempfile.NamedTemporaryFile("w", delete=False, encoding="utf-8") as temp:
+            writer = csv.writer(temp, delimiter=",")
             rows = self._db.execute(
                 text(self.inventory_report_query()),
                 {"library_id": library_id, "data_source_name": data_source_name},
@@ -2885,6 +2882,8 @@ class GenerateInventoryReports(Script):
                e.language,
                e.publisher,
                e.medium as format,
+               w.audience,
+               wg.genres,
                ic.name collection_name,
                DATE_PART('day', l.expires::date) - DATE_PART('day',lp.availability_time::date) as license_duration_days,
                l.expires license_expiration_date,
@@ -2905,6 +2904,11 @@ class GenerateInventoryReports(Script):
              integration_configurations ic,
              integration_library_configurations il,
              libraries lib,
+             works w left outer join (select wg.work_id, string_agg(g.name, ',' order by g.name) as genres
+                                     from genres g,
+                                     workgenres wg
+                                     where g.id = wg.genre_id
+                                     group by wg.work_id) wg on w.id = wg.work_id,
              editions e left outer join (select lp.presentation_edition_id,
                                          p.library_id,
                                          count(h.id) active_hold_count
@@ -2942,6 +2946,7 @@ class GenerateInventoryReports(Script):
              licensepools lp left outer join licenses l on lp.id = l.license_pool_id
         where lp.identifier_id = i.id and
               e.primary_identifier_id = i.id and
+              e.id = w.presentation_edition_id and
               d.id = e.data_source_id and
               c.id = lp.collection_id and
               c.integration_configuration_id = ic.id and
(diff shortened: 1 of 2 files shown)


core/scripts.py
```diff
@@ -2804,46 +2798,37 @@ def do_run(self, cmd_args: list[str] | None = None) -> None:
...
-                    not_(
-                        IntegrationConfiguration.settings_dict.contains(
-                            {"include_in_inventory_report": False}
-                        )
-                    ),
-                )
-            ).all()
+            attachments: dict[str, str] = {}
+            library = get_one(self._db, Library, id=data.library_id)
+            assert library
```
jonathangreen, line 2807: If a library is deleted, between when the task is queued, and when its run, library will be `None`. Which will trigger this assertion. Maybe we want to handle this more gracefully then just dying with an `AssertionError`?

core/scripts.py
```diff
@@ -2861,16 +2846,28 @@ def process_task(self, task: DeferredTask):
             for file in files:
                 os.remove(file.name)
 
-    def generate_report(self, data_source_name: str, library_id: int) -> str:
-        """Generate a csv file and return the file path"""
-        with tempfile.NamedTemporaryFile(
-            "w",
-            delete=False,
-        ) as temp:
+    def prepare_email_attachment(self, attachments, file_path, attachment_name, files):
```
jonathangreen, line 2849: Minor: Missing type hints on this function

core/scripts.py
```diff
@@ -2861,16 +2846,28 @@ def process_task(self, task: DeferredTask):
...
+        """
+        extract contents of files and prepare in a dictionary of email attachments
+        """
+        with open(file_path) as the_file:
+            attachments[attachment_name] = the_file.read()
+            files.append(the_file)
```
jonathangreen, line 2855: After you exit the `with` here, this is no longer a valid file handle, so it seems dangerous to add the closed file handle to the files list.

core/scripts.py
```diff
@@ -2804,46 +2798,37 @@ def do_run(self, cmd_args: list[str] | None = None) -> None:
 
     def process_task(self, task: DeferredTask):
         data = InventoryReportTaskData(**task.data)
-        files = []
+        files: list[IO] = []
```
jonathangreen, line 2801: Minor: Since we only have a single file now, does this really need to be a list?

core/scripts.py
```diff
@@ -2861,16 +2846,28 @@ def process_task(self, task: DeferredTask):
...
-        with tempfile.NamedTemporaryFile(
-            "w",
-            delete=False,
-        ) as temp:
+    def prepare_email_attachment(self, attachments, file_path, attachment_name, files):
+        """
+        extract contents of files and prepare in a dictionary of email attachments
+        """
+        with open(file_path) as the_file:
+            attachments[attachment_name] = the_file.read()
```
jonathangreen, line 2854: This file can potentially be quite large right? I don't think we should read its whole contents into memory here, thats going to blow up the memory usage of this process.

core/scripts.py
```diff
@@ -2905,45 +2899,113 @@ def inventory_report_query(self) -> str:
...
+              e.id = lib_holds.presentation_edition_id and
+              e.id = w.presentation_edition_id and
               d.id = e.data_source_id and
               c.id = lp.collection_id and
               c.integration_configuration_id = ic.id and
               ic.id = il.parent_id and
               ic.id = collection_sharing.parent_id and
+              ic.goal = 'LICENSE_GOAL' and
+              (ic.settings -> 'include_in_inventory_report' = 'true' or
+              ic.settings -> 'include_in_inventory_report' is null) and
```
jonathangreen, line 3007: I don't think this is going to filter for what you want. Since you are just checking if this key isn't set, won't this select all of the collections in a CM, not just the collections we are interested in.

core/scripts.py
```diff
@@ -2905,45 +2899,113 @@ def inventory_report_query(self) -> str:
...
+              c.integration_configuration_id = ic.id and
+              ic.id = il.parent_id and
+              ic.id = collection_sharing.parent_id and
+              ic.goal = 'LICENSE_GOAL' and
+              (ic.settings -> 'include_in_inventory_report' = 'true' or
+              ic.settings -> 'include_in_inventory_report' is null) and
+              il.library_id = lib.id and
+              lib.id = :library_id
+         order by title, author
+        """
```
jonathangreen, line 2947: Minor: There is a lot of duplication between these queries, it would be nice to not have to duplicate so much code here.

tests/core/test_scripts.py
```diff
@@ -2589,7 +2590,7 @@ def test_do_run(self, db: DatabaseTransactionFixture):
         collection = db.collection(
             name="BiblioBoard Test Collection", settings=settings.dict()
         )
-        collection.libraries = [library]
+        collection.libraries = [library, library2]
```
jonathangreen, line 2593: It would be good to include some other collection types here, ones that we don't expect to be picked up like an overdrive collection etc.

tests/core/test_scripts.py
```diff
@@ -2657,17 +2675,19 @@ def test_do_run(self, db: DatabaseTransactionFixture):
         assert kwargs["subject"].__contains__("Inventory Report")
```
jonathangreen, line 2675: why use `.__contains__` directly here instead of the `in` operator?

tests/core/test_scripts.py
```diff
@@ -2657,17 +2675,19 @@ def test_do_run(self, db: DatabaseTransactionFixture):
...
         attachments: dict = kwargs["attachments"]
 
-        assert len(attachments) == 1
-        key = [*attachments.keys()][0]
-        assert "biblioboard" in key
-        value = attachments[key]
-        assert len(value) > 0
-        csv_file = StringIO(value)
-        reader = csv.reader(csv_file, delimiter=",")
+        assert len(attachments) == 2
```
jonathangreen, line 2678: Why two attachments, aren't we supposed to be putting all the files into a single attachment?

tests/core/test_scripts.py
```diff
@@ -2657,17 +2675,19 @@ def test_do_run(self, db: DatabaseTransactionFixture):
...
+        inventory_report_key = [x for x in attachments.keys() if "inventory" in x][0]
+        assert inventory_report_key
+        assert "test_library" in inventory_report_key
+        inventory_report_value = attachments[inventory_report_key]
+        assert len(inventory_report_value) > 0
+        inventory_report_reader = csv.reader(
```
jonathangreen, line 2684: I think these tests would be a lot easier to read if you used a [csv.DictReader](https://docs.python.org/3/library/csv.html#csv.DictReader) here. Then you don't have to do all the manipulation with indexes and finding the first row etc.

core/scripts.py
```diff
@@ -2905,45 +2899,113 @@ def inventory_report_query(self) -> str:
...
                                   from loans ln,
+                                       licensepools lp,
                                        patrons p,
                                        libraries l
                                   where p.id = ln.patron_id and
                                         p.library_id = l.id and
+                                        ln.license_pool_id = lp.id and
                                         l.id = :library_id
-                                  group by p.library_id, ln.license_pool_id) lib_loans on lp.id = lib_holds.license_pool_id
+                                  group by p.library_id, lp.presentation_edition_id) lib_loans
```
jonathangreen, line 2918: Minor: Have you tested these queries against any of our production libraries?

core/scripts.py
dbernstein: I'll fix.

core/scripts.py
dbernstein: I'll fix.

core/scripts.py
dbernstein: We still have multiple files: holds and inventory.

jonathangreen (approved):
Looks good

PR #1762, opened 2024-04-04
Add __init__.py files

jonathangreen:
## Description

Some of our modules were missing `__init__.py` files to mark the directory as a module. I think this is causing some subtle problems with imports, since that makes these folder namespace packages instead of modules.

## Motivation and Context

Make sure we are consistently using `__init__.py` to mark modules in the package.

## How Has This Been Tested?

- Running tests in CI.

## Checklist


tdilauro (approved):
🚀 

Funny. I don't ever recall running into a problem because of this.

PR #1763, opened 2024-04-04
Fix interpolation issue with JSONFormatter

jonathangreen:
## Description

Calling logging with a format string that expects a mapping is supported in Python, but this case wasn't handled correctly by `JSONFormatter`.

```python
logging.info('%(value)s', {"value": "bar"})
```

Would result an exception, which resulted in a log message saying:
```
Log message could not be formatted. Exception: %r.
```

Update the formatter to be more resilient, and add some test cases to make sure we are handling this. While I was adding the test cases, I cleaned up the tests a bit as well by adding a fixture to reduce the duplication there.

## Motivation and Context

While working on the Celery PR, I came across this bug with our default log formatter. Celery uses a lot of log messages in this style, so I was seeing A LOT of these errors trying to get log handlers setup with Celery.

## How Has This Been Tested?

- Running unit tests
- Testing with Celery locally

## Checklist

### DIFF
--- a/core/service/logging/log.py
+++ b/core/service/logging/log.py
@@ -3,14 +3,14 @@ from __future__ import annotations
 import json
 import logging
 import socket
-from collections.abc import Callable
+from collections.abc import Callable, Mapping, Sequence
 from logging import Handler
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 🛠️

PR #1764, opened 2024-04-05
Update type hint for send_email

jonathangreen:
## Description

Small change to update the type hint for `send_email`. Update attachments to be a Mapping rather then a dict.

## Motivation and Context

Since Mapping is immutable, it allows covariance for the value type. See https://mypy.readthedocs.io/en/stable/common_issues.html#variance. Updating this type hint to be a mapping allows us to pass `dict[bytes]` into `send_email` rather then always having to make sure `attachments` has all the types defined in the function type hint.

## How Has This Been Tested?

- Tested locally

## Checklist

### DIFF
--- a/core/service/email/email.py
+++ b/core/service/email/email.py
@@ -43,6 +43,12 @@ def send_email(
         receivers=receivers,
         text=text,
         html=html,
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
🎸

PR #1772, opened 2024-04-11
Remove setting to disable token auth

jonathangreen:
## Description

Now that token auth is rolled out everywhere, remove setting to disable it.

## Motivation and Context

Token auth was rolled out everywhere in PP-951.

## How Has This Been Tested?

- Running tests.

## Checklist


tdilauro (approved):
🔥

PR #1773, opened 2024-04-11
Fix flaky lane test

jonathangreen:
## Description

Updates assertion to compare a set, rather then a list, since order doesn't matter here, and can change.

## Motivation and Context

Saw this test failing randomly in CI.

## How Has This Been Tested?

Running tests.

## Checklist


tdilauro (approved):
👍🏽 Looks good!

PR #1774, opened 2024-04-11
Update the CustomListsController to use ProblemDetailException

jonathangreen:
## Description

Instead of using `Exception` and checking if it contains a `ProblemDetail`, just use the `ProblemDetailException` class.
 
## Motivation and Context

Noticed this as part of #1760, but I wanted to break it out since its not directly related.

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/api/admin/controller/custom_lists.py
+++ b/api/admin/controller/custom_lists.py
@@ -39,7 +39,7 @@ from core.model import (
 )
 from core.problem_details import INVALID_INPUT, METHOD_NOT_ALLOWED
 from core.query.customlist import CustomListQueries
-from core.util.problem_detail import ProblemDetail
+from core.util.problem_detail import ProblemDetail, ProblemDetailException
 
 
 class CustomListsController(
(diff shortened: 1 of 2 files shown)


api/admin/controller/custom_lists.py
```diff
@@ -180,42 +180,41 @@ def _create_or_update_list(
             if auto_update_query is not None:
                 try:
                     auto_update_query_str = json.dumps(auto_update_query)
-                except json.JSONDecodeError:
-                    raise Exception(
+                except TypeError:
```
jonathangreen, line 183: This is updated from `JSONDecodeError` to `TypeError`, because previously it had no test coverage and when I did this update I decided to add a test here. It turns out that it was catching the wrong exception. `json.dumps` doesn't raise `JSONDecodeError`.

jonathangreen:
Missed one exception when I was updating to `TypeError`first time around. Should be good to go now.

tests/api/admin/controller/test_custom_lists.py
```diff
@@ -1160,3 +1161,64 @@ def test_auto_update_edit(self, admin_librarian_fixture: AdminLibrarianFixture):
...
+            auto_update=True,
+            auto_update_query={"query": "foo"},
+            auto_update_facets={"foo": object()},  # type: ignore[dict-item]
+        )
+
+        assert isinstance(response, ProblemDetail)
+        assert response.status_code == 400
+        assert response.detail == "auto_update_facets is not JSON serializable"
+
+    def test_auto_update_deleted_entries(
```
tdilauro, line 1208: Do we want a test for entries here, as well? Or maybe there is one somewhere else already?

tdilauro (approved):
A good catch and some nice clean up in here.

One question about a possible additional test below.

tests/api/admin/controller/test_custom_lists.py
jonathangreen: Entries gets some test coverage here:
https://github.com/ThePalaceProject/circulation/blob/965d4225ee09ce95454e72bb0015c6ab8081df37/tests/api/admin/controller/test_custom_lists.py#L368-L395

PR #1776, opened 2024-04-11
Don't fail to delete work just because it's missing from search index.

tdilauro:
## Description

- Changes work deletion behavior, such that:
  - deletion does not fail just because the work is missing from the search index (for whatever reason); and
  - a warning is emitted if the work is missing from the search index, so that we can track how often that is happening, if we need to.

## Motivation and Context

Work deletion was failing when we attempted to delete the work from the search index, if the work was not present there. Since our goal was to delete the work anyway, we still ended up in the desired state, so it doesn't make sense to fail here.

The failure to delete the work could prevent other parts of the `database_reaper` from completing successfully.

## How Has This Been Tested?

- Created new tests for this behavior.
- Ran tests locally.
- [CI tests passed](https://github.com/ThePalaceProject/circulation/actions/runs/8653273302) for associated branch.

## Checklist

### DIFF
--- a/tests/core/models/test_work.py
+++ b/tests/core/models/test_work.py
@@ -5,8 +5,8 @@ from unittest.mock import MagicMock
 import opensearchpy
 import pytest
 import pytz
-from _pytest.logging import LogCaptureFixture
 from psycopg2.extras import NumericRange
+from pytest import LogCaptureFixture
 from sqlalchemy import select
 
 from core.classifier import Classifier, Fantasy, Romance, Science_Fiction
--- a/tests/core/models/test_work.py
+++ b/tests/core/models/test_work.py
@@ -1,8 +1,11 @@
 import datetime
+from contextlib import nullcontext
 from unittest.mock import MagicMock
 
+import opensearchpy
 import pytest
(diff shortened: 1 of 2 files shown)


tests/core/models/test_work.py
```diff
@@ -1,8 +1,11 @@
 import datetime
+from contextlib import nullcontext
 from unittest.mock import MagicMock
 
+import opensearchpy
 import pytest
 import pytz
+from _pytest.logging import LogCaptureFixture
```
jonathangreen, line 8: Minor: I noticed just today that `pytest` started exporting these symbols for use. So we can import for type checking `from pytest import LogCaptureFixture` to avoid the private package. I was thinking I'd make a PR to clean this up in the code base, might be worth changing here.

jonathangreen (approved):
Looks great!

tests/core/models/test_work.py
jonathangreen: Made a quick PR for this since I was thinking about it:
https://github.com/ThePalaceProject/circulation/pull/1777

PR #1777, opened 2024-04-11
Replace _pytest imports with public pytest types

jonathangreen:
## Description

Previously pytest didn't export the types for their fixtures in the public package. This made type checking hard. It looks like at some point they did start doing so in this PR: https://github.com/pytest-dev/pytest/pull/8017

## Motivation and Context

Update to use the public types rather then the internal `_pytest` package.

## How Has This Been Tested?

- Running in CI

## Checklist


tdilauro (approved):
🚀 Yay!

PR #1782, opened 2024-04-12
Update fcm to use service container (PP-1154)

jonathangreen:
## Description

Update FCM to load via service container, and load its settings via a pydantic settings class.

Updated the env variables for FCM from `SIMPLIFIED_*` to `PALACE_*`.

🚨 This one can't get merged until there is a hosting playbook PR to update the env variable name.

PP-1154

## Motivation and Context

Two reasons:
- Trying to move all our env vars to be prefixed with `PALACE_`
- Trying to make sure all our config is loaded via Pydantic settings classes

I took the opportunity while doing this to move FCM service into the service container, so we can more easily mock it.

## How Has This Been Tested?

- Running tests in CI

## Checklist


core/service/fcm/container.py
```diff
@@ -0,0 +1,31 @@
...
+
+from dependency_injector import providers
+from dependency_injector.containers import DeclarativeContainer
+from firebase_admin import App, initialize_app
+from firebase_admin.credentials import Certificate
+
+from core.service.fcm.fcm import credentials
+
+if TYPE_CHECKING:
+    pass
```
tdilauro, line 13: Do we need this for a side effect?

tdilauro (approved):
Looks good! 🚀 

One question below, but not a show stopper.

core/service/fcm/container.py
jonathangreen: Nope, just a leftover that should get cleaned up. I'll remove it.

PR #1783, opened 2024-04-12
PP-1002: Improve the reliability of Overdrive New Titles Monitor

dbernstein:
by checking for multiple consecutive out of scope dateAdded values before stopping.

## Description
Since Overdrive doesn't guarantee that the 'dateAdded' field will be strictly chronological when you specify 'dateAdded:desc', this update ensures that multiple (500) consecutive books have a dateAdded property that is out of scope before stopping the script.  Out of order dates should appear towards the start of the list.  There is an explanation in the code for why this is the case.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1002

## How Has This Been Tested?
I wrote a new test confirming the new behavior.

## Checklist

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -727,8 +727,11 @@ class OverdriveAPI(
 
         next_link = self.endpoint(
             self.EVENTS_ENDPOINT,
+            # From https://developer.overdrive.com/apis/search:
+            # "**Note: When you search using the lastTitleUpdateTime or
+            # lastUpdateTime parameters, your results will be automatically
+            # sorted in ascending order (and all other sort options will be ignored)."
             lastupdatetime=last_update,
-            sort="popularity:desc",
             limit=self.PAGE_SIZE_LIMIT,
             collection_token=self.collection_token,
         )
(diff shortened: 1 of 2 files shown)


jonathangreen:
@dbernstein the code here looks good to me.

My question is more around the design here. Where did `MAX_CONSECUTIVE_OUT_OF_SCOPE_DATES = 1000` come from? Did you derive this value from looking at the feeds from our production libraries or is this a value you guessed at?

I'd like to know if this value will ever be breached when harvesting the feeds from our production libraries, or if this change is effectively just going to make this monitor run through all the overdrive ids on every run?

dbernstein:
@jonathangreen :  This is based on what I observed.  The entries that were out of order appeared at the top of the list and did not  appear to exceed a few dozen  in a row.  With this change, I would not expect the list to process more than two thousand items.  I could probably make the number smaller, but I figured better to overcompensate in this case.   So in a nutshell,  yes, the number should be breached on every feed and quickly (assuming there are more than 1000 books in it).

jonathangreen (approved):
Thanks for the additional info 👍🏻

PR #1793, opened 2024-04-17
Fix thread safety issue in patron_activity (PP-1155)

jonathangreen:
## Description

This fixes a thread safety issue, where SQLAlchemy objects that are part of one thread's session are used in another thread. Leading to unpredictable behavior and deadlocks among the threads when under load. This was causing issues in the California CM instance when it was under increased load due to Aspen integration.

## Motivation and Context

Resolve issues where threads and workers deadlock with each other, trying to update the patron information.

## How Has This Been Tested?

- Tested locally with CA DB and UWSGI
- I wasn't able to make it deadlock, the way I see in production but I did see worker threads flushing changes from the main thread's Session.
- This seems to indicate my hypothesis of session mixing is at least a possible cause here.

## Checklist

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -15,6 +15,7 @@ from flask import Response
 from flask_babel import lazy_gettext as _
 from pydantic import PositiveInt
 from sqlalchemy.orm import Query
+from sqlalchemy.orm.scoping import ScopedSession
 
 from api.circulation_exceptions import (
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
🧵🚨 Looks great!

I made a couple minor edits to the main PR description.

PR #1794, opened 2024-04-17
[PP-1157] convert inventory and hold reports to celery

dbernstein:
## Description

This update converts the inventory and hold reports to use the celery task processing infrastructure.
Since nothing else is using the deferredtasks,  I've removed all related code with the exception of the deferredtasks table itself which I'll remove in a separate PR (via a migration).

One note:  I am no longer checking if a similar task exists or is process before issuing the request for a report to be generated.   @jonathangreen or @tdilauro:  do you think this is a requirement for this feature?  If so,  I can do a little research into how best to do this.  The reason it seems less important to me now is that the reports will be generated more or less immediately as opposed to possibly waiting several minutes before starting up (when we were using deferredtasks).  So the chance that duplicate requests will be made is less likely.

## Motivation and Context

https://ebce-lyrasis.atlassian.net/browse/PP-1157

## How Has This Been Tested?
New tests have been written.

## Checklist


api/admin/controller/report.py
```diff
@@ -31,21 +28,17 @@ def generate_inventory_report(self) -> Response | ProblemDetail:
...
-                f"An inventory report request was {'already' if not is_new else ''} received at {task.created}. "
-                f"When processing is complete, the report will be sent to {admin.email}."
+                f"An inventory and hold report request was received. Report processing can take a few minutes to "
+                f"finish depending on current server load. The completed reports will be sent to {admin.email}."
             )
 
-            self.log.info(msg + f" {task}")
-            http_status = HTTPStatus.ACCEPTED if is_new else HTTPStatus.CONFLICT
-            return Response(json.dumps(dict(message=msg)), http_status)
+            self.log.info(msg)
```
jonathangreen, line 40: Minor: If you wanted to log the task id, you could get the task from the `.delay` call and log its `.id` here. I only mention that because we were logging the task ID before. Might not be necessary though.

core/celery/tasks/generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,292 @@
...
+    Library,
+    get_one,
+)
+from core.opds_import import OPDSImporterSettings
+from core.service.celery.celery import QueueNames
+from core.service.container import Services, container_instance
+
+
+class GenerateInventoryAndHoldsReportsJob(Job):
+    _delete_attachments = True
```
jonathangreen, line 30: Minor: Might be better to make this a instance variable, and pass it into the constructor instead.

core/celery/tasks/generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,292 @@
...
+    def __init__(
+        self, session_maker: sessionmaker[Session], library_id: int, email_address: str
+    ):
+        super().__init__(session_maker)
+        self.library_id = library_id
+        self.email_address = email_address
+
+    @property
+    def services(self) -> Services:
+        return container_instance()
```
jonathangreen, line 41: Instead of loading the container instance here in the job, can you pass the `send_email` function into `GenerateInventoryAndHoldsReportsJob` constructor from the task? Loading the container this way works... but I'd like to try not to have the Jobs depend directly on the container if we can help it.

core/celery/tasks/generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,292 @@
...
+                    holds_report_file_path
+                )
+
+                self.services.email.send_email(
+                    subject=f"Inventory and Holds Reports {current_time}",
+                    receivers=[self.email_address],
+                    text="",
+                    attachments=attachments,
+                )
+            except Exception as e:
```
jonathangreen, line 111: Instead of catching the exception here, and logging a failure. I think we should drop the `except` block and just let the exception raise. Celery will log the exception and mark the job as failed. That way when we start collecting information about tasks this would count towards the failed job stat.

tests/api/admin/controller/test_report.py
```diff
@@ -36,15 +39,3 @@ def test_generate_inventory_report(self, report_fixture: ReportControllerFixture
             assert response.status_code == HTTPStatus.ACCEPTED
             body = json.loads(response.data)  # type: ignore
             assert body and body["message"].__contains__("admin@email.com")
```
jonathangreen, line 41: Minor: Probably better to use the `in` operator here, rather then directly calling `__contains__`.

tests/api/admin/controller/test_report.py
```diff
@@ -36,15 +39,3 @@ def test_generate_inventory_report(self, report_fixture: ReportControllerFixture
             assert response.status_code == HTTPStatus.ACCEPTED
             body = json.loads(response.data)  # type: ignore
```
jonathangreen, line 40: Minor: you can directly get the json from a response by calling `response.json`.

tests/api/admin/controller/test_report.py
```diff
@@ -22,7 +23,9 @@ def report_fixture(
 
 
 class TestReportController:
-    def test_generate_inventory_report(self, report_fixture: ReportControllerFixture):
+    def test_generate_inventory_and_hold_reports(
+        self, report_fixture: ReportControllerFixture, celery_fixture: CeleryFixture
```
jonathangreen, line 27: Minor: This works, but in the other controller tests I've written so far, I tried to isolate the controller from the queue, by mocking out the task. This way we don't have the bring `celery_fixture` into the controller tests. 

Something like: https://github.com/ThePalaceProject/circulation/blob/8e976536b16a2930c8032ead7a8dde5d34a0e3a2/tests/api/admin/controller/test_collections.py#L711-L713

Mocking out celery in web controllers like this aligns with what the celery docs suggest for testing strategies: https://docs.celeryq.dev/en/stable/userguide/testing.html#tasks-and-unit-tests

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,279 @@
...
+        position=0,
+    )
+
+    shared_patrons_in_hold_queue = 4
+    licensepool.patrons_in_hold_queue = shared_patrons_in_hold_queue
+
+    assert library.id
+
+    job = GenerateInventoryAndHoldsReportsJob(
+        mock_session_maker, library.id, email_address=email
```
jonathangreen, line 182: If you took `delete_attachments` and `send_email` as args here, this could be re-written as:
```python
    send_email_mock = create_autospec(job.services.email.container.send_email)
    # for testing, don't delete the files associated with the attachments so we can read them after the script
    # runs
    job = GenerateInventoryAndHoldsReportsJob(
        mock_session_maker, library.id, email_address=email, send_email=send_email_mock, delete_attachments=False
    )
```

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+
+    send_email_mock = create_autospec(job.services.email.container.send_email)
+    # for testing, don't delete the files associated with the attachments so we can read them after the script
+    # runs
+    GenerateInventoryAndHoldsReportsJob._delete_attachments = False
+    job.services.email.container.send_email = send_email_mock
+    job.run()
+    send_email_mock.assert_called_once()
+    args, kwargs = send_email_mock.call_args
```
jonathangreen, line 192: Minor: `call_args` is a named tuple, so you can directly access `kwargs` without unpacking it like this: `send_email_mock.call_args.kwargs`

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,279 @@
...
+        assert row["publisher"] == publisher
+        assert row["audience"] == "young adult"
+        assert row["genres"] == "genre_a,genre_z"
+        assert row["format"] == edition.BOOK_MEDIUM
+        assert row["data_source"] == data_source
+        assert row["collection_name"] == collection_name
+        assert int(row["shared_active_hold_count"]) == shared_patrons_in_hold_queue
+        assert int(row["library_active_hold_count"]) == 3
+
+        assert row_count == 1
```
jonathangreen, line 250: Minor: This doesn't matter so much because there is only a single row here, but this line and the following lines should be out-dented so they are not being run in the `for` loop.

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,279 @@
...
+
+        # clean up files
+        for f in attachments.values():
+            os.remove(f)
+
+
+def create_test_opds_collection(collection_name, data_source, db, library):
```
jonathangreen, line 257: Minor: Would be nice to have some type hints on this function.

jonathangreen: Since this function is defined here, it might be nice to use it to create the collection for `test_job_run` as well.

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def test_generate_inventory_and_hold_reports_task(
+    db: DatabaseTransactionFixture, celery_fixture: CeleryFixture
+):
+    library = db.library(short_name="test_library")
+    # there must be at least one opds collection associated with the library for this to work
+    create_test_opds_collection("c1", "d1", db, library)
+    GenerateInventoryAndHoldsReportsJob._delete_attachments = True
+    services = container_instance()
```
jonathangreen, line 275: Instead of directly calling `container_instance` it would be better to use the `ServicesFixture` pytest fixture. 

Then you could override `send_email` like this:
```python
def test_generate_inventory_and_hold_reports_task(
    db: DatabaseTransactionFixture, celery_fixture: CeleryFixture, services_fixture: ServicesFixture
):
   ...
   send_email_mock = create_autospec(services_fixture.services.email.send_email)
   services_fixture.services.email.send_email.override(send_email_mock)
```

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,279 @@
...
+    attachments: dict = kwargs["attachments"]
+
+    assert len(attachments) == 2
+    inventory_report_key = [x for x in attachments.keys() if "inventory" in x][0]
+    assert inventory_report_key
+    assert "test_library" in inventory_report_key
+    inventory_report_value = attachments[inventory_report_key]
+    assert inventory_report_value
+    inventory_report_reader = csv.DictReader(open(inventory_report_value))
+    inventory_row_count = 0
```
jonathangreen, line 204: Instead of doing `inventory_row_count` like this, it would be nice to test it before going into the loop, so that we know if we get the wrong number of results before. You could do something like this I think:

```python
inventory_report_csv = list(csv.DictReader(open(inventory_report_value)))

assert len(inventory_report_csv) == 1

for row in inventory_report_csv:
   ...
```

jonathangreen (commented):
Generally looks good, added a few comments to address.

jonathangreen:
> One note: I am no longer checking if a similar task exists or is process before issuing the request for a report to be generated. @jonathangreen or @tdilauro: do you think this is a requirement for this feature? If so, I can do a little research into how best to do this. The reason it seems less important to me now is that the reports will be generated more or less immediately as opposed to possibly waiting several minutes before starting up (when we were using deferredtasks). So the chance that duplicate requests will be made is less likely.

I think this is fine. If we find we need de-duplication we can add it in the future.

dbernstein:
@jonathangreen : the updates addressing your feedback are ready to go.

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,288 @@
...
+def test_generate_inventory_and_hold_reports_task(
+    db: DatabaseTransactionFixture,
+    services_fixture: ServicesFixture,
+    celery_fixture: CeleryFixture,
+):
+    library = db.library(short_name="test_library")
+    # there must be at least one opds collection associated with the library for this to work
+    create_test_opds_collection("c1", "d1", db, library)
+    send_email_mock = create_autospec(
+        services_fixture.services.email.container.send_email
```
jonathangreen, line 284: Don't need `container` here as well.

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
```diff
@@ -0,0 +1,288 @@
...
+    )
+    services_fixture.services.email.container.send_email = send_email_mock
+    generate_inventory_and_hold_reports.delay(library.id, "test@email").wait()
+    send_email_mock.assert_called_once()
```
jonathangreen, line 288: You could rewrite the mocks here, so that they check on the mock email sender provided by the services fixture. That way we know that the correct `send_email` gets passed in from the container. That way we don't run into a situation like we have here where the mock provides something the real container doesn't have.

```python
def test_generate_inventory_and_hold_reports_task(
    services_fixture: ServicesFixture,
    celery_fixture: CeleryFixture,
):
    library = db.library(short_name="test_library")
    # there must be at least one opds collection associated with the library for this to work
    create_test_opds_collection("c1", "d1", db, library)
    generate_inventory_and_hold_reports.delay(library.id, "test@email").wait()
    services_fixture.email_fixture.mock_emailer.send.assert_called_once()
    assert "Inventory and Holds Reports" in services_fixture.email_fixture.mock_emailer.send.call_args.kwargs["subject"]
    assert services_fixture.email_fixture.mock_emailer.send.call_args.kwargs["receivers"] == ["test@email"]
```

jonathangreen (commented):
This is close, but the email service from the container needs to be updated or it won't actually work outside of the tests.

tests/core/celery/tasks/test_generate_inventory_and_hold_reports.py
jonathangreen: Nice to add some assertions as well about what gets passed into `send` to make sure the email address etc is making the full trip into the emailer.

jonathangreen (approved):
Looks good!

PR #1797, opened 2024-04-18
Load overdrive tests file from files fixture

jonathangreen:
## Description

Update Overdrive test to load test file from the files fixture.

## Motivation and Context

Loading the file directly as it was, prevents the tests from succeeding when running from a different path. This was causing me not to be able to run the tests in my IDE. Slight annoyance, but its nice to have fixed.

## How Has This Been Tested?

- Running tests

## Checklist


tdilauro (approved):
Looks good!

PR #1799, opened 2024-04-18
Update to xmlsec v1.3.14

jonathangreen:
## Description

This PR pulls in the latest `xmlsec`, and removes all the annoying stuff we had in the builds to get around the incompatibility of `xmlsec` and `lxml` and the dependencies we installed just to build `xmlsec`.

## Motivation and Context

The Python `xmlsec` module just released v1.3.14, which has a couple big changes:
- They upload wheels for linux and osx
- Built against the newest version of the xmlsec library

This should make installing on a mac much less painful as well. Although you may need to reinstall `lxml` after this PR goes in. Since you probably have a locally compiled version of `lxml` installed.

## How Has This Been Tested?

- Running in CI

## Checklist


dbernstein (approved):
🚀

tdilauro (approved):
Hooray! 🥳 🎉

PR #1800, opened 2024-04-19
Fix docker migration test CI workflow (PP-1171)

jonathangreen:
## Description

Update `docker/ci/test_migrations.sh`:
- Output more debug information, so this is easier to troubleshoot next time
- Format that output, so its easy to see whats going on in gh actions
- Don't try to build a container for the old version of the CM that we need to run
  - Instead of building a container, we just use an old container image that was already built
  - To make this easier a new docker-compose file `test_migrations.yml` was added for this purpose

## Motivation and Context

Previously we tried to build a container based on the old version of the codebase for these tests. That is an error prone endeavor, and failed in this case because of the change in dependencies in the container build introduced in https://github.com/ThePalaceProject/circulation/pull/1799.

In order to reduce the chances of this failing in the future, we just use an old previously build version of the container. This also simplifies the script, since it no longer has to move back and forth between old and new git commits in order to run the tests.

## How Has This Been Tested?

- Running script locally
- Running in CI

## Checklist

### DIFF
--- a/docker/ci/test_migrations.sh
+++ b/docker/ci/test_migrations.sh
@@ -81,14 +81,16 @@ success() {
 
 debug_echo() {
   if [[ $DEBUG_ECHO_ENABLED -eq 1 ]]; then
-    echo "$@"
+    printf "%q " "$@"
+    printf "\n"
   fi
 }
 
 # Run a docker-compose command
 compose_cmd() {
-  debug_echo "++ docker compose $*"
-  docker compose -f docker-compose.yml -f docker/ci/test_migrations.yml --progress quiet "$@"
+  args=(docker compose -f docker-compose.yml -f docker/ci/test_migrations.yml --progress quiet)
+  debug_echo "++" "${args[@]}" "$@"
+  "${args[@]}" "$@"
 }
 
 # Run a command in a particular container using docker-compose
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
@jonathangreen :  I just never stop admiring your skillz, dog. 🥇 You set a high bar.

tdilauro (approved):
Looks good! ⬆️ ⬇️ ⬆️ ▶️ 🥳

docker/ci/test_migrations.sh
```diff
@@ -3,139 +3,212 @@
...
-echo ""
+gh_endgroup
 
 # Migrate up to the current commit and check if the database is in sync
-echo "Testing upgrade migrations on branch ${current_branch}"
-git checkout -q "${current_branch}"
-run_migrations "upgrade head"
-check_db "A new migration is required or a up migration is broken."
+gh_group "Testing upgrade migrations"
+run_migrations "webapp" "upgrade head"
```
dbernstein, line 193: Should this be" webapp-old" here?

dbernstein (commented):
Not sure if the issue I pointed out is a problem.  Feel free to merge if it is not.

docker/ci/test_migrations.sh
jonathangreen: It can't be `webapp-old` because the old container doesn't have the migrations in it in order to move to the latest DB version.

dbernstein: okay - right of course.

PR #1802, opened 2024-04-22
Move code into palace.manager package (PP-1161)

jonathangreen:
## Description

Move all the CM files into a namespace package called `palace`. The cm code is in the `palace.manager` sub-module. 

I started moving some of the code from `core` into the main `palace.manager` package, but there is still more work to do there. We will probably want to continue moving things around after this PR, but this one is the most disruptive, since it moves all the code into the package, and moves around our database models.

After this PR all our database models live at: `palace.manager.sqlalchemy.model.*`. I broke the import cycles that caused the `lanes` and `saml` models to previously live elsewhere, so now all our DB models live in that module and importing `palace.manager.sqlalchemy.model` is enough to initialize them.

The separate `customlists` package has also been moved into the main `palace.manager` package.

The large number of removals on this PR are due to fixture files that were removed, since they were not being used anywhere. Since we use relative paths to lookup the fixtures, I had to go through all these files anyway.

I went through and made comments on the PR in the parts of the code that I changed. I attempted to primarily move things around and not change much logic. The only logic changes that were made as part of this PR had to do with loading files based on paths, since the paths changed, and some of them moved into the package.

## Motivation and Context

See PP-1161. Its helpful to be able to import our CM code, but previously this was impossible without messing with import paths since we didn't have a reliable package name.

Note: One major downside of this change is we no longer have the nice property of having two packages `api` and `core` to allow us to run tests in parallel. So all the tests have to run sequentially now, meaning CI runs take longer. I'm going to make a follow up ticket to see if there is anything we can do to make this a bit faster, since the test runs now take quite a lot longer.

## How Has This Been Tested?

- Tests in CI
- Testing locally
  - Note: I'm still doing some local testing to try to make sure everything is working okay

## Checklist


jonathangreen:
I'm still doing some testing locally to try to verify everything is still functional after the re-naming, but I think this is ready for a first round of review when folks have time.

src/palace/manager/sqlalchemy/session.py
```diff
@@ -0,0 +1,182 @@
...
+            select([literal_column("proname")])
+            .select_from(table("pg_proc"))
+            .where(literal_column("proname") == "fn_recursive_equivalents")
+        )
+        result = session.execute(query).all()
+
+        # If it doesn't, create it.
+        if not result:
+            resource_file = (
+                resources_dir("sqlalchemy") / cls.RECURSIVE_EQUIVALENTS_FUNCTION
```
jonathangreen, line 112: Only code changes here are to load this from the package, rather then manipulating file paths here.

src/palace/manager/sqlalchemy/util.py
```diff
@@ -0,0 +1,178 @@
...
+from sqlalchemy.engine import Connection
+from sqlalchemy.exc import IntegrityError, MultipleResultsFound, NoResultFound
+from sqlalchemy.orm import Session
+
+# This is the lock ID used to ensure that only one circulation manager
+# initializes or migrates the database at a time.
+LOCK_ID_DB_INIT = 1000000001
+
+
+@contextmanager
```
jonathangreen, line 19: broken out of `model/__init__.py`. This is where the bulk of the non-import content in that file was moved to.

src/palace/manager/util/resources.py
```diff
@@ -0,0 +1,7 @@
+from importlib.abc import Traversable
+from importlib.resources import files
+
+
+def resources_dir(subdir: str) -> Traversable:
```
jonathangreen, line 5: We use this new function to load any content needed from the package. It uses `importlib` to get the path to the package. This handles the various cases where our code could end up in different package locations.

tests/fixtures/files.py
```diff
@@ -1,76 +1,82 @@
...
 import pytest
 
+from palace.manager.sqlalchemy.model.resource import Representation
+from tests.fixtures.database import DatabaseTransactionFixture
 
-class ResourcesFilesFixture:
-    """A fixture providing access to resource files."""
 
-    """These are files that are served by the CM in production, and aren't test suite specific."""
+class FilesFixture:
```
jonathangreen, line 9: This fixture is updated for the new import paths, and a number of the other separate files fixtures have all been centralized in this file.

tests/manager/api/test_app.py
```diff
@@ -0,0 +1,20 @@
...
+):
+    # Use the db transaction fixture so that we don't use the production settings by mistake
+    with patch.object(
+        HTTP, "set_quick_failure_settings"
+    ) as mock_set_quick_failure_settings:
+        # Initialize the app, which will set the HTTP configuration
+        initialize_application()
+
+    # Make sure that the HTTP configuration was set
+    mock_set_quick_failure_settings.assert_called_once()
```
jonathangreen, line 20: This test was updated to use a mock. Previously the call it was making to `initialize_application` was changing the global state for HTTP module, which caused other tests to fail. Using a mock here does the same test, but prevents this side effect.

tests/mocks/flask.py
```diff
@@ -0,0 +1,19 @@
+from flask_pydantic_spec.flask_backend import Context
+from flask_pydantic_spec.utils import parse_multi_dict
+
+
+def add_request_context(request, model, form=None) -> None:
```
jonathangreen, line 5: This isn't new, it was just moved out of the tests package into mocks.

tests/mocks/sip.py
```diff
@@ -0,0 +1,45 @@
+from palace.manager.api.sip import SIPClient
+from palace.manager.api.sip.client import Constants
+
+
+class MockSIPClient(SIPClient):
```
jonathangreen, line 5: This isn't new, but it was in the main package, so it got moved out to `mocks` as part of this re-org.

src/palace/manager/customlists/explain.py
```diff
@@ -31,12 +31,6 @@ def _parse_arguments(args: list[str]) -> argparse.Namespace:
             default=0,
             help="Increase verbosity",
         )
-        parser.add_argument(
-            "--report-schema-file",
```
jonathangreen, line 35: Since these schema files are part of the package, it doesn't really make sense to pass them into the script anymore. Instead we load them from the package directly.

jonathangreen:
Codecov is reporting that coverage has decreased due to this PR. It looks like that is because we are now reporting coverage on all the files in the `palace.manager` package, which adds an additional 39 files that coverage is monitoring. I these additional files account for the decrease in coverage.

jonathangreen:
Pushed one small commit just now https://github.com/ThePalaceProject/circulation/pull/1802/commits/d30fc4ce01a0d9fc8514b8ec49751c444e8be6ac that resolves some issues I found when testing in the docker containers locally. These issues were causing the scripts not to run in the scripts container.

bin/run
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 #
-# Ensures that Library Simplified Circulation Manager scripts don't run
+# Ensures that Palace Manager scripts don't run
```
dbernstein, line 3: Is there any reason for us not to abandon this script and instead using something that already does this like [flock](https://man7.org/linux/man-pages/man1/flock.1.html)?

dbernstein (approved):
Screenshot 2024-04-26 at 10 38 18 AM

core/resources/lcc_one_level.json~
tdilauro, line 1: This looks like a temporary editing file. Should it be here?

core/resources/random_isbns.txt~
tdilauro, line 1: This looks like a temporary editing file. Should it be here?

bin/run
tdilauro: Probably not worth changing it at this point, since it's used to run scripts via cron and those are going to be moved to the queue-based approach.

tdilauro: Probably not worth changing it at this point, since it's used to run scripts via cron and those are going to be moved to the queue-based approach.

core/resources/lcc_one_level.json~
jonathangreen: No it shouldn't. That is why the PR is deleting it 😄

Looks like its been in the repo for a looong time.

core/resources/random_isbns.txt~
jonathangreen: No it shouldn't. That is why the PR is deleting it 😄

Looks like its been in the repo for a looong time.

bin/run
jonathangreen: I think it makes sense to keep whats here like @tdilauro mentioned, since this is going away soon. Nice to know flock exists though, I hadn't heard of it before, so I appreciate the tip.

PR #1809, opened 2024-04-29
Override Celery automatic task naming

jonathangreen:
## Description

Override Celery automatic task naming, so that all of our tasks are not prefixed with: `palace.manager.celery.tasks`. 

See Celery docs: https://docs.celeryq.dev/en/stable/userguide/tasks.html#changing-the-automatic-naming-behavior

I'd like to make sure this one gets merged before we cut a release containing: https://github.com/ThePalaceProject/circulation/pull/1802.

## Motivation and Context

Make it easier to reference our tasks on the command line, in config and move them around in the package in the future if we want to.

## How Has This Been Tested?

- Tested locally
- Tested in CI

## Checklist

### DIFF
--- a/src/palace/manager/celery/celery.py
+++ b/src/palace/manager/celery/celery.py
@@ -1,0 +1,14 @@
+import celery
+
+
+class Celery(celery.Celery):
+    def gen_task_name(self, name: str, module: str):
+        """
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 🎸

PR #1814, opened 2024-05-01
Remove threaded collection coverage provider (PP-1212)

jonathangreen:
## Description

Removes the unused script `RunThreadedCollectionCoverageProviderScript` and some unused functions in `utils` that were previously only used by `RunThreadedCollectionCoverageProviderScript`.

## Motivation and Context

While working on PP-1212 I was troubleshooting some tests failures in this class. When I went to see where the class was actually used, it turned out to only be used in tests. I'd say its been dead for some time, so it makes sense to 🔥 it instead of fixing the test failures.

## How Has This Been Tested?

- Running tests in CI

## Checklist


dbernstein (approved):
Begone cruft!

PR #1817, opened 2024-05-01
Run tests in parallel with pytest-xdist (PP-1212)

jonathangreen:
## Description

Uses `pytest-xdist` to bring down the time it takes to run our test suite. It takes the time in CI from 15 minutes to 7 minutes or so. Locally for me it brings a run down to about 3 minutes, which is much easier to deal with then the 13 minutes that it used to take.

Because this involves running the tests in parallel, some of the test fixtures needed to be updated to allow multiple workers to be calling them at the same time. The biggest change here is that the tests used to run in the database name that is passed into the tests, but now the tests create their own database, then clean it up afterwords, so that each worker has its own database.

## Motivation and Context

Make it less painful to run our test suite and debug problems.

## How Has This Been Tested?

- Running locally
- Running in CI

## Checklist


.github/workflows/test.yml
```diff
@@ -52,49 +52,7 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage.xml
-          flags: manager
-          name: manager-${{ matrix.python-version }}
-          verbose: true
-
-  test-migrations:
```
jonathangreen, line 59: Removed because migrations are not run as part of the main test suite.

pyproject.toml
```diff
@@ -9,11 +9,14 @@ exclude_also = [
...
 ]
+include_namespace_packages = true
 
 [tool.coverage.run]
 branch = true
+concurrency = ["multiprocessing", "thread"]
+parallel = true
 relative_files = true
-source = ["palace.manager"]
+source = ["src"]
```
jonathangreen, line 19: This needed to be updated to get coverage and xdist working together.

src/palace/manager/core/config.py
```diff
@@ -72,46 +71,26 @@ class Configuration(ConfigurationConstants):
...
-        # controls which database is used, and it's set by the
-        # package_setup() function called in every component's
-        # tests/__init__.py.
-        test = os.environ.get("TESTING", False)
-        if test:
-            environment_variable = cls.DATABASE_TEST_ENVIRONMENT_VARIABLE
-        else:
-            environment_variable = cls.DATABASE_PRODUCTION_ENVIRONMENT_VARIABLE
-
-        url = os.environ.get(environment_variable)
```
jonathangreen, line 95: Setting up the DB for testing is entirely handled by the pytest mocks now and doesn't depend on the `TESTING` env var.

src/palace/manager/search/service.py
```diff
@@ -110,10 +110,6 @@ def read_pointer_set_empty(self) -> None:
     def index_create(self, revision: SearchSchemaRevision) -> None:
         """Atomically create an index for the given base name and revision."""
 
-    @abstractmethod
-    def indexes_created(self) -> list[str]:
-        """A log of all the indexes that have been created by this client service."""
-
```
jonathangreen, line 116: This method was only ever used in tests to delete created indexes. This is now handled by the search fixture, so this method can come out.

src/palace/manager/sqlalchemy/session.py
```diff
@@ -62,18 +63,8 @@ def setup_event_listener(
...
-            url = Configuration.database_url()
-        if url:
-            bind_obj = cls.engine(url)
-        elif session:
-            bind_obj = session.get_bind()
-            if not os.environ.get("TESTING"):
-                # If a factory is being created from a session in test mode,
-                # use the same Connection for all of the tests so objects can
-                # be accessed. Otherwise, bind against an Engine object.
-                bind_obj = bind_obj.engine
```
jonathangreen, line 76: This was moved to pytest fixtures as well, so we no longer have to rely on TESTING env var.

tests/fixtures/api_controller.py
```diff
@@ -83,7 +83,6 @@ def __init__(
         self,
         db: DatabaseTransactionFixture,
         services_fixture: ServicesFixture,
-        setup_cm: bool,
```
jonathangreen, line 86: The `setup_cm` arg was only used in the scopedsession test, and its not needed there anymore due to the fixture changes. So we remove this options to simply the fixture a bit.

tests/fixtures/search.py
```diff
@@ -38,41 +38,31 @@ class ExternalSearchFixture(LoggerMixin):
...
 
         # Set up our testing search instance in the services container
         self.search_container = Search()
-        self.search_container.config.from_dict(self.search_config.dict())
+        search_config_dict = self.search_config.dict()
+        search_config_dict["index_prefix"] = self.index_prefix
+        self.search_container.config.from_dict(search_config_dict)
         self.services_container.search.override(self.search_container)
 
-        self._indexes_created: list[str] = []
```
jonathangreen, line 50: Previously this fixture tracked all the indexes created and then deleted them. Now it gets a unique index prefix per fixture invocation, and all indexes are created with this prefix. When the fixture is cleaned up, all the indexes with that prefix are deleted. 

This allows the fixture to run in parallel without collisions between names, and makes the cleanup logic easier and less likely to leak an index.

tests/fixtures/webserver.py
```diff
@@ -180,7 +180,7 @@ def requests(self) -> list[MockAPIServerRequest]:
 @pytest.fixture
 def mock_web_server() -> Generator[MockAPIServer, None, None]:
     """A test fixture that yields a usable mock web server for the lifetime of the test."""
-    _server = MockAPIServer("127.0.0.1", 10256)
+    _server = MockAPIServer("127.0.0.1", 0)
```
jonathangreen, line 183: This binds to an OS provided available random port. This was done so that this fixture can run in parallel, with multiple tests having this fixture running at once.

tests/manager/api/controller/test_scopedsession.py
```diff
@@ -1,68 +1,62 @@
...
-    to a database transaction fixture in all of the various methods that will be called.
-    """
+from tests.fixtures.database import DatabaseFixture
+from tests.fixtures.services import ServicesFixture
+from tests.mocks.circulation import MockCirculationManager
+
+
+class ScopedSessionFixture:
+    def __init__(
+        self, db_fixture: DatabaseFixture, services: ServicesFixture, session: Session
```
jonathangreen, line 21: This tests was largely re-written. It had to jump through a lot of hoops before, since the assumption was that it ran inside of a transaction that could be rolled back at the end. That made the functionality hard to test. 

Now that we have a fixture `function_database` that provides a temporary database for the test function, we can re-write this to do the test more directly. This fixture is needed for the migration tests anyway, as they can't be written to work in a transaction, so it makes sense to use it here as well.

tests/manager/api/test_app.py
```diff
@@ -10,9 +10,14 @@ def test_initialize_application_http(
...
-        HTTP, "set_quick_failure_settings"
-    ) as mock_set_quick_failure_settings:
+    with (
+        patch.object(
+            HTTP, "set_quick_failure_settings"
+        ) as mock_set_quick_failure_settings,
+        patch("palace.manager.api.app.initialize_database"),
+        patch("palace.manager.api.app.initialize_circulation_manager"),
+        patch("palace.manager.api.app.app"),
+    ):
```
jonathangreen, line 20: We need some mocks here, otherwise this test trips over the patch we apply that raises an exception if a test tries to create an engine directly. The test wasn't very safe before because it was doing a full app initialization, but leaking the database connection.

tests/manager/core/test_config.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+from sqlalchemy.exc import ArgumentError
+
+from palace.manager.core.config import CannotLoadConfiguration, Configuration
+from palace.manager.service.logging.configuration import LogLevel
+
+
+class TestConfiguration:
+    def test_database_url(
```
jonathangreen, line 9: Since `database_url` is mocked out now for most of our tests, add a small test for it, so we don't lose coverage of it.

tests/manager/search/test_external_search.py
```diff
@@ -2040,34 +2033,45 @@ def assert_featured(description, worklist, facets, expect):
...
         #
-        # The order of the works when using random depends on 4 things:
-        # - The seed
-        # - The id (work_id)
-        # - The index name
-        # - The shard id
-        # If any of those change the order of works in this result may change,
-        # and hence the order of works in this assert must also change
-        # E.g. If the index version changes from v5 to v6, this may affect the order of works queried
-        # Keeping everything else the same, the order of works will remain reproducible across test runs
```
jonathangreen, line 2057: Previously this test would fail every time we changed the index name, since the index name is part of the seed for the random function. The changes in this PR make the index name effectively random, since each call to the fixture gets its own index name. So this test was re-written to just test what we need to test here, which is that the random function re-orders the high quality and low quality titles, but doesn't cause them to mix.

tests/manager/sqlalchemy/test_session.py
```diff
@@ -0,0 +1,77 @@
...
+        timestamp = get_one(
+            db.session,
+            Timestamp,
+            collection=None,
+            service=Configuration.SITE_CONFIGURATION_CHANGED,
+        )
+        assert timestamp is not None
+        old_timestamp = timestamp.finish
+        SessionManager.initialize_data(db.session)
+        assert old_timestamp == timestamp.finish
```
jonathangreen, line 31: This first test is just moved from tests/manager/sqlalchemy/test_util.py since this is a more appropriate location for it.

tests/manager/sqlalchemy/test_session.py
```diff
@@ -0,0 +1,77 @@
...
+@patch.object(SessionManager, "session")
+@patch.object(Configuration, "database_url")
+def test_production_session(mock_database_url, mock_session):
+    # Make sure production_session() calls session() with the URL from the
+    # configuration.
+    mock_database_url.return_value = "test-url"
+    session = production_session()
+    mock_database_url.assert_called_once()
+    mock_session.assert_called_once_with("test-url", initialize_data=True)
+    assert session == mock_session.return_value
```
jonathangreen, line 77: Since the DB fixture mocks these for most tests now, add some small tests for these functions to make sure they are doing what we expect and to maintain coverage of them.

tests/fixtures/services.py
```diff
@@ -201,7 +201,7 @@ def wired(self) -> Generator[None, None, None]:
         self.services.unwire()
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture
```
jonathangreen, line 204: This fixture was set to `autouse` before, but that isn't necessary. It just needs to be added to the alembic fixture. I'd like to avoid `autouse` fixtures if we can, so you always know the fixtures that are active when calling a function.

tests/migration/conftest.py
```diff
@@ -21,60 +21,41 @@
...
-        config_options={
-            "file": str(Path(__file__).parent.parent.parent.absolute() / "alembic.ini")
-        }
-    )
+    return Config(config_options={"file": str(alembic_config_path)})
 
 
 @pytest.fixture
-def alembic_engine(database: DatabaseFixture) -> Engine:
+def alembic_engine(function_database: DatabaseFixture) -> Engine:
```
jonathangreen, line 39: Now that we have a fixture `function_database` that can provide a per test function database in the main test suite, we can use it here to run the alembic tests in an isolated environment as part of the main test suite.

tests/migration/conftest.py
```diff
@@ -21,60 +21,41 @@
...
     """
-    return database._engine
+    return function_database.engine
 
 
 @pytest.fixture
 def alembic_runner(
     alembic_config: dict[str, Any] | alembic.config.Config | Config,
     alembic_engine: Engine,
+    services_fixture: ServicesFixture,
```
jonathangreen, line 50: services_fixture was added here, so that it no longer needs to be an autouse fixture.

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ x-cm-variables: &cm
     PALACE_CELERY_CLOUDWATCH_STATISTICS_DRYRUN: "true"
 
     # Set up the environment variables used for testing as well
-    SIMPLIFIED_TEST_DATABASE: "postgresql://palace:test@pg:5432/circ"
+    PALACE_TEST_DATABASE_URL: "postgresql://palace:test@pg:5432/circ"
```
tdilauro, line 25: Minor: Do we want to omit the database name part from this URL, since we're running tests in parallel?

tdilauro (approved):
Looks great and is a huge time saver! 🚀 

A couple of minor comments, one of which is not even related to this PR, per se.

src/palace/manager/core/config.py
```diff
@@ -72,46 +71,26 @@ class Configuration(ConfigurationConstants):
...
 
-        url_obj = None
         try:
             url_obj = make_url(url)
         except ArgumentError as e:
-            # Improve the error message by giving a guide as to what's
-            # likely to work.
+            # Improve the error message by giving a guide as to what's likely to work.
             raise ArgumentError(
                 "Bad format for database URL (%s). Expected something like postgresql://[username]:[password]@[hostname]:[port]/[database name]"
```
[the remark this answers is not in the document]
jonathangreen: I can update that as part of this PR. Does just dropping the `(%s)` part from the string make sense?

docker-compose.yml
jonathangreen: This does work generally, but we can't do it in this case. We connect to whatever DB is passed into the env var, to create the worker DBs. If no DB is passed in, PG defaults to connecting to a DB with the same name as the users. But in this case when we create the container, we tell PG to name the users DB something different, `circ` instead. So if we don't pass in `circ` here, the calls will fail with an error something like `Unable to connect DB palace does not exist.`

PR #1819, opened 2024-05-02
PP-1219: Fixes overdrive monitor failures.

dbernstein:
## Description
This update fixes a bug introduced in https://github.com/ThePalaceProject/circulation/pull/1783.  It appears that the way I have overridden the NewTitlesOverdriveMonitor construct is resulting in a failure in the dependency injection code when running in production.  Additionally there was a string interpolation bug in an area of the code that was not previously covered.

When this PR has been approved I'll deploy a hotfix as well as create a PR against main with these changes and I'll create a 20.0.1 release before deploying out to staging.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1219

## Checklist

### DIFF
--- a/tests/api/test_overdrive.py
+++ b/tests/api/test_overdrive.py
@@ -2233,6 +2233,15 @@ class TestOverdriveAPI:
         assert fulfill.content_link_redirect is True
         assert fulfill.content_link == "https://example.org/epub-redirect"
 
+    def test_bug_fix_pp_1219(self, overdrive_api_fixture: OverdriveAPIFixture):
+        with patch.object(
+            overdrive_api_fixture.api, "_get_book_list_page"
+        ) as get_book_list:
+            get_book_list.return_value = ([], None)
+            result = overdrive_api_fixture.api.recently_changed_ids(utc_now(), None)
+            # this should no longer fail.
+            assert [i for i in result] == []
+
 
 class TestOverdriveAPICredentials:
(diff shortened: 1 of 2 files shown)


tests/api/test_overdrive.py
```diff
@@ -2233,6 +2233,15 @@ def test_no_drm_fulfillment(self, overdrive_api_fixture: OverdriveAPIFixture):
         assert fulfill.content_link_redirect is True
         assert fulfill.content_link == "https://example.org/epub-redirect"
 
+    def test_bug_fix_pp_1219(self, overdrive_api_fixture: OverdriveAPIFixture):
```
tdilauro, line 2236: Could we give this test method a more descriptive name? I'm guessing something like `test_no_recently_changed_books`.

tests/api/test_overdrive.py
```diff
@@ -2233,6 +2233,15 @@ def test_no_drm_fulfillment(self, overdrive_api_fixture: OverdriveAPIFixture):
...
+        with patch.object(
+            overdrive_api_fixture.api, "_get_book_list_page"
+        ) as get_book_list:
+            get_book_list.return_value = ([], None)
+            result = overdrive_api_fixture.api.recently_changed_ids(utc_now(), None)
+            # this should no longer fail.
```
tdilauro, line 2242: Leftover comment?

tdilauro (approved):
Looks good! Just a couple minor comments in the test.

PR #1821, opened 2024-05-03
Move docker migration test into the docker workflow

jonathangreen:
## Description

Move the docker migration test into the docker workflow.

## Motivation and Context

When I split up the workflows, I left the docker migration test as part of the main test workflow, but it should really exist as part of the docker build. That way it can benefit from the caching done there, and make sure it uses the correct base image. This change would have let us find the issue in https://github.com/ThePalaceProject/circulation/pull/1800 before it got merged to main.

## How Has This Been Tested?

Running tests in CI.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -205,10 +205,36 @@ jobs:
         if: always()
         run: docker compose down
 
+  migration-test:
+    name: Migration test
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
👍🏽

PR #1822, opened 2024-05-03
Rename fixture classes that were causing warnings

jonathangreen:
## Description

Rename two fixture classes to avoid warnings. And make pytest treat `PytestWarning` as an error, so that we notice these sooner in the future. Since its easy to name fixture classes incorrectly by accident.

## Motivation and Context

Two fixture classes `TestHttpFixture` and `TestLicenseFixture` were causing Pytest warnings because they were being loaded as tests because the class name started with `Test`. 

## How Has This Been Tested?

Running tests in CI

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -314,6 +314,9 @@ addopts = [
     "--numprocesses=auto",
     "--strict-markers",
 ]
+filterwarnings = [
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 👍🏽

PR #1823, opened 2024-05-03
Fix potentially flaky saml test

jonathangreen:
## Description

Update the SAML tests not to depend on a particular domain being configured in the app for their test run. This eliminates warnings from flask and should make the tests safer.

## Motivation and Context

The SAML tests are updating the global flask apps configuration to set `SERVER_NAME`, which could impact other tests, since the app is shared across tests. This can cause some tests to fail, especially when running in parallel. I think this caused one of the failures @tdilauro saw yesterday.

## How Has This Been Tested?

Running tests in CI.

## Checklist


tdilauro (approved):
Looks good! 🎸🤘🏽

PR #1828, opened 2024-05-03
Set application name in PG connection string

jonathangreen:
## Description

Set the `application_name` parameter in the PG connection string.

## Motivation and Context

AWS RDS metrics break stats out by the `application_name` parameter. We don't currently fill anything in there, but being able to know where the queries are coming from. 

## How Has This Been Tested?

Running tests in CI.

## Checklist


tdilauro (approved):
Looks good! 📓 🚀

PR #1836, opened 2024-05-10
Search index rebuild script undefined variable

jonathangreen:
## Description

Small fix to search index rebuild script. It raising a undefined variable exception because `self.services` was being accessed before the parent `__init__` function had been called to set it up.

## Motivation and Context

Script was failing when trying to rebuild index.

## How Has This Been Tested?

Tested locally.

## Checklist


tdilauro (approved):
🐛🦶🏽

PR #1837, opened 2024-05-10
Add support for stack_info parameter in JSONFormatter

jonathangreen:
## Description

As of [Python 3.3 logging calls take a `stack_info` parameter](https://docs.python.org/3.10/library/logging.html#logging.Logger.debug), which adds a stack trace to the log message. This PR adds support for this parameter to the `JSONFormatter`, so the stack trace is included in a `stack` key of the log message, when it is available.

## Motivation and Context

While debugging I needed to include a stack trace in my log messages, and before this change `JSONFormatter` did not include this information when a log record requested it.

## How Has This Been Tested?

- Tested locally
- Added unit test

## Checklist


tdilauro (approved):
Looks good!

PR #1838, opened 2024-05-10
Update classes to use LoggerMixin

jonathangreen:
## Description

Update a few classes to use `LoggerMixin`, so we get logger names that are formatted consistently.

## Motivation and Context

While working on PP-1225, I needed to enable and disable some loggers at debug log level, and these inconsistently named logging classes were causing me some issues.

## How Has This Been Tested?

- Running locally

## Checklist


tdilauro (approved):
Looks good! 🪵

PR #1839, opened 2024-05-10
Add logger utility functions

jonathangreen:
## Description

Add a new logger utility function: `logger_for_function` and move `logger_for_cls` from a static method on `LoggerMixin` out to its own utility function.

## Motivation and Context

Similar to https://github.com/ThePalaceProject/circulation/pull/1838, I needed a couple of the lane creation functions to have a consistent logger name. It seemed helpful to add a new `logger_for_function` call so we can get a logger for this purpose.

## How Has This Been Tested?

- Tested running locally

## Checklist

### DIFF
--- a/src/palace/manager/api/lanes.py
+++ b/src/palace/manager/api/lanes.py
@@ -1,5 +1,3 @@
-import logging
-
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
🪵

PR #1840, opened 2024-05-12
Adds API to get library-specific info on inventory reports. (PP-1210)

tdilauro:
## Description

- Adds the `/admin/reports/generate_inventory_report/:library_short_name` GET method endpoint to provide information about the collections that will be included if an inventory report is generated.
- Includes API documentation at:
  - Swagger: `.../apidoc/swagger#/admin.inventory`
  - ReDoc: `.../apidoc/redoc#tag/admin.inventory`
- Small refactoring in `...api_routes.MockControllerMethod` class to make it a little easier to verify calls to controller methods on routes for which content is validated on the way out.

## Motivation and Context

It is useful to know which collections will be included in the inventory reports. This is primarily for users of the Palace Manage admin user interface.

[Jira [PP-1210](https://ebce-lyrasis.atlassian.net/browse/PP-1210)]

## How Has This Been Tested?

- Adds new tests for the new feature.
- Tested locally via `curl` and with an updated Admin UI client.
- Added new tests to the test suite.
- [CI tests for the associated feature branch pass](https://github.com/ThePalaceProject/circulation/actions/runs/9052055225).

## Checklist

[PP-1210]: https://ebce-lyrasis.atlassian.net/browse/PP-1210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/api/admin/controller/report.py
+++ b/src/palace/manager/api/admin/controller/report.py
@@ -3,7 +3,6 @@ from http import HTTPStatus
 
 import flask
 from flask import Response
-from sqlalchemy import not_, select
 from sqlalchemy.orm import Session
 
 from palace.manager.api.admin.model.inventory_report import (
@@ -13,16 +12,11 @@ from palace.manager.api.admin.model.inventory_report import (
 from palace.manager.api.admin.problem_details import ADMIN_NOT_AUTHORIZED
 from palace.manager.celery.tasks.generate_inventory_and_hold_reports import (
     generate_inventory_and_hold_reports,
+    library_report_integrations,
 )
 from palace.manager.core.problem_details import INTERNAL_SERVER_ERROR
-from palace.manager.integration.goals import Goals
 from palace.manager.sqlalchemy.constants import MediaTypes
 from palace.manager.sqlalchemy.model.admin import Admin
-from palace.manager.sqlalchemy.model.collection import Collection
-from palace.manager.sqlalchemy.model.integration import (
-    IntegrationConfiguration,
-    IntegrationLibraryConfiguration,
-)
 from palace.manager.sqlalchemy.model.library import Library
 from palace.manager.util.log import LoggerMixin
 from palace.manager.util.problem_detail import ProblemDetail, ProblemDetailException
@@ -49,20 +43,12 @@ class ReportController(LoggerMixin):
         if not admin.is_librarian(library):
             raise ProblemDetailException(ADMIN_NOT_AUTHORIZED)
 
-        collections = self._db.scalars(
-            select(Collection)
-            .join(IntegrationConfiguration)
-            .join(IntegrationLibraryConfiguration)
-            .where(
-                IntegrationLibraryConfiguration.library_id == library.id,
-                IntegrationConfiguration.goal == Goals.LICENSE_GOAL,
-                not_(
-                    IntegrationConfiguration.settings_dict.contains(
-                        {"include_in_inventory_report": False}
-                    )
-                ),
+        collections = [
+            integration.collection
+            for integration in library_report_integrations(
+                library=library, session=self._db
             )
-        ).all()
+        ]
         info = InventoryReportInfo(
             collections=[
                 InventoryReportCollectionInfo(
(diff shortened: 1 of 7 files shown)


src/palace/manager/api/admin/controller/report.py
```diff
@@ -3,23 +3,80 @@
...
+        collections = self._db.scalars(
+            select(Collection)
+            .join(IntegrationConfiguration)
+            .join(IntegrationLibraryConfiguration)
+            .where(
+                IntegrationLibraryConfiguration.library_id == library.id,
+                IntegrationConfiguration.goal == Goals.LICENSE_GOAL,
+                not_(
+                    IntegrationConfiguration.settings_dict.contains(
+                        {"include_in_inventory_report": False}
```
tdilauro, line 61: I got most of this `select` from the code for generating the reports, but I'm not sure where the "include_in_inventory_report" setting is coming from. I do see it [here in OPDS collection settings](https://github.com/ThePalaceProject/circulation/blob/b1ab61ac9a9506d4ccad11efaeb4df5e3998a84b/src/palace/manager/core/opds_import.py#L131-L141), but I don't see it getting set to false on other collection types that probably shouldn't be included in inventory reports. I looked around a bit, but didn't see it set to either `true` or `false`.  Here's the query I used:

```sql
SELECT *
FROM integration_configurations ic
WHERE ic.goal = 'LICENSE_GOAL'
  AND (
    ic.settings @> '{"include_in_inventory_report": true}' OR  
    ic.settings @> '{"include_in_inventory_report": false}'
  )
```

jonathangreen: So getting this info in a pure SQL query is hard, mostly due to our use of inheritance to set the `include_in_inventory_report` setting. 

Some background on how this is set: 

You are correct this setting is getting set in the OPDS collection settings:
https://github.com/ThePalaceProject/circulation/blob/b1ab61ac9a9506d4ccad11efaeb4df5e3998a84b/src/palace/manager/core/opds_import.py#L131-L141

We don't store the default values for our serialized settings in the database. This is done in the `dict()` method for the Pydantic settings class:
https://github.com/ThePalaceProject/circulation/blob/530f25be8e64f4ef6a62294bcf2b25be8edd225b/src/palace/manager/integration/settings.py#L337-L341

I thought this was better called out in the comments, but I can't find it right now, so we should probably add a better comment adding these details. But we're not storing the defaults when we serialize to the database, so it is easier to update the defaults in the future.

When we are interacting with the settings via the Pydantic class (how I was envisioning we would mostly interact with them, rather then querying, this is transparent, since the defaults get re-filled when the values are loaded from the database).

So the only value that ever gets saved out to the DB is `include_in_inventory_report == False`. So that is why the code that finds the integrations when generating the reports is structured the way it is. 

First it finds any license integrations where `include_in_inventory_report == False` is not in the settings via a query. 
https://github.com/ThePalaceProject/circulation/blob/530f25be8e64f4ef6a62294bcf2b25be8edd225b/src/palace/manager/celery/tasks/generate_inventory_and_hold_reports.py#L66-L79

Then it filters down this list, to only those integrations that actually contain this setting:
https://github.com/ThePalaceProject/circulation/blob/530f25be8e64f4ef6a62294bcf2b25be8edd225b/src/palace/manager/celery/tasks/generate_inventory_and_hold_reports.py#L80-L86

jonathangreen: Since this code is missing that last filtering step, it will return collections without the setting, which I don't think is what we want here. 

Perhaps since this query is so complicated, it should get broken out to a utility function that both the report generation and the controller can call.

jonathangreen (commented):
I think the query in `inventory_report_info` is going to return unrelated collections, so some additional filtering is needed. I added some explanation in a comment. After that change this PR looks good to me.

tdilauro:
@jonathangreen I think this is ready for another look, when you have a chance.

PR #1841, opened 2024-05-13
Add search classes to strict type checking

jonathangreen:
## Description

Our search document classes were already typed, they were just missing a few hints necessary to add them to the list of classes that opt into strict type checking. 

## Motivation and Context

Since I'm working on these classes in PP-1225 I added the missing hints here.

## How Has This Been Tested?

- Running tests

## Checklist


src/palace/manager/search/service.py
```diff
@@ -336,20 +340,20 @@ def write_pointer_set(self, revision: SearchSchemaRevision) -> None:
 
     def read_pointer(self) -> str | None:
         try:
-            result: dict = self._client.indices.get_alias(name=self.read_pointer_name())
+            result = self._client.indices.get_alias(name=self.read_pointer_name())
             for name in result.keys():
                 if name.startswith(f"{self.base_revision_name}-"):
-                    return name
+                    return name  # type: ignore[no-any-return]
```
tdilauro, line 346: Hmmm. It looks like `name` should be a `str` here.

tdilauro (approved):
Looks good! 👍🏽 

One small comment.

src/palace/manager/search/service.py
jonathangreen: From what I could see client is untyped so `self._client.indices.get_alias` ends up giving us an `Any` type, which means `name` ends up being `Any`. I type hinted the returned dict, so we get a better type then `Any`.

PR #1847, opened 2024-05-14
Break scripts.py and core/scripts.py out

jonathangreen:
## Description

Both `scripts.py` and `core/scripts.py` are huge, so this PR breaks them out into smaller files. This makes no changes other then moving the classes around.

## Motivation and Context

With the Celery work, a lot of our scripts are going to end up being updated or removed. It seems like a good time break up the files, before these changes go in.

## How Has This Been Tested?

- Running tests

## Checklist


jonathangreen:
This is failing on patch coverage, but its really just highlighting our current lack of coverage for some of our scripts.

tdilauro (approved):
Looks good to me, based on which files were deleted and the fact that tests are passing. 🤞🏽🚀

PR #1849, opened 2024-05-14
Update search indexing to use celery rather then coverage records (PP-1225)

jonathangreen:
## Description

This PR updates our search indexing to happen via a celery queue, vs the existing work coverage record approach. The existing search index work coverage records are left in the DB and will be removed in a follow up PR.

All responsibility for the search index initialization and migration has been moved into `InstanceInitializationScript`. This script now handles both the DB and search index initialization. When a migration needs to happen to a new search index this now happens via a celery task, previously this was handled in the search coverage monitor.

There was some refactoring of responsibilities between the external search classes and the search client. The search client is now responsible for making sure the correct search alias is used when requests are being made.

## Motivation and Context

On our larger CMs, a large amount of the DB load is due to the work coverage records being generated for search. This is mostly due to deadlocks happening between these records when two processes try to update the same work. Moving search indexing to a queue like this should drastically reduce the DB load, and improve our performance.

## How Has This Been Tested?

- Tested locally
- Running unit tests

## Checklist


jonathangreen:
I'm still doing some final local testing on this one, but I think this is ready for review whenever anyone has the cycles to take a look.

tests/manager/search/test_external_search.py
```diff
@@ -131,6 +122,85 @@ def query_works_multi(self, queries, debug=False):
...
         assert pagination.size == default.size
 
+    def test_remove_work(
+        self,
+        end_to_end_search_fixture: EndToEndSearchFixture,
+        db: DatabaseTransactionFixture,
+        caplog: pytest.LogCaptureFixture,
+    ):
+        moby_duck = db.work(title="Moby Duck", with_open_access_download=True)
+        moby_dick = db.work(title="Moby Dick", with_open_access_download=True)
```
tdilauro, line 132: I had to look at this a couple of times because I didn't catch the difference. Moby Duck. 🦆

tests/mocks/search.py
```diff
@@ -162,46 +148,43 @@ def index_submit_documents(
...
-                items.remove(item)
-
-    def is_pointer_empty(*args):
-        return False
+        pointer = self.write_pointer_name()
+        items = self._documents_by_index[pointer]
+        to_remove = []
+        for item in items:
+            if item.get("_id") == doc_id:
+                to_remove.append(item)
```
tdilauro, line 185: I see this came in from older code, but could do this with a list comprehension:
```python
 to_remove = [item for item in items if item.get("_id") == doc_id]
```

tests/manager/sqlalchemy/model/test_work.py
```diff
@@ -1411,6 +1407,171 @@ def test_to_search_doc_datetime_cases(self, db: DatabaseTransactionFixture):
...
+                assert len(doc["customlists"]) == len(work.custom_list_entries)
+                for idx, custom_list in enumerate(work.custom_list_entries):
+                    actual = doc["customlists"][idx]
+                    assert actual["list_id"] == custom_list.list_id
+                    assert actual["featured"] == custom_list.featured
+            else:
+                assert doc["customlists"] is None
+
+            if work.presentation_edition.contributions:
+                assert len(doc["contributors"]) is len(
```
tdilauro, line 1508: '==' instead `is` here?

tdilauro (approved):
This looks great! 🎸🤘🏽

Just a few small comments. I'm gonna go ahead and approve.

tdilauro:
@jonathangreen Also, I forgot to mention that your pre-review comments were very handy. Thanks!

src/palace/manager/util/backoff.py
```diff
@@ -0,0 +1,19 @@
...
+
+def exponential_backoff(retries: int) -> int:
+    """
+    Exponential backoff time, based on number of retries.
+
+    The backoff includes some random jitter to prevent thundering herd, if
+    many items are retrying at the same time.
+
+    :param retries: The number of retries that have already been attempted.
+    :return: The number of seconds to wait before the next retry.
```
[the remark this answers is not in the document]
jonathangreen: In the instances we use this function right now, this is limited by the max retries of the task, but it does make sense to have a parameter like this. I'll add it in.

tests/manager/sqlalchemy/model/test_work.py
jonathangreen: This whole test was just moved in from the coverage providers tests. Not sure why it was using `is`. I updated it.

tests/manager/search/test_external_search.py
jonathangreen: I think I pulled this from an older test. It is hard to see the diff though. I renamed things a bit to make it easier to see whats happening at a glance.

PR #1851, opened 2024-05-17
Add some better logging when we get unexpected responses from ODL (PP-1283)

jonathangreen:
## Description

Gives some more information in our logs when we get unexpected status back from ODL feeds, so that its easier to pass the information on to DeMarque.

## Motivation and Context

This information would have helped while troubleshooting PP-1283. 

## How Has This Been Tested?

- Ran tests

## Checklist


tdilauro (approved):
Looks good! 🪵

PR #1859, opened 2024-05-22
[PP-1280] Capture User-Agent header value in S3 Analytics.

dbernstein:
## Description

Adds user-agent field to the S3 Analytics.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1280

## How Has This Been Tested?
Unit tests have been updated.

## Checklist

### DIFF
--- a/src/palace/manager/api/s3_analytics_provider.py
+++ b/src/palace/manager/api/s3_analytics_provider.py
@@ -31,6 +31,7 @@ class S3AnalyticsProvider(LocalAnalyticsProvider):
         old_value,
         new_value,
         neighborhood: str | None = None,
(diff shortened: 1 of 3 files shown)


jonathangreen:
According to the [Flask docs](https://flask.palletsprojects.com/en/2.3.x/api/#flask.Request.user_agent) `request.user_agent` returns [`UserAgent`](https://werkzeug.palletsprojects.com/en/2.3.x/utils/#werkzeug.user_agent.UserAgent) not a string. So you should either update the test and type hints to so we are passing around a `UserAgent` or get the string representation of it and pass that around. It looks like `.string` gets the original header value.

PR #1862, opened 2024-05-22
Remove ActivitySync notification type 🔥 (PP-1304)

jonathangreen:
## Description

This removes the `ActivitySync` notification type that was never handled in the mobile apps.

## Motivation and Context

Removing it because it conflicts with the work I'm doing for PP-1113, and its not being used anyway. We may want to introduce a similar endpoint in the future, but its not necessary right now.

## How Has This Been Tested?

- Running tests

## Checklist


tdilauro (approved):
Looks good! 🔥

PR #1863, opened 2024-05-23
Fixes a bug in retrieving the user agent header.  For whatever reason…

dbernstein:
…, flask.request.user_agent is evaluating to False as the predicate of the if statement (possibly because it is a cached property) causing us not to capture the user agent value.

## Description
This fixes a bug in https://github.com/ThePalaceProject/circulation/pull/1859 which was not evident to me until was examining the user agent values on minotaur.  Initially I thought it has something to do with the headers not being passed to flask from uwsgi or nginx but it turned out to be another issue.  Since flask.request.user_agent is never None and flask.request.user_agent.string will return an empty string  if no header is set,  this change ensures that  None is passed to S3 Analytics if there is no value set.  The ETL process will be responsible for translating the null in to a default value.

## How Has This Been Tested?
Manually.

## Checklist

### DIFF
--- a/src/palace/manager/service/analytics/analytics.py
+++ b/src/palace/manager/service/analytics/analytics.py
@@ -35,9 +35,13 @@ class Analytics(LoggerMixin):
         if not time:
             time = utc_now()
 
-        user_agent = (
-            flask.request.user_agent.string if flask.request.user_agent else None
-        )
+        user_agent: str | None = None
+        try:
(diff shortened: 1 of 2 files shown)


dbernstein:
@jonathangreen :  I added some error handling in case analytics do get called outside of the request context either now or down the road.    I also added test coverage for analytics.collect_event

PR #1865, opened 2024-05-23
Add type hints for the circulation loans controller

jonathangreen:
## Description

Add type hints for `palace.manager.api.controller.circulation_manager` and `palace.manager.api.controller.loan`.

## Motivation and Context

As part of the refactoring I'm doing to the loans controller, having some type hints makes things a lot easier to reason about.

## How Has This Been Tested?

Running unit tests

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/loan.py
+++ b/src/palace/manager/api/controller/loan.py
@@ -23,24 +23,31 @@ from palace.manager.api.problem_details import (
     NO_ACTIVE_LOAN_OR_HOLD,
     NO_LICENSES,
 )
+from palace.manager.core.problem_details import INTERNAL_SERVER_ERROR
 from palace.manager.feed.acquisition import OPDSAcquisitionFeed
 from palace.manager.sqlalchemy.model.datasource import DataSource
-from palace.manager.sqlalchemy.model.licensing import DeliveryMechanism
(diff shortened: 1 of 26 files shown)


src/palace/manager/api/controller/loan.py
```diff
@@ -467,39 +504,39 @@ def revoke(self, license_pool_id):
...
+            return INTERNAL_SERVER_ERROR
         return OPDSAcquisitionFeed.entry_as_response(
-            OPDSAcquisitionFeed.single_entry(work, annotator),
+            single_entry_feed,
             mime_types=flask.request.accept_mimetypes,
         )
 
-    def detail(self, identifier_type, identifier):
-        if flask.request.method == "DELETE":
-            return self.revoke_loan_or_hold(identifier_type, identifier)
```
jonathangreen, line 477: I removed the `DELETE` endpoint here, because `revoke_loan_or_hold` is undefined. Looking through our logs, it seems that this endpoint never gets called with `DELETE`, so it doesn't seem worth it to fix.

jonathangreen:
This is failing on patch coverage, but since I didn't change much other then type hints, I think it should be fine as is.

tdilauro (approved):
Looks good!

PR #1868, opened 2024-05-25
Ensure configuration up-to-date on each request.

tdilauro:
## Description

Moves the check to reload configuration if it has changed to a Flask `before_request` lifecycle function.

## Motivation and Context

I happened to notice the location of this call when working on something else. It seemed that is should be moved to a more "expected" place in the codebase for a couple of reasons:
- The check was being performed in only a subset of cases. It would probably be triggered on a fairly regular basis, but there might be some weird edge cases in which it could lead to unexpected and hard to reproduce behavior. 
- It was in an unexpected place.

## How Has This Been Tested?

- Added new tests.
- CI tests for associated feature branch pass.

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/base.py
+++ b/src/palace/manager/api/controller/base.py
@@ -108,15 +108,20 @@ class BaseCirculationManagerController(LoggerMixin):
         data = self.manager.authentication_for_opds_document
         return Response(data, 401, headers)
 
-    def library_for_request(self, library_short_name):
-        """Look up the library the user is trying to access.
+    def library_for_request(
+        self, library_short_name: str | None
+    ) -> Library | ProblemDetail:
+        """Look up the library the user is trying to use.
 
-        Since this is called on pretty much every request, it's also
-        an appropriate time to check whether the site configuration
-        has been changed and needs to be updated.
+        :param library_short_name: Optional key for looking up the library.
+        :return: The Library for the provided key, if found; a default library,
+            if available; or the `LIBRARY_NOT_FOUND` ProblemDetail.
+
+        If the key is present, use it to look up the library. If it is not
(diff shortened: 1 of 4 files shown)


src/palace/manager/api/controller/base.py
```diff
@@ -108,21 +108,26 @@ def authenticate(self):
...
+        :param library_short_name: Optional key for looking up the library.
+        :return: The Library for the provided key, if found; a default library,
+            if available; or the `LIBRARY_NOT_FOUND` ProblemDetail.
+
+        If the key is present, use it to look up the library. If it is not
+        found, return the `LIBRARY_NOT_FOUND` ProblemDetail. If the key is
+        not present, try to get a default library. If one is not found, return
+        the `LIBRARY_NOT_FOUND` ProblemDetail.
         """
-        self.manager.reload_settings_if_changed()
```
tdilauro, line 118: This deletion is the key change. The additions are just adding some type hints and hopefully improving documentation.

tests/manager/api/controller/test_base.py
```diff
@@ -578,6 +578,10 @@ def test_library_for_request(
             assert expect_default == value
             assert expect_default == flask.request.library  # type: ignore
 
+    # TODO: Remove this test. It's just prove that it passes once in the PR.
+    #  The `reload_settings_if_changed` call has been moved from the relative
+    #  obscurity of `BaseCirculationManagerController.library_for_request`
+    #  into a Flask `before_request` function.
```
tdilauro, line 584: Plan to remove this test once the PR has been reviewed.

jonathangreen (approved):
Looks good to me 🚀

tests/manager/api/controller/test_base.py
tdilauro: This test has been removed.

PR #1870, opened 2024-05-27
Query optimization for the /admin/stats endpoint (PP-1311)

jonathangreen:
## Description

Reduce the number of sql queries we are doing when hitting the stats endpoint by doing some of the aggregation on the database side. 

Jira: PP-1311

## Motivation and Context

Using a copy of the production CA database on my laptop, the `/admin/stats` endpoint takes about 30 - 40 seconds for a admin user to load and about 15 - 20 seconds for a normal non-admin user. This is without the database under load, I would guess the performance is worse in production.

After these changes for me locally the admin user takes about 3 seconds to load, and a non-admin user is under a second. The admin user is still slower then I would like, but 10x speedup is a good start here.

## How Has This Been Tested?

- Tested locally
  - Before and after these changes the /admin/stats endpoint produces the same json with the production DB i was using to test
- Unit tests 

## Checklist


dbernstein (approved):
@jonathangreen : noice.

PR #1871, opened 2024-05-28
Fix issue with monitoring queue stats (PP-1312)

jonathangreen:
## Description

Update our metrics agent to better handle tasks that do not have a routing key. This also adds more information when there is an error to the logs, so that we have more info if we continue to have issues.

## Motivation and Context

Because of the way Celery tracks tasks, its possible for our monitor to get a task without `routing_key` when the task is complete. This happens if a task is started in one snapshot and completed in another, which is a common case when there are a lot of tasks in the queue. 

I'm not sure that this has solved the issue, because its a hard one to reproduce locally, but the changes here should help and give more info in the logs if we do continue to see issues. 

## How Has This Been Tested?

- Tested locally

## Checklist

### DIFF
--- a/src/palace/manager/celery/monitoring.py
+++ b/src/palace/manager/celery/monitoring.py
@@ -175,7 +175,14 @@ class Cloudwatch(Polaroid):
             QueueStats,
             {queue.name: QueueStats() for queue in self.app.conf.get("task_queues")},
         )
-        self.tasks: defaultdict[str, TaskStats] = defaultdict(TaskStats)
+        self.tasks: defaultdict[str, TaskStats] = defaultdict(
+            TaskStats,
+            {
+                task: TaskStats()
+                for task in self.app.tasks.keys()
+                if not self.is_celery_task(task)
+            },
+        )
 
     @staticmethod
     def is_celery_task(task_name: str) -> bool:
--- a/src/palace/manager/celery/monitoring.py
+++ b/src/palace/manager/celery/monitoring.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections import defaultdict
 from collections.abc import Sequence
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


src/palace/manager/celery/monitoring.py
```diff
@@ -116,18 +116,17 @@ class QueueStats(LoggerMixin):
...
-        if task.uuid in self.queued:
-            if task.started:
-                self.log.debug("Task %s started.", task.uuid)
-                self.queued.remove(task.uuid)
-        else:
-            if task.sent and not task.started:
-                self.log.debug("Task %s queued.", task.uuid)
-                self.queued.add(task.uuid)
+        if task.uuid in self.queued and task.started:
+            self.log.debug("Task %s started.", task.uuid)
```
dbernstein, line 125: maybe use an f string for the log message : ie f""Task {task.uuid} started."

src/palace/manager/celery/monitoring.py
```diff
@@ -116,18 +116,17 @@ class QueueStats(LoggerMixin):
...
+            self.queued.remove(task.uuid)
+        elif task.routing_key == self.name and task.sent and not task.started:
+            self.log.debug("Task %s queued.", task.uuid)
```
dbernstein, line 128: same

dbernstein (approved):
Looks good other than the minor comments about formatted strings.

src/palace/manager/celery/monitoring.py
```diff
@@ -182,20 +181,25 @@ def on_shutter(self, state: State) -> None:
...
+            except:
+                task_info = ", ".join(
+                    [
+                        f"[{k}]:{v}"
+                        for k, v in task.info(
+                            extra=["name", "sent", "started", "uuid"]
+                        ).items()
+                    ]
                 )
+                self.logger.exception(f"Error processing task: {task_info}.")
```
tdilauro, line 195: It might be useful to track what exceptions we're getting here, in case we eventually want to narrow the reach of the `except` above.

jonathangreen: `logger.exception` logs the full exception stack trace, along with the information about the exception.

tdilauro (approved):
Latest changes look good!

PR #1872, opened 2024-05-28
[PP-978] Add work suppression for library api calls.

dbernstein:
## Description
This update adds support to the HTTP API for suppressing/unsuppresessing a work for a given library.

The API end points are as follows:

`<host>/admin/works/<identifier_type>/<identifier>/suppress-for-library`
`<host>/admin/works/<identifier_type>/<identifier>/unsuppress-for-library`

These calls require LIBRARY_MANAGER level privileges.

Additionally,  I have updated the suppress/unsuppress work across a CM to require LIBRARY_MANAGER level privileges since suppression of titles across license pools vs across libraries seem to me to be similar in terms of the level of responsibility associated with the action.  If I'm wrong about that I'm happy to revert those changes.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-978

## How Has This Been Tested?

New tests have been added,  broken tests have been fixed.
## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/work_editor.py
+++ b/src/palace/manager/api/admin/controller/work_editor.py
@@ -367,13 +367,14 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
 
         return Response("", 200)
 
-    def suppress(self, identifier_type, identifier):
+    def suppress(self, identifier_type, identifier) -> Response | ProblemDetail:
         """Suppress the license pool associated with a book."""
-        self.require_library_manager(flask.request.library)
+        library = flask.request.library  # type: ignore
+        self.require_library_manager(library)
 
         # Turn source + identifier into a LicensePool
         pools = self.load_licensepools(
-            flask.request.library, identifier_type, identifier
+            flask.request.library, identifier_type, identifier  # type: ignore
         )
         if isinstance(pools, ProblemDetail):
             # Something went wrong.
@@ -386,19 +387,18 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
             pool.suppressed = True
         return Response("", 200)
 
-    def unsuppress(self, identifier_type, identifier):
+    def unsuppress(self, identifier_type, identifier) -> Response | ProblemDetail:
         """Unsuppress all license pools associated with a book.
 
         TODO: This will need to be revisited when we distinguish
         between complaints about a work and complaints about a
         LicensePoool.
         """
-        self.require_library_manager(flask.request.library)
+        library = flask.request.library  # type: ignore
+        self.require_library_manager(library)
 
         # Turn source + identifier into a group of LicensePools
-        pools = self.load_licensepools(
-            flask.request.library, identifier_type, identifier
-        )
+        pools = self.load_licensepools(library, identifier_type, identifier)
         if isinstance(pools, ProblemDetail):
             # Something went wrong.
             return pools
@@ -408,7 +408,9 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
             pool.suppressed = False
         return Response("", 200)
 
-    def suppress_for_library(self, identifier_type: str, identifier: str):
+    def suppress_for_library(
+        self, identifier_type: str, identifier: str
+    ) -> Response | ProblemDetail:
         """Suppress a book at the level of a library."""
 
         library = flask.request.library  # type: ignore
@@ -430,7 +432,9 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
 
         return Response("", 200)
 
-    def unsuppress_for_library(self, identifier_type: str, identifier: str):
+    def unsuppress_for_library(
+        self, identifier_type: str, identifier: str
+    ) -> Response | ProblemDetail:
         """Unsuppress book at the level of a library"""
 
         library = flask.request.library  # type: ignore
--- a/src/palace/manager/api/admin/controller/work_editor.py
+++ b/src/palace/manager/api/admin/controller/work_editor.py
@@ -23,14 +23,16 @@ from palace.manager.api.admin.problem_details import (
 from palace.manager.api.controller.circulation_manager import (
     CirculationManagerController,
 )
-from palace.manager.api.problem_details import REMOTE_INTEGRATION_FAILED
+from palace.manager.api.problem_details import (
+    LIBRARY_NOT_FOUND,
+    REMOTE_INTEGRATION_FAILED,
+)
 from palace.manager.core.classifier import (
     NO_NUMBER,
     NO_VALUE,
     SimplifiedGenreClassifier,
     genres,
 )
-from palace.manager.core.problem_details import INVALID_INPUT
 from palace.manager.feed.acquisition import OPDSAcquisitionFeed
 from palace.manager.feed.annotator.admin import AdminAnnotator
 from palace.manager.sqlalchemy.model.classification import (
@@ -738,6 +740,6 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
             return Response(str(_("Success")), 200)
 
     def no_library_response(self) -> ProblemDetail:
-        return INVALID_INPUT.detailed(
+        return LIBRARY_NOT_FOUND.detailed(
             "A library must be specified in the request. No library specified."
         )
--- a/src/palace/manager/api/admin/controller/work_editor.py
+++ b/src/palace/manager/api/admin/controller/work_editor.py
@@ -369,7 +369,7 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
 
     def suppress(self, identifier_type, identifier):
         """Suppress the license pool associated with a book."""
-        self.require_librarian(flask.request.library)
+        self.require_library_manager(flask.request.library)
 
         # Turn source + identifier into a LicensePool
         pools = self.load_licensepools(
@@ -393,7 +393,7 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
         between complaints about a work and complaints about a
         LicensePoool.
         """
-        self.require_librarian(flask.request.library)
+        self.require_library_manager(flask.request.library)
 
         # Turn source + identifier into a group of LicensePools
         pools = self.load_licensepools(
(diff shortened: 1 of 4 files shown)


dbernstein:
@tdilauro :  I thought you might want to review this one since you're working on the UI for this feature.

src/palace/manager/api/admin/routes.py
```diff
@@ -219,6 +219,34 @@ def unsuppress(identifier_type, identifier):
     return app.manager.admin_work_controller.unsuppress(identifier_type, identifier)
 
 
+@library_route(
+    "/admin/works/<identifier_type>/<path:identifier>/suppress-for-library",
+    methods=["POST"],
+)
+@has_library
```
tdilauro, line 226: I'm concerned about using the `has_library` decorator here, because if no library is provided, it will use the `default` library, which is arbitrary (usually, it's the first library created in a PM instance). In any instance that has more than one library, I don't think this is safe.

We could use the `allows_library` decorator, which will set `flask.request.library` to `None`, if no library is provided. In that case, we should probably handle the `library is None` case.

src/palace/manager/api/admin/routes.py
```diff
@@ -219,6 +219,34 @@ def unsuppress(identifier_type, identifier):
...
+    return app.manager.admin_work_controller.suppress_for_library(
+        identifier_type, identifier
+    )
+
+
+@library_route(
+    "/admin/works/<identifier_type>/<path:identifier>/unsuppress-for-library",
+    methods=["POST"],
+)
+@has_library
```
tdilauro, line 240: Same concern about `has_library` as above.

tests/manager/api/admin/test_routes.py
```diff
@@ -383,6 +383,26 @@ def test_unsuppress(self, fixture: AdminRouteFixture):
             http_method="POST",
         )
 
+    def test_suppress_for_library(self, fixture: AdminRouteFixture):
+        url = "/admin/works/<identifier_type>/an/identifier/suppress-for-library"
+        fixture.assert_authenticated_request_calls(
+            url,
+            fixture.controller.suppress_for_library,  # type: ignore
```
tdilauro, line 390: Same minor type checker suggestion here: # type: ignore[attr-defined].

tests/manager/api/admin/test_routes.py
```diff
@@ -383,6 +383,26 @@ def test_unsuppress(self, fixture: AdminRouteFixture):
...
+            "<identifier_type>",
+            "an/identifier",
+            http_method="POST",
+        )
+
+    def test_unsuppress_for_library(self, fixture: AdminRouteFixture):
+        url = "/admin/works/<identifier_type>/an/identifier/unsuppress-for-library"
+        fixture.assert_authenticated_request_calls(
+            url,
+            fixture.controller.unsuppress_for_library,  # type: ignore
```
tdilauro, line 400: Same minor type checker suggestion here: # type: ignore[attr-defined].

src/palace/manager/api/admin/controller/work_editor.py
```diff
@@ -386,19 +387,18 @@ def suppress(self, identifier_type, identifier):
...
-    def unsuppress(self, identifier_type, identifier):
+    def unsuppress(self, identifier_type, identifier) -> Response | ProblemDetail:
         """Unsuppress all license pools associated with a book.
 
         TODO: This will need to be revisited when we distinguish
         between complaints about a work and complaints about a
         LicensePoool.
         """
-        self.require_librarian(flask.request.library)
+        library = flask.request.library  # type: ignore
```
tdilauro, line 397: Same minor type checker suggestion here: # type: ignore[attr-defined].

dbernstein:
@tdilauro : thanks for the review - I'll integrate your suggestions shortly.

jonathangreen:
Minor comment so feel free to ignore. I was wondering about the endpoints for this one. Because of the decorators this ends up looking like:
`/<library_short_name>/admin/works/<identifier_type>/<identifier>/suppress-for-library`

A couple suggestions:
- Drop `-for-library` and just re-purpose the `suppress` route since, as I understand it, we don't want to support global suppression anyway.
  - As part of this we could remove the code for global suppression (this could be in a follow up PR as well).
- Then instead of having an endpoint to suppress and one to unsuppress, would it make sense to just suppress with a `POST` and unsuppress with a `DELETE`.

dbernstein:
@jonathangreen :  I like it. I don't mind integrating those changes into the current PR.

dbernstein:
Assuming we want to move forward (pending @courte 's approval):
I propose do the following:
1.  Remove old suppression code 
2. Create migration to remove old data elements. 
3. Rewire the suppression end points to the new implementation.
4. Add  DELETE method support to /suppress (in order to support unsuppress.
5. Deprecate /unsuppress but keep it in there until the front end is no longer using it.

dbernstein:
I held off on removing `LicensePools.suppressed` because I was unsure of the impact on the feed for @tdilauro .  So I created https://ebce-lyrasis.atlassian.net/browse/PP-1320 to cover that work when we get around to it.  For the time being, I don't think anything should be impacted by leaving it in there since nobody is using the feature anyway (except for a couple of books in California).

src/palace/manager/api/admin/controller/work_editor.py
```diff
@@ -367,45 +368,58 @@ def edit(self, identifier_type, identifier):
...
-        self.require_librarian(flask.request.library)
+    def suppress(
+        self, identifier_type: str, identifier: str
+    ) -> Response | ProblemDetail:
+        """Suppress a book at the level of a library."""
+
+        library = flask.request.library  # type: ignore[attr-defined]
+
+        if library is None:
+            return self.no_library_response()
```
tdilauro, line 379: This is fine, though it returns a 400. In other places (e.g., [inventory reports](https://github.com/ThePalaceProject/circulation/blob/d213661bd86185ad32127410837e6526f12474ad/src/palace/manager/api/admin/controller/report.py#L35-L37) and [the `has_library` decorator via `BaseCirculationController.library_for_request`](https://github.com/ThePalaceProject/circulation/blob/85615864fea7bf0906fd1321dbf2f06cf0602065/src/palace/manager/api/controller/base.py#L130-L131)), we return the `LIBRARY_NOT_FOUND` problem detail or just raise a ProblemDetailException with it. These return 404.

```python
...
from palace.manager.api.problem_details import LIBRARY_NOT_FOUND
...
    library: Library | None = getattr(request, "library")
    if library is None:
        raise ProblemDetailException(LIBRARY_NOT_FOUND)
```

src/palace/manager/api/admin/controller/work_editor.py
```diff
@@ -367,45 +368,58 @@ def edit(self, identifier_type, identifier):
...
 
-        TODO: This will need to be revisited when we distinguish
-        between complaints about a work and complaints about a
-        LicensePoool.
-        """
-        self.require_librarian(flask.request.library)
+        library = flask.request.library  # type: ignore[attr-defined]
+
+        if library is None:
+            return self.no_library_response()
```
tdilauro, line 406: Same here re: `LIBRARY_NOT_FOUND` problem detail.

tests/manager/api/admin/controller/test_work_editor.py
```diff
@@ -74,64 +75,6 @@ def work_fixture(
...
-            AdminRole.LIBRARIAN, work_fixture.ctrl.db.default_library()
-        )
-        with work_fixture.request_context_with_library_and_admin("/"):
-            pytest.raises(
-                AdminNotAuthorized,
-                work_fixture.manager.admin_work_controller.details,
-                lp.identifier.type,
-                lp.identifier.identifier,
-            )
-
```
tdilauro, line 134: The `details` endpoint is not going away, so these tests will be needed, just in a different form. 

It is okay for them to be removed for the moment, though. I will add them back in some form as part of work on PP-1318.

tests/manager/api/admin/test_routes.py
```diff
@@ -373,6 +384,28 @@ def test_suppress(self, fixture: AdminRouteFixture):
...
+
+        # TODO: when /suppress end point is removed (currently deprecated we'll want to remove the following two lines
+        # since these tests will start failing here.
+        url = "/admin/works/<identifier_type>/an/identifier/suppress"
+        fixture.assert_authenticated_request_calls(
+            url,
+            fixture.controller.unsuppress,  # type: ignore
+            "<identifier_type>",
+            "an/identifier",
+            http_method="DELETE",
```
tdilauro, line 405: The `/suppress` endpoint doesn't have a "DELETE" method. Copypasta? Or did you mean to POST to the `/unsuppress` endpoint here? There's a test below that does that.

tdilauro (commented):
This looks pretty close. Just a few comments. Also, I haven't investigated, but a bunch of tests are failing with
```
TypeError: CirculationManager.__init__() got multiple values for argument 'services'
```

tests/manager/api/admin/test_routes.py
dbernstein: Good catch - yes.  Copypasta indeed.  Fixed.

jonathangreen:
@dbernstein re: the test failures here. https://github.com/ThePalaceProject/circulation/pull/1880 should help the situation here. A test fixture is failing to unload due to a context manager which is causing issues with lots of tests. Once that PR goes in its much more clear which tests are failing on this PR.

src/palace/manager/api/admin/controller/work_editor.py
tdilauro: I actually think that this should return a 404, but fine to leave this as-is here and I can update in the in-progress follow-up PR.

src/palace/manager/api/admin/controller/work_editor.py
tdilauro: I actually think that this should return a 404, but fine to leave this as-is here and I can update in the in-progress follow-up PR.

tdilauro (approved):
I think this one is fine to go in now. 🎸🤘🏽

I noted a few changes that I expect to make in a follow up PR.

src/palace/manager/api/admin/controller/work_editor.py
dbernstein: Fixed.

PR #1876, opened 2024-05-30
Revert customlists to run from a script rather then a celery task (PP-1279)

jonathangreen:
## Description

Revert the celery tasks added in https://github.com/ThePalaceProject/circulation/pull/1760 that updates custom lists. This adds back in the `custom_list_update_new_entries` script, and supporting code.

## Motivation and Context

The custom lists scripts seem to be adding a LOT of search index update tasks, and putting a load of load on the Open Search server which is causing problems with indexing not being able to keep up. 

As part of PP-1279, I want to roll back these celery jobs, and see if that helps the situation. If it does I'm going to make a ticket for next sprint to look into these custom lists jobs, and try to resolve the issues there. Resolving the issue with custom lists looks like it will be a large task, so I'd like to handle it next sprint.

## How Has This Been Tested?

Tested by running script locally.

## Checklist


jonathangreen:
This one is failing on coverage, but since its just reverting some work, I think its fine. We had those holes in our tests before. Hopefully will be able to undo this revert next sprint anyway.

dbernstein (approved):
Looks good.

PR #1878, opened 2024-05-30
Fix run script path for custom list update script

jonathangreen:
## Description

Fix the run script path for the custom list update script. 

## Motivation and Context

I missed when I did the revert that the path for the run script has changed from `core/bin/run` to `bin/run`, so the custom list update script isn't running.

## How Has This Been Tested?

N/A

## Checklist


tdilauro (approved):
👍🏽

PR #1879, opened 2024-05-30
[PP-1274] Removes ODL Hold Reaper and replaces with ODL 2 Hold Reaper.

dbernstein:
## Description
This update removes the ODLHoldsReaper which currently is not doing anything because there are no longer any ODL based collections.

I added in the ODL2HoldsReaper odl2.py which reaps holds associated with  ODL2 collections.

Furthermore, in order to keep the dashboard stats as up to date as possible, I modified the active holds query to hide holds that are expired (and will be reaped within 8 hours when the reaper runs).
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1274

## Checklist

### DIFF
--- a/bin/odl2_hold_reaper
+++ b/bin/odl2_hold_reaper
@@ -1,0 +1,8 @@
+#!/usr/bin/env python
+"""Check for ODL 2 holds that have expired and delete them."""
+
+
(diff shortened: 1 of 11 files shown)


jonathangreen (approved):
Looks good

PR #1886, opened 2024-06-03
Ensures that fiction metadata element is being stored in the search i…

dbernstein:
…ndex in lowercase so that searches for "fiction" and "nonfiction" produce the expected results.

> Resolves: https://ebce-lyrasis.atlassian.net/browse/PP-1332

## Description
This is the story as I understand it: 

The "Fiction" and "Nonfiction" values associated with "fiction" search document property have been stored in a mix case way since at least 3/23/2022.  At that time we did not have a classifier list filter in the custom list builder in circulation-admin.  When that feature was introduced on 3/28/2024,  the list values of "fiction" and "nonfiction" were defined.  The client sends those values to the backend for processing.  When @jonathangreen  made that addition we had fuzzy filtering turned on so "fiction" and "nonfiction" in the query matched for "Fiction" and "Nonfiction" in the search index.  However, when we disabled fuzzy they are no longer exact matches.   I believe we are both normalizing our queries to use lowercase matching and lowercasing any text going into the search index in the author, title, summary, genre, audience etc fields.  It appears that fiction was not being lowercased and thus the non-matching once fuzziness was disabled.

This update simply lowercases the fiction field value on the way into the search index.   This way,  search queries coming into palace using "fiction" and "Fiction" will produce the expected matches. 

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1332

## How Has This Been Tested?

I verified that the matches were not occurring.  Then I verified that changing the incoming value from the client to mix case caused search results to appear.  I wrote a failing end to end test that demonstrated the problem.  I then put in the fix and confirmed the test was passing.  Finally I verified that the UI was giving the expected results post fix.

I'll be adding an end to end test to bring this one to completion.

## Checklist

### DIFF
--- a/tests/manager/celery/tasks/test_search.py
+++ b/tests/manager/celery/tasks/test_search.py
@@ -75,7 +75,7 @@ def test_search_reindex(
     end_to_end_search_fixture.expect_results([work1, work2, work4], "", ordered=False)
 
 
-def test_PP_1332_fiction_returns_results(
+def test_fiction_query_returns_results(
     db: DatabaseTransactionFixture, end_to_end_search_fixture: EndToEndSearchFixture
 ) -> None:
     work1 = db.work(with_open_access_download=True, fiction=True)
@@ -87,20 +87,20 @@ def test_PP_1332_fiction_returns_results(
     end_to_end_search_fixture.expect_results(
         expect=[work1, work2], ordered=False, query_string=""
     )
-    default_filter = Filter()
-    default_filter.search_type = "json"
+    json_filter = Filter()
+    json_filter.search_type = "json"
     qs = {"query": {"key": "fiction", "value": "fiction"}}
     end_to_end_search_fixture.expect_results(
         expect=[work1],
         ordered=False,
-        filter=default_filter,
+        filter=json_filter,
         query_string=json.dumps(qs),
     )
     qs["query"]["value"] = "nonfiction"
     end_to_end_search_fixture.expect_results(
         expect=[work2],
         ordered=False,
-        filter=default_filter,
+        filter=json_filter,
         query_string=json.dumps(qs),
     )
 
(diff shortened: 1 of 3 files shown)


tests/manager/celery/tasks/test_search.py
```diff
@@ -73,6 +75,36 @@ def test_search_reindex(
     end_to_end_search_fixture.expect_results([work1, work2, work4], "", ordered=False)
 
 
+def test_PP_1332_fiction_returns_results(
```
jonathangreen, line 78: I don't think the test name here has to refer to a ticket. It would be nice to have the test name stand on its own, so we don't have to go digging in jira in the future when we are refactoring this test.

tests/manager/celery/tasks/test_search.py
```diff
@@ -73,6 +75,36 @@ def test_search_reindex(
...
+    work1 = db.work(with_open_access_download=True, fiction=True)
+    work2 = db.work(with_open_access_download=True, fiction=False)
+    documents = get_work_search_documents(db.session, 2, 0)
+    assert {doc["_id"] for doc in documents} == {work2.id, work1.id}
+
+    end_to_end_search_fixture.populate_search_index()
+    end_to_end_search_fixture.expect_results(
+        expect=[work1, work2], ordered=False, query_string=""
+    )
+    default_filter = Filter()
```
jonathangreen, line 90: This isn't really the default filter, the default filter is what you get when you don't pass a filter into the query. This should probably be called `filter` or `json_filter`.

jonathangreen (approved):
Made a couple comments on this one that would be nice to clean up before merging, but it looks good. 

Since we've had this field in mixed case for some time, I'm a bit concerned that this change is going to break some other use of this field in the search index, where we are currently depending on it to be mixed case, instead of lower case. I'm assuming you've looked for uses like this already.

dbernstein:
@jonathangreen :  yes, I reviewed all the instances of "Fiction" that may be affected by this change and I didn't see anything that would present a problem.

PR #1888, opened 2024-06-05
ODL2 collections can disallow holds at the collection level. (PP-1316)

tdilauro:
## Description

- ODL2 collections can disable holds by setting `holds_limit` to `0`.
- Attempting to place a hold on a book in that collection with this setting in effect results in a (new) `HoldsNotPermitted` problem detail exception.
- As with library-level `dont_display_reserves` setting, books for collections with this configuration will be filtered from work lists when there is no availability; so, in most cases, holds for these books won't even be attempted and we should not even need to return a problem detail.

## Motivation and Context

Need to be able to prevent holds on an entire collection, but without forcing a library to disable holds for all of its collections.

[Jira [PP-1316](https://ebce-lyrasis.atlassian.net/browse/PP-1316)]

## How Has This Been Tested?

- Function tested locally.
- Added new tests and updated others.
- [CI tests for associated branch](https://github.com/ThePalaceProject/circulation/actions/runs/9386205068) pass.

## Checklist

[PP-1316]: https://ebce-lyrasis.atlassian.net/browse/PP-1316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/api/circulation_exceptions.py
+++ b/src/palace/manager/api/circulation_exceptions.py
@@ -13,6 +13,7 @@ from palace.manager.api.problem_details import (
     EXPIRED_CREDENTIALS,
     HOLD_FAILED,
     HOLD_LIMIT_REACHED,
+    HOLDS_NOT_PERMITTED,
     INVALID_CREDENTIALS,
     LOAN_LIMIT_REACHED,
     NO_ACCEPTABLE_FORMAT,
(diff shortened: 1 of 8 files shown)


jonathangreen (approved):
Looks good! 🚀

PR #1889, opened 2024-06-05
Update and type hint the circulation API Mocks (PP-1113)

jonathangreen:
## Description

This adds type hints to our circulation api mocks, and cleans up some of the circulation api tests. It shouldn't change our code coverage, and doesn't change any non-test code.

## Motivation and Context

As part of PP-1113, I've had to do a lot of changes to these tests. I wanted to break some of that refactoring out into its own PR, so the main PR is a bit easier to read.

## How Has This Been Tested?

- Running unit tests
- Looking at coverage

## Checklist

### DIFF
--- a/tests/manager/api/test_circulationapi.py
+++ b/tests/manager/api/test_circulationapi.py
@@ -2,7 +2,7 @@
 import datetime
 from datetime import timedelta
 from typing import cast
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, create_autospec, patch
 
 import flask
 import pytest
@@ -34,7 +34,6 @@ from palace.manager.api.circulation_exceptions import (
     PatronHoldLimitReached,
     PatronLoanLimitReached,
 )
-from palace.manager.core.opds_import import OPDSAPI
 from palace.manager.service.analytics.analytics import Analytics
 from palace.manager.sqlalchemy.model.circulationevent import CirculationEvent
 from palace.manager.sqlalchemy.model.datasource import DataSource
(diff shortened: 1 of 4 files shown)


tests/manager/api/test_circulationapi.py
```diff
@@ -183,15 +178,6 @@ def test_attempt_borrow_with_existing_remote_loan(
...
         # Remote loan.
-        circulation_api.circulation.add_remote_loan(
-            circulation_api.pool.collection,
-            circulation_api.pool.data_source,
-            circulation_api.identifier.type,
-            circulation_api.identifier.identifier,
-            self.YESTERDAY,
-            self.IN_TWO_WEEKS,
-        )
-
```
jonathangreen, line 194: This call to `add_remote_loan` was here, but because we never call `patron_activity`, it makes no difference. So we just remove them without impacting the tests.

A number of these are removed. I'm just going to call out this one, but they were all removed for the same reason.

tests/manager/api/test_circulationapi.py
```diff
@@ -893,92 +851,23 @@ def test_fulfill_errors(self, circulation_api: CirculationAPIFixture):
         circulation_api.pool.open_access = True
         circulation_api.pool.collection = collection
 
-        circulation_api.circulation.remotes[
-            circulation_api.pool.data_source.name
-        ] = OPDSAPI(circulation_api.db.session, collection)
```
jonathangreen, line 898: This test didn't really make sense, because it was injecting a `OPDSAPI` into the `MockCirculationApi`. All the functionality that this tested in `OPDSAPI` is tested directly in that class, so most of this test was removed. The only part that was left is making sure when we call `FormatNotAvailable` bubbles up when calling `fulfill`.

The coverage for `OPDSAPI` shouldn't change at all with this removal.

dbernstein (approved):
Looks good.

PR #1898, opened 2024-06-12
Feature flag to make inventory reports sysadmin only. (PP-1329)

tdilauro:
## Description

- Adds an Admin UI feature flag to restrict inventory reports to sysadmins (default: `true` = restrict)
- Adds a settings mechanism that can selectively restrict override from the environment.

## Motivation and Context

Need ability to restrict inventory report requests from non-sysadmins until we've been able to test that functionality.

There were two things I was trying to accomplish:
- Have one home for configuration of feature flags; and
- Avoid accidentally overriding configuration that we have decided is fixed (see the hard-coded values I pulled back into the feature flags object).

Being able to prevent override by the environment variables preserves the existing behavior of those pre-existing feature flags (i.e., they could not be overridden by the environment -- they were fixed).

I believe the main use case for this is to avoid accidentally reconfiguring some functionality that we have decided is now 'fixed' into the system. I think should mostly apply only to functionality that is in transition. That said, we haven't gone back to the admin UI and removed these fixed feature flags. I'm not sure if that's because we want them there or because we just haven't done anything about them.

I made a separate class for this primarily to make updates to the feature flags a little clearer and less cluttered.

[Jira [PP-1329](https://ebce-lyrasis.atlassian.net/browse/PP-1329)]

## How Has This Been Tested?

- Manually tested locally via API requests.
- All tests pass locally.
- CI tests pass.

## Checklist

[PP-1329]: https://ebce-lyrasis.atlassian.net/browse/PP-1329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/palace/manager/util/settings.py
tdilauro, line 1: Not sure that this is the best place for this, but it cannot go in the same file with `ServiceConfiguration` because of an import loop due to `LoggerMixin`.

src/palace/manager/api/admin/controller/view.py
```diff
@@ -67,8 +67,8 @@ def __call__(self, collection, book, path=None):
         admin_js = AdminClientConfig.lookup_asset_url(key="admin_js")
         admin_css = AdminClientConfig.lookup_asset_url(key="admin_css")
 
-        # We always have local_analytics
-        show_circ_events_download = True
+        # # We always have local_analytics
+        # show_circ_events_download = True
```
jonathangreen, line 71: Minor: Should be able to remove this comment.

src/palace/manager/util/settings.py
```diff
@@ -0,0 +1,134 @@
...
+
+from pydantic.env_settings import BaseSettings, SettingsSourceCallable
+from pydantic.fields import ModelField
+
+from palace.manager.core.config import CannotLoadConfiguration
+from palace.manager.service.configuration import ServiceConfiguration
+from palace.manager.util.log import LoggerMixin
+
+
+class ServiceConfigurationWithLimitedEnvOverride(ServiceConfiguration, LoggerMixin):
```
jonathangreen, line 13: Would it make sense for this to live in `palace.manager.service.configuration` with the parent `ServiceConfiguration` class. Rather then in `palace.manager.util.settings`?

jonathangreen (approved):
Added a couple minor comments, but the code here looks good to me.

One question I have though, reading though this is what is the use case for flags that we can't override with environment variables? Why not just use `ServiceConfiguration` and allow `enable_auto_list` and `show_circ_events_download` to be overridden if desired?

src/palace/manager/util/settings.py
tdilauro: Yeah. Probably. I wasn't entirely sure where to put it, but that makes sense.

src/palace/manager/api/admin/controller/view.py
tdilauro: There were two things I was trying to accomplish:
- Have one home for configuration of feature flags; and
- Avoid accidentally overriding configuration that we have decided is fixed (see the hard-coded values I pulled back into the feature flags object).

Being able to prevent override by the environment variables preserves the existing behavior of those pre-existing feature flags (i.e., they could not be overridden by the environment -- they were fixed).

I believe the main use case for this is to avoid accidentally reconfiguring some functionality that we have decided is now 'fixed' into the system. I think should mostly apply only to functionality that is in transition. That said, we haven't gone back to the admin UI and removed these fixed feature flags. I'm not sure if that's because we want them there or because we just haven't done anything about them.

I made a separate class for this primarily to make updates to the feature flags a littel clearer and less cluttered.

src/palace/manager/util/settings.py
tdilauro: Though, as I noted above, it cannot live in the same file:
> Not sure that this is the best place for this, but it cannot go in the same file with ServiceConfiguration because of an import loop due to LoggerMixin.

src/palace/manager/api/admin/controller/view.py
jonathangreen: I'm fine with this going in as is, so feel free either way. 

My opinion on this though is that we should allow the feature flags to be overridden by env vars and any of the flags we want to be set as fixed, we can put in tickets to remove the flag and clean up the config in the admin ui and CM.

src/palace/manager/util/settings.py
jonathangreen: Maybe we could move `palace.manager.service.configuration.ServiceConfiguration` to `palace.manager.service.configuration.base.ServiceConfiguration` and put this in the `palace.manager.service.configuration` package. We can just leave it where it is for now as well.

tdilauro: @jonathangreen I didn't see this before I moved them. I put both of them in `palace.manager.service.configuration` package directory. I can move them, though, if you'd prefer.

jonathangreen: No whats here looks good to me.

PR #1899, opened 2024-06-12
Add a refresh query parameter to the loans feed (PP-1368)

jonathangreen:
## Description

Aspen doesn't need us to go out and refresh our loans feed every time they make a request. They are going to add this query parameter to the loans feed, so we add an additional check to support it.

JIRA: PP-1368

## Motivation and Context

Reduce the load of generating the loans feed when we don't actually need to refresh it.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/loan.py
+++ b/src/palace/manager/api/controller/loan.py
@@ -6,6 +6,7 @@ import flask
 from flask import Response, redirect
 from flask_babel import lazy_gettext as _
 from lxml import etree
+from pydantic import parse_obj_as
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good!

tdilauro:
Code coverage is probably complaining about not testing the exception case.

jonathangreen:
@tdilauro just pushed a commit that add some more testing, so we test the exception case as well as some true and false values.

PR #1904, opened 2024-06-13
Celery logging config

jonathangreen:
## Description

Its possible for the log level config to be `None` so we get an exception when accessing `levelno`. This just adds a None check.

## Motivation and Context

Fix a minor bug I noticed in while testing https://github.com/ThePalaceProject/circulation/pull/1894.

## How Has This Been Tested?

Running tests.

## Checklist


jonathangreen:
Codecov is complaining about this one, but the line it wants tested is pretty difficult to actually test. I think this one is fine to go in as is.

PR #1905, opened 2024-06-14
Per-library work suppression is idempotent. (PP-1378)

tdilauro:
## Description

Makes per library work suppression (and unsuppression) idempotent. 
- Suppressing a book for a library that is already suppressed for that library does not result in an error. 
- Unsuppressing a book for a library that is already unsuppressed for that library does not result in an error.

## Motivation and Context

We were not handling an exception when unsuppressing more than once. This resulted in a 500 internal server error.

[Jira [PP-1378](https://ebce-lyrasis.atlassian.net/browse/PP-1378)]

## How Has This Been Tested?

- New test case that failed before the fix and passed afterward.
- Manual testing locally.

## Checklist

[PP-1378]: https://ebce-lyrasis.atlassian.net/browse/PP-1378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/api/admin/controller/work_editor.py
+++ b/src/palace/manager/api/admin/controller/work_editor.py
@@ -390,12 +390,18 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
             # Something went wrong.
             return work
 
-        work.suppressed_for.append(library)
-        self.log.info(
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
🚀 Looks great!

PR #1911, opened 2024-06-19
Add TaskLock to search reindex task

jonathangreen:
## Description

Add `TaskLock` to protect the search reindex task. That way we are sure there is only one search reindex task running at a time.

The task holds the lock the whole time the reindex is happening, even when the task requeues itself, so we are fairly sure that only one long running full reindex task runs at a time.

## Motivation and Context

In https://github.com/ThePalaceProject/circulation/pull/1909 I added the `TaskLock` class. This is well suited to make sure that only one reindex task is happening at a time. 

## How Has This Been Tested?

- Running tests locally

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/search.py
+++ b/src/palace/manager/celery/tasks/search.py
@@ -4,6 +4,7 @@ from collections.abc import Sequence
 from typing import Any
 
 from celery import chain, shared_task
+from celery.exceptions import Ignore, Retry
 from opensearchpy import OpenSearchException
 from sqlalchemy import select
 from sqlalchemy.orm import Session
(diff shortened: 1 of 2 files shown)


src/palace/manager/celery/tasks/search.py
```diff
@@ -69,31 +70,40 @@ def search_reindex(task: Task, offset: int = 0, batch_size: int = 500) -> None:
...
 
-    if len(documents) == batch_size:
-        # This task is complete, but there are more works waiting to be indexed. Requeue ourselves
-        # to process the next batch.
-        raise task.replace(
-            search_reindex.s(offset=offset + batch_size, batch_size=batch_size)
-        )
+        add_documents_to_index(task, index, documents)
+
+        if len(documents) == batch_size:
```
tdilauro, line 98: Same comments here as on #1909, but indented. 😄

tdilauro (approved):
Looks good! 🎸

src/palace/manager/celery/tasks/search.py
jonathangreen: This is a similar case. It can happen, but if it does, the task just gracefully handles having 0 items to process and then exits.

jonathangreen: I added some tests with different batch sizes to make sure this is fully exercised

PR #1912, opened 2024-06-20
Overdrive API patron_activity returning None

jonathangreen:
## Description

Add a few type hints to the OD API class and make sure the `patron_activity` function don't return `None`.

## Motivation and Context

While monitoring the production roll out for https://github.com/ThePalaceProject/circulation/pull/1894 I saw a few tasks failing with the following exception: 

```
remote_holds_and_loans
    key = self.IdentifierKey(activity.identifier_type, activity.identifier)
AttributeError: 'NoneType' object has no attribute 'identifier_type'
```

It turns out this is because the OD API can return None, which wasn't expected.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/src/palace/manager/api/overdrive.py
+++ b/src/palace/manager/api/overdrive.py
@@ -8,6 +8,7 @@ import logging
 import re
 import time
 import urllib.parse
+from collections.abc import Iterable
 from threading import RLock
 from typing import Any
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🛠️

PR #1914, opened 2024-06-20
Add support for SQS to our Celery configuration (PP-1397)

jonathangreen:
## Description

Add support for deploying via SQS to our Celery configuration. The default broker we use is still Redis, but this allows us to experiment with using SQS if we want to. This also lets us set arbitrary celery settings that aren't currently part of our configuration model, so we are able to test and customize the SQS behaviour.

## Motivation and Context

Since SQS is a managed service, it provides nice metrics about queue length etc that we could use if we were using SQS. It might be worth doing a test deploy with SQS to see if it works for us. This will let us do some experiments with it.

## How Has This Been Tested?

- Tested locally
- Running unit tests

## Checklist

### DIFF
--- a/src/palace/manager/service/celery/configuration.py
+++ b/src/palace/manager/service/celery/configuration.py
@@ -1,10 +1,17 @@
+import os
 from typing import Any
 
-from pydantic import RedisDsn
+from pydantic import AnyUrl, Extra
+from pydantic.env_settings import BaseSettings, SettingsSourceCallable
 
 from palace.manager.service.configuration.service_configuration import (
     ServiceConfiguration,
 )
+
+
+class CeleryBrokerUrl(AnyUrl):
+    host_required: bool = False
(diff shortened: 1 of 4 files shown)


src/palace/manager/service/celery/configuration.py
```diff
@@ -43,6 +56,27 @@ class CeleryConfiguration(ServiceConfiguration):
...
+        ) -> tuple[SettingsSourceCallable, ...]:
+            # We return an additional function that will parse the environment
+            # variables and extract any that are not part of the settings model,
+            # so that we can set additional configuration options for Celery at
+            # deployment time if needed.
+            return (
+                init_settings,
+                env_settings,
+                file_secret_settings,
+                additional_fields_from_env,
```
tdilauro, line 78: Nice use of pydantic's custom sources capabilities! 🚀

tdilauro (approved):
Looks good! 🏅

PR #1920, opened 2024-06-27
Fix bug with OPDS2 schema validation (PP-1414)

jonathangreen:
## Description

This PR makes a couple improvements here while fixing this bug:
- Replace the use of the deprecated `RefResolver` class with the new `Registry` class.
- Previously we kept several of the schema documents locally, because we had patched them to remove a problematic regex. These documents were what was referring to the old path for `metadata.schema.json`.
  - Instead of this, we create a small validation function that patches the documents automatically, so we don't have to keep a modified copy in this repo. 
- Update our resolver to only resolve `file://` URIs and update our ODL schema to use these URIs so it is clear what we should be loading locally vs fetching remotely.
- Add an explicit dependency on `jsonschema`, we were previously using the library, but we didn't explicitly depend on it, it was coming in via a transitive dependency.
- Remove the dependency on `types-jsonschema` since these type stubs are out of date. They stopped updating them, and are concentrating on adding type info into `jsonschema` directly. This work isn't done yet, so they don't have a `py.typed` marker published yet. The outdated type stubs are more hassle then they are worth right now while things are in flux.

Hopefully this will reduce the likelihood that we have this problem again in the future.

## Motivation and Context

Our tests started failing because we were no longer able to resolve: 
https://readium.org/webpub-manifest/schema/extensions/presentation/metadata.schema.json

The URL of this file was changed here: https://github.com/readium/webpub-manifest/pull/89

## How Has This Been Tested?

- Running tests locally

## Checklist


tdilauro (approved):
Looks good! 🏁

PR #1928, opened 2024-07-15
Sirsi configuration was missing item type (PP-1415)

jonathangreen:
## Description

For PP-1415 I was testing Sirsi configuration, and I noticed that even though we have options set, the type of the form item isn't set to select, so they weren't being displayed. 

## Motivation and Context

This updates the form so it works as I would expect.

## How Has This Been Tested?

Tested locally.

## Checklist


tdilauro (approved):
Looks good!

PR #1931, opened 2024-07-16
Return more information from sirsi self tests (PP-1415)

jonathangreen:
## Description

This adds some additional information to the self test results for the sirsi integration. The self test results now output the full data we get back from the Sirsi api for a patron as part of the results.

## Motivation and Context

This should make it easier to troubleshoot when sirsi auth is failing, or when we need to validate particular information for a patron like the btype request in PP-1415.

This PR has two commits on it, looking at them separately might make it easier to review. The first one https://github.com/ThePalaceProject/circulation/pull/1931/commits/b7873ebb666a4088cea33369227c0d200dc059d8 refactors tests to make the changes easier to test, and the second one https://github.com/ThePalaceProject/circulation/pull/1931/commits/187fd421d24bfbb0b0dae28a7e6d69ef3d6a3408 updates the self tests and adds tests for that change.

## How Has This Been Tested?

- Unit tests
- Tested locally with config from PP-1415
  - This helped to validate that `patronType` contains the data that they were referring to as `btype`.

## Checklist

### DIFF
--- a/src/palace/manager/api/sirsidynix_authentication_provider.py
+++ b/src/palace/manager/api/sirsidynix_authentication_provider.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
+import json
 import os
+from collections.abc import Callable, Generator
 from gettext import gettext as _
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Any, Literal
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Nice work! 🥇

PR #1933, opened 2024-07-16
[PP-1431] Ensures that audience classifiers with multiple words (such…

dbernstein:
… as "Young Adult" and "Adults Only") returns results in the list search interface.

## Description
Before audience is indexed in the search index, the spaces are removed. So "Young Adult" is being put in the search index as "YoungAdult".  When incoming queries come in from the user interface however "Young Adult" was not being transformed.  That is why searching on "YoungAdult" produced results as well as "Young" and "Adult", but not "Young Adult".   I fixed the problem by adding a missing value transform for audience.  Incidentally this will also fix the problem for lists intended for "Adults Only".

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1431

## How Has This Been Tested?

Manually tested.  Unit test added.
## Checklist

### DIFF
--- a/src/palace/manager/search/external_search.py
+++ b/src/palace/manager/search/external_search.py
@@ -1097,10 +1097,16 @@ class JSONQuery(Query):
             value = list(transformed)[0] if len(transformed) > 0 else value
             return value
 
+        @staticmethod
+        def audience(value: str) -> str:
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
🚀 looks good

PR #1934, opened 2024-07-17
Fix patron activity sync not persisting (PP-1448)

jonathangreen:
## Description

This fixes a bug where the patron activity sync task, is not committing its changes to the DB. The main change here is to update the task to use `task.transaction()`, so that the data from the sync is automatically committed to the database.

The PR also adds some additional testing to verify that a successful sync creates a loan and hold. 

## Motivation and Context

Looking into PP-1448, I saw that loans were syncing from B&T, but that the loan never updated the loans feed. This was because the updated loans were not ever being committed into the database.

## How Has This Been Tested?

- Tested locally
- Updated unit test

## Checklist

### DIFF
--- a/src/palace/manager/api/circulation.py
+++ b/src/palace/manager/api/circulation.py
@@ -822,8 +822,8 @@ class PatronActivityCirculationAPI(
                 # This was just created, we shouldn't delete it.
                 continue
             self.log.info(
-                "Deleting %s (%r) for patron (%s)"
-                % (item.__class__.__name__, item, item.patron.authorization_identifier)
+                f"Deleting {item.__class__.__name__} ({item.id}) on license pool (id: {item.license_pool_id}) "
+                f"for patron ({item.patron.authorization_identifier})"
             )
             self._db.delete(item)
 
--- a/src/palace/manager/celery/tasks/patron_activity.py
+++ b/src/palace/manager/celery/tasks/patron_activity.py
@@ -37,7 +37,7 @@ def sync_patron_activity(
             )
             return
 
-        with task.session() as session:
+        with task.transaction() as session:
             patron = get_one(session, Patron, id=patron_id)
             collection = get_one(session, Collection, id=collection_id)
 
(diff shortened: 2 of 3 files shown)


src/palace/manager/api/circulation.py
```diff
@@ -822,8 +822,8 @@ def delete_loans_or_holds(self, loans_or_holds: Iterable[LoanOrHoldT]) -> None:
                 # This was just created, we shouldn't delete it.
                 continue
             self.log.info(
-                "Deleting %s (%r) for patron (%s)"
-                % (item.__class__.__name__, item, item.patron.authorization_identifier)
+                f"Deleting {item.__class__.__name__} ({item.id}) on license pool (id: {item.license_pool_id}) "
+                f"for patron ({item.patron.authorization_identifier})"
```
jonathangreen, line 826: This is a small change to logging that I made while troubleshooting this. Previously were we printing the `__repr__` for the `Loan` or `Hold` object. Since neither of them defines `__repr__`, we were just getting something like `Deleting Loan (<palace.manager.sqlalchemy.model.patron.Loan object at 0x7fa324d5e9e0>) for patron`, which is pretty useless for debugging. The updated log line prints the LP ID, so its a bit easier to track down what is happening.

src/palace/manager/celery/tasks/patron_activity.py
```diff
@@ -37,7 +37,7 @@ def sync_patron_activity(
             )
             return
 
-        with task.session() as session:
+        with task.transaction() as session:
```
jonathangreen, line 40: This is the actual bugfix.

tests/manager/celery/tasks/test_patron_activity.py
```diff
@@ -162,7 +163,37 @@ def test_not_supported(
...
+            HoldInfo(
+                collection=sync_task_fixture.collection,
+                data_source_name=hold_pool.data_source,
+                identifier_type=hold_pool.identifier.type,
+                identifier=hold_pool.identifier.identifier,
+                hold_position=1,
+                start_date=None,
+                end_date=None,
+            )
+        )
```
jonathangreen, line 192: The test now adds a remote loan and hold, and makes sure they are synced.

tdilauro (approved):
Looks good to me! 🛠️

PR #1936, opened 2024-07-17
Update suppressed books feed. (PP-1462)

tdilauro:
## Description

Changes the feed of suppressed books to include only books suppressed at the library level for the designated library.

## Motivation and Context

Since we've changed book un/suppression to work on a per-library basis, the feed of suppressed books needs to work that way, too.

[Jira [PP-1462](https://ebce-lyrasis.atlassian.net/browse/PP-1462)]

## How Has This Been Tested?

- Functionality manually tested in local environment.
- Tests passing in local environment.
- CI tests for associated branch pass.

## Checklist

[PP-1462]: https://ebce-lyrasis.atlassian.net/browse/PP-1462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/feed/admin.py
+++ b/src/palace/manager/feed/admin.py
@@ -25,8 +25,8 @@ class AdminFeed(OPDSAcquisitionFeed):
         _pagination = pagination or Pagination.default()
 
         q = (
-            _db.query(LicensePool)
-            .join(Work)
+            _db.query(Work)
+            .join(LicensePool)
             .join(Edition)
             .filter(
                 and_(
@@ -35,11 +35,10 @@ class AdminFeed(OPDSAcquisitionFeed):
                     Work.suppressed_for.contains(library),
                 )
             )
-            .order_by(Edition.title)
+            .order_by(Edition.sort_title)
         )
-        pools = _pagination.modify_database_query(_db, q).all()
+        works = _pagination.modify_database_query(_db, q).all()
 
-        works = [pool.work for pool in pools]
         feed = cls(title, url, works, annotator, pagination=_pagination)
         feed.generate_feed()
 
@@ -65,8 +64,7 @@ class AdminFeed(OPDSAcquisitionFeed):
                 rel="first",
             )
 
-        previous_page = _pagination.previous_page
-        if previous_page:
+        if previous_page := _pagination.previous_page:
             feed.add_link(
                 annotator.suppressed_url(previous_page),
                 rel="previous",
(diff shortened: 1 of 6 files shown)


src/palace/manager/feed/admin.py
```diff
@@ -22,13 +26,16 @@ def suppressed(
...
         q = (
             _db.query(LicensePool)
+            .join(Work)
+            .join(Edition)
             .filter(
                 and_(
-                    LicensePool.suppressed == True,
-                    LicensePool.superceded == False,
+                    LicensePool.suppressed == false(),
+                    LicensePool.superceded == false(),
```
jonathangreen, line 34: @tdilauro is there a reason to use `false()` vs `False` here? I've not seen the `false()` function used in a comparison like this before.

src/palace/manager/feed/admin.py
```diff
@@ -22,13 +26,16 @@ def suppressed(
 
         q = (
             _db.query(LicensePool)
```
jonathangreen, line 28: Minor: Does this query actually need to join against `LicensePool` anymore? It seems like we're only using the pool to get the `Work`.

jonathangreen (approved):
Couple minor questions, but it looks good to me.

src/palace/manager/feed/admin.py
tdilauro: It's just a way to avoid PEP 8 complaints about comparison (`==` vs. `is`) to False. It reduces the number of squiggles in my IDE. 😅

I've used it in the past for, e.g., [dashboard stats queries](https://github.com/ThePalaceProject/circulation/blob/8cfd6599b0cc6f0b9e24928958ee7890f2806ee9/src/palace/manager/api/admin/dashboard_stats.py#L35-L44).

jonathangreen: Cool! TIL 😄

tdilauro: Nice to return the favor every once in a blue moon!

src/palace/manager/feed/admin.py
tdilauro: We need the license pool for the filter conditions. But I could probably return the works directly and at least avoid the additional assignment to get them.

tdilauro: Updated.

PR #1946, opened 2024-07-24
Rename classes for OPDS2 + ODL integration

jonathangreen:
## Description

All of our classes for OPDS2 + ODL integration are named ODL2 right now, which is somewhat confusing, since ODL2 isn't really a thing. This renames all the classes and fixtures etc to be more in line with what they are actually processing. 

The one thing that isn't renamed is the label on the `OPDS2WithODLApi` class. It would be nice to rename this label at some point, but because the protocol for the class is named after the label and this protocol is stored in the database, some care has to be taken renaming this, so I'd like to leave that for another time.

All the changes here were done with the rename refactor action in Pycharm, no logic has been updated as part of this PR.

## Motivation and Context

This seemed reasonably easy to do after the work in https://github.com/ThePalaceProject/circulation/pull/1944, and it seems desirable to have these classes better named. 

## How Has This Been Tested?

- Running unit tests

## Checklist


tdilauro (approved):
Looks good. Nice to see this naming change! 🚀

PR #1948, opened 2024-07-24
[PP-1491] Introduce retries to prevent ObjectDeletedError

dbernstein:
## Description
This update attempts to solve the problem of occasional monitor failures due to stale data.  I believe the problem arises when two different scripts are updating rows at the same time.   Because the SweepMonitor processes updates in batches of 50-100 books,  a change can be made by one of the circulation monitors to a database resource that is in a member of a batch that has been retrieved from the database but not yet modified.  This can results in both a StaleDataError (not observed but possible) or an ObjectDeletedError (observed).

The following monitors that were showing the behavior:  OverdriveFormatSweep, AxisCollectionReaper, and the Axis360CirculationMonitor.

This update adds retry functionality to the SweepMonitor (and all of its subclasses not overriding `process_batch`).
The retrier will try a maximum of 10 times, each time exponentially backing off until the sleep time hits 60s.

Similarly I added retry functionality to the Axis360CirculationMonitor with the same retry parameters to the `process_book` and removed the batched `commit` calls since batch the commits complicate the retry logic.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1491

## Checklist

### DIFF
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -780,7 +780,6 @@ class Axis360CirculationMonitor(CollectionMonitor, TimelineMonitor):
         count = 0
         for bibliographic, circulation in self.api.recent_activity(start):
             self.process_book(bibliographic, circulation)
-            self._db.commit()
             count += 1
         progress.achievements = "Modified titles: %d." % count
 
@@ -796,6 +795,14 @@ class Axis360CirculationMonitor(CollectionMonitor, TimelineMonitor):
     def process_book(
         self, bibliographic: Metadata, circulation: CirculationData
     ) -> tuple[Edition, LicensePool]:
+        if self._db.dirty:
+            # If an update failed on a previous attempt due to stale or missing data, the session should be dirty.
+            # Therefore, we should roll it back before trying again.
+            # I've put the rollback here rather than in a @retry before_sleep hook because I could neither
+            # figure out how to cleanly reference the db session from within class or module  level method on
+            # the one hand, nor pass an instance method reference to the hook.
+            self._db.rollback()
+
         edition, new_edition, license_pool, new_license_pool = self.api.update_book(
             bibliographic, circulation
         )
@@ -807,6 +814,7 @@ class Axis360CirculationMonitor(CollectionMonitor, TimelineMonitor):
             self.bibliographic_coverage_provider.handle_success(identifier)
             self.bibliographic_coverage_provider.add_coverage_record_for(identifier)
 
+        self._db.commit()
         return edition, license_pool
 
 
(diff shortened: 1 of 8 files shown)


src/palace/manager/api/axis.py
```diff
@@ -776,13 +781,28 @@ def catch_up_from(
...
+            | retry_if_exception_type(ObjectDeletedError)
+        ),
+        stop=stop_after_attempt(MAX_RETRIES),
+        wait=wait_exponential(multiplier=1, min=1, max=60),
+        reraise=True,
+    )
     def process_book(
         self, bibliographic: Metadata, circulation: CirculationData
     ) -> tuple[Edition, LicensePool]:
+        if self._db.dirty:
```
dbernstein, line 798: I don't love this check here, but I couldn't think of a cleaner way to do it and still use `tenacity` which is a nice library.  Any suggestions here would be appreciated.

jonathangreen: I don't think this catches all the conditions you want to catch here... dirty just means that changes haven't been flushed, not that there are changes within the transaction. 

Does it make more sense to take a similar approach to overdrive and not bring in the `tenacity` library here?

src/palace/manager/core/monitor.py
```diff
@@ -505,12 +514,31 @@ def run_once(self, *ignore):
...
 
+    @retry(
+        retry=(
+            retry_if_exception_type(StaleDataError)
+            | retry_if_exception_type(ObjectDeletedError)
+        ),
+        stop=stop_after_attempt(MAXIMUM_BATCH_RETRIES),
+        wait=wait_exponential(multiplier=1, min=1, max=60),
+        reraise=True,
+    )
```
jonathangreen, line 525: Why add this code to both the monitor case class and to the axis monitor directly?

dbernstein:
@jonathangreen :  I'm going to rework the PR not to use tenacity as you suggested.  I liked tenacity because it seemed like it would be cleaner but it is more complicated.

src/palace/manager/core/monitor.py
dbernstein: Not sure I understand the question.   If I understand it the reason was that they don't share the same "process" methods.  SweepMonitor and Axis360CirculationMonitor share a common ancestor (CollectionMonitor) but the processing is handled by subclasses.

dbernstein:
@jonathangreen :  I decided to stick with tenacity (esp because it offers nice methods for doing exponential backoff and I think it makes the code easier to read).  I was able to get around the issue I was running into by using a nested transaction.  I believe this solution will perform more or less equally well with what came before.  The only dramatic increase in commits will come with the Axis360CirculationMonitor since commits where being batched 50 at a time.  I don't think this will result in significant performance degradation because almost all of the latency with Axis appears to be due to http communications with the service (about 2-3 per book).   The SweeperMonitor commits are still being batched in groups of 100 and the OverdriveCirculationMonitor still does one commit per book.

jonathangreen (approved):
Looks good

PR #1954, opened 2024-07-29
[PP-1511] Update Palace Marketplace Feed domain.

dbernstein:
## Description

This PR includes a database migration that replaces all integration_configuration.settings['external_account_id'] values containing "market.feedbooks.com" with "market.thepalaceproject.org"
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1511

## How Has This Been Tested?
I confirmed that ./docker/ci/test_migrations.sh passed locally.

## Checklist


tdilauro:
@dbernstein I think we've decided to do this via local db changes, rather than a project migration to avoid affecting any non-Palace instances.

If so, we should probably close this one.

dbernstein:
Closing:  changes applied via ansible script.

PR #1955, opened 2024-07-30
Fix flaky test comparison

jonathangreen:
## Description

It looks like `main` has a failed test right now, due to the comparison order in `test_sync_patron_activity_creates_local_loans` being non-deterministic. This PR updates the test to use a set instead, so that order doesn't matter.

See this test run: https://github.com/ThePalaceProject/circulation/actions/runs/10151142851/job/28069801915

## Motivation and Context

I was looking into why the build failed for `main` and noticed this.

## How Has This Been Tested?

- Ran test locally

## Checklist


tdilauro (approved):
🐛🦶🏽

PR #1962, opened 2024-07-30
[PP-1551] Ensure EPUB is always first preview link when present as an…

dbernstein:
… option.

## Description
This update ensures that the sample links are sorted such that EPUB, if present, will appear at the top of the list.

If the media type contains `/epub` the link will appear first in the list of sample urls. Otherwise the media type will be sorted alphabetically.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1551

## How Has This Been Tested?
Two new unit tests added.

## Checklist

### DIFF
--- a/src/palace/manager/feed/annotator/base.py
+++ b/src/palace/manager/feed/annotator/base.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 from collections import defaultdict
 from decimal import Decimal
+from functools import cmp_to_key
 from typing import Any
 from urllib.parse import quote
 
@@ -243,6 +244,21 @@ class ToFeedEntry:
 
 
 class Annotator(ToFeedEntry):
+    @classmethod
+    def _sample_link_comparator(cls, link1: Link, link2: Link):
+        link_1_type = link1.type.lower()
+        link_2_type = link2.type.lower()
+        if "/epub" in link_1_type and "/epub" not in link_2_type:
+            return -1
+        elif "/epub" in link_2_type and "/epub" not in link_1_type:
+            return 1
+        elif link_1_type > link_2_type:
+            return 1
+        elif link_1_type < link_2_type:
+            return -1
+        else:
+            return 0
+
     def annotate_work_entry(
         self, entry: WorkEntry, updated: datetime.datetime | None = None
     ) -> None:
@@ -284,6 +300,8 @@ class Annotator(ToFeedEntry):
                     type=sample.resource.representation.media_type,
                 )
             )
+
+        other_links.sort(key=cmp_to_key(self._sample_link_comparator))
 
         if edition.medium:
             additional_type = Edition.medium_to_additional_type.get(str(edition.medium))
--- a/src/palace/manager/feed/annotator/base.py
+++ b/src/palace/manager/feed/annotator/base.py
@@ -258,17 +258,6 @@ class Annotator(ToFeedEntry):
             return -1
         elif link_1_type != epub_type and link_2_type == epub_type:
             return 1
-        elif link_1_type > link_2_type:
-            return 1
-        elif link_1_type < link_2_type:
-            return -1
-        elif link1.href and link2.href:
-            if link1.href > link2.href:
-                return 1
-            elif link1.href < link2.href:
-                return -1
-            else:
-                return 0
         else:
             return 0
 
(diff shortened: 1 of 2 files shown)


src/palace/manager/feed/annotator/base.py
```diff
@@ -243,6 +244,21 @@ def content(cls, work: Work | None) -> str:
 
 
 class Annotator(ToFeedEntry):
+    @classmethod
+    def _sample_link_comparator(cls, link1: Link, link2: Link):
```
tdilauro, line 248: Since `cls` isn't used, this could be a `@staticmethod` or pulled out of the class as a function.

src/palace/manager/feed/annotator/base.py
```diff
@@ -243,6 +244,21 @@ def content(cls, work: Work | None) -> str:
...
+        link_1_type = link1.type.lower()
+        link_2_type = link2.type.lower()
+        if "/epub" in link_1_type and "/epub" not in link_2_type:
```
tdilauro, line 251: Could we use the type constants, rather than strings, for this?

This might also obviate the need for `...type.lower()` and it's associated type checking errors.

src/palace/manager/feed/annotator/base.py
tdilauro: Also, we need a return type annotation.

tests/manager/feed/test_annotators.py
```diff
@@ -409,6 +410,32 @@ def test_annotate_work_entry(self, db: DatabaseTransactionFixture):
...
+                data_source=edition.data_source,
+                rel=Hyperlink.SAMPLE,
+                resource=resource,
+            )
+            return link
+
+        create_sample_link(MediaTypes.PDF_MEDIA_TYPE, "http://pdf")
+        create_sample_link(MediaTypes.EPUB_MEDIA_TYPE, "http://epub")
+        create_sample_link(MediaTypes.TEXT_HTML_MEDIA_TYPE, "http://html")
+        create_sample_link(MediaTypes.APPLICATION_JSON_MEDIA_TYPE, "http://json")
```
tdilauro, line 437: It might be useful to include multiple `epub` and multiple of some other type to completely test the comparitor. In that case, the assertions below might need to check the `href`s, as well, to ensure that we get the order we're expecting.

tests/manager/feed/test_annotators.py
```diff
@@ -694,3 +728,26 @@ def __init__(self, last_update):
         result._hit = hit
         entry = entry_for(result)
         assert "2017-01-01" in entry.get("updated")
+
+    def test_sample_link_sort(self):
+        epub_link = Link(rel=None, href=None, type="application/epub+zip")
+        html_link = Link(rel=None, href=None, type="text/html")
+        pdf_link = Link(rel=None, href=None, type="application/pdf")
+        kepub_link = Link(rel=None, href=None, type="z-application/kepub+zip")
```
tdilauro, line 736: Similar comment here re:  multiples of some types to fully test commparitor.

tdilauro (commented):
A few comments below for consideration.

tests/manager/feed/test_annotators.py
```diff
@@ -455,6 +483,16 @@ def test_annotate_work_entry(self, db: DatabaseTransactionFixture):
         assert data.language == None
         assert data.updated == FeedEntryType(text=strftime(now))
 
+        # other links
+        other_links = data.other_links
+        assert other_links[0].type == MediaTypes.EPUB_MEDIA_TYPE
+        assert other_links[0].href.startswith("http://epub-a")
+        assert other_links[1].type == MediaTypes.EPUB_MEDIA_TYPE
+        assert other_links[1].href.startswith("http://epub-b")
```
tdilauro, line 491: The `href`s here have type `str | None`, so we need to handle the None case. A couple ways to do that:
```python
assert other_links[0].href is not None
assert other_links[0].href.startswith("http://epub-a")
```
or 
```python
assert other_links[0].href and other_links[0].href.startswith("http://epub-a")
```

tests/manager/feed/test_annotators.py
tdilauro: Revising this comment, based on my recent comment about not needing to do anything except sort the epubs to the top. I don't think we need to have more than one of any type except epub, so I think this one is resolved.

tdilauro (approved):
This one is pretty close. 🎸🤘🏽

A couple more suggestions below.

dbernstein:
Okay - I think we're there now:   "Guards! Make sure the Prince doesn't leave this room..." :)

PR #1966, opened 2024-08-05
[PP-1525] Add support for bypassing user blocks on Sirsidynix Authent…

dbernstein:
…ication Provider.

## Description

This update adds a setting to the Sirsidynix Horizon Authentication provider for ignoring user blocks due to any block other than an expired card.  By default blocks are respected.

When blocks are ignored, we are skipping the call for patron block information. 

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1525

## How Has This Been Tested?
I added three unit tests and verified manually.

## Checklist

### DIFF
--- a/src/palace/manager/api/sirsidynix_authentication_provider.py
+++ b/src/palace/manager/api/sirsidynix_authentication_provider.py
@@ -66,7 +66,6 @@ class SirsiDynixHorizonAuthSettings(BasicAuthProviderSettings):
             type=ConfigurationFormItemType.SELECT,
             options={"true": "Yes, block.", "false": "No, do not block."},
         ),
-        alias="patron block status",
     )
 
 
--- a/src/palace/manager/api/sirsidynix_authentication_provider.py
+++ b/src/palace/manager/api/sirsidynix_authentication_provider.py
@@ -165,7 +165,7 @@ class SirsiDynixHorizonAuthenticationProvider(
         self.sirsi_disallowed_suffixes = library_settings.library_disallowed_suffixes
         self.sirsi_library_id = library_settings.library_id
 
-        # Check if patrons should be blocked based on SIP status
+        # Check if patrons should be blocked based on ILS status
         self.patron_status_should_block = settings.patron_status_block
 
     def remote_authenticate(
--- a/src/palace/manager/api/sirsidynix_authentication_provider.py
+++ b/src/palace/manager/api/sirsidynix_authentication_provider.py
@@ -167,10 +167,7 @@ class SirsiDynixHorizonAuthenticationProvider(
         self.sirsi_library_id = library_settings.library_id
 
         # Check if patrons should be blocked based on SIP status
-        if settings.patron_status_block:
-            self.patron_status_should_block = True
-        else:
-            self.patron_status_should_block = False
+        self.patron_status_should_block = settings.patron_status_block
 
     def remote_authenticate(
         self, username: str | None, password: str | None
--- a/src/palace/manager/api/sirsidynix_authentication_provider.py
+++ b/src/palace/manager/api/sirsidynix_authentication_provider.py
@@ -56,6 +56,19 @@ class SirsiDynixHorizonAuthSettings(BasicAuthProviderSettings):
         alias="CLIENT_ID",
     )
 
+    patron_status_block: bool = FormField(
+        True,
+        form=ConfigurationFormItem(
+            label="Patron Status Block",
+            description=(
+                "Block patrons from borrowing based on the status of the ILS's patron block field?"
+            ),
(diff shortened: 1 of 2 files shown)


tests/manager/api/test_sirsidynix_auth_provider.py
```diff
@@ -203,6 +216,64 @@ def test_remote_patron_lookup(self, sirsi_auth_fixture: SirsiAuthFixture):
         assert patrondata.block_reason == PatronData.NO_VALUE
         assert patrondata.library_identifier == "testtype"
 
+    @pytest.mark.parametrize(
```
dbernstein, line 219: @jonathangreen or @tdilauro :  I used the pytest.mark.parametrize decorator here in order to send optional parameters to the sirsi_auth_fixture test fixture.   I couldn't see a cleaner way to override the patron_status_block default that is set in the fixture.  It works, but I fear the readability is suboptimal.   I'd be curious if either of you have suggestions for accomplishing what I'm trying to do here in a more readable way.

dbernstein:
This PR I believe be ready is functional, though I haven't tested it with a test user from Vancouver yet (still waiting on them).  I just have one question about a couple of parameterized tests.

src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -153,6 +166,12 @@ def __init__(
         self.sirsi_disallowed_suffixes = library_settings.library_disallowed_suffixes
         self.sirsi_library_id = library_settings.library_id
 
+        # Check if patrons should be blocked based on SIP status
+        if settings.patron_status_block:
+            self.patron_status_should_block = True
+        else:
+            self.patron_status_should_block = False
```
jonathangreen, line 173: Minor: Since `patron_status_block` is already a `bool`, it probably makes sense to collapse this down to:

```suggestion
        # Check if patrons should be blocked based on SIP status
        self.patron_status_should_block = settings.patron_status_block:
```

tests/manager/api/test_sirsidynix_auth_provider.py
jonathangreen: I made some suggestions on how you can accomplish this. I think the only change you need to the fixture is to change @dbernstein I made some suggestions how you can accomplish this. I think all the changes you need to the fixture are to change the `provider_mocked_api` function.

src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -56,6 +56,19 @@ class SirsiDynixHorizonAuthSettings(BasicAuthProviderSettings):
...
+        True,
+        form=ConfigurationFormItem(
+            label="Patron Status Block",
+            description=(
+                "Block patrons from borrowing based on the status of the ILS's patron block field?"
+            ),
+            type=ConfigurationFormItemType.SELECT,
+            options={"true": "Yes, block.", "false": "No, do not block."},
+        ),
+        alias="patron block status",
```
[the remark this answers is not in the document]
dbernstein: I suppose not - I pulled this code from the sip authentication provider.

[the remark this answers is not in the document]
jonathangreen: That makes sense. We should probably clean up those aliases in SIP. The aliases were used when we migrated the configuration settings to load any that didn't have valid python names (spaces in them etc).

tests/manager/api/test_sirsidynix_auth_provider.py
```diff
@@ -77,8 +77,13 @@ def headers(self, api: SirsiDynixHorizonAuthenticationProvider) -> dict[str, str
             "x-sirs-clientID": api.sirsi_client_id,
         }
 
-    def provider_mocked_api(self) -> MockedSirsiApi:
-        provider = self.provider()
+    def provider_mocked_api(
+        self,
+        provider: SirsiDynixHorizonAuthenticationProvider | None = None,
+        patron_status_info: dict | None = None,
```
jonathangreen, line 83: Minor: Would be good to fill in the generic info for the `dict` type hint here, probably `dict[str, Any]`.

jonathangreen (approved):
One minor comment, but this looks good!

src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -153,6 +165,9 @@ def __init__(
         self.sirsi_disallowed_suffixes = library_settings.library_disallowed_suffixes
         self.sirsi_library_id = library_settings.library_id
 
+        # Check if patrons should be blocked based on SIP status
```
tdilauro, line 168: Minor: Since this isn't a SIP integration, maybe rephrase to "...based on their status from ILS."

PR #1974, opened 2024-08-07
[PP-1377] Ensure all eligible collections are displayed in report pre…

dbernstein:
…view.
## Description
This update addresses the bug that was causing the list of collections to be included in the inventory report to fail when the library in question as associated with at least one overdrive advantage account.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1377

## How Has This Been Tested?
I tested it manually and added a new test that covers the conditions required to reproduce the bug.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/generate_inventory_and_hold_reports.py
+++ b/src/palace/manager/celery/tasks/generate_inventory_and_hold_reports.py
@@ -36,7 +36,15 @@ def eligible_integrations(
     def is_eligible(integration: IntegrationConfiguration) -> bool:
         if integration.protocol is None:
             return False
-        settings = registry[integration.protocol].settings_load(integration)
+
+        if integration.collection.parent:
+            parent_integration = integration.collection.parent.integration_configuration
+        else:
+            parent_integration = None
+
+        settings = registry[integration.protocol].settings_load(
+            integration, parent_integration
+        )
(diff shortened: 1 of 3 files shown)


src/palace/manager/celery/tasks/generate_inventory_and_hold_reports.py
```diff
@@ -36,7 +36,15 @@ def eligible_integrations(
...
+
+        if integration.collection.parent:
+            parent_integration = integration.collection.parent.integration_configuration
+        else:
+            parent_integration = None
+
+        settings = registry[integration.protocol].settings_load(
+            integration, parent_integration
+        )
         return isinstance(settings, OPDSImporterSettings)
```
jonathangreen, line 48: I'd like to avoid adding an unused `parent` parameter to the `HasIntegrationConfiguration` class, since parent isn't really meaningful in that context.

Instead of that, I think, we can just make this a `issubclass` check, which is simpler and will be faster since it doesn't have to instantiate a settings object.

```suggestion
        settings_cls = registry[integration.protocol].settings_class()
        return issubclass(settings_cls, OPDSImporterSettings)
```

dbernstein:
@jonathangreen :  Thanks for the feedback.  I merged in your suggestions.

jonathangreen (approved):
Looks good!

PR #1977, opened 2024-08-07
Fix docker warning annotation

jonathangreen:
## Description

Recent PRs have had the following warning annotation on them: 

```
GitHub Actions / Push circ-exec

Legacy key/value format with whitespace separator should not be used

LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format
More info: https://docs.docker.com/go/dockerfile/rule/legacy-key-value-format/
```

This PR should fix that, so we're not seeing that annotation on every PR.

## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,7 +32,7 @@ RUN . env/bin/activate && \
 
 FROM common AS exec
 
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
👍🏽

PR #1978, opened 2024-08-07
Update how contributor roles are normalized via OPDS2 importer (PP-1530)

jonathangreen:
## Description

Updates the OPDS2 importer so that the roles it sets on contributors are normalized to be the same internal contributor roles that the rest of the importers use. Previously it was just directly using whatever was given as the role string. It also updates the importer to ignore contributors with a blank name set. This seems like a bug on the biblio side, but I think its also worth handling reasonably in our code.

The OPDS2 spec says that the [`contributor`](https://github.com/readium/webpub-manifest/tree/master/contexts/default#contributors) element:
> **should** be used with an object that contains a role. All values for the role element **should** be based on [MARC relator codes](https://www.loc.gov/marc/relators/relaterm.html)

Since its not a **must** I updated the parsing to take either: the MARC codes we support, or case insensitive versions of the role strings we use internally. If the role isn't found, it will log a warning and use the default.

**Note**: For review most of the changes here are just a big find and replace to replace all the roles we used to use with an enum, so its easier to iterate over them. The bulk of the changes for this PR are much smaller and are in this commit: https://github.com/ThePalaceProject/circulation/pull/1978/commits/3ea31cd31542ba0b8084e12a679038a6f54465c5 hopefully that will make it a bit easier to review.

## Motivation and Context

After importing the biblio OPDS2 feed, it looked generally good, but didn't have author information. Spending some time debugging, it looks like none of our other feeds use the generic `contributor` tag, with specific `role` set, rather they use the more specific `author`, `artist`, `editor` elements, which were already being normalized.

## How Has This Been Tested?

- Tested locally with biblio feeds

## Checklist


dbernstein (approved):
Looks good.

PR #1981, opened 2024-08-08
[PP-1556] Fix broken time tracking summary script.

dbernstein:
## Description

Since it is possible for multiple editions to be returned  when looking up an edition by primary identifier (ie when there are multiple collections with overlapping contents on the same CM), we should grab the first result rather than expecting one and only one result.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1556

## Checklist


jonathangreen (approved):
Looks good

jonathangreen:
Minor: Might be nice to add a test for this case, so we don't break it again when refactoring in the future.

PR #1982, opened 2024-08-08
Fix bug in HTTP._request_with_timeout when headers are None (PP-1582)

jonathangreen:
## Description

`HTTP._request_with_timeout` was failing when `None` was passed in as a headers value. The function was also modifying the passed in headers dict, which might be unexpected by the caller. This PR fixes both of those issues and adds tests for them. 

## Motivation and Context

The code in https://github.com/ThePalaceProject/circulation/pull/1960 was calling `HTTP._request_with_timeout` causing an exception.

## How Has This Been Tested?

- Tested locally
- Tested with unit tests

## Checklist

### DIFF
--- a/src/palace/manager/util/http.py
+++ b/src/palace/manager/util/http.py
@@ -261,29 +261,29 @@ class HTTP(LoggerMixin):
         )
         backoff_factor: float = float(kwargs.pop("backoff_factor", 1.0))
 
-        # Unicode data can't be sent over the wire. Convert it
-        # to UTF-8.
+        # Unicode data can't be sent over the wire. Convert it to UTF-8.
         if "data" in kwargs and isinstance(kwargs["data"], str):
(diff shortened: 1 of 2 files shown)


src/palace/manager/util/http.py
```diff
@@ -261,29 +261,29 @@ def _request_with_timeout(
...
-                v = v.encode("utf8")
-            new_headers[k] = v
-        kwargs["headers"] = new_headers
+        version = (
+            manager.__version__
+            if manager.__version__
+            else cls.DEFAULT_USER_AGENT_VERSION
+        )
+        ua_header = {"User-Agent": f"Palace Manager/{version}"}
+        headers = ua_header | (kwargs.get("headers") or {})
```
dbernstein, line 276: I love the compact elegance of this line.

dbernstein (approved):
🥇

PR #1987, opened 2024-08-12
[PP-1589] Always block expired cards for Sirsidynix Horizon regardles…

dbernstein:
…s of ignore blocks flag

## Description

This update ensures that when the sirsidynix api indicates a patron is expired, the user will be blocked from performing checkout operations even if patron blocks are ignored for the sirsi auth provider.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1589

## How Has This Been Tested?
New unit tests created.

## Checklist

### DIFF
--- a/src/palace/manager/api/sirsidynix_authentication_provider.py
+++ b/src/palace/manager/api/sirsidynix_authentication_provider.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import datetime
 import json
 import os
 from collections.abc import Callable, Generator
@@ -8,7 +7,6 @@ from gettext import gettext as _
 from typing import TYPE_CHECKING, Any, Literal
 from urllib.parse import urljoin
 
-import dateutil
 from pydantic import HttpUrl
 from sqlalchemy.orm import Session
 
@@ -63,8 +61,9 @@ class SirsiDynixHorizonAuthSettings(BasicAuthProviderSettings):
         form=ConfigurationFormItem(
             label="Enforce Patron ILS Blocks",
             description=(
-                "Block patrons from borrowing based on the status of the ILS's patron block field? "
-                "(Note: expired accounts are always blocked.)"
+                "Block patrons from borrowing based on the approved, hasMaxDaysWithFines, hasMaxFines, hasMaxLostItem, "
+                "hasMaxOverdueDays, hasMaxItemsCheckedOut fields from the ILS.(Note: expired accounts are always "
+                "blocked)."
             ),
             type=ConfigurationFormItemType.SELECT,
             options={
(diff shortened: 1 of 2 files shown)


src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -228,44 +230,53 @@ def remote_patron_lookup(
...
+
+        expires_date: str = status_fields.get("privilegeExpiresDate")
+        expires_date_obj = None
+
+        if expires_date:
+            expires_date_obj = dateutil.parser.parse(expires_date)
+        elif status_fields.get("expired"):
+            patrondata.block_reason = SirsiBlockReasons.EXPIRED
+        # Blockable statuses
+        if expires_date_obj and datetime.datetime.utcnow() >= expires_date_obj:
```
jonathangreen, line 256: This is assuming that the datetime we get is in the server local time, which I don't think we can safely assume here. Looking at the API responses, it doesn't look like it gives us any info about the timezone to expect, although we could reach out to sirsi and confirm this. 

So we don't have to deal with all this ugly timezone math, maybe we are better off just using the `expired` boolean here, rather then trying to determine if its expired ourselves.

src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -228,44 +230,53 @@ def remote_patron_lookup(
...
+        ):
+            patrondata.block_reason = PatronData.TOO_MANY_OVERDUE
+        elif status_fields.get("hasMaxItemsCheckedOut"):
+            patrondata.block_reason = PatronData.TOO_MANY_LOANS
+
+        # If previously, the patron was blocked this should unset the value in the DB
+        if patrondata.block_reason is None:
             patrondata.block_reason = PatronData.NO_VALUE
 
+        if not self.patron_status_should_block:
```
jonathangreen, line 275: This variable seems poorly named given the new changes.

jonathangreen (commented):
I think this one needs a couple changes before it goes in. 

1. I'm worried about the timezone math we are doing to compare the datetime not working reliably around the expiration date. I think we can avoid all that if we just rely on the 'expired' field that the ILS gives us. Since it presumably knows the timezone.
2. We need to update the options in the configuration to accurately reflect what they are doing, since we are no longer disabling all blocks.

dbernstein:
@jonathangreen : I see what I did re sirsi: I was looking at the wrong endpoint.  What threw me off is that both `api_read_patron_data` and `api_patron_status_info`  return the privilegeExpiresDate field.   I'm now ignoring that field and going off of the "expired" field alone.  

However since we get `privilegeExpiresDate` in the "read patron  I could use that for determining expiration status and if `blocks are not enforced` we could avoid making the `api_patron_status_info` call and perhaps making the interaction faster.   But then there is the issue of the time zone.   What do you think should I go off of the `privilegeExpiresDate` in the `api_read_patron_data` and avoid the call to `api_patron_status_info`  when blocks are not enforced or go off of "expired" flag which will require a call to `api_patron_status_info` regardless of the block status?

jonathangreen:
@dbernstein i think it makes sense to just go off the `expired` flag, even though it takes an extra request.

dbernstein:
@jonathangreen :  are we good to go then?

src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -56,15 +56,19 @@ class SirsiDynixHorizonAuthSettings(BasicAuthProviderSettings):
...
 
-    patron_status_block: bool = FormField(
+    patron_blocks_enforced: bool = FormField(
         True,
         form=ConfigurationFormItem(
-            label="Patron Status Block",
+            label="Enforce Patron ILS Blocks",
             description=(
-                "Block patrons from borrowing based on the status of the ILS's patron block field?"
+                "Block patrons from borrowing based on the status of the ILS's patron block field? "
```
jonathangreen, line 64: It would be nice to be really explicit here. I'd suggest something like:
> Block patrons from borrowing based on the `approved`, `hasMaxDaysWithFines`, `hasMaxFines`, `hasMaxLostItem`, `hasMaxOverdueDays`, `hasMaxItemsCheckedOut` fields from the ILS. (Note: expired accounts are always blocked).

jonathangreen (approved):
One comment about the description of the setting.

PR #1995, opened 2024-08-19
Refactor s3 multipart uploads (PP-1472)

jonathangreen:
## Description

Refactor multipart s3 uploads, so that the context manager is just a wrapper around logic in the main S3Service class. 

## Motivation and Context

In PP-1472, I need to do multipart s3 uploads outside of a context manager. The way this functionality was previously implemented this was impossible. This PR just moves the logic around so the functions are accessible outside of a context manager.

## How Has This Been Tested?

- Running units tests

## Checklist

### DIFF
--- a/src/palace/manager/service/storage/s3.py
+++ b/src/palace/manager/service/storage/s3.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import dataclasses
 import sys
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 🚀

PR #1999, opened 2024-08-20
Ensure that borrowing privileges are allowed when patron.block_reason…

dbernstein:
… == PatronData.NO_VALUE

## Description

There was some privilege checks outside for the sirsidynix horizon auth provider that I missed on the last round of changes.
This update ensures that `patron.block_reason = NO_VALUE` is  understood to mean do not block even if there are fines.

## Motivation and Context

Follow on ticket for https://ebce-lyrasis.atlassian.net/browse/PP-1589

## Checklist

### DIFF
--- a/src/palace/manager/api/util/patron.py
+++ b/src/palace/manager/api/util/patron.py
@@ -3,6 +3,7 @@ import datetime
 import dateutil
 from money import Money
 
+from palace.manager.api.authentication.base import PatronData
 from palace.manager.api.circulation_exceptions import (
     AuthorizationBlocked,
     AuthorizationExpired,
@@ -69,26 +70,20 @@ class PatronUtility:
         :raises OutstandingFines: If the patron has too many outstanding fines.
 
         """
+        if patron.block_reason == PatronData.NO_VALUE:
+            return
+
         if not cls.authorization_is_active(patron):
             # The patron's card has expired.
             raise AuthorizationExpired()
 
-        if cls.has_excess_fines(patron):
+        if patron.block_reason is PatronData.EXCESSIVE_FINES or cls.has_excess_fines(
+            patron
+        ):
             raise OutstandingFines(fines=patron.fines)
-
-        from palace.manager.api.authentication.base import PatronData
 
         if patron.block_reason is None:
             return
-
-        if patron.block_reason == PatronData.NO_VALUE:
-            return
-
-        if patron.block_reason is PatronData.EXCESSIVE_FINES:
-            # The authentication mechanism itself may know that
-            # the patron has outstanding fines, even if the circulation
-            # manager is not configured to make that deduction.
-            raise OutstandingFines(fines=patron.fines)
 
         raise AuthorizationBlocked()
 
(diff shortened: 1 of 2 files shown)


src/palace/manager/api/util/patron.py
```diff
@@ -69,27 +70,21 @@ def assert_borrowing_privileges(cls, patron):
         :raises OutstandingFines: If the patron has too many outstanding fines.
 
         """
+        if patron.block_reason == PatronData.NO_VALUE:
+            return
+
```
tdilauro, line 75: It's not obvious to me why it is okay to move this ahead of checking whether the patron's card has expired. It seems unlikely that a patron with an expired card would have borrowing privileges.

If this case is actually allowed for some reason, it would be good to both (1) add a test with this scenario, and (2) update the docstring with some explanatory text.

tdilauro (commented):
One comment below...

jonathangreen (changes requested):
I don't think its safe to assume that `PatronData.NO_VALUE` for blocks means that we shouldn't block on fines. 

We have a separate setting  `max_outstanding_fines` https://github.com/ThePalaceProject/circulation/blob/b1ab61ac9a9506d4ccad11efaeb4df5e3998a84b/src/palace/manager/integration/configuration/library.py#L436-L443 that this change would break. Especially for ILS that don't send block information, but do send fine information. 

I think what you are trying to accomplished can be done by setting `max_outstanding_fines` to `None` (no value in the config form), without any code changes. 

Am I understanding what you are trying to accomplish here correctly? If not can you give some more context for this change so I can understand?

dbernstein:
@jonathangreen and @tdilauro :  The context here is that Vancouver uses SirsiDynixHorizon which supports the ability to ignore blocks on non-expired cards.  When that setting is enabled we are setting the patron.block_reason to PatronData.NO_VALUE when we want the circulation manager to ignore block reasons.   So if there are excessive fines, we just leave them be and set the block_reason to NO_VALUE. So when the code reaches the borrow check we are looking only at the fine info and not taking into account that upstream we overwrote the block reason to avoid being blocked on excessive fines. 

So yes - I believe you're right @jonathangreen  that setting `max_outstanding_fines` to None should accomplish what we want there, but we'll have to make sure to have the user both ignore blocks AND set max fines to zero.  In other words it's not clear that the ignore blocks will be overridden by a `max_outstanding_fines` value being exceeded.

dbernstein:
I see now what you are suggesting  is also how SIP2 works - ie max_outstanding_fines is set at the library level.   So maybe I can close this PR.

jonathangreen:
@dbernstein what we currently do in the sirsi provider is the same thing we do in the SIP provider. 
https://github.com/ThePalaceProject/circulation/blob/e384f764882447d19550eacff78fad0fe4db0b75/src/palace/manager/api/sip/__init__.py#L452-L475

We set the blocks based on the configuration, and set the fines as well, with the fines being enforced later. To me this makes sense. I think it would be more confusing if setting `Enforce Patron ILS Blocks` to `false` in the patron auth settings would also stop enforcing the max fines setting in the library configuration. @dwilcox might have a opinion on this one, since its question about admin facing settings.

dbernstein:
I was confused.  This PR does not need to be.

PR #2002, opened 2024-08-21
[PP-1456] Add loan count column to audio book playback time reports.

dbernstein:
## Description

As the title indicates, this PR adds a loan count column to the audio book playback time reports.

When playback time data comes in through the API,  the code attempts to resolve the loan associated with the audio book and patron.  If the loan cannot be found (ie because it was purged) we use the user's patron id.  Then we create a sha1 of the loan or patron ID in order to uniquely identify the loan.

Then on the reporting side, we perform a count of the unique loans within the time range and join that count to the seconds summation query.

The default value for the loan_identifier is an empty string.  Since we don't have a way of associating loans or patrons with existing playtime records, I don't see a good way of counting old playtime entries and summaries.  Once the start date of the reporting window passes the date of the release of this update to production the loan counting will be correct.
Before that, it will be one loan per unique isbn, collection, library.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1456

## How Has This Been Tested?
Updated the existing tests and added addition assertions.

## Checklist

### DIFF
--- a/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
+++ b/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
@@ -32,7 +32,7 @@ def upgrade() -> None:
     )
 
     op.create_unique_constraint(
-        "unique_playtime_entry_with_loan_identifier",
+        "unique_playtime_entry",
         "playtime_entries",
         [
             "tracking_id",
@@ -55,7 +55,7 @@ def upgrade() -> None:
     )
 
     op.create_unique_constraint(
-        "unique_playtime_summary_with_loan_identifier",
+        "unique_playtime_summary",
         "playtime_summaries",
         [
             "timestamp",
@@ -68,11 +68,8 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    session = Session(bind=op.get_bind())
-    conn = session.connection()
-
     op.drop_constraint(
-        "unique_playtime_entry_with_loan_identifier",
+        "unique_playtime_entry",
         "playtime_entries",
         type_="unique",
     )
@@ -92,7 +89,7 @@ def downgrade() -> None:
     )
 
     op.drop_constraint(
-        "unique_playtime_summary_with_loan_identifier",
+        "unique_playtime_summary",
         "playtime_summaries",
         type_="unique",
     )
--- a/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
+++ b/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
@@ -30,7 +30,30 @@ def upgrade() -> None:
         sa.Column("loan_identifier", sa.String(length=40), nullable=False, default=""),
     )
 
+    op.drop_index("unique_playtime_summary", "playtime_summaries")
+
+    op.create_unique_constraint(
+        "unique_playtime_summary",
+        "playtime_summaries",
+        [
+            "timestamp",
+            "identifier_str",
+            "collection_name",
+            "library_name",
+            "loan_identifier",
+        ],
+    )
+
 
 def downgrade() -> None:
     op.drop_column("playtime_entries", "loan_identifier")
+
+    op.drop_index("unique_playtime_summary", "playtime_summaries")
+
     op.drop_column("playtime_summaries", "loan_identifier")
+
+    op.create_unique_constraint(
+        "unique_playtime_summary",
+        "playtime_summaries",
+        ["timestamp", "identifier_str", "collection_name", "library_name"],
+    )
--- a/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
+++ b/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
@@ -30,7 +30,7 @@ def upgrade() -> None:
         sa.Column("loan_identifier", sa.String(length=40), nullable=False, default=""),
     )
 
-    op.drop_index("unique_playtime_summary", "playtime_summaries")
+    op.drop_constraint("unique_playtime_summary", "playtime_summaries", type_="unique")
 
     op.create_unique_constraint(
         "unique_playtime_summary",
@@ -48,7 +48,7 @@ def upgrade() -> None:
 def downgrade() -> None:
     op.drop_column("playtime_entries", "loan_identifier")
 
-    op.drop_index("unique_playtime_summary", "playtime_summaries")
+    op.drop_constraint("unique_playtime_summary", "playtime_summaries", type_="unique")
 
     op.drop_column("playtime_summaries", "loan_identifier")
 
--- a/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
+++ b/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
@@ -1,0 +1,102 @@
+"""Add loan_identifier column to playtime tables.
+
+Revision ID: 7a2fcaac8b63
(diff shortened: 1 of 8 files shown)


alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+    op.drop_column("playtime_entries", "loan_identifier")
+
+    op.create_unique_constraint(
+        "unique_playtime_entry",
+        "playtime_entries",
+        [
+            "tracking_id",
+            "timestamp",
```
tdilauro, line 84: The appearance of `timestamp` in this downgrade is unexpected, since the previous index did not include it...
```
__table_args__ = (
        UniqueConstraint(
            "tracking_id",
            "identifier_str",
            "collection_name",
            "library_name",
            name="unique_playtime_entry",
        ),
    )```

File: src/palace/manager/sqlalchemy/model/time_tracking.py [hunk 2]
```diff
@@ -74,12 +74,15 @@ class PlaytimeEntry(Base):
...
 
+    loan_identifier = Column(String(50), nullable=False)
+
     __table_args__ = (
         UniqueConstraint(
             "tracking_id",
             "identifier_str",
             "collection_name",
             "library_name",
+            "loan_identifier",
```
tdilauro, line 85: I'm not entirely sure that we need to update the constraint for playtime_entries, since a given patron shouldn't be "reading" the same book both with and without a loan_identifier within the already-existing constraints (`tracking_id`/timestamp, `identifier_str`, etc.).

That said, there's probably no harm in it (other than increasing the size of the index, of course).

src/palace/manager/sqlalchemy/model/time_tracking.py
```diff
@@ -74,12 +74,15 @@ class PlaytimeEntry(Base):
     collection: Mapped[Collection] = relationship("Collection", uselist=False)
     library: Mapped[Library] = relationship("Library", uselist=False)
 
+    loan_identifier = Column(String(50), nullable=False)
```
tdilauro, line 77: Do we need 50 characters here? I don't think the value returned by `resolve_loan_identifier` should ever be more than 40.

src/palace/manager/scripts/playtime_entries.py
tdilauro, line 1: I am worried about including the isbn and title as grouping criteria when counting loans. Say, for example, get the isbn and/or title in some cases, but not in others, for a given loan. That could result in a single loan being counted as more than one loan.

This is a a bit of a thorny problem. I isn't an issue when summing up playback time, but it is when counting unique loans.

tdilauro (commented):
I have some concerns here about how isbn/title could affect loan counts. A few other concerns, as well. I didn't really get into the tests yet, since they might be affected by how we decide to deal with these other concerns.

src/palace/manager/api/controller/playtime_entries.py
```diff
@@ -16,9 +19,18 @@
...
+from palace.manager.sqlalchemy.model.licensing import LicensePool
+from palace.manager.sqlalchemy.model.patron import Loan, Patron
 from palace.manager.sqlalchemy.util import get_one
 
 
+def resolve_loan_identifier(patron: Patron, loan: Loan | None) -> str:
+    def sha1(msg):
+        return hashlib.sha1(msg.encode()).hexdigest()
+
+    return sha1(f"loan: {loan.id}") if loan else sha1(f"patron:{patron.id}")
```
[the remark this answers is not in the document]
dwilcox: I don't know if this helps so feel free to disregard, but we don't need to associate loans with patrons to satisfy this use case. It is sufficient to count the number of loans of the title that occurred during the reporting period, including instances where the same patron borrowed a title more than once.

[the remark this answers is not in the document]
dbernstein: @tdilauro :  the use of the patron_id is just a fallback in the very rare case that the loan is not present.  For example if the playtime update did not happen for a long long time and the loan data had since been purged.  Otherwise the loan ID will be used in the vast majority of cases.   

We could, as you suggested discard individual entries.  I"m okay with making that change.

alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
dbernstein: Very good catch.

src/palace/manager/sqlalchemy/model/time_tracking.py
dbernstein: Maybe we should leave it in there as a sanity check?  I don't know. 

I defer to your judgement on this one,  not only because I respect your opinion but also because you are the Playtime Tracking Man!!! :)

src/palace/manager/scripts/playtime_entries.py
dbernstein: @tdilauro  - I think I've addressed this issue.  I added some checks in one of the existing tests to confirm that it is working as expected.

src/palace/manager/api/controller/playtime_entries.py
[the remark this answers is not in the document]
dbernstein: I think it's better not to discard because we'll  lose some (if a small amount)  playtime and a few loans.  Or am I mistaken?

src/palace/manager/api/controller/playtime_entries.py
```diff
@@ -49,8 +62,37 @@ def track_playtimes(self, collection_id, identifier_type, identifier_idn):
...
+
+        loan = self._db.execute(
+            select(Loan)
+            .select_from(Loan)
+            .join(LicensePool)
+            .where(
+                LicensePool.identifier == identifier,
+                Loan.patron == flask.request.patron,
+                Loan.start >= min_time_entry,
+                Loan.start + default_loan_period < max_time_entry,
```
tdilauro, line 82: It doesn't look like we create any loans in our tests. If not, I'm not sure how the `loan` lookup gets tested.

tdilauro (commented):
👏🏽 Nice solution to the isbn and/or title mismatch issue. Additional comments below.

src/palace/manager/scripts/playtime_entries.py
```diff
@@ -214,33 +220,80 @@ def do_run(self):
...
+            )
+            .where(PlaytimeSummary.timestamp.between(start, until))
+            .group_by(
                 PlaytimeSummary.identifier_str,
                 PlaytimeSummary.collection_name,
                 PlaytimeSummary.library_name,
-                PlaytimeSummary.isbn,
-                PlaytimeSummary.title,
-                sum(PlaytimeSummary.total_seconds_played),
+                PlaytimeSummary.identifier_id,
```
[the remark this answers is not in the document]
dbernstein: (blush)

src/palace/manager/api/controller/playtime_entries.py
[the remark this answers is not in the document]
dbernstein: Good point - I didn't see the end field.

dbernstein: I'll fix that.

tests/manager/api/controller/test_playtime_entries.py
```diff
@@ -104,15 +106,16 @@ def test_track_playtime_duplicate_id_ok(
...
         # Attach the identifier to the collection
         pool = db.licensepool(
             db.edition(
                 identifier_type=identifier.type, identifier_id=identifier.identifier
             ),
             collection=collection,
         )
-        patron = db.patron()
+
+        loan_identifier = resolve_loan_identifier(loan=None)
```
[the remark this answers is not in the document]
dbernstein: Sure why not.

dbernstein:
@tdilauro :  I've made updates to the PR to address the issues you raised.

tdilauro (approved):
I think this one is ready to go!! 🚀

PR #2008, opened 2024-08-26
Prune old migrations

jonathangreen:
## Description

Similar to https://github.com/ThePalaceProject/circulation/pull/1540 remove some old migrations that we don't need to be dragging along anymore. 

This also removes some helper code and fixtures that were only used for these migrations.

In order to upgrade to whatever release this lands in, you will have to be running v17.0.0+, which shouldn't be a problem at this point, and won't impact us, because all of our environments are already running versions later then that. Its probably worth mentioning in our release notes this is the case though.

## Motivation and Context

Remove the old migrations and helper code for moving from configurationsettings to integration configurations, since we don't have any more of these migrations to do, and these tables have now been removed from our database schema for months.

I had to make some typing changes to these files for https://github.com/ThePalaceProject/circulation/pull/1963, which is what triggered me to write up this PR.

## How Has This Been Tested?

- Running tests locally

## Checklist


tdilauro (approved):
LGTM! 🔥

PR #2017, opened 2024-08-28
Convert MARC Export to use Celery (PP-1472)

jonathangreen:
## Checklist


README.md
```diff
@@ -149,7 +149,7 @@ grant all privileges on database circ to palace;
 Redis is used as the broker for Celery and the caching layer. You can run Redis with docker using the following command:
 
 ```sh
-docker run -d --name redis -p 6379:6379 redis
+docker run -d --name redis -p 6379:6379 redis/redis-stack-server
```
jonathangreen, line 152: Need to update the docker image we use for redis because the one maintained by docker doesn't have the json extension installed.

src/palace/manager/marc/annotator.py
```diff
@@ -580,244 +567,3 @@ def add_web_client_urls(
                     subfields=[Subfield(code="u", value=url)],
                 )
             )
-
-
-class MarcExporterSettings(BaseSettings):
```
jonathangreen, line 585: The settings classes still exist, they just got moved out to a different file.

src/palace/manager/api/admin/controller/catalog_services.py
```diff
@@ -21,7 +21,7 @@
 
 
 class CatalogServicesController(
-    IntegrationSettingsController[MARCExporter],
+    IntegrationSettingsController[MarcExporter],
```
jonathangreen, line 24: I'm partial to lower-casing acronyms when using CamelCase, giving `MarcExporter` instead of `MARCExporter`, so we don't end up with abominations like `MARCXMLExporter`. This change was unintentional in my refactoring however, so I'm open to the argument that it should be changed back to `MARCExporter`.

src/palace/manager/marc/settings.py
```diff
@@ -0,0 +1,73 @@
...
+    BaseSettings,
+    ConfigurationFormItem,
+    ConfigurationFormItemType,
+    FormField,
+)
+
+
+class MarcExporterSettings(BaseSettings):
+    # This setting (in days) controls how often MARC files should be
+    # automatically updated. Since the crontab in docker isn't easily
```
dbernstein, line 15: The comment should be updated to reflect that we're not using crontab anymore.

dbernstein (approved):
That is an absolute beast.  I like your batch + re-queueing approach.  I think that pattern will likely come in handy in other places.

src/palace/manager/marc/settings.py
jonathangreen: Good catch, thanks!

PR #2019, opened 2024-08-29
Add a script to clean up old marc file exports (PP-348)

jonathangreen:
## Description

Create a Celery task that runs once a day cleaning up old MARC exports stored in S3. The script leaves the most recent full export and the 12 most recent delta exports. It also cleans up any files for libraries and collections that have been deleted or had their MARC exporter disabled.

## Motivation and Context

The MARC export functionality creates records for libraries and lanes, but never removes any previous exports. So old, useless files are accumulating endlessly, using storage, increasing costs, and worsening user experience.

## How Has This Been Tested?

- Running tests

## Checklist


tdilauro (approved):
Looks good! 🚀

PR #2026, opened 2024-09-03
[PP-1650] Allows SirsiDynix Horizon Auth Provider users to reset Adob…

dbernstein:
…e IDs.

## Description

This update brings SirsiDynix Horizon Auth Provider in line with other  Auth Providers which do not require users to authenticate before resetting their Adobe IDs. 

When a CM admin tries to reset a patron's the Adobe ID, the CM makes an attempt to verify that the patron exists.   For SIP2 and Millenium Providers, this kind of unauthenticated query is supported   For Kansas, it is not supported so the KansasAuthenticationAPI just quietly passes the patron object back to the caller.   However for SirsiDynix Horizon an authenticated session is required to pull patron information.  

This update therefore allows a Sirsidynix Auth Provider caller to return an incomplete PatronData object rather than no result at all.  In this way, CM admins will be able to reset adobe ids for sirsidynix-using library patrons without needing the patrons credentials.

I moved this PR out of draft mode because I haven't yet heard back from SirsiDynix and I'm not holding my breath.  I doubt that their API supports what we ideally want to be able to do here: namely, to retrieve some minimal user info without having to authenticate.  If it turns out that we can,  I can always put in a new PR to take advantage of pulling patron info without authenticating.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1630

## How Has This Been Tested?
Tests have been updated and coverage improved.

## Checklist

### DIFF
--- a/src/palace/manager/api/sirsidynix_authentication_provider.py
+++ b/src/palace/manager/api/sirsidynix_authentication_provider.py
@@ -193,7 +193,7 @@ class SirsiDynixHorizonAuthenticationProvider(
 
     def remote_patron_lookup(
         self, patron_or_patrondata: Patron | PatronData
-    ) -> None | SirsiDynixPatronData | PatronData:
+    ) -> None | PatronData:
         """Do a remote patron lookup, this method can only look up a patron with a patrondata object
         with a session_token already setup within it.
         This method also checks all the reasons that a patron may be blocked for.
@@ -206,7 +206,7 @@ class SirsiDynixHorizonAuthenticationProvider(
         # if the patron data object is not authenticated just pass it back after ensuring the
         # complete flag is set to False
         if (
-            not hasattr(patron_or_patrondata, "session_token")
+            not isinstance(patron_or_patrondata, SirsiDynixPatronData)
             or patron_or_patrondata.session_token is None
         ):
             patron_or_patrondata.complete = False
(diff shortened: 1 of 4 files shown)


src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -194,22 +193,30 @@ def remote_authenticate(
 
     def remote_patron_lookup(
         self, patron_or_patrondata: Patron | PatronData
-    ) -> None | SirsiDynixPatronData:
+    ) -> None | SirsiDynixPatronData | PatronData:
```
jonathangreen, line 196: Since `SirsiDynixPatronData` is a `PatronData` this can be simplified to `None | PatronData`

src/palace/manager/api/sirsidynix_authentication_provider.py
```diff
@@ -194,22 +193,30 @@ def remote_authenticate(
...
+        # Return None if a patron object is passed in.
+        if isinstance(patron_or_patrondata, Patron):
             return None
 
+        # if the patron data object is not authenticated just pass it back after ensuring the
+        # complete flag is set to False
+        if (
+            not hasattr(patron_or_patrondata, "session_token")
+            or patron_or_patrondata.session_token is None
+        ):
```
jonathangreen, line 211: You can eliminate all the `# type: ignore` comments below by updating this to be:
```python
  if (
      not isinstance(patron_or_patrondata, SirsiDynixPatronData)
      or patron_or_patrondata.session_token is None
  ):
      patron_or_patrondata.complete = False
      return patron_or_patrondata
```

jonathangreen (approved):
Looks good. 

I made some minor comments about type checking that would be nice to fix up before this one gets merged though.

PR #2028, opened 2024-09-05
Fix bug with playtime migration with existing data

jonathangreen:
## Description

Update the migration so that we add the new columns as nullable, fill them with the default data, then set them to not nullable.

## Motivation and Context

Running `main` against an existing DB today I ran into this issue with the `20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py` migration.

```
{"host": "Syrah.woh.ca", "name": "alembic.runtime.migration", "level": "INFO", "filename": "migration.py", "message": "Running upgrade 7ba553f3f80d -> 7a2fcaac8b63, Add loan_identifier column to playtime tables.", "timestamp": "2024-09-05T14:20:18.621451+00:00", "process": 15725}
  File "/Users/jgreen/.pyenv/versions/circulation/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1910, in _execute_context
    self.dialect.do_execute(
  File "/Users/jgreen/.pyenv/versions/circulation/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
psycopg2.errors.NotNullViolation: column "loan_identifier" contains null values
```

## How Has This Been Tested?

- Tested locally with existing DB

## Checklist

### DIFF
--- a/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
+++ b/alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
@@ -22,12 +22,29 @@ def upgrade() -> None:
 
     op.add_column(
         "playtime_entries",
-        sa.Column("loan_identifier", sa.String(length=40), nullable=False, default=""),
+        sa.Column("loan_identifier", sa.String(length=40), nullable=True, default=""),
     )
 
     op.add_column(
         "playtime_summaries",
-        sa.Column("loan_identifier", sa.String(length=40), nullable=False, default=""),
+        sa.Column("loan_identifier", sa.String(length=40), nullable=True, default=""),
+    )
+
+    # Migrate the existing playtime records before we set the new columns to not nullable.
+    conn.execute("UPDATE playtime_entries SET loan_identifier = ''")
+    conn.execute("UPDATE playtime_summaries SET loan_identifier = ''")
+
+    op.alter_column(
+        "playtime_entries",
+        "loan_identifier",
+        existing_type=sa.String(length=40),
+        nullable=False,
+    )
+    op.alter_column(
+        "playtime_summaries",
+        "loan_identifier",
+        existing_type=sa.String(length=40),
+        nullable=False,
     )
 
     op.drop_constraint("unique_playtime_summary", "playtime_summaries", type_="unique")


alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
```diff
@@ -22,12 +22,29 @@ def upgrade() -> None:
...
     op.add_column(
         "playtime_entries",
-        sa.Column("loan_identifier", sa.String(length=40), nullable=False, default=""),
+        sa.Column("loan_identifier", sa.String(length=40), nullable=True, default=""),
     )
 
     op.add_column(
         "playtime_summaries",
-        sa.Column("loan_identifier", sa.String(length=40), nullable=False, default=""),
+        sa.Column("loan_identifier", sa.String(length=40), nullable=True, default=""),
```
tdilauro, line 30: This looks good and will definitely do the job.

I do think we could resolve the issue without the explicit `UPDATE`s or additional `ALTER COLUMN`s by setting the server_default, as follows:
```python
    op.add_column(
        "playtime_entries",
        sa.Column(
            "loan_identifier", 
            sa.String(length=40), 
            nullable=False,  # <-
            default="",
            server_default="",  # <-
        ),
    )
    op.add_column(
        "playtime_summaries",
        sa.Column(
            "loan_identifier", 
            sa.String(length=40), 
            nullable=False,  # <-
            default="",
            server_default="",  # <-
        ),
    )
```

The `server_default` should affect only the existing rows.

tdilauro (approved):
Looks good! 🏁

This is my bad for missing this issue in the original PR.

A possible alternative is suggested below.

alembic/versions/20240821_7a2fcaac8b63_add_loan_identifier_column_to_playtime_tables.py
jonathangreen: This suggestion would work as well. 

In order to not have alembic show a DB difference, we would have to alter the column after to remove the `server_default` or update the sqlalchemy model to also contain a `server_default`. Since the it gets set on the column in the DB, so adding a default is actually a change to the schema.

PR #2029, opened 2024-09-05
Fix non-canonical protocol not showing settings (PP-1472)

jonathangreen:
## Description

If a setting is stored in the database with a non-canonical protocol, that is the protocol that is being sent to the admin UI. Because of this the admin UI isn't showing settings for these integrations.

This is a good thing to fix anyway, since its the first time we've tried migrating the name of a protocol, but it likely won't be the last time we have to migrate a protocol name.

## Motivation and Context

In https://github.com/ThePalaceProject/circulation/pull/2017, I updated the protocol for the MARC integration from `MARCExporter` to `MarcExporter`. This triggered the problem on the dev server.

## How Has This Been Tested?

- Tested locally with existing DB.

## Checklist

### DIFF
--- a/tests/manager/service/integration_registry/test_base.py
+++ b/tests/manager/service/integration_registry/test_base.py
@@ -5,7 +5,8 @@ import pytest
 from palace.manager.integration.goals import Goals
 from palace.manager.service.integration_registry.base import (
     IntegrationRegistry,
-    IntegrationRegistryException,
+    LookupException,
+    RegistrationException,
 )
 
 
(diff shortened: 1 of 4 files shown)


tests/manager/service/integration_registry/test_base.py
```diff
@@ -100,7 +101,7 @@ def test_registry_get_returns_default_if_name_not_registered(
     assert registry.get("test_class", "default") == "default"
 
     # __get__ throws KeyError
-    with pytest.raises(KeyError):
+    with pytest.raises(LookupException):
```
tdilauro, line 104: Minor: ^ Want to adjust comment above?

tdilauro (approved):
LGTM! 🥇

tests/manager/service/integration_registry/test_base.py
jonathangreen: Missed that one 🤦🏻, thank you!

PR #2034, opened 2024-09-06
Refactor BadResponseException (PP-1641)

jonathangreen:
## Description

Refactor `BadResponseException`, so it has a reference to the response that can be used to deal with the exception.

This required changing a couple references to `RemoteIntegrationException`, where the response wasn't readily available. These exceptions have the same base class, so this should be handled okay.

I updated the OPDS2 + ODL importer to use the new exception, rather then doing its own status code parsing. I also updated mock OPDS2 + ODL importer to use `MockHTTPClient` instead of having its own mock client.

## Motivation and Context

- This work was done as a pre-requisite for the work in https://github.com/ThePalaceProject/circulation/pull/2033. I figured breaking this out to a separate PR would make it easier to review.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/src/palace/manager/util/http.py
+++ b/src/palace/manager/util/http.py
@@ -4,7 +4,7 @@ import logging
 import time
 from collections.abc import Callable, Mapping, Sequence
 from json import JSONDecodeError
-from typing import TYPE_CHECKING, Any
+from typing import Any
 from urllib.parse import urlparse
 
 import requests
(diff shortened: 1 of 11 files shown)


src/palace/manager/util/http.py
```diff
@@ -99,55 +96,28 @@ def __init__(
...
-            url,
-            message,
-            status_code=status_code,
-            debug_message="Status code: %s\nContent: %s"
-            % (
-                status_code,
-                content,
-            ),
-        )
+        self.response = response
```
tdilauro, line 114: Minor: ^ Does the above comment make sense anymore?

tdilauro (approved):
Looks good. 🛠️

src/palace/manager/util/http.py
jonathangreen: Good call, will 🪓

PR #2035, opened 2024-09-06
Delegate Fulfillment instead of using FulfillmentInfo (PP-1641)

jonathangreen:
## Description

Refactor `FulfillmentInfo` class which just contained information about a fufillment, but had multiple methods to override how fulfillment was done, to be a new class called `Fulfillment` which the loans controller delegates fulfillment to. 

This lets each API control how fulfillment is done, and reduces the amount of if / then we need to have in the loan controller. We define several types of Fulfillment, corresponding to the logic that used to exist in the loans controller, and update the API classes to use them.

The goal of this PR is the have fulfillment for each API act the same way before and after, just refactoring to use the new structure.

## Motivation and Context

This refactor builds on https://github.com/ThePalaceProject/circulation/pull/2034 and is part of the work needed for https://github.com/ThePalaceProject/circulation/pull/2033. We are already effectively delegating control of the fulfillment, but not being explicit about it, so I think this refactor makes it easier to reason about how fulfillment will work.

## How Has This Been Tested?

- Running unit tests

## Checklist


src/palace/manager/api/axis.py
```diff
@@ -1967,27 +1995,19 @@ class Axis360AcsFulfillmentInfo(FulfillmentInfo, LoggerMixin):
...
-    def problem_detail_document(self, error_details: str) -> ProblemDetail:
-        service_name = urlparse(self.content_link).netloc
-        self.log.warning(error_details)
-        return INTEGRATION_ERROR.detailed(
-            _(RequestNetworkException.detail, service=service_name),
-            title=RequestNetworkException.title,
-            debug_message=error_details,
-        )
+    def __init__(
+        self, content_link: str, content_type: str | None, verify: bool
```
tdilauro, line 2005: Could this `content_type` default to `DeliveryMechanism.ADOBE_DRM`? Would we be calling this with any others?

tdilauro (approved):
This looks good to me.

There are a couple of small comments below, but nothing super important.

A broader comment is that some of the test and fixture classes still have "FulfillmentInfo" in their names.

jonathangreen:
> A broader comment is that some of the test and fixture classes still have "FulfillmentInfo" in their names.

@tdilauro I think I got all the ones that need to be changed. Its a bit annoying, the axis API has an API call named `FulfillmentInfo`, so there are some references in the Axis classes that are valid, and some that need to be changed. I'll go go through a search again and see if I missed any.

src/palace/manager/api/axis.py
jonathangreen: Makes sense. I'll change that.

jonathangreen:
@tdilauro you were right, I did miss some references to `FulfillmentInfo` that should have been updated. I think I got them all now in https://github.com/ThePalaceProject/circulation/pull/2035/commits/86acbb01a3b925fcff2f29394ad07cfe18c67ab8.

PR #2036, opened 2024-09-06
Update Enki non-DRM to use RedirectFulfillment

jonathangreen:
## Description

In https://github.com/ThePalaceProject/circulation/pull/2035 I noticed that we appear to be streaming full Enki ebooks though the CM, which isn't ideal. This updates to use a redirect fulfillment. 

🚨 ~~From what I see this *should* work, but I need to find a way to do a real world test on this. Going to leave this in draft until I figure that bit out.~~

Edit: I did some testing with this, it does look like we are currently proxying these titles though the CM. Its not a huge number of titles, but we are better off using a redirect for them. All the local testing I can do says this should work, but can't fully test until its deployed somewhere and we can pull in a title with the mobile app. 

I think this one can get merged, and we'll do some testing on minotaur then roll back if we need to.

I was testing with the title: "All I Really Need to Know I Learned from KISS" by Chris Epting.

## Motivation and Context

Looks like this code came in way back as part of this PR: https://github.com/NYPL-Simplified/circulation/pull/717. Looks like it was a band-aid fix that lasted ~7 years, so pretty good bandaid actually 😆. 

## How Has This Been Tested?

- Added unit test
- Tested that previously the CM acted as a proxy
- Tested that after this change, the CM serves a redirect link
  - Tested that I can access that redirect link
  - This should be fine for the apps, this is how other content works

## Checklist

### DIFF
--- a/src/palace/manager/api/enki.py
+++ b/src/palace/manager/api/enki.py
@@ -19,6 +19,8 @@ from palace.manager.api.circulation import (
     HoldInfo,
     LoanInfo,
     PatronActivityCirculationAPI,
+    RedirectFulfillment,
+    UrlFulfillment,
 )
 from palace.manager.api.circulation_exceptions import (
     CannotFulfill,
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 🚀

PR #2039, opened 2024-09-06
[PP-1686] Add test that ensures null loan end dates are included in l…

dbernstein:
…oan resolution query.

## Description
In QA'ing [PP-1456] I noticed we were not resolving loans with an unspecified end date where we should have been.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1686

## How Has This Been Tested?
I added a unit test to verify this case.

## Checklist

[PP-1456]: https://ebce-lyrasis.atlassian.net/browse/PP-1456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/api/controller/playtime_entries.py
+++ b/src/palace/manager/api/controller/playtime_entries.py
@@ -4,7 +4,7 @@ import hashlib
 
 import flask
 from pydantic import ValidationError
-from sqlalchemy import select
+from sqlalchemy import or_, select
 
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🔧

PR #2041, opened 2024-09-09
Add OPDS2 reaper as bin script. (PP-1521)

tdilauro:
## Description

Adds a script to reap books from an OPDS 2 collection, if the book no longer appears anywhere in the feed.

This work can help inform future refactoring to support feed-based reaping of OPDS collections.

## Motivation and Context

Mainly need to clean up ProQuest collections.

[Jira [PP-1521](https://ebce-lyrasis.atlassian.net/browse/PP-1521)]

## How Has This Been Tested?

Manual testing in local development environment.

## Checklist

[PP-1521]: https://ebce-lyrasis.atlassian.net/browse/PP-1521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/opds2_reaper_monitor
+++ b/bin/opds2_reaper_monitor
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 """Remove availability of items no longer present in OPDS 2.0 import collections."""
 import json
-from typing import Any
+from collections.abc import Generator
+from typing import Any, cast
 
 from webpub_manifest_parser.opds2 import OPDS2FeedParserFactory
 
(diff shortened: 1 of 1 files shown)


jonathangreen (approved):
This looks good. 

I'm a bit concerned about memory usage and DB query performance when this is run on a very large collection, but we can see if that concern is warranted or not when we run this against the PQ collection.

tdilauro:
> This looks good.
> 
> I'm a bit concerned about memory usage and DB query performance when this is run on a very large collection, but we can see if that concern is warranted or not when we run this against the PQ collection.

I was thinking about the memory issue, too. I have an idea for resolving it.

tdilauro:
I switched to generating the feed pages in order with a generator, rather than a list, so that those pages shouldn't stick around as long. This should be fine, since the reaper needs to visit every page and they don't need to be in any particular order. 

I did some testing with [Memray](https://bloomberg.github.io/memray/index.html) and didn't see much difference in a relatively small feed of < 4K entries, but I haven't done a test against a very large feed.

PR #2045, opened 2024-09-10
Fix issues caused by differences between redis and elasticache (PP-1693)

jonathangreen:
## Description

Resolves issues surfaced on Minotaur when running against ElastiCache instead of Redis. These issues were around the escaping of characters in object key names. This seems to be handled differently in Redis vs Elasticache. In order to work around this, I ended up creating a scheme where we escaping the problematic characters. 

I don't like this code, or that we are depending on behaviour that is hard to test on CI, but it does fix the issue. If this continues to be an issue, we may either have to rewrite the MARC redis model, so that it doesn't depend on JSON or try to figure out how we can test directly with ElastiCache in CI.

## Motivation and Context

Fix issues that surfaced after deployment to Minotaur.

## How Has This Been Tested?

- Tested against ElastiCache and Redis locally
- Full test with minotaur DB and ElastiCache to make sure we generate marc files

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/marc.py
+++ b/src/palace/manager/celery/tasks/marc.py
@@ -133,32 +133,37 @@ def marc_export_collection(
         # Sync the upload_manager to ensure that all the data is written to storage.
         upload_manager.sync()
 
-        if len(works) == batch_size:
-            # This task is complete, but there are more works waiting to be exported. So we requeue ourselves
-            # to process the next batch.
-            raise task.replace(
-                marc_export_collection.s(
-                    collection_id=collection_id,
-                    start_time=start_time,
-                    libraries=[l.dict() for l in libraries_info],
-                    batch_size=batch_size,
-                    last_work_id=works[-1].id,
-                    update_number=upload_manager.update_number,
+        if len(works) != batch_size:
+            # We have finished generating MARC records. Cleanup and exit.
+            with task.transaction() as session:
+                collection = MarcExporter.collection(session, collection_id)
+                collection_name = collection.name if collection else "unknown"
+                completed_uploads = upload_manager.complete()
+                MarcExporter.create_marc_upload_records(
+                    session,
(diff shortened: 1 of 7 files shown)


src/palace/manager/celery/tasks/marc.py
```diff
@@ -133,32 +133,37 @@ def marc_export_collection(
...
+    raise task.replace(
+        marc_export_collection.s(
+            collection_id=collection_id,
+            start_time=start_time,
+            libraries=[l.dict() for l in libraries_info],
+            batch_size=batch_size,
+            last_work_id=works[-1].id,
+            update_number=upload_manager.update_number,
         )
+    )
```
jonathangreen, line 166: The changes in `src/palace/manager/celery/tasks/marc.py` could come in as a separate PR if desired. They fix a concurrency issue I found while doing other testing. Occasionally, since we were re-queuing while holding the lock, another worker would pick up the new task before this one had released the lock. So this updates the order of operations, so we release the lock before calling `task.replace`.

tests/manager/service/redis/models/test_marc.py
```diff
@@ -21,9 +23,10 @@ def __init__(self, redis_fixture: RedisFixture):
             self._redis_fixture.client, self.mock_collection_id
         )
 
-        self.mock_upload_key_1 = "test1"
-        self.mock_upload_key_2 = "test2"
-        self.mock_upload_key_3 = "test3"
+        # Some keys with special characters to make sure they are handled correctly.
+        self.mock_upload_key_1 = "test/test1/?$xyz.abc"
+        self.mock_upload_key_2 = "t'est💣/tëst2.\"ext`"
```
tdilauro, line 28: Love the 💣 here! 🤣

tdilauro (approved):
Looks great! 💣🚀

A couple of minor comments below.

PR #2048, opened 2024-09-10
Add user agent to logs

jonathangreen:
## Description

If we are logging in the context of a flask request, include the user agent in the logs.

## Motivation and Context

When troubleshooting with the logs, there has been a couple times where it would have been helpful to know what user agent triggered a particular log line.

## How Has This Been Tested?

- Added unit test

## Checklist


tdilauro (approved):
📓 🎸

PR #2049, opened 2024-09-10
Allow restriction to multiple patron locations. (PP-1648)

tdilauro:
## Description

Support more than a single patron location when restricting which patrons can be associated with a given library.

This PR does this by adding an additional Library Identifier Field, "Patron Location", to the fields that can be used to restrict library membership.

Note:
Setting this PR as draft for now, since I still need to remove the recently added single-valued patron location restriction and add/update some more tests, but would be happy to get an initial review on the approach here.

## Motivation and Context

Some libraries have multiple locations, so we need a mechanism to support those.

[Jira [PP-1648](https://ebce-lyrasis.atlassian.net/browse/PP-1648)]

## How Has This Been Tested?

- Manually tested locally.
- Fixed existing tests.
- [CI tests pass for associated](https://github.com/ThePalaceProject/circulation/actions/runs/10801475003) branch.

## Checklist

[PP-1648]: https://ebce-lyrasis.atlassian.net/browse/PP-1648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/api/authentication/basic.py
+++ b/src/palace/manager/api/authentication/basic.py
@@ -764,13 +764,16 @@ class BasicAuthenticationProvider(
     def get_library_identifier_field_data(
         self, patrondata: PatronData
     ) -> tuple[PatronData, str | None]:
-        supported_fields = {
-            LibraryIdenfitierRestrictionField.BARCODE.value: patrondata.authorization_identifier,
-            LibraryIdenfitierRestrictionField.PATRON_LIBRARY.value: patrondata.library_identifier,
+        predefined_field = {
+            k.lower(): v
+            for k, v in {
+                LibraryIdenfitierRestrictionField.BARCODE.value: patrondata.authorization_identifier,
+                LibraryIdenfitierRestrictionField.PATRON_LIBRARY.value: patrondata.library_identifier,
+            }.items()
         }
-        library_verification_field = self.library_identifier_field.lower()
-        if library_verification_field in supported_fields:
-            return patrondata, supported_fields[library_verification_field]
+        id_field = self.library_identifier_field.lower()
+        if (field_value := predefined_field.get(id_field)) is not None:
+            return patrondata, field_value
 
         if not patrondata.complete:
             remote_patrondata = self.remote_patron_lookup(patrondata)
@@ -779,6 +782,10 @@ class BasicAuthenticationProvider(
                 return patrondata, None
             patrondata = remote_patrondata
 
+        if (field_value := predefined_field.get(id_field)) is not None:
+            return patrondata, field_value
+
+        # By default, we'll return the `library_identifier` value.
         return patrondata, patrondata.library_identifier
 
     def enforce_library_identifier_restriction(
@@ -807,9 +814,9 @@ class BasicAuthenticationProvider(
             # Restriction field is blank, so everything matches.
             return patrondata
 
-        patrondata, field = self.get_library_identifier_field_data(patrondata)
+        patrondata, field_value = self.get_library_identifier_field_data(patrondata)
         isValid, reason = self._restriction_matches(
-            field,
+            field_value,
             self.library_identifier_restriction_criteria,
             self.library_identifier_restriction_type,
         )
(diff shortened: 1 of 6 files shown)


src/palace/manager/api/authentication/basic.py
```diff
@@ -223,7 +228,8 @@ class BasicAuthProviderLibrarySettings(AuthProviderLibrarySettings):
             "values here. This value is not used if <em>Library Identifier Restriction Type</em> "
```
jonathangreen, line 228: It might be worth changing the type of this field to `LibraryIdenfitierRestrictionFields` instead of a string, if we're going to enforce an enum of values here.

src/palace/manager/api/authentication/basic.py
```diff
@@ -716,39 +725,52 @@ def identifies_individuals(self):
...
+                    failure_reason = f"{value!r} does not exactly match {_string!r}"
+            case [LibraryIdentifierRestriction.LIST, *_]:
+                if value not in (_list := cast(list, restriction)):
+                    failure_reason = f"{value!r} not in list {restriction!r}"
+
+        return (False, failure_reason) if failure_reason else (True, "")
 
     def get_library_identifier_field_data(
         self, patrondata: PatronData
     ) -> tuple[PatronData, str | None]:
```
jonathangreen, line 766: I think this change will need some corresponding changes to how the millennium patron auth works:
https://github.com/ThePalaceProject/circulation/blob/49c534bb6e012c06e60f16f4c0045a4243df40e6/src/palace/manager/api/millenium_patron.py#L154-L164

src/palace/manager/api/authentication/basic.py
```diff
@@ -761,8 +783,16 @@ def get_library_identifier_field_data(
 
     def enforce_library_identifier_restriction(
         self, patrondata: PatronData
-    ) -> PatronData | None:
-        """Does the given patron match the configured library identifier restriction?"""
+    ) -> PatronData:
```
jonathangreen, line 786: Getting rid of the `None` here is great 🥇

jonathangreen (commented):
The approach here looks good to me!

src/palace/manager/api/authentication/basic.py
tdilauro: Could you say a little more about this? I'll do some manual testing; but right now, I can't see the problem.

tdilauro: I did some testing on this yesterday and it seemed to work as expected.

src/palace/manager/api/authentication/basic.py
tdilauro: We might not be able do that for all integrations (e.g., with the Millenium integration, since we need to support field names that can vary from library to library). But I'll see what we can do.

tdilauro:
@jonathangreen This could use another look when you have a moment. I'm doing some more manual testing now.

src/palace/manager/api/authentication/basic.py
```diff
@@ -777,15 +824,19 @@ def enforce_library_identifier_restriction(
             # Restriction field is blank, so everything matches.
             return patrondata
 
-        patrondata, field = self.get_library_identifier_field_data(patrondata)
-        if self._restriction_matches(
-            field,
+        patrondata, field_value = self.get_library_identifier_field_data(patrondata)
+        isValid, reason = self._restriction_matches(
```
jonathangreen, line 828: Minor: `isValid` is very javascript of you 😄, should probably be `is_valid`.

jonathangreen (approved):
Looks good to me!

src/palace/manager/api/authentication/basic.py
tdilauro: Hahaha. 💯

tdilauro: Okay. Updated that! Can you tell where my mind is right now?

PR #2040, opened 2024-09-06
Fix the sweeper monitor commit failures.

dbernstein:
## Checklist

- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.

PR #2050, opened 2024-09-11
Ensures that the overdrive new titles script does not fail when cover…

dbernstein:
…age records are added.

## Description
The overdrive new titles  and overdrive recent titles monitors are currently broken due to a hidden commit when coverage records are added.   I discovered this issue while fixing https://github.com/ThePalaceProject/circulation/pull/2040

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1708

## How Has This Been Tested?
I tested this manually to make sure that it is fixed and changed a single unit test that was affected.

## Checklist

### DIFF
--- a/tests/manager/core/test_coverage.py
+++ b/tests/manager/core/test_coverage.py
@@ -535,9 +535,6 @@ class TestBaseCoverageProvider:
         class MockProvider1(AlwaysSuccessfulCoverageProvider):
             OPERATION = "i succeed"
 
-            def finalize_batch(self):
-                self.finalized = True
-
         success_provider = MockProvider1(db.session)
 
         batch = [i1, i2]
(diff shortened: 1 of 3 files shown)


tests/manager/core/test_coverage.py
```diff
@@ -546,9 +543,6 @@ def finalize_batch(self):
         # Two successes.
         assert (2, 0, 0) == counts
 
-        # finalize_batch() was called.
-        assert True == success_provider.finalized
-
```
tdilauro, line 551: Seems like we would want to ensure that `finalize_batch` is called, since, in hindsight, it was easy to miss. Could you make a test that fails without the fixes you made above, but passes with them?

tdilauro (approved):
Nice catch here! 🎉 

One suggestion below related to the test code that was removed.

PR #2052, opened 2024-09-11
Fix S3 multipart upload bug in MARC integration (PP-1693)

jonathangreen:
## Description

Fix error on S3 upload seen on Minotaur. 

```
{
  "host": "6e56971b8591",
  "name": "celery.app.trace",
  "level": "ERROR",
  "filename": "trace.py",
  "message": "Task marc.marc_export_collection[482f2b12-98f4-43a3-bfa0-780f90670aac] raised unexpected: ClientError('An error occurred (InvalidArgument) when calling the UploadPart operation: Part number must be an integer between 1 and 10000, inclusive')",
  "timestamp": "2024-09-11T13:19:38.712993+00:00",
  "traceback": "Traceback (most recent call last):\n  File \"/var/www/circulation/env/lib/python3.10/site-packages/celery/app/trace.py\", line 453, in trace_task\n    R = retval = fun(*args, **kwargs)\n  File \"/var/www/circulation/env/lib/python3.10/site-packages/celery/app/trace.py\", line 736, in __protected_call__\n    return self.run(*args, **kwargs)\n  File \"/var/www/circulation/src/palace/manager/celery/tasks/marc.py\", line 134, in marc_export_collection\n    upload_manager.sync()\n  File \"/var/www/circulation/src/palace/manager/marc/uploader.py\", line 74, in sync\n    self._s3_sync(needs_upload)\n  File \"/var/www/circulation/src/palace/manager/marc/uploader.py\", line 54, in _s3_sync\n    upload_part = self.storage_service.multipart_upload(\n  File \"/var/www/circulation/src/palace/manager/service/storage/s3.py\", line 221, in multipart_upload\n    result = self.client.upload_part(\n  File \"/var/www/circulation/env/lib/python3.10/site-packages/botocore/client.py\", line 569, in _api_call\n    return self._make_api_call(operation_name, kwargs)\n  File \"/var/www/circulation/env/lib/python3.10/site-packages/botocore/client.py\", line 1023, in _make_api_call\n    raise error_class(parsed_response, operation_name)\nbotocore.exceptions.ClientError: An error occurred (InvalidArgument) when calling the UploadPart operation: Part number must be an integer between 1 and 10000, inclusive",
  "process": 124,
  "celery": {
    "request_id": "482f2b12-98f4-43a3-bfa0-780f90670aac",
    "task_name": "marc.marc_export_collection"
  }
}
```

```
[...]

PR #2053, opened 2024-09-11
Fix another S3 multipart upload issue with marc exporter (PP-1693)

jonathangreen:
## Description

- Update an offset that I forgot to update in https://github.com/ThePalaceProject/circulation/pull/2052
- Update fixture, so it would find this error in the future
- Add a more extensive integration test for `MarcUploadManager`

## Motivation and Context

I forgot an offset that needed to be updated in https://github.com/ThePalaceProject/circulation/pull/2052.

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/src/palace/manager/marc/uploader.py
+++ b/src/palace/manager/marc/uploader.py
@@ -56,16 +56,17 @@ class MarcUploadManager(LoggerMixin):
             )
             self.upload_session.add_part_and_clear_buffer(key, upload_part)
 
-    def sync(self) -> None:
+    def sync(self, *, complete: bool = False) -> None:
         # First sync our buffers to redis
         buffer_lengths = self.upload_session.append_buffers(self._buffers)
         self._buffers.clear()
 
-        # Then, if any of our redis buffers are large enough, upload them to S3
+        # Then, if any of our redis buffers are large enough, or the upload is complete
+        # sync them to S3.
         needs_upload = [
             key
             for key, length in buffer_lengths.items()
-            if length > self.storage_service.MINIMUM_MULTIPART_UPLOAD_SIZE
+            if length > self.storage_service.MINIMUM_MULTIPART_UPLOAD_SIZE or complete
         ]
 
         if not needs_upload:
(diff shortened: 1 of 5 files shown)


tests/manager/marc/test_uploader.py
```diff
@@ -332,3 +332,64 @@ def test__abort(
...
+            # 1. A small record that doesn't need to be uploaded in parts, it just
+            #    gets uploaded directly when complete is called (test1).
+            # 2. A large record that needs to be uploaded in parts, on the first sync
+            #    call its buffer is large enough to trigger the upload. When complete
+            #    is called, there is no data in the buffer, so no final part needs to be
+            #    uploaded (test2).
+            # 3. A large record that needs to be uploaded in parts, on the first sync
+            #    call its buffer is large enough to trigger the upload. When complete
+            #    is called, there is data in the buffer, so a final part needs to be
+            #    uploaded (test3).
```
tdilauro, line 363: Minor/pedantic: It took me a minute to process this explanation because of the sentence structure. Maybe start a new sentence or add a semicolon after "... in parts" at the beginning of each case explanation.

tdilauro (approved):
This all looks good. A couple of suggestions below.

src/palace/manager/marc/uploader.py
```diff
@@ -109,7 +109,10 @@ def complete(self) -> set[str]:
                     # Upload the last chunk if the buffer is not empty, the final part has no
                     # minimum size requirement.
                     upload_part = self.storage_service.multipart_upload(
-                        key, upload.upload_id, len(upload.parts), upload.buffer.encode()
+                        key,
+                        upload.upload_id,
+                        len(upload.parts) + 1,
```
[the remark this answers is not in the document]
jonathangreen: I simplified the logic here so its easier to follow, and the part logic with offsets is all encapsulated in the same place. 

The simplification comes at the expense of efficiency, since it will require a couple more calls to redis and another call to s3, but that inefficiency probably doesn't make much a difference here and its much more readable.

tests/manager/marc/test_uploader.py
jonathangreen: I re-worked this comment.

PR #2055, opened 2024-09-11
Don't proxy the full response headers (PP-1713)

jonathangreen:
## Description

When we proxy requests through to the client, don't forward the headers from the response. 

## Motivation and Context

This solves an issue we were seeing where the `Content-Encoding` header is set on the request, but since Python requests automatically decodes the transfer encoding, we were mistakenly forwarding on the `Content-Encoding` header, without encoded data. 

I'm open to the suggestion we should just drop the `Content-Encoding` header, instead of not forwarding any headers, but it seemed safer to me to just not pass on any headers, to avoid surprises like this in the future. The only header we pass on is `Content-Type`, if it is set.

JIRA: [PP-1713](https://ebce-lyrasis.atlassian.net/browse/PP-1713)

## How Has This Been Tested?

- Did some local testing to reproduce the problem we were seeing in production

## Checklist

[PP-1713]: https://ebce-lyrasis.atlassian.net/browse/PP-1713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/api/circulation.py
+++ b/src/palace/manager/api/circulation.py
@@ -271,6 +271,14 @@ class RedirectFulfillment(UrlFulfillment):
         )
 
 
+class FetchResponse(Response):
+    """
+    Response object that defaults to no mimetype if none is provided.
+    """
+
+    default_mimetype = None
+
+
 class FetchFulfillment(UrlFulfillment, LoggerMixin):
     """
     Fulfill a loan by fetching a URL and returning the content. This should be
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 🔨 

I agree with your reasoning. It makes sense to filter out as much as possible to start. If we eventually find that we need to pass other headers through, we can customize as needed and we'll know what's getting through and why. And, if we let everything through now, it might be hard to filter more later without breaking something.

PR #2060, opened 2024-09-16
Don't cache overdrive token in credentials table (PP-1717)

jonathangreen:
## Description

Don't cache the overdrive authentication token in the credentials table. Instead just refresh it when we need it. This prevents multiple threads / processes from getting into a deadlock waiting on postgres transactions that are all trying to update the same credential object.

## Motivation and Context

Debugging the deadlock in PP-1717, and this is a factor in that deadlock. It may not be the only issue, but its one of the factors causing this deadlock.

This will mean calling out to get a session token slightly more often, but it shouldn't matter doing one extra call to their API now and then. If we were really concerned we could cache these in redis, but it doesn't seem worth the effort.

## How Has This Been Tested?

- Tested locally

## Checklist

### DIFF
--- a/src/palace/manager/api/overdrive.py
+++ b/src/palace/manager/api/overdrive.py
@@ -9,7 +9,7 @@ import re
 import urllib.parse
 from collections.abc import Iterable
 from threading import RLock
-from typing import Any
+from typing import Any, NamedTuple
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 🚀

PR #2062, opened 2024-09-16
Better app name reporting for collection monitors

jonathangreen:
## Description

Right now all of our `CollectionMonitor` report their application name in PG as `RunCollectionMonitorScript`. This updates the `RunCollectionMonitorScript` to report the app name as the name of the `CollectionMonitor` class.

## Motivation and Context

While troubleshooting the deadlock I'm seeing in PP-1717, knowing the app name to figure out what script was causing issues was very helpful.

## How Has This Been Tested?

- Running tests
- Local troubleshooting

## Checklist

### DIFF
--- a/src/palace/manager/scripts/monitor.py
+++ b/src/palace/manager/scripts/monitor.py
@@ -1,8 +1,11 @@
 import logging
 
+from sqlalchemy.orm import Session
+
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Looks good to me! 🪵

PR #2064, opened 2024-09-17
Update batch sizes for overdrive monitors (PP-1717)

jonathangreen:
## Description

This lowers the size of the batch of work being done by `OverdriveCollectionReaper` and `OverdriveFormatSweep`. Previously `OverdriveCollectionReaper` defaulted to 100 and `OverdriveFormatSweep` defaulted to 25. Since each item of the batch takes one (or multiple) OD API requests, that means these batches can take a significant amount of time. We only commit a transaction at the end of a batch, so lock conflicts can block for a significant amount of time. This should ease those conflicts.

## Motivation and Context

The lock conflicts can impact web workers, if they need a lock on the same row being updated by the script. This can cause timeouts, or even deadlocks like in the case of PP-1717.

## How Has This Been Tested?

- Local testing

## Checklist


tdilauro (approved):
👍🏽

PR #2065, opened 2024-09-17
Add a configuration to enable a debug faulthandler (PP-1717)

jonathangreen:
## Description

Adds a new env variable `PALACE_LOG_DEBUG_TRACEBACK_INTERVAL`. This option is disabled when it is zero (the default). When it is non-zero it uses the python [faulthandler](https://docs.python.org/3/library/faulthandler.html) module to print stack traces of every thread to stderr every `INTERVAL` seconds.

## Motivation and Context

This is useful to figure out the state of the app when a thread is deadlocked, watching the stack traces after the deadlock occurs gives an idea of what is happening. I though this would be useful to add to the main codebase as it might help assist if / when we get deadlocks happening in environments that are not easily reproduced. This could be enabled with a env var to get a bit of additional information to assist with debugging.

## How Has This Been Tested?

- Tested locally, I was using it to help track down a deadlock in PG.

## Checklist

### DIFF
--- a/src/palace/manager/service/logging/configuration.py
+++ b/src/palace/manager/service/logging/configuration.py
@@ -6,7 +6,7 @@ from enum import auto
 from typing import Any
 
 import boto3
-from pydantic import PositiveInt, validator
+from pydantic import NonNegativeInt, PositiveInt, validator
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
🪵👀

PR #2066, opened 2024-09-17
Replace parametrize parameters with pytest.param

jonathangreen:
## Description

A number of our tests that use `@pytest.mark.parametrize` have test case ids, but since they were converted after they were written to use `parametrize` the id isn't being used. This updates them all to use `pytest.param` with an `id` set.

## Motivation and Context

I've tried to run our tests a couple times now using `-v`, but because that prints out the full test name, and the test name for some of these tests is a 1000 line xml string, it makes the verbose output impossible to read. Converting them to use `pytest.param` fixes this.

## How Has This Been Tested?

- Running tests locally

## Checklist


tdilauro (approved):
Only gave this a cursory scan, but looks good! 🛠️

PR #2057, opened 2024-09-13
OPDS2 + ODL Handle case where no licenses are available (PP-1716)

jonathangreen:
## Description

Handle the case where we think there are available licenses, but when we reach out to the content provider, it turns out there are no licenses available.

## Motivation and Context

This was found as part of release testing in PP-1716.

## How Has This Been Tested?

- New unit tests

## Checklist

- [X] I have updated the documentation accordingly.
- [X] All new and existing tests passed.

PR #2069, opened 2024-09-17
ODL feed error when license already checked out (PP-1716)

jonathangreen:
## Description

In https://github.com/ThePalaceProject/circulation/pull/2057 I partially solved this one. We are doing the correct thing raising `NoAvailableCopies`. But then when we try to place a hold, because we didn't update the availability, the we were raising `CurrentlyAvailable`.

## Motivation and Context

JIRA: PP-1716

## How Has This Been Tested?

- Unit tests

## Checklist


tdilauro (approved):
Looks good! 🏁

PR #2070, opened 2024-09-18
Fix flaky test TestFeaturedFacets:test_run (PP-1726)

jonathangreen:
## Description

I've been doing some testing with this locally, and I think this test depended on the commit removed in https://github.com/ThePalaceProject/circulation/pull/2040. Adding an `expire_all` seems like it has the test working reliably for me on my local machine.

Its hard to say for sure this fixes it, but it seems like it does for me locally. Time will tell in CI.

## Motivation and Context

This one has been failing fairly frequently in CI, and has been driving me crazy (its not that far a drive 😆). I had one instance where it failed re-runs in a row today. 

## How Has This Been Tested?

- Running tests in a loop on my computer hah

## Checklist


jonathangreen:
I ran the tests overnight in a loop on my laptop and they ran all night without issue, so I'm fairly confident this does fix the issue.

tdilauro (approved):
🚽 🚀

PR #2075, opened 2024-09-18
Fix RunCollectionMonitorScript PG application name

jonathangreen:
## Description

Update the class that we pass into `production_session` to use `self.monitor_class` instead of `self.monitor_class.__class__`.

## Motivation and Context

When I added this in https://github.com/ThePalaceProject/circulation/pull/2062, I mistakenly thought `self.monitor_class` was an instance.

## How Has This Been Tested?

- Tested locally

## Checklist


tdilauro (approved):
🔨

PR #2076, opened 2024-09-18
Remove call to expire_all in ODL code

jonathangreen:
## Description

Move the call to `expire_all` into the test code.

## Motivation and Context

I noticed this while working on https://github.com/ThePalaceProject/circulation/pull/2074. I was wondering if you considered this approach @dbernstein of putting this call into the tests instead of the main codebase. Generally, I think, mucking about with the session should be avoided as much as we can.

I left this in draft because I wanted to get your feedback on it @dbernstein, in case there is some pitfall here that I might be falling into.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/src/palace/manager/api/odl/api.py
+++ b/src/palace/manager/api/odl/api.py
@@ -769,8 +769,6 @@ class OPDS2WithODLApi(
                 days=default_reservation_period
             )
 
-        _db.expire_all()
-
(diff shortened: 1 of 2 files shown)


jonathangreen:
I'm going to close this one and roll it into changes I need to make for holds in PP-1728.

PR #2077, opened 2024-09-19
Fix uwsgi log file permissions (PP-1667)

jonathangreen:
## Description

Updates our docker build to make sure that:
- `/var/log/uwsgi` 
  - owned by root:adm
- `/var/log/uwsgi/uwsgi.log` 
  - owned by simplified:adm
  - has 644 permissions, so cloudwatchd can read and upload the files

## Motivation and Context

I'm not sure when this got broken, but I realized when trying to work on PP-1667 this morning that we were not geting logs uploaded from staging, and then I went looking and all recently deployed CMs are in the same state.

Since `/var/log` is mounted as a volume, this will fix things for any new CM, but existing ones will to have their permissions updated. I'm planning to make a follow up PR to do this.

## How Has This Been Tested?

- Ran a `docker compose up` locally, and checked things have the correct permissions

## Checklist

### DIFF
--- a/docker/services/logrotate/logrotate.d/celery.conf
+++ b/docker/services/logrotate/logrotate.d/celery.conf
@@ -2,7 +2,7 @@
     missingok
     daily
     create 0660 simplified adm
-    rotate 30
+    rotate 13
     compress
     delaycompress
     notifempty
--- a/docker/services/logrotate/logrotate.d/simplified.conf
+++ b/docker/services/logrotate/logrotate.d/simplified.conf
@@ -1,7 +1,6 @@
 /var/log/simplified/*.log {
     missingok
     daily
-    create 0700 root root
     rotate 13
     copytruncate
     compress
@@ -13,8 +12,7 @@
 /var/log/uwsgi/*.log {
     missingok
     daily
-    create 0660 simplified adm
-    rotate 30
+    rotate 13
     copytruncate
     compress
     delaycompress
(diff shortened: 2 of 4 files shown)


docker/services/logrotate/logrotate.d/celery.conf
```diff
@@ -2,7 +2,7 @@
     missingok
     daily
     create 0660 simplified adm
-    rotate 30
+    rotate 13
```
jonathangreen, line 5: Align this with the retention we have for our other logs

docker/services/logrotate/logrotate.d/simplified.conf
```diff
@@ -1,7 +1,6 @@
 /var/log/simplified/*.log {
     missingok
     daily
-    create 0700 root root
```
jonathangreen, line 4: Because we use `copytruncate` this line is actually ignored, so we might as well drop it.

docker/services/logrotate/logrotate.d/simplified.conf
```diff
@@ -13,8 +12,7 @@
 /var/log/uwsgi/*.log {
     missingok
     daily
-    create 0660 simplified adm
-    rotate 30
+    rotate 13
```
jonathangreen, line 15: Align this with the configuration we push out in our playbook.

docker/services/uwsgi/uwsgi.d/40_log.ini
```diff
@@ -1,5 +1,4 @@
 [uwsgi]
 log-format = [uwsgi] %(var.HTTP_X_FORWARDED_FOR) (%(addr)) - - [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)" host_hdr=%(host) req_time_elapsed=%(msecs) process=%(pid) worker=%(wid)
-logfile-chmod = 644
```
jonathangreen, line 3: This configuration option isn't used when using `logger`, its only used when using the `log-to` configuration. So we might as well drop it, since its deceiving that it is here.

When we switched to using `logger` in this config is probably when it broke.

tdilauro (approved):
Looks good! 🎸

PR #2080, opened 2024-09-21
Update to Pydantic 2.9 (PP-890)

jonathangreen:
## Description

- Update to use Pydantic 2.x
- Remove any calls to deprecated Pydantic methods
- I didn't do any changes here to make use of any new methods in Pydantic, this is pretty much the minimal work necessary to do the update.
- The biggest change here is dropping all the settings source overrides we had, because they were more difficult to convert since Pydantic updated how settings sources work. You now have to implement an abstract class instead of a callable see: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#customise-settings-sources.
  - `ServiceConfigurationWithLimitedEnvOverride` class was replaced with `Literal[True]` annotations in the AdminClientFeatureFlags class. I think this accomplishes the same goal, but I need to check in with @tdilauro about this.
  - `CeleryConfiguration` removed the ability to specify extra variables from the environment. This isn't being used in production and was added to make it easier to test features. I think it can be dropped for now and re-added later.

## Motivation and Context

Keeping our dependencies up to date. This is a big one that was out of date.

## How Has This Been Tested?

- Run tests in CI

## Checklist


src/palace/manager/api/admin/config.py
```diff
@@ -24,38 +26,20 @@ class AdminClientFeatureFlags(ServiceConfigurationWithLimitedEnvOverride):
...
-        # - environment_override_warning_fields: warnings and ignore environment
-        # - environment_override_error_fields: raise exception
-        environment_override_warning_fields: set[str] = {
-            "enable_auto_list",
-            "show_circ_events_download",
-        }
-        environment_override_error_fields: set[str] = set()
+    model_config = SettingsConfigDict(
+        env_prefix="PALACE_ADMINUI_FEATURE_",
+        alias_generator=AliasGenerator(serialization_alias=to_camel),
```
jonathangreen, line 41: Pydantic has its own alias generator `to_camel` that I switched this to use, and removed `_snake_to_camel_case`.

src/palace/manager/api/admin/controller/custom_lists.py
```diff
@@ -37,18 +36,18 @@
...
-from palace.manager.util.flask_util import parse_multi_dict
+from palace.manager.util.flask_util import CustomBaseModel, parse_multi_dict
 from palace.manager.util.problem_detail import ProblemDetail, ProblemDetailException
 
 
 class CustomListsController(
     CirculationManagerController, AdminPermissionsControllerMixin
 ):
-    class CustomListSharePostResponse(BaseModel):
+    class CustomListSharePostResponse(CustomBaseModel):
```
jonathangreen, line 46: These seem like they should have been using `CustomBaseModel` all along.

src/palace/manager/api/admin/form_data.py
```diff
@@ -39,23 +39,26 @@ def get_settings_dict(
...
         """
         return_data: dict[str, Any] = {}
-        for field in settings_class.__fields__.values():
-            if not isinstance(field.field_info, FormFieldInfo):
-                continue
-            form_item = field.field_info.form
+        for field_name, field_info in settings_class.model_fields.items():
+            assert isinstance(
+                field_info, FormFieldInfo
+            ), f"Expected FormFieldInfo, got {field_info.__class__}"
```
jonathangreen, line 45: This used to just silently not include non `FormFieldInfo` settings. I switched this to be an assertion so we fail in this case. It shouldn't ever be the case, and if it is, I think we probably want to fail so we can understand what happened.

src/palace/manager/api/millenium_patron.py
```diff
@@ -105,7 +107,6 @@ def validate_neighborhood_mode(cls, v):
                 AuthenticationMode.FAMILY_NAME: "Family Name",
             },
         ),
-        alias="auth_mode",
```
jonathangreen, line 108: I removed a bunch of old unused aliases, that were left over from when we migrated from the settings table. These weren't used anymore, and it seemed like a good time to remove them.

src/palace/manager/service/sitewide.py
```diff
@@ -1,28 +1,24 @@
...
+    patron_web_hostnames: list[HttpUrl] | Literal["*"] = []
     authentication_document_cache_time: NonNegativeInt = 3600
     quicksight_authorized_arns: dict[str, list[str]] | None = None
 
-    @validator("base_url")
-    def validate_base_url(cls, v: AnyHttpUrl | None) -> AnyHttpUrl | None:
-        # Our base url should not end with a slash, if it does we remove it.
-        if v is not None and v.endswith("/"):
-            return cast(AnyHttpUrl, v.rstrip("/"))
-        return v
```
jonathangreen, line 23: The new pydantic URL validation does this automatically, so we just drop this validator.

src/palace/manager/api/admin/config.py
```diff
@@ -24,38 +26,20 @@ class AdminClientFeatureFlags(ServiceConfigurationWithLimitedEnvOverride):
...
-    # Setting `const=True` ensures that the default value is not overridden.
-    # Add them to the one of the `environment_override_*` Config settings
-    # below to prevent them from being overridden.
-    # NB: Overriding the `env_prefix` with `env=...` here may lead to
-    #     incorrect values in warnings and exceptions, since `env_prefix`
-    #     is used to generate the full environment variable name.
-    enable_auto_list: bool = Field(
+    # The following fields CANNOT be overridden by environment variables, so
+    # their types are set to Literal[True] to prevent them from being overridden.
+    enable_auto_list: Literal[True] = Field(
```
tdilauro, line 31: I think this will behave more like an `environment_override_error_fields` (which we're not currently using) field, which would probably prevent the admin UI working.

On the contrary, the way `environment_override_[warning]_fields` (which we are using) behaves is that it:
- prevents a change to the variable; and
- emits a warning, so that we might notice that something was being set that should not be set, so that we could go clean it up.

I might be missing it, but I didn't see a test for this case.

tdilauro (commented):
This is good shape, but I do have concerns about the issue you flagged me for in the main PR comment. Right now, we lazy load the configuration, so we might not see the problem until someone uses that admin UI. We might be able to mitigate this by eagerly loading these settings on deployment, so that we would fail faster and could fix the issue while it has our attention.

tests/manager/integration/test_base.py
```diff
@@ -38,8 +38,10 @@ def base_fixture():
 
 def test_integration_settings_load(base_fixture: BaseFixture) -> None:
     return_value: BaseSettings = base_fixture.load()
-    base_fixture.mock_settings_cls.assert_called_once_with(test="test", number=123)
-    assert return_value is base_fixture.mock_settings_cls.return_value
+    base_fixture.mock_settings_cls.model_validate.assert_called_once_with(
+        {"test": "test", "number": 123}
```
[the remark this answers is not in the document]
jonathangreen: Done

src/palace/manager/api/model/time_tracking.py
```diff
@@ -45,10 +47,10 @@ def validate_seconds_played(cls, value: int):
 
 class PlaytimeEntriesPost(CustomBaseModel):
     book_id: str | None = Field(
-        description="An identifier of a book (currently ignored)."
+        None, description="An identifier of a book (currently ignored)."
     )
     library_id: str | None = Field(
-        description="And identifier for the library (currently ignored)."
+        None, description="And identifier for the library (currently ignored)."
```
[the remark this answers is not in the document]
jonathangreen: Done

src/palace/manager/api/admin/config.py
jonathangreen: How would you feel about removing `Literal[True]` as I have here, and just adding a todo comment to go in and remove these feature flags if we no longer need them?

I implemented this in https://github.com/ThePalaceProject/circulation/pull/2080/commits/d890d47937b905fc01abeb00b7ee1c35552a77be.

tdilauro: I think that would be better than delayed failing. Tradeoffs. 🤷🏽‍♂️

tdilauro (approved):
I think this one is good to go! 🚀

PR #2083, opened 2024-09-24
fix: opds2-reaper-monitor-script (PP-1756)

tdilauro:
## Description

Makes a few improvements and fixes a couple of bugs found with practical testing:
- Changes the service name, so that "OPDS Import Monitor" timestamp records are not clobbered by the reaper.
- Changes the script name, so that the logged name is less verbose.
- Fixes formatting bugs.
- Splits up logging, so that better log level can be applied.
- Emits an error message when individual collections are specified along with the all collections flag (rather than silently doing nothing).

## Motivation and Context

[Jira [PP-1756](https://ebce-lyrasis.atlassian.net/browse/PP-1756)]

## How Has This Been Tested?

Manually tested in the staging environment.

## Checklist

[PP-1756]: https://ebce-lyrasis.atlassian.net/browse/PP-1756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/opds2_reaper_monitor
+++ b/bin/opds2_reaper_monitor
@@ -35,6 +35,9 @@ def main():
 class OPDS2ReaperMonitor(OPDS2ImportMonitor):
     """Monitor to make unavailable any license pools without a matching identifier in the feed."""
 
+    # TODO: This should really go in a base `OPDSReaperMonitor` class.
+    SERVICE_NAME = "OPDS Reaper Monitor"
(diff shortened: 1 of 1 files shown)


jonathangreen (approved):
Looks good!

PR #2085, opened 2024-09-24
Update license pools for ODL licenses that are loan + concurrency limited

jonathangreen:
## Description

The rest of the code seems to treat `licenses_owned` as the maximum number of concurrent lends that can be expected. We were setting that value to `checkouts_left`, which is the case only if `terms_concurrency` isn't set, otherwise its the minimum of the two values.

## Motivation and Context

JIRA: PP-1728

There are some other issues with holds that I'm planning to address in a follow up PR, but this is a major cause of holds that can't be fulfilled.

## How Has This Been Tested?

- Tested locally

## Checklist


dbernstein (approved):
Looks good.

PR #2092, opened 2024-09-27
Bump version of our pre-commit hooks

jonathangreen:
## Description

This one touches a lot of files, but it shouldn't be changing any behaviour. Everything in this PR was automatically generated by pre-commit.

All I did to create this PR was:
```
pre-commit autoupdate
pre-commit run -a
```

## Motivation and Context

While working on PP-1769 I ran into an issue with the `black` pre-commit hook:
```
error: cannot format src/palace/manager/opds/base.py: Cannot parse: 102:12:             match (rel, type):
```

It looks like the older version of `black` we were using couldn't handle that match statement, I could have tried to work around it, but I figured I might as well just update the version we were using.

## How Has This Been Tested?

- Tests in CI

## Checklist


bin/informational/adobe-id-for-short-client-token
```diff
@@ -23,8 +23,7 @@ REGISTRY_RESPONSE_RE = re.compile(
 )
 
 
-class InvalidTokenException(Exception):
-    ...
+class InvalidTokenException(Exception): ...
```
tdilauro, line 26: Yay!

src/palace/manager/api/axis.py
```diff
@@ -262,7 +262,7 @@ def authorization_headers(self) -> dict[str, str]:
         )
         return dict(Authorization="Basic " + authorization_b64)
 
-    def _run_self_tests(self, _db: Session) -> Generator[SelfTestResult, None, None]:
+    def _run_self_tests(self, _db: Session) -> Generator[SelfTestResult]:
```
tdilauro, line 265: Type defaults! Assuming this came from pyupgrade.

tdilauro (approved):
This looks fine to me. 🏆 

PEP 696 type defaults FTW!

src/palace/manager/api/axis.py
jonathangreen: It did! I was happy to see this as well!

PR #2094, opened 2024-10-01
Fix broken playtime entries test.

dbernstein:
## Description
This update fixes a broken test.  The test started failing due to a date addition issue that was interacting badly with the current month.  I changed it to ensure that the playtime entry in question is always in scope.

## Motivation and Context
Build was broken.

## Checklist

### DIFF
--- a/src/palace/manager/scripts/playtime_entries.py
+++ b/src/palace/manager/scripts/playtime_entries.py
@@ -238,7 +238,12 @@ class PlaytimeEntriesEmailReportsScript(Script):
                 sql_max(coalesce(PlaytimeSummary.title, "")).label("title2"),
                 count(distinct(PlaytimeSummary.loan_identifier)).label("loan_count"),
             )
-            .where(PlaytimeSummary.timestamp.between(start, until))
+            .where(
+                and_(
+                    PlaytimeSummary.timestamp >= start,
+                    PlaytimeSummary.timestamp < until,
+                )
+            )
             .group_by(
                 PlaytimeSummary.identifier_str,
                 PlaytimeSummary.collection_name,
@@ -257,7 +262,12 @@ class PlaytimeEntriesEmailReportsScript(Script):
                 coalesce(PlaytimeSummary.title, "").label("title"),
                 sum(PlaytimeSummary.total_seconds_played).label("total_seconds_played"),
             )
-            .where(PlaytimeSummary.timestamp.between(start, until))
+            .where(
+                and_(
+                    PlaytimeSummary.timestamp >= start,
+                    PlaytimeSummary.timestamp < until,
+                )
+            )
             .group_by(
                 PlaytimeSummary.identifier_str,
                 PlaytimeSummary.collection_name,
--- a/tests/manager/scripts/test_playtime_entries.py
+++ b/tests/manager/scripts/test_playtime_entries.py
@@ -584,9 +584,9 @@ class TestPlaytimeEntriesEmailReportsScript:
         playtime(
             db.session, identifier, collection, library, dt1m(4), 2, loan_identifier
         )
-        # one month + 32 days ago : in the future and therefore out of range.
+        # out of range: after end of default reporting period
         playtime(
-            db.session, identifier, collection, library, dt1m(32), 3, loan_identifier
+            db.session, identifier, collection, library, dt1m(31), 6, loan_identifier
         )
         playtime(
             db.session, identifier, collection, library, dt1m(-31), 60, loan_identifier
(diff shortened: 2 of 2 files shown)


tests/manager/scripts/test_playtime_entries.py
```diff
@@ -576,11 +576,13 @@ def test_do_run(
         # We're using the RecursiveEquivalencyCache, so must refresh it.
         EquivalentIdentifiersCoverageProvider(db.session).run()
 
+        # one month + 3 days ago
         playtime(
             db.session, identifier, collection, library, dt1m(3), 1, loan_identifier
         )
+        # one month + 4 days ago
         playtime(
-            db.session, identifier, collection, library, dt1m(31), 2, loan_identifier
```
tdilauro, line 583: I think this entry is meant to be near the beginning of the month during which the report is being run and, thus, should not be included in the report. `dt1m` is calculating an offset from the first day of the previous month. 

I believe that:
- the `dt1m(31)` entry is meant to represent "out of range - after reporting period", and
- the `dt1m(-31) entry, "out of range - before reporting period" (it is helpfully labeled to that effect)

dbernstein: Right I think that test wants to verify that two identifiers on the same library, collection and loan sum the have their seconds summed together.   When it was dt1m(31),  they were both being summed up to 3 (1+2) because dt1m(31) was in the past.  However today is October 1st.   dt1m(31) is  Sept 1st + 31 days = October 2 in the future. 

So the expected values were wrong, but it was also my intent to verify that two invocations with the same parameters at different but value times are summed.  I'll update the test.

tdilauro: But the defaults for that script are to run the previous month, as indicated by the `cutoff` and `until` [dates in the test](https://github.com/ThePalaceProject/circulation/blob/3f5d5aeba8c3ec2cd101ecae6465454b08545091/tests/manager/scripts/test_playtime_entries.py#L678-L679); so no entry from the current month should be counted. The `until` date is **after** what should be included in the report.

```python
    cutoff = date1m(0).replace(day=1)
    until = utc_now().date().replace(day=1)
```

Here's [the explanation of the `until` date from the scripts `argparser`](https://github.com/ThePalaceProject/circulation/blob/ebcddcab60487e7f9a763b27296c39aca04f2649/src/palace/manager/scripts/playtime_entries.py#L139-L147):

```
        help="'Until' date for report in ISO 8601 'yyyy-mm-dd' format."
        " The report will represent entries from the 'start' date up until,"
        " but not including, this date.",
    )
```

dbernstein: Okay PR updates.  It turns out sqlalchemy's between function is inclusive which is not what we want to align with `until` definition in the argparser.

I verified that one month back plus 30 days (which is Oct 1st) failed before the change to the where clause.   Then it passed with the updates.  Then I changed it to 31 so that it will succeed in a 31 day month like October.

src/palace/manager/scripts/playtime_entries.py
```diff
@@ -238,7 +238,12 @@ def _fetch_report_records(self, start: datetime, until: datetime) -> Query:
                 sql_max(coalesce(PlaytimeSummary.title, "")).label("title2"),
                 count(distinct(PlaytimeSummary.loan_identifier)).label("loan_count"),
             )
-            .where(PlaytimeSummary.timestamp.between(start, until))
+            .where(
+                and_(
```
dbernstein, line 242: This where clause is exclusive whereas between is inclusive.

dbernstein: Good freakin catch.  I don't know how you do it.

tests/manager/scripts/test_playtime_entries.py
```diff
@@ -576,11 +576,17 @@ def test_do_run(
...
+        # one month + 3 days ago : in scope
         playtime(
             db.session, identifier, collection, library, dt1m(3), 1, loan_identifier
         )
+        # one month + 4 days ago : in scope  should be added
         playtime(
-            db.session, identifier, collection, library, dt1m(31), 2, loan_identifier
+            db.session, identifier, collection, library, dt1m(4), 2, loan_identifier
+        )
+        # one month + 31 days ago : out of range of reporting period
```
tdilauro, line 587: Minor quibble about this comment: it sounds like it's over two months ago. Could we re-word to something like:

```
# out of range: after end of default reporting period
```

But, good to have this labeled, so that we know what's up next time! For future us! 🥂

src/palace/manager/scripts/playtime_entries.py
tdilauro: 😊 Thanks!

The tests really caught it. I just noticed that we were needing to relax a test that should've passed.

tdilauro (approved):
Looks good! 🎉 

I requested change to the comment in the test.

PR #2095, opened 2024-10-01
OPDS ODL API changes for UL Feed (PP-1769)

jonathangreen:
## Description

Some major updates to the OPDS2 + ODL API code. 

The highlights are:
- Switch the validation of the status documents we get back to use Pydantic
  - There is some work here to start building up models of OPDS2 and ODL that can be the basis of replacing webpub manifest parser when we get to it.
- Update the checkout / checkin / fulfillment logic, to reflect the differences between the documents we get back from DeMarque and those returned by the Open Source LCP server stack. 
  - While I was in here, I updated the fulfillment logic to support ODL without LCP encryption.
- Using the correct HTTP methods as defined in the spec, rather then using GET for everything
  - DeMarque supports using GET, but LCP server doesn't, and we should be following the spec

## Motivation and Context

Starting to pull in ODL feeds from other distributors.

JIRA: PP-1769

## How Has This Been Tested?

- [x] Running unit tests
- [x] Manually testing all the different DRM / content type interactions from marketplace via the app
  - This is complete, the interactions were tested in Cerberus.
    - LCP Ebook
    - LCP Audiobook
    - Adobe Ebook
    - DeMarque Audiobook
    - UL Abook
    - UL Ebook  

## Checklist


tests/manager/scripts/test_playtime_entries.py
```diff
@@ -529,6 +529,7 @@ def playtime(
 
 
 class TestPlaytimeEntriesEmailReportsScript:
+    @freeze_time("2021-01-01")
```
jonathangreen, line 532: 🚨 this is just a placeholder to get tests passing until https://github.com/ThePalaceProject/circulation/pull/2094 gets merged. This line should be removed before this gets merged in.

src/palace/manager/api/odl/api.py
```diff
@@ -300,46 +221,42 @@ def checkin(self, patron: Patron, pin: str, licensepool: LicensePool) -> None:
...
-            if link.get("rel") == "return":
-                return_url = link.get("href")
-                break
+        doc = self._request_loan_status("GET", loan.external_identifier)
+        if not doc.active:
+            self.log.warning(
+                f"Loan {loan.id} was already returned early, revoked by the distributor, or it expired."
+            )
+            self.update_loan(loan, doc)
+            return False
```
jonathangreen, line 237: This is a functionality change. We used to raise `NotCheckedOut` in this case.

I think the updated behavior is correct. If you are trying to return an already returned item, we should just process the return and delete the local loan, instead of raising an exception and keeping the loan on the users shelf, in a state where it cannot be returned.

src/palace/manager/api/odl/api.py
```diff
@@ -300,46 +221,42 @@ def checkin(self, patron: Patron, pin: str, licensepool: LicensePool) -> None:
...
-        # However, it might be because the loan has already been fulfilled
-        # and must be returned through the DRM system, which the app will
-        # do on its own, so we can ignore the problem.
-        new_loan = get_one(_db, Loan, id=loan.id)
-        if new_loan:
-            return False
+        # The parameters for this link (if its templated) are defined here:
+        # https://readium.org/lcp-specs/releases/lsd/latest.html#34-returning-a-publication
+        # None of them are required, and often the link is not templated.
+        return_url = return_link.href_templated({"name": "Palace Manager"})
```
jonathangreen, line 254: UL feeds have a templated link, so we need this update to expand the template before making the request.

src/palace/manager/api/odl/api.py
```diff
@@ -422,26 +339,59 @@ def _checkout(
...
+        checkout_url = url_template.expand(
+            id=str(identifier),
+            checkout_id=checkout_id,
+            patron_id=patron_id,
+            expires=requested_expiry.isoformat(),
+            notification_url=notification_url,
+            passphrase=encoded_pass,
+            hint=self.settings.passphrase_hint,
+            hint_url=self.settings.passphrase_hint_url,
+        )
```
jonathangreen, line 381: This code is just moved around and has not been changed. It was in `get_license_status_document` previously, but only used when that function was called from `_checkout`.

src/palace/manager/api/odl/api.py
```diff
@@ -450,35 +400,48 @@ def _checkout(
...
+            rel="self", type=LoanStatus.content_type()
+        )
+
+        # The ODL spec requires that a 'self' link be present in the links section of the response.
+        # See: https://drafts.opds.io/odl-1.0.html#54-interacting-with-a-checkout-link
+        # However, the open source LCP license status server does not provide this link, so we make
+        # an extra request to try to get the information we need from the 'status' link in the license
+        # document, which the LCP server does provide.
+        # TODO: Raise this issue with LCP server maintainers, and try to get a fix in place.
+        #   once that is done, we should be able to remove this fallback.
```
jonathangreen, line 422: Started a slack thread about this here: https://readium.slack.com/archives/C0SGTMMR6/p1727814143241779

src/palace/manager/api/odl/api.py
```diff
@@ -564,43 +484,66 @@ def _find_matching_delivery_mechanism(
...
-        )
+        drm_scheme = delivery_mechanism.delivery_mechanism.drm_scheme
+        fulfill_cls: Callable[[str, str | None], UrlFulfillment]
+        if drm_scheme == DeliveryMechanism.NO_DRM:
+            # If we have no DRM, we can just redirect to the content link and let the patron download the book.
+            fulfill_link = doc.links.get(
+                rel="publication",
+                type=delivery_mechanism.delivery_mechanism.content_type,
+            )
+            fulfill_cls = RedirectFulfillment
```
jonathangreen, line 526: This supports the no-DRM case: 
https://www.notion.so/lyrasis/OPDS2-ODL-Feed-without-LCP-a69bb6024fce4a71b50280d9fa8a4e84

tests/manager/scripts/test_playtime_entries.py
jonathangreen: This has been removed.

src/palace/manager/api/odl/api.py
```diff
@@ -564,43 +489,66 @@ def _find_matching_delivery_mechanism(
...
+        elif drm_scheme == DeliveryMechanism.FEEDBOOKS_AUDIOBOOK_DRM:
+            # For DeMarque audiobook content, the link we are looking is stored in the 'manifest' rel.
```
tdilauro, line 533: Minor: missing word in the comment? "...link we are looking [up? for? is..."

tests/manager/api/controller/test_odl_notify.py
```diff
@@ -1,132 +1,147 @@
...
+        assert response == INVALID_LOAN_FOR_ODL_NOTIFICATION
+
+        # No loan, but distributor thinks it isn't active
+        with flask_app_fixture.test_request_context(
+            "/",
+            method="POST",
+            library=odl_fixture.library,
+            data=odl_fixture.loan_status_document("returned").model_dump_json(),
+        ):
+            response = odl_fixture.controller.notify(-55)
```
tdilauro, line 133: I think this `-55` is meant to represent a loan that can't exist, but it's not super clear in this context. Would be useful to add a comment here (and a few lines down) making that explicit. Or we could use a constant named something like "NON_EXISTENT_LOAN".

tdilauro (approved):
This looks good to me. 🍄 

A couple small comments below.

PR #2100, opened 2024-10-03
[PP-1507] Present sorts as links with palaceproject.io namespaced sort rels.

dbernstein:
## Description
This is a first pass at the ticket below: 

I added two new elements to the [terms namespace](https://www.notion.so/lyrasis/palaceproject-io-terms-namespace-572089bd44404cf395f02b6b78361fe4). 

Then I implemented them here for the OPDS 1 feed and OPDS 2 feeds.
I'm adding in new links for sorts based on the new namespace terms while leaving the current approach with facets in place for backward compatibility.   

The OPDS2 feed adds a new sort link and removes the old "Sort by" facet group since as far as I know, no one is using the OPDS2 feed.

The new sort links will appear just below the root.   The old sorting links formatted using the facet groups will remain in place until we we're ready to remove them.  You can expect something like the following:
```
<feed ... xmlns:palace="http://palaceproject.io/terms/properties/"... >
...
   <link  url="https://link-to-sorted-set-by-x" rel="http://palaceproject.io/terms/rel/sort" title="Sort by X"  />
   <link  url="https://link-to-sorted-set-by-y" rel="http://palaceproject.io/terms/rel/sort" title="Sort by Y"  />
   <link  url="https://link-to-sorted-set-by-z" rel="http://palaceproject.io/terms/rel/sort" title="Sort by Z"  palace:active-sort="true"  />

...
</feed>
```
I'm not sure if this is the right approach or whether it would be cleaner to do a more extensive refactor.  

I'd like feedback  on 1) the new terms and how they appear in the feed: is this what the clients would find useful?  2) should I be refactoring the FeedData class to remove the "Sort by" facet group all together?  Or should we do that as a later revision once the client apps start using OPDS2? 

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1507

## Checklist

### DIFF
--- a/src/palace/manager/feed/serializer/opds.py
+++ b/src/palace/manager/feed/serializer/opds.py
@@ -55,6 +55,15 @@ AUTHOR_MAPPING = {
 }
 
 
+def is_sort_link(link: Link) -> bool:
+    """A until method that determines if the specified link is a sort link"""
+    return (
+        hasattr(link, "facetGroup")
+        and link.facetGroup
+        == FacetConstants.GROUP_DISPLAY_TITLES[FacetConstants.ORDER_FACET_GROUP_NAME]
+    )
+
+
 class OPDS1Serializer(SerializerInterface[etree._Element], OPDSFeed):
     """An OPDS 1.2 Atom feed serializer"""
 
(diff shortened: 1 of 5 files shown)


dbernstein:
@mauricecarrier7 and @io7m :  I've updated the description above to include what you can expect to see in the revised feed re new sorting links.  Let me know if the approach I took is going to work for you on the client end.  An of course if you'd like me to change anything, don't hold back.

io7m:
I can't see any obvious reason why this would be a problem.

mauricecarrier7:
Same, looks good to me. I'm not clear on the extent of what will need to change when we move to OPDS2, but holding on major refactoring until then sounds like a good approach.

dbernstein:
I created a follow on ticket to refactor when we remove sorting facets altogether.

jonathangreen (approved):

PR #2101, opened 2024-10-04
Update the fixture for creating collections

jonathangreen:
## Description

Update the fixture for creating collections, similar to the integration configuration fixture, so its easy to just pass a class name in as the protocol when creating a collection. Also remove the set_settings function, and replace it with the new collection functionality.

## Motivation and Context

Needed to create a number of collections for testing and getting the protocol names was more difficult then it should have been.

## How Has This Been Tested?

- Tested in CI

## Checklist


tdilauro (approved):
Looks good to me! 🛠️

PR #2117, opened 2024-10-16
Update ODL hold calculation logic (PP-1728)

jonathangreen:
## Description

Completely re-work how we are calculating the hold queue for OPDS2 + ODL items. 

All the work to calculate the holds is moved out to live in a Celery task. When a hold is invoked through the API the hold is always placed in the queue, and the queue is later recalculated by the celery task.

## Motivation and Context

Previously the holds calculation was based on the number of loans that this CM had for an item. This is a problem is a loan gets lost, or if the collection is added to multiple CM. Now we use the number of licenses available, which should be safer and less error prone. However it may still be incorrect if the collection is on multiple CM, as these CM will be fighting for the same hold.

## How Has This Been Tested?

- Did some minor local testing
- Unit tests

## Checklist


dbernstein (approved):
Whoa:  big PR.  Looks good to the extent that I fully grocked it.  Way to move another task out of the scripts and into celery.

PR #2122, opened 2024-10-18
Update classifier code for mypy

jonathangreen:
## Description

- Remove all star imports for classifier code `from palace.manager.core.classifier import *`
  - These were the last star imports we had 🎉 
- Add a `__getattr__` method to `classifier/__init__.py` to let mypy know to expect dynamic imports
- Move around some of the classifier code to break import cycle
- Remove duplicated `GradeLevelClassifier` class

## Motivation and Context

- Trying to get to a place where we can toggle on `check_untyped_defs` mypy option.

## How Has This Been Tested?

- Running in CI

## Checklist


src/palace/manager/core/classifier/__init__.py
```diff
@@ -316,139 +311,6 @@ def and_up(cls, young, keyword):
         return old
 
 
-class GradeLevelClassifier(Classifier):
```
jonathangreen, line 319: This class was duplicated, and this copy wasn't being used, so I deleted it.

src/palace/manager/core/classifier/__init__.py
```diff
@@ -791,14 +653,25 @@ def add_genre(
...
+# to make them available as attributes of the module. This lets mypy know that
+# unknown attributes are actually GenreData objects.
+#
+# See this stackoverflow answer for more details:
+# https://stackoverflow.com/questions/60739889/bypass-mypys-module-has-no-attribute-on-dynamic-attribute-setting
+#
+# TODO: Eventually I'd like to refactor this, so we don't have to use __getattr__
+#   here, and can just import the GenreData objects directly.
+def __getattr__(name: str) -> GenreData:
+    return genres_by_variable_name[name]
```
jonathangreen, line 674: This is the core of the change in this PR.

src/palace/manager/core/classifier/__init__.py
```diff
@@ -821,710 +694,3 @@ def __new__(cls, value):
     def scrub_identifier(cls, identifier):
         if not identifier:
             return identifier
-
-
-class AgeOrGradeClassifier(Classifier):
```
jonathangreen, line 826: These classes were just moved out to `age.py`, to break an import cycle.

tests/manager/core/classifiers/test_classifier.py
```diff
@@ -1187,10 +1183,8 @@ def test_staff_target_age_overrides_others(
         staff_source = DataSource.lookup(session, DataSource.LIBRARY_STAFF)
         subject1 = data.transaction.subject(type="type1", identifier="subject1")
         subject1.target_age = NumericRange(6, 8, "[)")
-        subject1.weight_as_indicator_of_target_age = 1
         subject2 = data.transaction.subject(type="type2", identifier="subject2")
         subject2.target_age = NumericRange(6, 8, "[)")
-        subject2.weight_as_indicator_of_target_age = 1
```
jonathangreen, line 1193: `weight_as_indicator_of_target_age` isn't a property that is defined, and wasn't necessary to set for the test, so I just removed these properties.

jonathangreen:
Codecov is complaining, but any code that isn't covered by this, wasn't covered before, so I think this is fine for review without adding more coverage.

tdilauro (approved):
Cleaning up! 🧹🎸

Nice to get most of that stuff out of the `__init__` and to get rid of those `*` imports!

PR #2123, opened 2024-10-18
Replace sqlalchemy backref parameter with back_populates

jonathangreen:
## Description

This PR updates all references to `backref` to use `back_populates` as sqlalchmey recommends as the current best practice. This PR also tweaks some of the flags we use for mypy only in our tests, to disable some of the mypy warnings that aren't very meaningful for tests.

## Motivation and Context

The sqlalchemy `backref` parameter is considered legacy see: https://docs.sqlalchemy.org/en/14/orm/backref.html. Using this construct means that mypy (and IDE) can't infer the type of the dynamically generated parameter. 

This is part of the effort to get the code to a place where we can enable `check_untyped_defs` in mypy.

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/tests/manager/api/controller/test_work.py
+++ b/tests/manager/api/controller/test_work.py
@@ -79,12 +79,11 @@ class TestWorkController:
         # Find a real Contributor put in the system through the setup
         # process.
         [contribution] = work_fixture.english_1.presentation_edition.contributions
-        contributor = contribution.contributor
 
         # The contributor is created with both .sort_name and
         # .display_name, but we want to test what happens when both
         # pieces of data aren't avaiable, so unset .sort_name.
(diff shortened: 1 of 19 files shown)


tests/manager/api/controller/test_work.py
tdilauro, line 1: These changes didn't seem strictly necessary, but I don't see any problem with them.

tdilauro (approved):
Looks good! 🔧

tests/manager/api/controller/test_work.py
jonathangreen: Hmm... I forget why i changed this file... maybe for mypy? I'm not sure now, i guess i should have put a comment on this one.

PR #2124, opened 2024-10-18
Update mypy configuration for tests

jonathangreen:
## Description

Update mypy configuration for our tests to:
- Disable `union-attr` check
- Disable `warn-unreachable`

## Motivation and Context

Both of these flags don't provide a lot of value in our tests, and cause us to have to add a bunch of ignore comments, so just disabling them seems like a better option.

## How Has This Been Tested?

- Mypy / tests in CI

## Checklist


tdilauro (approved):
LGTM!

PR #2125, opened 2024-10-18
Run autoflake without ignore-init-module-imports

jonathangreen:
## Description

Now that we aren't using star imports for the classifier code we can run [autoflake](https://github.com/PyCQA/autoflake) without the `ignore-init-module-imports` option. This cleaned up some unneeded imports, and will catch future changes.

## Motivation and Context

🧹 

## How Has This Been Tested?

- Running tests

## Checklist


tdilauro (approved):
🚀

PR #2129, opened 2024-10-21
Handle StaleDataError for ODLNotificationController

jonathangreen:
## Description

Catch `StaleDataError` when processing ODL update. This can happen when a loan gets deleted by another process while the ODL notification controller is running. This has the possiblity of happening when we are locally deleting the loan for a returned book, and also receive a callback from the LCP server.

## Motivation and Context

Handle a fairly rare race condition with the `ODLNotificationController`.  I've seen it happen in my testing a couple of times, but not every time.

## How Has This Been Tested?

- Running tests locally

## Checklist


tdilauro (approved):
LGTM!

PR #2130, opened 2024-10-22
Use PG 16 in our CI workflow

jonathangreen:
## Description

Since we are updating all our environments to use PG 16 instead of PG 12, we should also update the docker image we are using for PG in CI.

## Motivation and Context

Making sure our CI workflows match what we have deployed.

## How Has This Been Tested?

- Running CI

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ docker restart opensearch
 #### Docker
 
 ```sh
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
🏃🏽

PR #2132, opened 2024-10-23
Handle two exceptions causing sync_patron_activity job to fail (PP-1838)

jonathangreen:
## Description

Handle `PatronAuthorizationFailedException` and `RemoteIntegrationException` so they do not cause the `sync_patron_activity` job to fail.

The `PatronAuthorizationFailedException` that we are getting come mostly from the Enki collection. This appears to happen when the patron doesn't have any loans for Enki items.

The `RemoteIntegrationException` errors that we are seeing are mostly 504 responses coming back from Overdrive. These responses tend to come in waves, so I think it makes sense to do a few retries to see if we can successfully perform a sync.

## Motivation and Context

When looking at setting up alerting for failing celery tasks, I saw that a large number of tasks were failing. In the last week I saw these failures:

| Task Name | Exception | Count |
| --- | --- | --- | 
| patron_activity.sync_patron_activity | PatronAuthorizationFailedException | 1719 |
| patron_activity.sync_patron_activity | RemoteIntegrationException | 231 |

Since these errors are mostly unavoidable, we don't want the job to fail because of them.

## How Has This Been Tested?

- Tests running in CI

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/patron_activity.py
+++ b/src/palace/manager/celery/tasks/patron_activity.py
@@ -1,6 +1,7 @@
 from celery import shared_task
 
 from palace.manager.api.circulation import PatronActivityCirculationAPI
+from palace.manager.api.circulation_exceptions import PatronAuthorizationFailedException
 from palace.manager.celery.task import Task
 from palace.manager.service.celery.celery import QueueNames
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! ✅

PR #2134, opened 2024-10-23
Fix get_one_or_create IntegrityError handler (PP-1838)

jonathangreen:
## Description

Fix an issue with the Exception handler for an `IntegrityError` in the `get_one_or_create` function. 

If one of the parameters for this function is a sqlalchemy class, then the logging function can cause a `PendingRollbackError`, since the rollback is happening after the log message. 

Update the function to use db.begin_nested() as a context manager, and write a test for this case. 

I added type hints to a couple related functions while I was debugging this before i narrowed in on exactly the cause. These hints are included in this PR. They aren't really related to this fix, but they seem harmless to come in with this one. I can break them out to separate PR if desired.
 
## Motivation and Context

We are seeing a number of `PendingRollbackError` happening via the `patron_activity.sync_patron_activity` celery task. 282 in the last week. Trying to clean up these errors that are causing tasks to fail, so we can add alerts for failed celery tasks.

## How Has This Been Tested?

- Added test for this case
- Running tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/api/overdrive.py
+++ b/src/palace/manager/api/overdrive.py
@@ -904,7 +904,7 @@ class OverdriveAPI(
             return response
 
     def get_patron_credential(
-        self, patron: Patron, pin: str | None, is_fulfillment=False
+        self, patron: Patron, pin: str | None, is_fulfillment: bool = False
     ) -> Credential:
         """Create an OAuth token for the given patron.
 
@@ -913,7 +913,7 @@ class OverdriveAPI(
         :param is_fulfillment: Boolean indicating whether we need a fulfillment credential.
         """
 
-        def refresh(credential):
+        def refresh(credential: Credential) -> Credential:
             return self.refresh_patron_access_token(
                 credential, patron, pin, is_fulfillment=is_fulfillment
             )
--- a/src/palace/manager/sqlalchemy/model/credential.py
+++ b/src/palace/manager/sqlalchemy/model/credential.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import datetime
 import uuid
-from typing import TYPE_CHECKING
+from collections.abc import Callable
+from typing import TYPE_CHECKING, Any
 
 from sqlalchemy import Column, DateTime, ForeignKey, Index, Integer, String
 from sqlalchemy.orm import Mapped, relationship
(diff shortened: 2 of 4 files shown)


src/palace/manager/api/overdrive.py
```diff
@@ -913,7 +913,7 @@ def get_patron_credential(
         :param is_fulfillment: Boolean indicating whether we need a fulfillment credential.
         """
 
-        def refresh(credential):
+        def refresh(credential: Credential) -> Credential:
```
jonathangreen, line 916: Unrelated type hint changes.

src/palace/manager/sqlalchemy/model/credential.py
```diff
@@ -120,15 +121,15 @@ def _filter_invalid_credential(
...
-        force_refresh=False,
+        _db: Session,
+        data_source: DataSource | str,
+        token_type: str,
+        patron: Patron | None,
+        refresher_method: Callable[[Credential], Any] | None,
+        allow_persistent_token: bool = False,
+        allow_empty_token: bool = False,
+        collection: Collection | None = None,
+        force_refresh: bool = False,
```
jonathangreen, line 132: Unrelated type hint changes.

src/palace/manager/sqlalchemy/util.py
```diff
@@ -121,25 +123,23 @@ def get_one_or_create(
...
-            __transaction.commit()
-            return obj
+            with db.begin_nested():
+                # These kwargs are supported by get_one() but not by create().
+                get_one_keys = ["on_multiple", "constraint"]
+                for key in get_one_keys:
+                    if key in kwargs:
+                        del kwargs[key]
+                obj = create(db, model, create_method, create_method_kwargs, **kwargs)
+                return obj
```
jonathangreen, line 134: this is the actual fix

tests/manager/sqlalchemy/test_util.py
```diff
@@ -45,6 +52,46 @@ def test_get_one(self, db: DatabaseTransactionFixture):
...
+        # it will be handled, an error message logged, and the existing
+        # object returned
+        caplog.clear()
+        caplog.set_level(LogLevel.debug)
+        with patch.object(util, "get_one", return_value=None):
+            credential_existing, is_new = get_one_or_create()
+        assert not is_new
+        assert isinstance(credential_existing, Credential)
+        assert credential_existing == credential
+        assert "INTEGRITY ERRROR" in caplog.text
```
jonathangreen, line 93: Test for the `IntegrityError` case

tdilauro (approved):
🚀

PR #2142, opened 2024-10-28
[PP-1509] Add support for default facet

dbernstein:
## Description

With this update,  the "default" facet or sort link will have a http://palaceproject.io/terms/properties/default property to indicate that it is a default property.  This default property will be present in both OPDS1 and OPDS2 feeds. However in order to see them, clients must specify an api-version of 2  in the mimetype.  

For OPDS 1: 
application/atom+xml;api-version=2
For OPDS 2:
application/opds+json;api-version=2

Otherwise version 1 of both formats will be delivered.  Version 1 is what is currently deployed in production.

Here is an [update to the documentation](https://www.notion.so/lyrasis/OPDS-Clients-3942ac215afb4b3aa4229df74aed938f?pvs=4#12e24d4ececa80728143cadd7ccf974e) that covers these updates.

I do have an open question about whether or not a default facet that happens to be active should have an "activeFacet" or "active-sort" set to true.

## Motivation and Context

https://ebce-lyrasis.atlassian.net/browse/PP-1506

## Checklist

### DIFF
--- a/src/palace/manager/feed/serializer/opds.py
+++ b/src/palace/manager/feed/serializer/opds.py
@@ -64,7 +64,7 @@ def is_sort_link(link: Link) -> bool:
     )
 
 
-class OPDS1Serializer(SerializerInterface[etree._Element], OPDSFeed):
+class OPDS1Version1Serializer(SerializerInterface[etree._Element], OPDSFeed):
     """An OPDS 1.2 Atom feed serializer"""
 
     def __init__(self) -> None:
@@ -415,3 +415,11 @@ class OPDS1Serializer(SerializerInterface[etree._Element], OPDSFeed):
         if link.get("activeFacet", False):
             sort_link.add_attributes(dict(activeSort="true"))
         return self._serialize_feed_entry("link", sort_link)
+
+
+class OPDS1Version2Serializer(OPDS1Version1Serializer):
+    """An OPDS 1.2 Atom feed serializer with Palace specific modifications (version 2) to support
+    new IOS and Android client features."""
+
+    def __init__(self) -> None:
+        pass
--- a/src/palace/manager/feed/serializer/opds.py
+++ b/src/palace/manager/feed/serializer/opds.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import abc
 import datetime
 from functools import partial
 from typing import Any, cast
 
 from lxml import etree
+from typing_extensions import override
 
 from palace.manager.core.facets import FacetConstants
 from palace.manager.feed.serializer.base import SerializerInterface
@@ -33,7 +35,7 @@ TAG_MAPPING = {
     "hashed_passphrase": f"{{{OPDSFeed.LCP_NS}}}hashed_passphrase",
 }
 
-ATTRIBUTE_MAPPING = {
+V1_ATTRIBUTE_MAPPING = {
     "vendor": f"{{{OPDSFeed.DRM_NS}}}vendor",
     "scheme": f"{{{OPDSFeed.DRM_NS}}}scheme",
     "username": f"{{{OPDSFeed.SIMPLIFIED_NS}}}username",
@@ -43,8 +45,12 @@ ATTRIBUTE_MAPPING = {
     "facetGroup": f"{{{OPDSFeed.OPDS_NS}}}facetGroup",
     "facetGroupType": f"{{{OPDSFeed.SIMPLIFIED_NS}}}facetGroupType",
     "activeFacet": f"{{{OPDSFeed.OPDS_NS}}}activeFacet",
-    "defaultFacet": f"{{{OPDSFeed.PALACE_PROPS_NS}}}default",
     "ratingValue": f"{{{OPDSFeed.SCHEMA_NS}}}ratingValue",
+}
+
+V2_ATTRIBUTE_MAPPING = {
+    **V1_ATTRIBUTE_MAPPING,
+    "defaultFacet": f"{{{OPDSFeed.PALACE_PROPS_NS}}}default",
     "activeSort": f"{{{OPDSFeed.PALACE_PROPS_NS}}}active-sort",
 }
 
@@ -56,8 +62,8 @@ AUTHOR_MAPPING = {
 }
 
 
-def is_sort_link(link: Link) -> bool:
-    """A until method that determines if the specified link is a sort link"""
+def is_sort_facet(link: Link) -> bool:
+    """A until method that determines if the specified link is part of a sort facet"""
     return (
         hasattr(link, "facetGroup")
         and link.facetGroup
@@ -65,9 +71,7 @@ def is_sort_link(link: Link) -> bool:
     )
 
 
-class OPDS1Version1Serializer(SerializerInterface[etree._Element], OPDSFeed):
-    """An OPDS 1.2 Atom feed serializer"""
-
+class BaseOPDS1Serializer(SerializerInterface[etree._Element], OPDSFeed):
     def __init__(self) -> None:
         pass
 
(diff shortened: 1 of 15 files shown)


jonathangreen (commented):
Just a few comments to look at before this one goes in.

Can you also update the PR with a small example of what these links look like in the different feed versions?

dbernstein:
@jonathangreen :  thanks for the review.  All your suggests look reasonable to me.  I'll update the PR  today.

tests/manager/feed/test_opds_base.py
```diff
@@ -13,45 +19,93 @@ def test_get_serializer(self):
...
+            headers=dict(Accept="application/opds+json;api-version=1")
+        )
+
+        assert isinstance(
+            get_serializer(request.accept_mimetypes), OPDS2Version1Serializer
+        )
+
+        request = Request.from_values(
+            headers=dict(Accept="application/opds+json;api-version=2")
+        )
```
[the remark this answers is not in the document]
dbernstein: I added this one in, but it resolves to OPDS1Version1Serializer.  The accept header matching logic is in the werzeug.datastructures.accept pacakge.

src/palace/manager/feed/serializer/opds.py
```diff
@@ -55,18 +62,16 @@
...
         hasattr(link, "facetGroup")
         and link.facetGroup
         == FacetConstants.GROUP_DISPLAY_TITLES[FacetConstants.ORDER_FACET_GROUP_NAME]
     )
 
 
-class OPDS1Serializer(SerializerInterface[etree._Element], OPDSFeed):
-    """An OPDS 1.2 Atom feed serializer"""
-
+class BaseOPDS1Serializer(SerializerInterface[etree._Element], OPDSFeed):
```
jonathangreen, line 74: You also need to inherent from ABC in order to make this an abstract base class.

See: 
https://docs.python.org/3/library/abc.html#abc.ABC

src/palace/manager/feed/serializer/opds.py
```diff
@@ -79,7 +84,7 @@ def _tag(
 
     def _attr_name(self, attr_name: str, mapping: dict[str, str] | None = None) -> str:
         if not mapping:
-            mapping = ATTRIBUTE_MAPPING
+            mapping = V1_ATTRIBUTE_MAPPING
```
jonathangreen, line 87: Since we defined `_get_attribute_mapping`, shouldn't this be calling that function?

src/palace/manager/feed/serializer/opds.py
```diff
@@ -118,13 +119,11 @@ def serialize_feed(
...
-                serialized.append(self._serialize_sort_link(link))
-                # TODO once the clients are no longer relying on facet based sorting
-                # an "else" should be introduced here since we only need to provide one style of sorting links
-            serialized.append(self._serialize_feed_entry("link", link))
-            # TODO end else here
+        for link in self._serialize_facet_links(feed.facet_links):
+            serialized.append(link)
+
+        for link in self._serialize_sort_links(feed.facet_links):
+            serialized.append(link)
```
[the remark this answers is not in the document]
jonathangreen: Still doesn't look like there is any coverage here?

src/palace/manager/feed/serializer/opds.py
```diff
@@ -314,12 +317,17 @@ def _serialize_feed_entry(
...
+                    attribute_mapping = self._get_attribute_mapping()
                     entry.set(
-                        ATTRIBUTE_MAPPING.get(attrib, attrib),
+                        attribute_mapping.get(attrib, attrib),
                         value if value is not None else "",
                     )
         return entry
 
+    @abc.abstractmethod
+    def _get_attribute_mapping(self) -> dict[str, str]:
```
jonathangreen, line 328: Minor: Since subclasses are expected to implement these, its would be nice to have a doc comment with a small bit of explanation of what an implementation should provide for this function.

src/palace/manager/feed/serializer/opds.py
```diff
@@ -405,13 +413,95 @@ def _serialize_data_entry(self, entry: DataEntry) -> etree._Element:
     def to_string(cls, element: etree._Element) -> str:
         return cast(str, etree.tostring(element, encoding="unicode"))
 
+    @abc.abstractmethod
     def content_type(self) -> str:
```
jonathangreen, line 417: Minor: Since subclasses are expected to implement these, its would be nice to have a doc comment with a small bit of explanation of what an implementation should provide for this function.

src/palace/manager/feed/serializer/opds.py
```diff
@@ -405,13 +413,95 @@ def _serialize_data_entry(self, entry: DataEntry) -> etree._Element:
...
-        return OPDSFeed.ACQUISITION_FEED_TYPE
+        pass
+
+    @abc.abstractmethod
+    def _serialize_facet_links(self, feed: FeedData) -> list[Link]:
```
jonathangreen, line 421: Minor: Since subclasses are expected to implement these, its would be nice to have a doc comment with a small bit of explanation of what an implementation should provide for this function.

src/palace/manager/feed/serializer/opds.py
```diff
@@ -405,13 +413,95 @@ def _serialize_data_entry(self, entry: DataEntry) -> etree._Element:
...
+        pass
+
+    @abc.abstractmethod
+    def _serialize_sort_links(self, feed: FeedData) -> list[Link]:
```
jonathangreen, line 425: Minor: Since subclasses are expected to implement these, its would be nice to have a doc comment with a small bit of explanation of what an implementation should provide for this function.

src/palace/manager/feed/serializer/opds.py
```diff
@@ -405,13 +413,95 @@ def _serialize_data_entry(self, entry: DataEntry) -> etree._Element:
...
+        links: list[Link] = []
+        facet_links = feed.facet_links
+        if facet_links:
+            for link in feed.facet_links:
+                # select only the sort facets for serialization
+                if is_sort_facet(link):
+                    links.append(self._serialize_sort_link(link))
+        return links
 
     def _serialize_sort_link(self, link: Link) -> etree._Element:
```
jonathangreen, line 488: We aren't using the `@override` decorator anywhere else in the codebase currently. I think if we're going to apply it in these classes we should apply it consistently, so it should get added to this method as well.

tests/manager/feed/test_opds_base.py
```diff
@@ -1,57 +1,64 @@
...
+            [
+                "application/atom+xml;profile=opds-catalog;kind=acquisition;q=0.08, application/opds+json;api-version=2;q=0.84, application/atom+xml;profile=opds-catalog;kind=acquisition;api-version=2;q=0.9",
+                OPDS1Version1Serializer,
+            ],
+            # No valid accept mimetype should default to OPDS1.x
+            ["text/html", OPDS1Version1Serializer],
+        ],
+    )
+    def test_get_serializer1(self, accept_header: str, serializer):
+        # The q-value should take priority
```
jonathangreen, line 62: Minor: Seems like this comment is outdated.

tests/manager/api/controller/test_loan.py
```diff
@@ -131,7 +131,10 @@ class OPDSSerializationTestHelper:
             (None, OPDSFeed.ENTRY_TYPE),
             ("default-foo-bar", OPDSFeed.ENTRY_TYPE),
             (AtomFeed.ATOM_TYPE, OPDSFeed.ENTRY_TYPE),
-            (OPDS2Serializer.CONTENT_TYPE, OPDS2Serializer.CONTENT_TYPE),
+            (
+                OPDS2Serializer.CONTENT_TYPE,
+                OPDS2Serializer.CONTENT_TYPE,
+            ),
```
jonathangreen, line 137:
```suggestion
            (OPDS2Serializer.CONTENT_TYPE, OPDS2Serializer.CONTENT_TYPE),
```

Minor: This change seems unnecessary. Its nice to have these be consistent, instead of having this one entry on multiple lines.

jonathangreen (approved):
A couple more comments here that would be nice to get resolved before merging @dbernstein.

PR #2145, opened 2024-10-30
Stop using Python money

jonathangreen:
## Description

The Python [money](https://pypi.org/project/money/) package is the only one that is stopping us from testing with Python 3.12. Its effectively abandoned, its last update was in 2016. This PR moves to using the python `decimal.Decimal` objects instead, since Money was just a small wrapper around this class anyway.

## Motivation and Context

Allow us to run tests against Python 3.12.

## How Has This Been Tested?

- Running tests in CI

## Checklist


tdilauro (approved):
🔧

PR #2146, opened 2024-10-30
Add Python 3.12 to CI

jonathangreen:
## Description

Run tests against Python 3.12.

## Motivation and Context

Make sure we are compatible for upgrade when the time comes 🚀 

## How Has This Been Tested?

- Running tests in CI

## Checklist


tdilauro (approved):
🍄

PR #2147, opened 2024-10-31
Clear PatronData.authorization_expires if it is no longer set

jonathangreen:
## Description

If a ILS once reported that a patron has an expiry, but stops reporting that date, currently we don't clear the previous information. This change will let us clear that info, so patrons don't get stuck with an expired card that the ILS cannot clear.

## Motivation and Context

Probably the issue in: https://ebce-lyrasis.atlassian.net/browse/SUP-232

## How Has This Been Tested?

- Tests in CI

## Checklist


tdilauro (approved):
🎉

PR #2148, opened 2024-10-31
[PP-1818] implement month active user data

dbernstein:
## Description
This update includes the following changes: 
1. The patron object now has a uuid field.
2. Patrons can erase their activity history with a PUT to  the `/patrons/me/erase_activity_history` endpoint.
3. The uuid is passed to the s3 analytics provider.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1818

## Checklist

### DIFF
--- a/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
+++ b/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
@@ -22,7 +22,29 @@ depends_on = None
 def upgrade() -> None:
     op.add_column(
         "patrons",
-        sa.Column("uuid", UUID(as_uuid=True), nullable=False, default=uuid.uuid4),
+        sa.Column("uuid", UUID(as_uuid=True), nullable=True, default=uuid.uuid4),
+    )
+
+    conn = op.get_bind()
+    rows = conn.execute("SELECT id from patrons").all()
+
+    for row in rows:
+        conn.execute(
+            """
+            UPDATE patrons
+            SET uuid = %(uuid)s
+            WHERE id = %(id)s
+            """,
+            {
+                "id": row.id,
+                "uuid": uuid.uuid4(),
+            },
+        )
+
+    op.alter_column(
+        table_name="patrons",
+        column_name="uuid",
+        nullable=True,
     )
 
 
--- a/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
+++ b/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
@@ -1,0 +1,30 @@
+"""Add UUID to patron table
+
+Revision ID: 272da5f400de
+Revises: 1938277e993f
+Create Date: 2024-10-30 17:41:28.151677+00:00
+
+"""
+
+import uuid
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import UUID
+
+# revision identifiers, used by Alembic.
+revision = "272da5f400de"
+down_revision = "1938277e993f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "patrons",
+        sa.Column("uuid", UUID(as_uuid=True), nullable=False, default=uuid.uuid4),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("patrons", "uuid")
--- a/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
+++ b/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
@@ -44,7 +44,7 @@ def upgrade() -> None:
     op.alter_column(
         table_name="patrons",
         column_name="uuid",
-        nullable=True,
+        nullable=False,
     )
 
 
--- a/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
+++ b/alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
@@ -29,16 +29,13 @@ def upgrade() -> None:
     rows = conn.execute("SELECT id from patrons").all()
 
     for row in rows:
+        uid = str(uuid.uuid4())
         conn.execute(
-            """
-            UPDATE patrons
-            SET uuid = %(uuid)s
-            WHERE id = %(id)s
-            """,
-            {
-                "id": row.id,
-                "uuid": uuid.uuid4(),
-            },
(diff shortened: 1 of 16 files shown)


src/palace/manager/api/controller/patron_activity_history.py
```diff
@@ -0,0 +1,19 @@
...
+from flask import Response
+
+from palace.manager.api.controller.circulation_manager import (
+    CirculationManagerController,
+)
+
+
+class PatronActivityHistoryController(CirculationManagerController):
+
+    def erase(self):
```
jonathangreen, line 15: Should have a type hint here: `-> Response:`

I've been trying to add any new code / controllers that we write to the pyproject.toml mypy config https://github.com/ThePalaceProject/circulation/blob/ebcddcab60487e7f9a763b27296c39aca04f2649/pyproject.toml#L58-L69

So they are strictly type checked from the beginning. Maybe this new controller can be added there.

src/palace/manager/api/routes.py
```diff
@@ -332,6 +332,15 @@ def patron_profile():
     return app.manager.profiles.protocol()
 
 
+@library_dir_route("/patrons/me/erase_activity_history", methods=["PUT"])
```
jonathangreen, line 335: This isn't really erasing anything. I think `patrons/me/reset_statistics_uuid` might be a more descriptive endpoint here.

src/palace/manager/api/s3_analytics_provider.py
```diff
@@ -25,13 +26,14 @@ def __init__(self, s3_service: S3Service | None):
...
     def _create_event_object(
         library: Library,
-        license_pool: LicensePool,
+        license_pool: LicensePool | None,
         event_type: str,
         time: datetime.datetime,
-        old_value,
-        new_value,
+        old_value: Any = None,
+        new_value: Any = None,
```
jonathangreen, line 33: Minor: It seems like we can probably do better for a type annotation then `Any` here. I really try to avoid `Any` as much as possible. It seems like these are probably `int | None` right?

src/palace/manager/api/s3_analytics_provider.py
```diff
@@ -132,26 +134,25 @@ def _create_event_object(
...
-        event_type,
-        time,
-        old_value=None,
-        new_value=None,
+        library: Library,
+        license_pool: LicensePool | None,
+        event_type: str,
+        time: datetime.datetime,
+        old_value: Any = None,
+        new_value: Any = None,
```
jonathangreen, line 149: Minor: Same comment as above about using `Any` here.

src/palace/manager/api/s3_analytics_provider.py
```diff
@@ -175,9 +173,12 @@ def collect_event(
...
         :param user_agent: The user_agent of the caller.
         :type user_agent:  str
+
+        :param patron: The patron associated with the event, where applicable
+        :type patron: Patron
         """
 
-        if not library and not license_pool:
+        if not library:
             raise ValueError("Either library or license_pool must be provided.")
```
jonathangreen, line 182: I'm not clear on why we are able to relax this conditional check. Was there some change that made it possible to drop the check if `license_pool` is None here? 

If we are going to change this condition, we should update the text for the exception that gets raised, since we are no longer checking for `license_pool` or even drop this `if` entirely like you did for the local analytics provider? If the type hints are accurate here (no idea if they are) then `library` cannot be `None`, so `if not library` is always going to be `False`.

tests/manager/api/controller/test_patron_activity_history.py
```diff
@@ -0,0 +1,45 @@
...
+
+import pytest
+
+from palace.manager.sqlalchemy.model.patron import Patron
+from tests.fixtures.api_controller import ControllerFixture
+from tests.fixtures.database import DatabaseTransactionFixture
+from tests.fixtures.services import ServicesFixture
+
+
+class PatronActivityHistoryFixture(ControllerFixture):
```
jonathangreen, line 11: `ControllerFixture` does a whole lot of setup that you probably don't need here. In any new fixtures we have been writing, I'm trying not to inherit from that class.

You should be able to use `FlaskAppFixture` with a small fixture that instantiates the controller to test this. A good example is the tests for [test_discovery_services.py](https://github.com/ThePalaceProject/circulation/blob/ebcddcab60487e7f9a763b27296c39aca04f2649/tests/manager/api/admin/controller/test_discovery_services.py).

tests/manager/api/controller/test_patron_activity_history.py
```diff
@@ -0,0 +1,45 @@
...
+            )
+            assert isinstance(patron, Patron)
+            uuid1 = patron.uuid
+            assert uuid1
+            response = (
+                patron_activity_history_fixture.manager.patron_activity_history.erase()
+            )
+            uuid2 = patron.uuid
+            assert uuid1 != uuid2
+            assert 200 == response.status_code
```
jonathangreen, line 45: I know we have these reversed assertions all over the place in our code, due to us migrating from a different testing framework, but for new tests pytest expects the first value given in the assertion to be the actual value, and the second value to be the expected value. 

```python
assert response.status_code == 200
```

alembic/versions/20241030_272da5f400de_add_uuid_to_patron_table.py
```diff
@@ -0,0 +1,52 @@
...
+                "id": row.id,
+                "uuid": uuid.uuid4(),
+            },
+        )
+
+    op.alter_column(
+        table_name="patrons",
+        column_name="uuid",
+        nullable=True,
+    )
```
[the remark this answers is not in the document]
dbernstein: Yes - good catch - it is supposed to be false.

tdilauro:
You may have considered this already, but while looking into PP-1863, I was reminded that we have a credential type of [Credential.IDENTIFIER_TO_REMOTE_SERVICE](https://github.com/ThePalaceProject/circulation/blob/a1d5c6149b43ed0043e15c5300861c2bb8569fec/src/palace/manager/sqlalchemy/model/credential.py#L91) that might be suitable for this purpose or as a template for associating an identifier with a patron, but without modifying the patron table.

Sorry to lob in a hand grenade 💣 at this stage. Feel free to do -- or not do ("There is no try?") -- what you will with this information.

dbernstein:
@tdilauro :  I appreciate your suggestion.  And I like it.  I can't think of a good reason not to do it that way other than that I did it this way and it works and is done and I feel like I need to move on.  But I suppose if we wanted to rework it in the future it would not be a heavy lift to do it.

jonathangreen (approved):
Looks good! 🎸

PR #2155, opened 2024-11-05
Fix Enki integration URL generation (PP-1876)

jonathangreen:
## Description

It looks like the URL setting changes when we upgraded Pydantic were causing the Enki integration to generate incorrect URLs (missing slash). I added a helper function to generate the URL and added a test for this case. 

## Motivation and Context

Enki integration not working, see: PP-1876.

## How Has This Been Tested?

- Unit tests
- Tested locally

## Checklist


tdilauro (approved):
/ 🔧

PR #2159, opened 2024-11-08
Use SQL formatting that aligns with other alembic scripts.

dbernstein:
## Description
Brings the sql parameter substitution style in line with other alembic scripts. 

## Motivation and Context
After trying to update the containers on minotaur, the most recent alembic script was failing, but only on minotaur.  The exception was complaining about the SQL statement.

I ensured that this update both passes the test_migrations.sh as well as ensured that when we upgrade the database to the current version locally when initializing with an existing database with at least one patron that this alembic script works as expected.

## Checklist


jonathangreen (commented):
👍

PR #2163, opened 2024-11-13
Replace webpub manifest parser with pydantic

jonathangreen:
## Checklist


src/palace/manager/core/opds2_import.py
```diff
@@ -396,53 +321,48 @@ def _extract_contributors(
...
-                continue
+            if isinstance(contributor, rwpm.ContributorWithRole):
+                roles = self._extract_contributor_roles(contributor.roles, default_role)
+            else:
+                roles = [default_role]
 
             contributor_metadata = ContributorData(
                 sort_name=contributor.sort_as,
-                display_name=contributor.name,
+                display_name=str(contributor.name),
```
tdilauro, line 334: Hmmm. Maybe just food for _future_ thought here, but this makes me realize that, at least for some properties (like contributors), it might not make sense for `name` to be a `LanguageMap`. But maybe it does (e.g., in the case of corporate authors, which might need a translation).

I note also that [the schema](https://readium.org/webpub-manifest/schema/contributor-object.schema.json) calls for a string or a language map.

tdilauro (approved):
This looks ready to roll! 🍣

PR #2168, opened 2024-11-14
Add Pydantic models for RWPM, OPDS2, and ODL (PP-1367)

jonathangreen:
## Description

Add pydantic models for Readium Webpub, OPDS2 and OPDS2 + ODL to CM codebase. 

I tried to call out in comments the spec document and json schema for every Pydantic model, so its easy to see why we are parsing things the way we are.

This PR just adds the models and some tests for them, it doesn't actually update the CM to use them. I'm going to update https://github.com/ThePalaceProject/circulation/pull/2163 so it just contains the update to use the models, so its a bit easier to review in two pieces.

Most of the type unions are setup to use [discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) for parsing speed and to make the errors easier to understand when parsing fails.

While doing this, I reorganized the pydantic models I started to add in https://github.com/ThePalaceProject/circulation/pull/2095:
- Updated `ListOfLinks` to `CompactCollection`
  - This better matches what the specs call this sort of object
- Updated some properties that used to return a `set` to return a `tuple` instead
  - Ordering shouldn't be important, but I wanted to standardize what we are returning across the OPDS pydantic models and a tuple seemed the most appropriate since it preserves order.
- I moved around some of the test fixture files I previously added to more appropriate locations.

## Motivation and Context

Replacing webpub manifest parser with Pydantic.

## How Has This Been Tested?

- Running tests in CI

## Checklist


src/palace/manager/opds/util.py
```diff
@@ -0,0 +1,91 @@
...
+    if isinstance(value, (list, tuple)):
+        return "list"
+    return None
+
+
+# Some type vars used in the TypeAlias definitions below.
+# Once https://github.com/pydantic/pydantic/issues/9418 is resolved
+# StrT could use a default as well as a bound, so we would only need
+# to specify it in cases where we are using a different string type.
+StrT = TypeVar("StrT", bound=str)
```
tdilauro, line 41: It looks like [the fix for this](https://github.com/pydantic/pydantic/pull/10789) has been merged, so soon, hopefully.

src/palace/manager/opds/palace.py
```diff
@@ -0,0 +1,91 @@
...
+    # Some DeMarque specific metadata we get in feeds from them
+    encoding_format: str | None = Field(None, alias="schema:encodingFormat")
+    work_example: list[WorkExample] = Field(
+        default_factory=list, alias="schema:workExample"
+    )
+
+    @field_validator("type")
+    @classmethod
+    def warning_when_type_is_not_valid(cls, type_: str) -> str:
+        if type_ not in [t for t in PublicationTypes]:
```
tdilauro, line 89: Minor: Looks like we could just use `list` here:
```python
    if type_ not in list(PublicationTypes):
        ...
```

tdilauro (approved):
Looks good! 🚀 

A few minor comments below, but could be merged as-is.

PR #2169, opened 2024-11-14
Ignore a couple more lines for codecov

jonathangreen:
## Description

It doesn't matter a great deal, but adding a `@overload` decorator is counted as a partial by codecov, since it never get actually called. Since they can never be called, it seems better to just exclude them.

## Motivation and Context

I noticed in https://github.com/ThePalaceProject/circulation/pull/2168 when I added a couple `@overload` for a function, these were counted as *partials* by coverage. 

## How Has This Been Tested?

Letting codecov run in CI.

## Checklist


tdilauro (approved):
🪛

PR #2173, opened 2024-11-19
Fix "Could not place hold" error with multiple licenses

jonathangreen:
## Description

When we have multiple licenses, it was possible for us to get out of sync with the provider, so we would try to place a hold, when a different license could have been tried instead. This updates the checkout to try all available licenses, in priority order, before resorting to a hold.

I wrote a new test for this case the exercises the new functionality. 

## Motivation and Context

JIRA: PP-1902 found when testing.

## How Has This Been Tested?

- Running tests in CI

## Checklist


tests/manager/api/odl/test_api.py
```diff
@@ -717,21 +717,99 @@ def test_checkout_no_available_copies_unknown_to_us(
...
+        assert license_2.checkouts_available == 0
+
+    def test_checkout_many_licenses(
+        self,
+        db: DatabaseTransactionFixture,
+        opds2_with_odl_api_fixture: OPDS2WithODLApiFixture,
+    ) -> None:
+        """
+        The title has 5 different licenses available, we are out of sync, so we think
+        some of them have copies available when they are not.
```
tdilauro, line 753: Found this comment a bit confusing.

Suggest something like:
> The title has 5 different licenses. Several of them seem to have copies available. But we are out of sync, so it turns out that not all of them do.

tdilauro (approved):
Looks good! 🥇 

This is a nice improvement to functionality!

One small suggestion on a comment below.

PR #2177, opened 2024-11-20
Helper function to get flask.request.library

jonathangreen:
## Description

Add a helper function `get_request_library()` to get `flask.request.library` in a type safe way. Update the existing code to use the new helper.

The bulk of the changes in this PR are updates to use the new function.

## Motivation and Context

While working on PP-1860, I'm handling some type errors, and I was trying to add type hints to make it clear if these errors happen again, but I keep getting type errors from `flask.request.library` attribute not existing. 

I don't love how we set and use `flask.request.library` all over the place, but its fairly embedded in the code at this point. Instead of constantly fighting with type errors from it, I thought it makes sense to isolate the untyped parts of it to this one function and make sure the edge cases of the function are well tested.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/work_editor.py
+++ b/src/palace/manager/api/admin/controller/work_editor.py
@@ -63,13 +63,14 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
 
         :return: An OPDSEntryResponse
         """
-        self.require_librarian(get_request_library())
+        library = get_request_library()
+        self.require_librarian(library)
 
-        work = self.load_work(get_request_library(), identifier_type, identifier)
+        work = self.load_work(library, identifier_type, identifier)
         if isinstance(work, ProblemDetail):
             return work
 
-        annotator = AdminAnnotator(self.circulation, get_request_library())
+        annotator = AdminAnnotator(self.circulation, library)
 
         # single_entry returns an OPDSEntryResponse that will not be
         # cached, which is perfect. We want the admin interface
@@ -137,7 +138,8 @@ class WorkController(CirculationManagerController, AdminPermissionsControllerMix
 
     def edit(self, identifier_type, identifier):
         """Edit a work's metadata."""
-        self.require_librarian(get_request_library())
+        library = get_request_library()
+        self.require_librarian(library)
 
         # TODO: It would be nice to use the metadata layer for this, but
         # this code handles empty values differently than other metadata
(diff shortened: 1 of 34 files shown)


src/palace/manager/api/admin/controller/work_editor.py
```diff
@@ -62,13 +63,13 @@ def details(self, identifier_type, identifier):
...
-        self.require_librarian(flask.request.library)
+        self.require_librarian(get_request_library())
 
-        work = self.load_work(flask.request.library, identifier_type, identifier)
+        work = self.load_work(get_request_library(), identifier_type, identifier)
         if isinstance(work, ProblemDetail):
             return work
 
-        annotator = AdminAnnotator(self.circulation, flask.request.library)
+        annotator = AdminAnnotator(self.circulation, get_request_library())
```
tdilauro, line 72: It might make sense to set a variable and use it here, rather than calling the `get_request_library` machinery multiple times here. It also makes it clearer that the value is the same in each case.

I saw this in a few other places in the PR, as well.

src/palace/manager/util/sentinel.py
```diff
@@ -0,0 +1,21 @@
...
+    We use this so we can differentiate between a variable that is not given
+    and a variable that is given as None. If https://peps.python.org/pep-0661/
+    is accepted, we should update this is use a proper sentinel value. For now,
+    we use this enum, since we can type check it.
+
+    This solution is based on discussion here:
+    https://github.com/python/typing/issues/236#issuecomment-227180301
+
+    It can be type hinted as: Literal[SentinelType.NotGiven]
+    """
```
tdilauro, line 19: Interesting. In the past, I've usually just used `Ellipsis` as a sentinel. But that is definitely overloading it's meaning.

tdilauro (approved):
One suggestion, but no showstoppers. Looks good! ⚒️

src/palace/manager/api/admin/controller/work_editor.py
jonathangreen: This was mostly an exercise in find / replace, so i missed these.

I pushed another commit that patched up the most egregious cases of repetition that i saw.

src/palace/manager/util/sentinel.py
jonathangreen: That is a good idea, I didn't think of using `Ellipsis` as a sentinel here.

src/palace/manager/api/admin/controller/work_editor.py
tdilauro: I thought as much, but figured it was worth dealing with these now, rather than having to hunt them down later.

PR #2179, opened 2024-11-20
Helper function to get flask.request.admin

jonathangreen:
## Description

Similar to https://github.com/ThePalaceProject/circulation/pull/2177 but for `flask.request.admin`.

Add a helper function get_request_admin() to get `flask.request.admin` in a type safe way. Update the existing code to use the new helper.

## Motivation and Context

Consistency in how we are accessing attributes set on `flask.request`.

## How Has This Been Tested?

- Running tests

## Checklist


src/palace/manager/api/admin/controller/library_settings.py
```diff
@@ -61,7 +62,8 @@ def process_get(self) -> Response:
 
         for library in libraries:
             # Only include libraries this admin has librarian access to.
-            if not flask.request.admin or not flask.request.admin.is_librarian(library):  # type: ignore[attr-defined]
+            admin = get_request_admin(default=None)
```
tdilauro, line 65: Change request: We should move this assignment outside of the loop.

tdilauro (approved):
Requested one change, but otherwise looks good. 🛠️

PR #2180, opened 2024-11-20
Add IP address to flask request logs

jonathangreen:
## Description

If we are in a flask request context, include `X-Forwarded-For` header and `request.remote_addr` in the log message, so we are able to query on them.

## Motivation and Context

When tracing issues via our logs, its helpful to be able to see the requests IP address. I was wishing we had this while troubleshooting the issues in PP-1860.

## How Has This Been Tested?

- Tests in CI

## Checklist


tdilauro (approved):
🪵🪓

PR #2181, opened 2024-11-20
[PP-1956] Ensure the context transaction manager is being used for cr…

dbernstein:
…eating identifiers in SweepMonitor runs.

## Description
This PR attempts to fix a problem we are seeing in production where the database session associated with the identifier object (obtained via the Session.object_session() method )  is sometimes not the same as the one associated with the context manager when the sweep monitor is running.   I'm not sure that this will fix it, but hopefully is will get us closer to a fix for the sweep monitors.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1956

## How Has This Been Tested?
Unit tests still pass.

## Checklist

### DIFF
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -629,7 +629,7 @@ class Axis360API(
         # NOTE: availability is bibliographic.circulation, so it's a
         # little redundant to call availability.apply() -- it's taken
         # care of inside bibliographic.apply().
-        bibliographic.apply(edition, self.collection, replace=policy)
+        bibliographic.apply(edition, self.collection, replace=policy, db=self._db)
         availability.apply(self._db, self.collection, replace=policy)
         return edition, new_edition, license_pool, new_license_pool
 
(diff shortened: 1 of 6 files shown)


jonathangreen:
@dbernstein why do think that we are getting the wrong session here? I'm trying to understand how that could happen.

jonathangreen:
This one has been going for a LONG time at this point, and it seems like we are adding fixes on top of other fixes to try to get this sorted out, but it hasn't been working, I'm not sure I'm comfortable adding another fix here without understanding why this is happening. 

This issue is related to at least these two other PRs: https://github.com/ThePalaceProject/circulation/pull/1948 (PP-1491) & https://github.com/ThePalaceProject/circulation/pull/2091 (PP-1755). Perhaps we need to go back to the original problem, and see if there is a better approach that we can take there so we get this actually sorted.

jonathangreen (approved):
@dbernstein i approved this one, because I don't think its going to cause any issues, but I also don't see how it will help the problem here.

dbernstein:
@jonathangreen :  My thinking here is that at least we're ensuring  that the session that is being used by the identifier class to add a link  is the same same that is being used by the context manager.  Unless I have fundamentally misunderstood something here.

PR #2182, opened 2024-11-21
[PP-1914] Remove collection facet from feed.

dbernstein:
## Description

This update removes the Collection Facet and all related artifacts from the code base.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1914

## How Has This Been Tested?

Unit tests have been updated.

## Checklist


jonathangreen (approved):
This looks good! 

We also need a follow up PR for this, that will go into the release after this one to clean the old settings out of our existing library configurations. They won't cause any issues, but its good to clean them out of the json field before we forget about them.

dbernstein:
Follow-on PR here: https://github.com/ThePalaceProject/circulation/pull/2204

PR #2184, opened 2024-11-21
Fix 'NoneType' object has no attribute 'computed' error (PP-1961)

jonathangreen:
## Description

There are two parts to this fix:
1. Trying a bit harder to get a work for the licensepools we are working with.
2. Tightening up some of our type hints and `None` handling, so that we log appropriate errors if we do get a `None` coming back.
  - I tried to make sure to add log statements for all of these so that we can figure out whats happening from the logs, if we see these happening in production. 

## Motivation and Context

This fixes an error we were seeing in production , where we couldn't generate a single item feed for a work. I was actually able to find an item this was happening to and reproduce it, which let me get a fix in place. 

Stack trace from production:
```
  File "/var/www/circulation/env/lib/python3.10/site-packages/flask/app.py", line 880, in full_dispatch_request
    rv = self.dispatch_request()
  File "/var/www/circulation/env/lib/python3.10/site-packages/flask/app.py", line 865, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
  File "/var/www/circulation/src/palace/manager/api/routes.py", line 120, in decorated
    return f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/api/routes.py", line 70, in decorated
    return f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/api/routes.py", line 93, in wrapped_function
    resp = make_response(f(*args, **kwargs))
  File "/var/www/circulation/src/palace/manager/core/app_server.py", line 87, in decorated
    v = f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/core/app_server.py", line 163, in compressor
    return f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/api/routes.py", line 510, in permalink
    return app.manager.work_controller.permalink(identifier_type, identifier)
  File "/var/www/circulation/src/palace/manager/api/controller/work.py", line 135, in permalink
    return OPDSAcquisitionFeed.entry_as_response(
  File "/var/www/circulation/src/palace/manager/feed/opds.py", line 88, in entry_as_response
    if not entry.computed:
AttributeError: 'NoneType' object has no attribute 'computed'
```

JIRA: 
- Parent: PP-1860 
- PP-1961

## How Has This Been Tested?

- Tested with item that was causing issue locally
- Running tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/circulation_manager.py
+++ b/src/palace/manager/api/controller/circulation_manager.py
@@ -155,6 +155,12 @@ class CirculationManagerController(BaseCirculationManagerController):
         if work is None:
             # We have no work for this license pool. Return a ProblemDetail
             # that will give a 404 status code.
+            self.log.warning(
+                "No work found for license pool %r %s/%s",
+                pools[0],
+                identifier_type,
+                identifier,
+            )
             return NOT_FOUND_ON_REMOTE
 
         if work and not work.age_appropriate_for_patron(self.request_patron):
--- a/src/palace/manager/api/controller/circulation_manager.py
+++ b/src/palace/manager/api/controller/circulation_manager.py
@@ -14,6 +14,7 @@ from palace.manager.api.problem_details import (
     NO_LICENSES,
     NO_SUCH_LANE,
     NOT_AGE_APPROPRIATE,
+    NOT_FOUND_ON_REMOTE,
     REMOTE_INTEGRATION_FAILED,
 )
 from palace.manager.api.util.flask import get_request_library
@@ -35,6 +36,7 @@ from palace.manager.sqlalchemy.model.licensing import (
 from palace.manager.sqlalchemy.model.patron import Hold, Loan, Patron
 from palace.manager.sqlalchemy.model.work import Work
 from palace.manager.sqlalchemy.util import get_one
+from palace.manager.util import first_or_default
 from palace.manager.util.problem_detail import ProblemDetail
 
 T = TypeVar("T", Loan, Hold)
@@ -147,9 +149,13 @@ class CirculationManagerController(BaseCirculationManagerController):
         if isinstance(pools, ProblemDetail):
             return pools
 
-        # We know there is at least one LicensePool, and all LicensePools
-        # for an Identifier have the same Work.
-        work: Work = pools[0].work
+        # We know there is at least one LicensePool. Find the first one with
+        # a work set on it.
+        work: Work | None = first_or_default([lp.work for lp in pools if lp.work])
+        if work is None:
+            # We have no work for this license pool. Return a ProblemDetail
+            # that will give a 404 status code.
+            return NOT_FOUND_ON_REMOTE
 
         if work and not work.age_appropriate_for_patron(self.request_patron):
             # This work is not age-appropriate for the authenticated
--- a/src/palace/manager/api/controller/loan.py
+++ b/src/palace/manager/api/controller/loan.py
@@ -25,7 +25,6 @@ from palace.manager.api.problem_details import (
 )
 from palace.manager.api.util.flask import get_request_library, get_request_patron
 from palace.manager.celery.tasks.patron_activity import sync_patron_activity
-from palace.manager.core.exceptions import PalaceValueError
 from palace.manager.feed.acquisition import OPDSAcquisitionFeed
 from palace.manager.service.redis.models.patron_activity import PatronActivity
 from palace.manager.sqlalchemy.model.library import Library
@@ -370,19 +369,9 @@ class LoanController(CirculationManagerController):
         ):
             # If this is a streaming delivery mechanism, create an OPDS entry
             # with a fulfillment link to the streaming reader url.
-            feed = OPDSAcquisitionFeed.single_entry_loans_feed(
+            return OPDSAcquisitionFeed.single_entry_loans_feed(
                 self.circulation, loan, fulfillment=fulfillment
             )
-            if isinstance(feed, ProblemDetail):
-                # This should typically never happen, since we've gone through the entire fulfill workflow
-                # But for the sake of return-type completeness we are adding this here
-                return feed
-            if isinstance(feed, Response):
-                return feed
-            else:
-                raise PalaceValueError(
-                    "Unexpected return type from OPDSAcquisitionFeed.single_entry_loans_feed"
-                )
 
         try:
             return fulfillment.response()
--- a/src/palace/manager/api/controller/loan.py
+++ b/src/palace/manager/api/controller/loan.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import flask
 from flask import Response
 from flask_babel import lazy_gettext as _
-from lxml import etree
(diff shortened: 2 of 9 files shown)


src/palace/manager/api/controller/circulation_manager.py
```diff
@@ -147,9 +149,13 @@ def load_work(
         if isinstance(pools, ProblemDetail):
             return pools
 
-        # We know there is at least one LicensePool, and all LicensePools
-        # for an Identifier have the same Work.
-        work: Work = pools[0].work
+        # We know there is at least one LicensePool. Find the first one with
+        # a work set on it.
+        work: Work | None = first_or_default([lp.work for lp in pools if lp.work])
```
jonathangreen, line 154: This is where the problem we were seeing in production stems from. `pools[0].work` was `None`, and that's where the trouble began. In that case there was another license pool that had a work set, so we try to find the first LP with a work set instead of just using the first one.

I also added some handling for the case where there is no work, since this case is still possible.

src/palace/manager/api/controller/loan.py
```diff
@@ -480,6 +481,14 @@ def revoke(self, license_pool_id: int) -> OPDSEntryResponse | ProblemDetail:
...
 
         work = pool.work
+        if not work:
+            # Somehow we just revoked a loan or hold for a LicensePool
+            # that has no Work. This shouldn't happen.
+            self.log.error(
+                "Revoked loan or hold for LicensePool %r which has no Work!",
+                pool,
+            )
+            return NOT_FOUND_ON_REMOTE
```
jonathangreen, line 491: It wasn't the case in this instance, but this could cause the same error, so I also added some error handling for `None` here as well.

src/palace/manager/api/controller/work.py
```diff
@@ -96,7 +98,9 @@ def contributor(
             max_age=lane.max_cache_age(), mime_types=flask.request.accept_mimetypes
         )
 
-    def permalink(self, identifier_type, identifier):
+    def permalink(
+        self, identifier_type: str, identifier: str
+    ) -> OPDSEntryResponse | ProblemDetail:
```
jonathangreen, line 103: Added some type hinting here so mypy checks this function, which would have given us an indication this could have been a problem.

src/palace/manager/feed/acquisition.py
```diff
@@ -558,15 +557,13 @@ def single_entry_loans_feed(
         if not annotator:
             annotator = LibraryLoanAndHoldAnnotator(circulation, None, library)
 
-        log = logging.getLogger(cls.__name__)
-
```
jonathangreen, line 562: The logging changes in this file are only tangentially related. I just cleaned up some of the logging here to use `LoggerMixin` while I was at it, so its easier to see where these messages are coming from in the future.

src/palace/manager/feed/acquisition.py
```diff
@@ -608,7 +605,7 @@ def single_entry_loans_feed(
     @classmethod
     def single_entry(
         cls,
-        work: Work | Edition | None,
+        work: Work | Edition,
```
jonathangreen, line 608: I looked at everyone calling this function to make sure they are not (or updated them) passing `None` here.

src/palace/manager/feed/acquisition.py
```diff
@@ -633,22 +627,23 @@ def single_entry(
...
                 active_edition = work.presentation_edition
                 identifier = active_edition.primary_identifier
+            else:
+                active_edition = None
 
         # There's no reason to present a book that has no active license pool.
         if not identifier:
-            logging.warning("%r HAS NO IDENTIFIER", work)
-            return None
+            raise PalaceValueError("Work has no identifier")
```
tdilauro: Do we not want the work info in this log message? Or are we getting that somehow?

tdilauro (approved):
Looks good! A couple small comments below.

src/palace/manager/feed/acquisition.py
```diff
@@ -660,18 +661,16 @@ def single_entry(
...
+                exc_info=e,
             )
             return cls.error_message(
                 identifier,
                 403,
                 "I know about this work but can offer no way of fulfilling it.",
             )
-        except Exception as e:
-            logging.error("Exception generating OPDS entry for %r", work, exc_info=e)
-            return None
```
jonathangreen, line 674: Drop this overly broad `except`. If we are getting random exceptions, we should handle them as they come up. Looking back at our logs, I don't see this being triggered recently, so it should be fine to drop.

tests/manager/feed/test_opds_acquisition_feed.py
```diff
@@ -929,15 +913,19 @@ def facet_link(cls, url, facet_title, group_title, selected, is_default):
...
         work = db.work(with_license_pool=True)
         hold, _ = work.active_license_pool().on_hold_to(patron)
 
-        feed = OPDSAcquisitionFeed.active_loans_for(
-            None, patron, LibraryAnnotator(None, None, db.default_library())
-        )
+        with services_fixture.wired():
+            feed = OPDSAcquisitionFeed.active_loans_for(
+                None, patron, LibraryAnnotator(None, None, db.default_library())
+            )
```
jonathangreen, line 928: This change is needed because without a wired service container we get an exception (and have been for awhile) that was being swallowed by the except clause I removed.

tests/manager/api/admin/controller/test_custom_lists.py
```diff
@@ -495,7 +498,10 @@ def test_custom_list_get_with_pagination(
             work = admin_librarian_fixture.ctrl.db.work(with_license_pool=True)
             list.add_entry(work)
 
-        with admin_librarian_fixture.request_context_with_library_and_admin("/"):
+        with (
+            admin_librarian_fixture.request_context_with_library_and_admin("/"),
+            services_fixture.wired(),
+        ):
```
jonathangreen, line 504: This change is needed because without a wired service container we get an exception (and have been for awhile) that was being swallowed by the except clause I removed.

src/palace/manager/api/controller/loan.py
```diff
@@ -381,12 +380,9 @@ def fulfill(
...
             else:
-                content = etree.tostring(feed)
-            return Response(
-                response=content,
-                status=200,
-                content_type=OPDSFeed.ACQUISITION_FEED_TYPE,
-            )
+                raise PalaceValueError(
+                    "Unexpected return type from OPDSAcquisitionFeed.single_entry_loans_feed"
+                )
```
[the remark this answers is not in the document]
jonathangreen: Agreed. I just completely removed the if statement and `raise` since it should never get hit anyway.

src/palace/manager/feed/acquisition.py
jonathangreen: Good call. I updated the exception to add the `repr` to the message:

```
PalaceValueError(f"Work has no associated identifier: {work!r}")
```

PR #2186, opened 2024-11-22
Works editor custom lists endpoints not handling library list ownership (PP-1962)

jonathangreen:
## Description

Previously if a work was in multiple libraries, in the admin UI, in the work list editor, you would see all the lists it was part of, even if those lists did not belong to the current library. And if you tried to edit the works list membership, it would fail with `AttributeError: 'NoneType' object has no attribute 'get'`.

This PR adds:
- Pydantic validation for the data submitted to the work editor custom list end point
- Does some refactoring to the works editor controller custom lists functions
  - Added types
  - Make sure the current library is being respected
- Added tests to cover the new checks

## Motivation and Context

Troubleshooting this trace-back from our logs:
```
Traceback (most recent call last):
  File "/var/www/circulation/env/lib/python3.10/site-packages/flask/app.py", line 880, in full_dispatch_request
    rv = self.dispatch_request()
  File "/var/www/circulation/env/lib/python3.10/site-packages/flask/app.py", line 865, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
  File "/var/www/circulation/src/palace/manager/api/routes.py", line 120, in decorated
    return f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/api/admin/routes.py", line 84, in decorated
    v = f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/api/admin/routes.py", line 57, in decorated
    return f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/api/admin/routes.py", line 75, in decorated
    return f(*args, **kwargs)
  File "/var/www/circulation/src/palace/manager/api/admin/routes.py", line 167, in work_custom_lists
    return app.manager.admin_work_controller.custom_lists(identifier_type, identifier)
  File "/var/www/circulation/src/palace/manager/api/admin/controller/work_editor.py", line 699, in custom_lists
    submitted_ids = [l.get("id") for l in lists if l.get("id")]
  File "/var/www/circulation/src/palace/manager/api/admin/controller/work_editor.py", line 699, in <listcomp>
    submitted_ids = [l.get("id") for l in lists if l.get("id")]
AttributeError: 'NoneType' object has no attribute 'get'
```

JIRA: 
- Parent: PP-1860 
- PP-1962

## How Has This Been Tested?

- Tested locally with production data
- Running unit tests in CI

## Checklist

- [X] I have updated the documentation accordingly.
[...]

PR #2187, opened 2024-11-22
Move custom lists pydantic model into models package

jonathangreen:
## Description

Move the custom lists controllers pydantic models into the `palace.manager.api.admin.model` package.

## Motivation and Context

I moved these while working on https://github.com/ThePalaceProject/circulation/pull/2186 because I thought I would be able to reuse them. Turns out the work editor uses different data, so I couldn't reuse them. I think it makes sense though to move these, so they are with the rest of the admin controller pydantic models.

## How Has This Been Tested?

- Running tests

## Checklist


tdilauro (approved):
🛠️

PR #2188, opened 2024-11-25
Handle Overdrive refresh_patron_access_token failues (PP-1964)

jonathangreen:
## Description

Previously in `refresh_patron_access_token` we were only handling the cases for status_code `200` and `400`. Everything else would just pass through the function without doing anything.

After this update any 2xx status code will be counted as a success, and any other status code will fail patron auth, log a useful error message and return a `ProblemDetail` document to the client.

I added a test case with a real API failure error from Overdrive, and added another test case to make sure we correctly handle unexpected responses.

## Motivation and Context

We were seeing unhandled exceptions in `Exception in web app: Something's wrong with the patron OAuth Bearer Token` on our webservers reasonably frequently. This appears to be happening when Overdrive is not happy with a patrons auth for some reason. 

## How Has This Been Tested?

- Tested locally

## Checklist

### DIFF
--- a/src/palace/manager/api/overdrive.py
+++ b/src/palace/manager/api/overdrive.py
@@ -8,6 +8,7 @@ import logging
 import re
 import urllib.parse
 from collections.abc import Iterable
+from json import JSONDecodeError
 from threading import RLock
 from typing import Any, NamedTuple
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
🎉

PR #2189, opened 2024-11-25
Log additional Palace request data

jonathangreen:
## Description

If we are in a request context when we log a message, we include information from `request.admin`, `request.library` and `request.patron` if any of them are set.

## Motivation and Context

While troubleshooting the issues in PP-1860, it would have been very helpful to have this additional information included with log messages.

## How Has This Been Tested?

- Tested locally
- New unit tests

## Checklist


tdilauro (approved):
Looks good. 🪵

PR #2191, opened 2024-11-25
Remove `Library.all_collections` property.

tdilauro:
## Description

Removes `Library.all_collections` (and its functionality), replaces its usage with `Library.collections`, and removes a test assertion specific to this method.

## Motivation and Context

While working on PP-1875, I ran into this attribute, which provides a generator that yields both the collections associated with a library and the parents of those collections, where present. 

It came in on this [commit](https://github.com/NYPL-Simplified/circulation/commit/d58ee6b83af1f57ae32a2acf8f7a01344271ed05) on [this PR](https://github.com/NYPL-Simplified/server_core/pull/570/commits/d58ee6b83af1f57ae32a2acf8f7a01344271ed05).

Here's some context from the PR description:

> When I tried setting up an Overdrive Advantage collection, I expected that I would be able to add the Overdrive Advantage collection to a library and have that library’s feeds also contain the books from the parent consortium collection. Without this change, I’d have to separately add the consortium collection to the library.
>
> The downside is that now there’s no way to get a feed that has the Overdrive Advantage books without the consortium’s books.

So, based on how we currently configure our collections that have parents -- specifically, we configure the library association for both OD main accounts and Advantage accounts, as appropriate -- I don't think we need the functionality it provides.

## How Has This Been Tested?

- Tests pass locally.
- [CI tests for associated branch](https://github.com/ThePalaceProject/circulation/actions/runs/12019497840) pass.

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/custom_lists.py
+++ b/src/palace/manager/api/admin/controller/custom_lists.py
@@ -114,7 +114,7 @@ class CustomListsController(
             .join(LicensePool, LicensePool.work_id == Work.id)
             .join(Collection, LicensePool.collection_id == Collection.id)
             .filter(LicensePool.identifier_id == identifier.id)
-            .filter(Collection.id.in_([c.id for c in library.all_collections]))
+            .filter(Collection.id.in_([c.id for c in library.collections]))
         )
         work = query.one()
         return work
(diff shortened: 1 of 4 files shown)


jonathangreen (approved):
Looks good!

PR #2194, opened 2024-11-26
Replace sqlalchemy backref parameter with back_populates

jonathangreen:
## Description

This is a follow up to: https://github.com/ThePalaceProject/circulation/pull/2123

I missed some `backref` parameters in that PR, this one replaces all `backref` with `back_populates` as sqlalchmey recommends as the current best practice.

## Motivation and Context

The sqlalchemy backref parameter is considered legacy see: https://docs.sqlalchemy.org/en/14/orm/backref.html. Using this construct means that mypy (and IDE) can't infer the type of the dynamically generated parameter.

## How Has This Been Tested?

- Running tests

## Checklist

### DIFF
--- a/src/palace/manager/feed/acquisition.py
+++ b/src/palace/manager/feed/acquisition.py
@@ -212,7 +212,7 @@ class OPDSAcquisitionFeed(BaseOPDSFeed):
         cls,
         work: Work,
         active_licensepool: LicensePool | None,
-        edition: Edition,
+        edition: Edition | None,
         identifier: Identifier,
         annotator: Annotator,
     ) -> WorkEntry:
@@ -926,10 +926,11 @@ class LookupAcquisitionFeed(OPDSAcquisitionFeed):
         if error_status:
             return cls.error_message(identifier, error_status, error_message or "")
 
-        if active_licensepool:
-            edition = active_licensepool.presentation_edition
-        else:
-            edition = _work.presentation_edition
+        edition = (
+            active_licensepool.presentation_edition
+            if active_licensepool
+            else _work.presentation_edition
+        )
         try:
             return cls._create_entry(
                 _work, active_licensepool, edition, identifier, annotator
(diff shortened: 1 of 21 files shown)


src/palace/manager/feed/acquisition.py
```diff
@@ -212,7 +212,7 @@ def _create_entry(
         cls,
         work: Work,
         active_licensepool: LicensePool | None,
-        edition: Edition,
+        edition: Edition | None,
```
jonathangreen, line 215: Because presentation_edition can be `None`, this type hint needed to be updated.

src/palace/manager/feed/annotator/base.py
```diff
@@ -236,9 +236,7 @@ def content(cls, work: Work | None) -> str:
                 and work.summary.representation
                 and work.summary.representation.content
             ):
-                content = work.summary.representation.content
-                if isinstance(content, bytes):
-                    content = content.decode("utf-8")
+                content = work.summary.representation.content.decode("utf-8")
```
jonathangreen, line 239: The typing changes here made it clear that `work.summary.representation.content` will always be `bytes`

src/palace/manager/feed/annotator/circulation.py
```diff
@@ -365,7 +365,7 @@ def annotate_work_entry(
         updated: datetime.datetime | None = None,
     ) -> None:
         work = entry.work
-        identifier = entry.identifier or work.presentation_edition.primary_identifier
+        identifier = entry.identifier
```
jonathangreen, line 368: `entry.identifier` cannot be `None`, so this `or` didn't do anything

src/palace/manager/sqlalchemy/model/work.py
```diff
@@ -1649,62 +1658,68 @@ def _set_value(parent, key, target):
 
         result["contributors"] = []
         if doc.presentation_edition and doc.presentation_edition.contributions:
-            for item in doc.presentation_edition.contributions:
+            for contribution in doc.presentation_edition.contributions:
```
jonathangreen, line 1661: All these for loop variable were updated because now that we have proper types reusing `item` in each loop gives us type errors. So I switched out the loop variable for each one.

tdilauro (approved):
Looks good! 🔙

PR #2197, opened 2024-11-27
Syncronize our SQLAlchemy type hints with db schema

jonathangreen:
## Description

Make sure what what is defined in our SQLAlchemy schema is accuratly reflected in the type hints:
- If a `relationship` is not a collection, and its foreign key column is nullable make sure the annotation is Optional
  - There were some cases where this caused a large number of mypy failures, since we were widely assuming that these relationships were not optional. In these cases I queried all our production instances, and if the the foreign key was never null, I added a `nullable=False` to the foreign key column.
- Appropriately set the columns annotation to not Optional if:
  - `nullable=False`
  - `primary_key=True`
- Add `nullable=False` and appropriately set annotation for columns that have `default` set

## Motivation and Context

Our sqlalchemy type hints have been added over time, and some of them did not accurately reflect what was in the DB schema. I noticed this while working on https://github.com/ThePalaceProject/circulation/pull/2194. 

These changes meant that I had to add some guard code for errors that mypy rightly flagged. It also meant that I could remove some guard code that was no longer necessary.

## How Has This Been Tested?

- Running unit tests
- Querying our production DBs to make sure our production data doesn't have any nulls set

## Checklist


jonathangreen:
This is okay to get a code review but I need to fix up the migrations parent before it gets merged.

tdilauro (approved):
This looks good! 🗺️

There are some cases we have not added an annotation. Was that intentional? What is the rationale for whether we are adding one or not?

jonathangreen:
> There are some cases we have not added an annotation. Was that intentional? What is the rationale for whether we are adding one or not?

@tdilauro every `relationship` mapped value should have a type hint after this change. If I missed any it wasn't intentional. 

Only `Column` mapped values that vary from the default introspection provided by the sqlalchemy mypy plugin (see [docs](https://docs.sqlalchemy.org/en/20/orm/extensions/mypy.html#introspection-of-columns-based-on-typeengine)) were annotated. In practice, this means I only annotated `Column` values that had `nullable=False` or `primary_key=True` set on them.

PR #2204, opened 2024-12-02
Add migration to remove obsolete settings (facets_enabled_collection …

dbernstein:
…and facets_default_collection).

## Description
This is a follow-up PR to clean-up obsolete settings removed in https://github.com/ThePalaceProject/circulation/pull/2182

## Checklist

### DIFF
--- a/alembic/versions/20241202_603b8ebd6daf_remove_collection_facet_settings.py
+++ b/alembic/versions/20241202_603b8ebd6daf_remove_collection_facet_settings.py
@@ -1,0 +1,29 @@
+"""Remove collection facet settings
+
+Revision ID: 603b8ebd6daf
+Revises: 272da5f400de
+Create Date: 2024-12-02 19:42:23.775579+00:00
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "603b8ebd6daf"
+down_revision = "272da5f400de"
+branch_labels = None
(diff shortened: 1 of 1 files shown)


alembic/versions/20241202_603b8ebd6daf_remove_collection_facet_settings.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    conn.execute(
+        "UPDATE integration_library_configurations set settings = settings::jsonb - 'facets_enabled_collection'"
+    )
+    conn.execute(
+        "UPDATE integration_library_configurations set settings = settings::jsonb - 'facets_default_collection'"
+    )
```
jonathangreen, line 25: These settings are stored in the `libraries` table, not the `integration_library_configurations` table.

Minor: You can also do the update in a single update statement and without using a cast to `jsonb` since the column already has that datatype.

```suggestion
    conn.execute(
        "UPDATE libraries set settings_dict = "
        "settings_dict - array['facets_enabled_collection', 'facets_default_collection']"
    )
```

jonathangreen (changes requested):
Need to update the table this migration is targeting, see comment on PR.

jonathangreen (approved):
Looks good

PR #2207, opened 2024-12-09
Retry celery patron sync on StaleDataError (PP-2005)

jonathangreen:
## Description

Retry when we get a `StaleDataError` while running a patron sync.

## Motivation and Context

Looking at the cloudwatch alarms we are getting for failed Celery tasks, it looks like almost all of them are:

```
sqlalchemy.orm.exc.StaleDataError: UPDATE statement on table 'loans' expected to update 1 row(s); 0 were matched.
```

These are probably mostly unavoidable, due to something happening (like a checkout or checkin) while the sync task is running. Just retrying should solve the issue.

## How Has This Been Tested?

- Added unit tests

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/patron_activity.py
+++ b/src/palace/manager/celery/tasks/patron_activity.py
@@ -1,4 +1,5 @@
 from celery import shared_task
+from sqlalchemy.orm.exc import StaleDataError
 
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
🚀

PR #2209, opened 2024-12-11
[PP-1947] convert attachments to s3 links

dbernstein:
## Description
This update changes the way we are delivering inventory and holds reports.   Instead of attaching the zipped reports to an email we are instead providing a link in the email to the zip which lives in public s3 bucket.  The zip file will be removed  automatically after 30 days. 

The url contains a UUID in order ensure that the email link is not guessable and to ensure that it is uniquely associated with the email request.  In addition the zip file name includes the library short name as well as the date range covered by the request. 
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1947

## How Has This Been Tested?
Unit tests cover it but it has not yet been tested in dev.

## Checklist


src/palace/manager/service/storage/s3.py
```diff
@@ -244,3 +244,60 @@ def multipart_abort(self, key: str, upload_id: str) -> None:
             Key=key,
             UploadId=upload_id,
         )
+
+    def update_bucket_expiration_rule(
+        self, prefix: str, expiration_in_days: int = 30
+    ) -> None:
```
jonathangreen, line 250: Minor: Seems like this might be something we want to just do in our terraform config, instead of doing it every time we do an inventory report. We already apply a lifecycle rule to [expire multipart uploads](https://github.com/ThePalaceProject/hosting-playbook/blob/main/terraform/modules/s3/main.tf#L71-L91).

src/palace/manager/celery/tasks/generate_inventory_and_hold_reports.py
```diff
@@ -148,13 +148,34 @@ def run(self) -> None:
                             arcname=f"palace-inventory-report-for-library-{file_name_modifier}.csv",
                         )
 
+                    reports_path = "inventory_and_holds"
+                    expiration_in_days = 30
+                    self.s3_service.update_bucket_expiration_rule(
+                        prefix=f"{reports_path}/", expiration_in_days=expiration_in_days
+                    )
+
+                    with zip_path.open(
```
jonathangreen, line 157: Minor: Instead of opening the file again here, I'm fairly sure that `report_zip` is a` IO[bytes]`, which `store_stream` can take directly.

jonathangreen (approved):
Looks good. I added a couple minor comments but feel free to merge as is if you don't want to address them.

src/palace/manager/service/storage/s3.py
dbernstein: 👍

PR #2210, opened 2024-12-12
Ignore axis now fulfillment (PP-1929)

jonathangreen:
## Description

Ignore any AxisNow/Blio fulfillment information we get back from B&T. 

## Motivation and Context

Right now, we are adding these items to a users bookshelf, but failing when they try to fulfill the item. It is preferable to not display it for now. See jira: PP-1929

## How Has This Been Tested?

Tested locally.

## Checklist

### DIFF
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -198,6 +198,7 @@ class Axis360API(
     # The name Axis 360 gives to its web interface. We use it as the
     # name for the underlying access control system.
     AXISNOW = "AxisNow"
+    BLIO = "Blio"
 
     delivery_mechanism_to_internal_format = {
         (epub, no_drm): "ePub",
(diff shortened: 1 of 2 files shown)


src/palace/manager/api/axis.py
```diff
@@ -1578,6 +1581,14 @@ def process_one(
                 or ""
             )
 
+            if not self.internal_format and (
+                checkout_format == self.api.AXISNOW or checkout_format == "Blio"
```
tdilauro, line 1585: Minor: Should we add "BLIO" as a constant to [the `Axis360API` class](https://github.com/ThePalaceProject/circulation/blob/05748ac3b5c1abf8672239526e4de7dbd9b5c76e/src/palace/manager/api/axis.py#L201-L200) and reference that here?

tests/manager/api/test_axis.py
```diff
@@ -551,6 +551,12 @@ def test_patron_activity(self, axis360: Axis360Fixture):
         assert isinstance(hold2, HoldInfo)
         assert isinstance(loan, LoanInfo)
 
+        # If the activity includes something with a Blio format, it is not included in the results.
+        data = axis360.sample_data("availability_with_axisnow_fulfillment.xml")
+        axis360.api.queue_response(200, content=data)
+        results = axis360.api.patron_activity(patron, "pin")
+        assert len(results) == 0
```
tdilauro, line 558: I'm a little confused by this one. The sample data looks like it has `ePub`, `Blio`, and `AxisNow` formats available. I thought we'd be able to use the `ePub` format, so I think I'm missing something here.

```xml
  ...
    <availableFormats>
      <formatName>Blio</formatName>
      <formatName>ePub</formatName>
      <formatName>AxisNow</formatName>
    </availableFormats>
  ...
```

tdilauro (approved):
I'm approving this because I'm pretty sure that I'm just not understanding something. If that is the case, it would be helpful to add a comment to the mainline source and maybe the test, as well, that addresses my confusion on the test result.

tests/manager/api/test_axis.py
jonathangreen: `availableFormats` is a bit misleading, those are the formats the item is available in before its checked out. Once its checked out, its format is locked to the format in `checkoutFormat` element in the response. So in this case `<checkoutFormat>Blio</checkoutFormat>`.

I'll add a clarifying comment to the code.

PR #2239, opened 2025-01-09
Set defaults for odl username and password

dbernstein:
## Description
A follow on PR to address an outstanding issue discovered while testing https://ebce-lyrasis.atlassian.net/browse/PP-1834
see comments on the ticket.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1834

## How Has This Been Tested?
I tested the changes manually to verify that the form looks right.

## Checklist


jonathangreen (approved):
👍🏻

PR #2240, opened 2025-01-09
Don't send hold notifications for Overdrive holds (PP-2009)

jonathangreen:
## Description

Ignore Overdrive holds when querying for holds to send notifications.

## Motivation and Context

See context in PP-2009.

## How Has This Been Tested?

- Added unit test

## Checklist


tdilauro (approved):
Looks good! 🚀

PR #2243, opened 2025-01-10
[PP-2043] Prevent None from displaying when there are no subtitles.

dbernstein:
## Description

Ensures that no subtitle element is returned in the OPDS 2 feed when the subtitle value is None thus preventing the rendering of "None" for the subtitle  in the user interface.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2043

## How Has This Been Tested?
Unit test added to verify the fix works.

## Checklist


tdilauro (approved):
Looks good! 🚀

PR #2245, opened 2025-01-13
Allow empty string for username and password in ODL settings.

dbernstein:
## Description
A follow-on PR for https://ebce-lyrasis.atlassian.net/browse/PP-1834 to address validation error in ODL 2.0 collection config when no specifying username and password.  Validation requires a string albeit empty for a string type control.  Changing the return type had undesirable ripple effects.   This changes the default value from None to an empty string which seemed to me the more straightforward solution.

## How Has This Been Tested?
I tested it manually on my local instance.

## Checklist

### DIFF
--- a/src/palace/manager/api/odl/settings.py
+++ b/src/palace/manager/api/odl/settings.py
@@ -78,14 +78,14 @@ class OPDS2WithODLSettings(OPDS2ImporterSettings):
         ),
     )
     password: str = FormField(
-        default=None,
+        default="",
         form=ConfigurationFormItem(
             label=_("Library's API password"),
             required=False,
         ),
     )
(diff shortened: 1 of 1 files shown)


jonathangreen (commented):
Looks fine. I would have thought that changing the type of password to be `str | None` would be a better solution for consistency since the base class for the OPDS settings defines them like that, since authentication is optional there:

https://github.com/ThePalaceProject/circulation/blob/b4d189be4ea43d77f6ebc0d44a4ef18378033877/src/palace/manager/core/opds_import.py#L157-L177

But I haven't tried making this type change, so maybe there are some consequences I'm not thinking about.

jonathangreen (approved):

PR #2246, opened 2025-01-13
[PP-2065] Address library not persistent within session errors by ens…

dbernstein:
…uring the

library and other objects accessed within the new transaction are attached to the session.

## Description
I was using session.refresh(obj) instead of session.merge(obj).  `merge` reattaches a detached object to the session where `refresh` refreshes an object that is already associated with the session. 

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2065

## Checklist


jonathangreen (approved):
Looks good 🚀

PR #2247, opened 2025-01-14
Fix uwsgi log file permissions (PP-2067)

jonathangreen:
## Description

This builds on the changes from https://github.com/ThePalaceProject/circulation/pull/2077. It moves setting permissions on the log files from container build time to run time for both uwsgi and nginx logs. That way if the logs volume is mounted in a host system, we still get the correct log permissions at run time.

## Motivation and Context

JIRA: PP-2067

After the recent docker host system ubuntu upgrade, our logs are not being created with the correct permissions.

## How Has This Been Tested?

- Tested locally with docker

## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -73,12 +73,7 @@ COPY docker/services/nginx /etc/nginx/
 
 # Setup uwsgi
 COPY docker/services/uwsgi /etc/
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good! 🎸

PR #2249, opened 2025-01-14
Ensure that empty subtitles - ie those with a value of None - are set…

dbernstein:
… to null in the database rather than the string "None"

## Description
This is a follow on PR:  Included is a database migration to sweep out the "None"s without requiring a complete reimport.  Though we'll still to wait for the search refresh.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2043

## Checklist


jonathangreen (approved):
🚀

PR #2262, opened 2025-01-23
Add support for suppressing config settings in admin ui. (PP-2064)

tdilauro:
## Description

Adds the "hidden" property to configuration form items to indicate to the admin client that the form element for the setting should be hidden from view, but should be transmitted back when the settings are saved.

## Motivation and Context

It is sometimes necessary to hide a configuration setting.

[Jira [PP-2064](https://ebce-lyrasis.atlassian.net/browse/PP-2064)]

## How Has This Been Tested?

- Tested behavior in local dev environment.
- All tests pass locally.
- [CI tests for associated branch](https://github.com/ThePalaceProject/circulation/actions/runs/12931626136) pass.

## Checklist

[PP-2064]: https://ebce-lyrasis.atlassian.net/browse/PP-2064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/integration/settings.py
+++ b/src/palace/manager/integration/settings.py
@@ -234,6 +234,9 @@ class ConfigurationFormItem(LoggerMixin):
     # weight will be displayed in the order they were added.
     weight: int = 0
 
+    # If set to True, the Admin UI will be directed to hide this field.
+    suppressed: bool = False
+
     @staticmethod
     def get_form_value(value: Any) -> Any:
         if value is None:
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
Looks good! 

I wonder if `hidden` might be a better name for this? I could see us wanting to hide fields in the future based on other configuration values. Not necessary to implement here, just food for thought when naming this.

tdilauro:
Yeah, I was thinking about this as longer-term functionality that was not just for this issue. I also thought about renaming the property, but didn't feel strongly enough about it. Happy to do it, though.

PR #2269, opened 2025-01-27
Add debug info to logs when NoActiveLoan exception occurs in axis 360.

dbernstein:
## Description

Add debug info when axis 360 API cannot find an active loan for a title.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2092

## Checklist

### DIFF
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -477,7 +477,7 @@ class Axis360API(
         identifier = licensepool.identifier
         # This should include only one 'activity'.
         internal_format = self.internal_format(delivery_mechanism)
-        debug_info: list[str] = [
+        log_messages: list[str] = [
             f"arguments for patron_activity method: "
             f"patron.id={patron.id},"
             f"internal_format={internal_format}, "
@@ -485,12 +485,13 @@ class Axis360API(
             f"patron_id={patron.id}"
         ]
         activities = self.patron_activity(
-            patron, pin, licensepool.identifier, internal_format, debug_info
+            patron, pin, licensepool.identifier, internal_format, log_messages
         )
 
-        debug_info.append(
+        log_messages.append(
             f"Patron activities returned from patron_activity method: {activities}"
         )
+
         for loan in activities:
             if not isinstance(loan, AxisLoanInfo):
                 continue
@@ -507,12 +508,11 @@ class Axis360API(
             return fulfillment
         # If we made it to this point, the patron does not have this
         # book checked out.
-        self.log.warning(
-            "Unable to fulfill because there is no active loan.  See info statements below for details."
+        log_messages.insert(
+            0,
+            "Unable to fulfill because there is no active loan. See info statements below for details:",
         )
-        for statement in debug_info:
-            self.log.info(statement)
-
+        self.log.error("\n  ".join(log_messages))
         raise NoActiveLoan()
 
     def place_hold(
@@ -564,7 +564,7 @@ class Axis360API(
         pin: str | None,
         identifier: Identifier | None = None,
         internal_format: str | None = None,
-        debug_info: list[str] | None = None,
+        log_messages: list[str] | None = None,
     ) -> list[AxisLoanInfo | HoldInfo]:
         if identifier:
             assert identifier.identifier is not None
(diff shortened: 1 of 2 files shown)


src/palace/manager/api/axis.py
```diff
@@ -498,6 +507,12 @@ def fulfill(
             return fulfillment
         # If we made it to this point, the patron does not have this
         # book checked out.
+        self.log.warning(
+            "Unable to fulfill because there is no active loan.  See info statements below for details."
+        )
+        for statement in debug_info:
+            self.log.info(statement)
```
jonathangreen, line 514: Minor: Variable naming here is kind of deceiving, since you are logging a variable named `debug_info` at log level INFO, I would have assumed it would go into the DEBUG logs.

src/palace/manager/api/axis.py
```diff
@@ -498,6 +507,12 @@ def fulfill(
             return fulfillment
         # If we made it to this point, the patron does not have this
         # book checked out.
+        self.log.warning(
+            "Unable to fulfill because there is no active loan.  See info statements below for details."
+        )
+        for statement in debug_info:
```
jonathangreen, line 513: Minor: If you are troubleshooting this in cloudwatch logs on a busy CM where multiple loans are happening, correlating this warning level log statement, with the other info level statements is doable, but will be difficult. It will probably be easier to understand what is happening if all the related information is logged in a single log statement.

src/palace/manager/api/axis.py
```diff
@@ -549,21 +564,32 @@ def patron_activity(
...
             title_ids = [identifier.identifier]
         else:
             title_ids = None
+
         availability = self.availability(
             patron_id=patron.authorization_identifier, title_ids=title_ids
         )
-        return list(
+
+        availability_content_str = availability.content.decode("utf-8")
```
jonathangreen, line 579: Minor: Instead of decoding this yourself and assuming the character encoding is utf-8, you can do `availability.text` https://docs.python-requests.org/en/latest/api/#requests.Response.text which decodes it according to the request headers.

src/palace/manager/api/axis.py
```diff
@@ -549,21 +564,32 @@ def patron_activity(
...
+        if debug_info:
+            debug_info.append(f"arguments to availability call: title_ids={title_ids}")
+            debug_info.append(
+                f"response to availability call: status={availability.status_code}, content={availability_content_str}"
+            )
+        loan_info_list = list(
             AvailabilityResponseParser(self, internal_format).process_all(
-                availability.content
+                availability_content_str
```
jonathangreen, line 587: Minor: The generic `process_all` method does take `bytes` or `str` so this change should be fine. All the tests for `AvailabilityResponseParser.process_all` pass in `bytes` via the `sample_data` method. Since you are changing what we pass into the function it might be worth updating the tests to match, so we are testing the same input type as the method gets in real usage.

https://github.com/ThePalaceProject/circulation/blob/3ea99eed70bf6a48bcd7cfd9121dfa4e997f65dd/tests/manager/api/test_axis.py#L1504-L1614

jonathangreen (approved):
Left a couple minor comments to consider

dbernstein:
Thanks @jonathangreen for the review.  I made the changes you suggested.

src/palace/manager/api/axis.py
```diff
@@ -498,6 +508,12 @@ def fulfill(
             return fulfillment
         # If we made it to this point, the patron does not have this
         # book checked out.
+        debug_info.insert(
+            0,
+            "Unable to fulfill because there is no active loan. See info statements below for details:",
+        )
+
+        self.log.debug("\n  ".join(debug_info))
```
jonathangreen, line 516: Sorry @dbernstein, I should have been more clear in my comment before. I think logging at the warning or even error level makes sense here, like you had before. Debug level logging is tricky to enable in production because of the huge volume of logs it generates. 

I wasn't very clear in my previous comment, but I was suggesting that you might want to change the variable name to something else, like `logging_info` or `log_messages`.

jonathangreen (approved):
Changes look good @dbernstein, I approved this one (I approved it last time as well). Left one more minor comment to consider.

src/palace/manager/api/axis.py
dbernstein: On second thought, I think error is the way to go since in almost every situation, if not every situation it arises, it would indicate a real problem.

PR #2270, opened 2025-01-28
Update runs-on to address segfault when building for arm64 (PP-2093)

jonathangreen:
## Description

Our base image builds are failing with a segfault coming from GCC. 

I believe this is the same issue noted here: https://github.com/actions/runner-images/issues/11471

Attempt the mitigation that is mentioned in that thread.

## Motivation and Context

Unblocking our docker builds.

I tried updating the base image build in my fork of circulation, and this change allowed the base image build to pass: https://github.com/jonathangreen/circulation/actions/runs/13013623180

PP-2093

## How Has This Been Tested?

- Running actions in CI

## Checklist

### DIFF
--- a/.github/workflows/build-base-image.yml
+++ b/.github/workflows/build-base-image.yml
@@ -14,7 +14,11 @@ concurrency:
 jobs:
   docker-build-baseimage:
     name: Build Base Image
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🎛️

PR #2277, opened 2025-02-05
[PP-1473] convert axis 360 import to celery

dbernstein:
## Description

This update does several things:
1) it  converts the axis 360 import script to a celery task running on the same schedule.  (every 15 minutes).  It accomplishes the task of running an import for all axis 360 collections on a cm by spawn new tasks for each collection.  As the collection task pulls new books from the import feed, it will return a list of all the identifiers to the backend and then spawn the import task.  The import task will then in batches of 25 IDs import the books.  If the target max execution time (2 minutes) is exceeded, it will spawn a new import task to process the remaining ids and exit.  So all the axis collections will be processed in parallel, but each collection will only be working on one set of ids at a time.

2) It converts the axis 360 reaper collection to a celery task running on the same schedule as before.  The reap_all_collections task will queue up each collection for reaping with a sub task.  Each reap_collection task will run through  the associated identifiers in a fixed order, and process in batches of 25 identifiers.  When a full batch is processed - 25 by default - the task will queue a new task with a new offset and finish.  In this way,  there should never be more than 1 reaper task per collection running at the same time.   Still, all the collections will be processed in parallel.

3) I removed the coverage monitor for axis 360 as well as any coverage dependencies associated with the axis 360 scripts.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1473

## How Has This Been Tested?
Unit tests added.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/axis.py
+++ b/src/palace/manager/celery/tasks/axis.py
@@ -29,7 +29,7 @@ DEFAULT_START_TIME = datetime_utc(1970, 1, 1)
 TARGET_MAX_EXECUTION_SECONDS = 120
 
 
-@shared_task(queue=QueueNames.default, bind=True)
+@shared_task(queue=QueueNames.default, bind=True, task_ignore_results=True)
 def import_all_collections(
     task: Task, import_all: bool = False, batch_size: int = DEFAULT_BATCH_SIZE
 ) -> None:
@@ -59,7 +59,7 @@ def list_identifiers_for_import(
     task: Task,
     collection_id: int,
     import_all: bool = False,
-) -> list[str] | None:
+) -> list[str]:
     """
     A task for resolving a list identifiers to import an axis collection based on the
      most recent timestamp's start date.
@@ -159,20 +159,22 @@ def timestamp(
     return timestamp
 
 
-@shared_task(queue=QueueNames.default, bind=True)
+@shared_task(queue=QueueNames.default, bind=True, task_ignore_results=True)
 def import_identifiers(
     task: Task,
     collection_id: int,
     identifiers: list[str] | None,
     processed_count: int = 0,
-    batch_size: int = 25,
+    batch_size: int = DEFAULT_BATCH_SIZE,
     target_max_execution_time_in_seconds: float = TARGET_MAX_EXECUTION_SECONDS,
-) -> list[str] | None:
+) -> None:
     """
-    This method creates new or updates new editions and license pools for each pair of metadata and circulation data in
-    the items list.
+    This method creates new or updates new editions and license pools for each identifier in the list of identifiers.
+    It will query the axis availability api in batches of {batch_size} IDs and process each result in a single database
+    transaction.  It will continue in this way until it has finished the list or exceeded the max execution time
+    which defaults to 2 minutes.  If it has not finished in the target time, it will requeue the task with the
+    remaining unprocessed identifiers.
     """
-
     if not identifiers:
         task.log.info(
             f"Identifiers list is None: the list_identifiers_for_"
@@ -183,7 +185,6 @@ def import_identifiers(
     with task.session() as session:
         collection = Collection.by_id(session, id=collection_id)
         api = create_api(session, collection)
-        batch: list[str] = []
         start_seconds = time.perf_counter()
         total_imported_in_current_task = 0
         identifiers_list_length = len(identifiers)
@@ -306,7 +307,7 @@ def get_collections_by_protocol(
     return collections
 
 
-@shared_task(queue=QueueNames.default, bind=True)
+@shared_task(queue=QueueNames.default, bind=True, task_ignore_results=True)
 def reap_all_collections(task: Task) -> None:
     """
     A shared task that loops through all Axis360 Api based collections and kick off an
@@ -322,7 +323,7 @@ def reap_all_collections(task: Task) -> None:
         task.log.info(f"Finished queuing collection for reaping.")
 
 
-@shared_task(queue=QueueNames.default, bind=True)
+@shared_task(queue=QueueNames.default, bind=True, task_ignore_results=True)
 def reap_collection(
     task: Task, collection_id: int, offset: int = 0, batch_size: int = 25
 ) -> None:
--- a/src/palace/manager/celery/tasks/axis.py
+++ b/src/palace/manager/celery/tasks/axis.py
@@ -1,0 +1,288 @@
+import time
+from datetime import datetime
+
+from celery import shared_task
+from sqlalchemy.orm import Session
+from sqlalchemy.orm.exc import ObjectDeletedError, StaleDataError
+from tenacity import (
+    retry,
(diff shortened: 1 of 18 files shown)


dbernstein:
I reworked this PR to use the results_backed per our discussion.   The results_backend config is in this PR and will need to be merged first and then I'll rebase this PR on it.  With the backend configured, the tests passed.

jonathangreen:
Its really hard to separate whats in this PR from https://github.com/ThePalaceProject/circulation/pull/2282. 

Usually when I have multiple related PRs, I try to make sure that each PR only contains the changes for that PR in the diff view so its reasonably easy to review. You can do that by changing which branch a PR targets. It looks like the commit history is separate as well though, so that didn't work for me in this case. 

I'm going to let you either merge https://github.com/ThePalaceProject/circulation/pull/2282 or sort out the commits on this PR so the diff is easier to view before I review this one.

dbernstein:
Okay - I think this one is in a better place with no conflicting changes between this PR and [the config PR](https://github.com/ThePalaceProject/circulation/pull/2282).

tests/manager/celery/tasks/test_axis.py
```diff
@@ -0,0 +1,401 @@
...
+
+    assert f"Edition (id={edition_1.id}" in caplog.text
+    assert (
+        f"Finished run importing identifiers for collection ({collection.name}, id={collection.id})"
+        in caplog.text
+    )
+    assert f"Imported {len(title_ids)} identifiers" in caplog.text
+
+
+def test_import_items_with_requeue(
```
jonathangreen, line 221: I'd rather see this test use less mocks, and actually allow the task to be re-queued, so we can test the full behavior when the task has to cycle more then once. This will be easier once you switch to using `task.replace`.

tests/manager/api/test_axis.py
```diff
@@ -1962,91 +1741,6 @@ def axis360provider(
...
-            Axis360BibliographicCoverageProvider,
-            axis360provider.db.session,
-            api_class=MockAxis360API,
-        )
-        [provider] = script.providers
-        assert isinstance(provider, Axis360BibliographicCoverageProvider)
-        assert isinstance(provider.api, MockAxis360API)
-
-    def test_process_item_creates_presentation_ready_work(
-        self, axis360provider: Axis360ProviderFixture
```
jonathangreen, line 1982: I think it would be nice to preserve a test like this, that tests the whole process, making sure that the imported item actually lands in the DB.

jonathangreen (commented):
This generally looks good. I added some comments to take a look at. As usual, feel free to ignore the minor / nit comments if you disagree. There are a couple non-minor comments that would be good to address before this one goes in.

src/palace/manager/celery/tasks/axis.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+def _redis_lock_queue_collection_import(client: Redis, collection_id: int) -> RedisLock:
+    return RedisLock(
+        client,
+        lock_name=f"Axis360QueueCollectionImport-{collection_id}",
+    )
+
+
+def get_collections_by_protocol(task: Task, session: Session, protocol_class) -> list[Collection]:  # type: ignore[no-untyped-def]
```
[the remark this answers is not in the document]
dbernstein: @jonathangreen : Fair comment: I wasn't sure how to type it:  What would the type be?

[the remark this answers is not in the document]
jonathangreen: It depends how generic you want to make it.

Looking at where you call that function, you always call it with `protocol_class=Axis360API`, so you could do:
```python
def get_collections_by_protocol(task: Task, session: Session, protocol_class: type[Axis360API]) -> list[Collection]:
```

or you could base it on what `registry.get_protocols` can support, which would give you something more generic that can take more types. 
```python
def get_collections_by_protocol(task: Task, session: Session, protocol_class: type[CirculationApiType]) -> list[Collection]:
```

src/palace/manager/celery/tasks/axis.py
```diff
@@ -0,0 +1,395 @@
...
+                f"Total imported {total_imported_in_current_task} identifiers in current task for collection ({collection.name}, id={collection_id})"
+            )
+
+            # remove identifiers processed in previous batch
+            identifiers = identifiers[len(batch) :]
+            identifiers_list_length = len(identifiers)
+            # measure elapsed seconds
+            elapsed_seconds = time.perf_counter() - start_seconds
+
+            if elapsed_seconds > target_max_execution_time_in_seconds:
```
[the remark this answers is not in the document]
dbernstein: @jonathangreen : My thinking here was to try to minimize the overhead of replacing the task with a potentially large list of identifiers being passed over the wire repeatedly. I thought it might perform better especially if the processing of a batch is fast relative to the cost of task replacement.  I'm happy to remove it if you think that's the way to go.

src/palace/manager/celery/tasks/axis.py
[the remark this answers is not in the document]
dbernstein: nevermind.

tests/manager/api/test_axis.py
dbernstein: I brought this one back in.

dbernstein: As well as the other one.

src/palace/manager/celery/tasks/axis.py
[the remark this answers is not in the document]
jonathangreen: What I was worried about is starvation of other tasks, because mostly they are very short, much less then 2 minutes. I've been treating 2 minutes as the upper bound. That all said, I'm fine with merging it and seeing what happens.

src/palace/manager/celery/tasks/axis.py
```diff
@@ -0,0 +1,404 @@
...
+        total_imported_in_current_task = 0
+        while len(identifiers) > 0:
+            batch = identifiers[:batch_size]
+
+            try:
+                for metadata, circulation in api.availability_by_title_ids(
+                    title_ids=batch
+                ):
+                    process_book(task, session, api, metadata, circulation)
+            except Exception as e:
```
jonathangreen, line 224: Minor: It would be nice to narrow this a bit instead of catching every exception. We were handling `ObjectDeletedError`, `StaleDataError` before, can we just catch those here?

src/palace/manager/celery/tasks/axis.py
```diff
@@ -0,0 +1,404 @@
...
+                wait_time = exponential_backoff(task.request.retries)
+                task.log.error(
```
jonathangreen, line 226: Minor: Might want to use `task.log.exception` here so we get a stack trace if we're trying to debug what caused us to have to retry.

jonathangreen (approved):
Looks good ✅ 

I went though and resolved all the comments that were addressed, and I added a couple more minor comments you might want to consider before merging.

PR #2280, opened 2025-02-10
Build arm containers with new github actions arm runner (PP-2139)

jonathangreen:
## Description

Build our images using the github ARM runners.

This makes several changes:
- Actually run out unit tests on the ARM image (I think this is a bit win, it was too slow with emulation, but since native libraries can cause issues, its nice to verify that our tests actually run in the built arm image).
- Technically our images get pushed every commit now, but they are not tagged until the tests pass. 
  - This lets us use the images in other workflow jobs easily, while making it unlikely anyone will come across a broken image.
  - It has the advantage that if desired, you could pull the image via its hash for debugging. 
- The build takes place in a (pretty ugly IMO) two step process, where native images are built for each platform, then they are combined together into a manifest and tagged. 
  - This is the process that that docker documentation recommends for a multipart build like this, so despite being kind of ugly, it is the officially blessed way to do things. 

## Motivation and Context

Github now has native arm runners: 
https://github.com/community/community/discussions/148648

This allows us to build our ARM images faster, and without relying on emulation which was causing issues (see: https://github.com/actions/runner-images/issues/11471). 

This work is mainly based on this documentation from docker: 
https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners

But adapted to our environment / workflow.

## How Has This Been Tested?

- I did some testing of these workflows on my fork 
- Workflows running in CI on this PR

## Checklist

### DIFF
--- a/.github/actions/poetry/action.yml
+++ b/.github/actions/poetry/action.yml
@@ -51,7 +51,7 @@ runs:
       id: cache
       with:
         path: ${{ steps.poetry-dir.outputs.home }}
-        key: ${{ runner.os }}-poetry${{ inputs.version }}-install-py${{ steps.python-version.outputs.version }}
+        key: ${{ runner.os }}-${{ runner.arch }}-poetry${{ inputs.version }}-install-py${{ steps.python-version.outputs.version }}
 
     - run: curl -sSL https://install.python-poetry.org | python - --yes --version ${{ inputs.version }}
       env:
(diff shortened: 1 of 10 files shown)


.github/actions/poetry/action.yml
```diff
@@ -51,7 +51,7 @@ runs:
       id: cache
       with:
         path: ${{ steps.poetry-dir.outputs.home }}
-        key: ${{ runner.os }}-poetry${{ inputs.version }}-install-py${{ steps.python-version.outputs.version }}
+        key: ${{ runner.os }}-${{ runner.arch }}-poetry${{ inputs.version }}-install-py${{ steps.python-version.outputs.version }}
```
jonathangreen, line 54: This could be broken out to a separate PR if we want. Our action to install poetry didn't take into account the runners architecture before, so it tried to use the same cache for both intel and arm, causing poetry installs to fail on whichever platform wasn't in the cache.

This adds the platform to the cache key, so it will work with new arm based runners.

.github/workflows/build-base-image.yml
```diff
@@ -3,6 +3,13 @@ on:
   schedule:
     # Every Monday at 6:22am Eastern Time
     - cron: '22 10 * * 1'
+  push:
+    # Build base image when the Dockerfile or the workflow file changes
+    branches:
+      - main
+    paths:
+      - .github/workflows/build-base-image.yml
+      - docker/Dockerfile.baseimage
```
jonathangreen, line 12: We now update the base image on pushes to `main` that modify the base-image workflow or dockerfile. Previously this was done as part of `build.yml`, but based on the new structure, pushing this within its own workflow made more sense to me.

.github/workflows/build.yml
```diff
@@ -120,98 +144,222 @@ jobs:
...
+    name: Integration test (${{ matrix.arch.name }})
+    runs-on: ${{ matrix.arch.runner }}
     needs: [build]
     permissions:
       contents: read
     strategy:
       fail-fast: false
       matrix:
-        platform: ["linux/amd64", "linux/arm64"]
-        image: ["scripts", "webapp"]
```
jonathangreen, line 139: The majority of the time in this workflow is spent pulling the image, rather then running the tests, so it made sense to combine the job for scripts and webapp, since running the tests themselves is quick.

docker/Dockerfile.ci
```diff
@@ -1,4 +1,4 @@
-FROM opensearchproject/opensearch:1 as opensearch
+FROM opensearchproject/opensearch:1 AS opensearch
```
jonathangreen, line 1: This isn't strictly necessary, but resolves a build warning about mixed cases in statements

docker/ci/test_migrations.yml
```diff
@@ -1,5 +1,3 @@
-version: "3.9"
-
```
jonathangreen, line 2: Again not strictly necessary, just resolves a build warning about `version` being deprecated

docker/ci/test_scripts.sh
```diff
@@ -9,11 +9,14 @@ container="$1"
...
 
+# Output the version file for debugging
+docker compose exec "$container" cat /var/www/circulation/src/palace/manager/_version.py
+
 # Wait for container to start
 wait_for_runit "$container"
 
 # Make sure database initialization completed successfully
-timeout 240s grep -q 'Initialization complete' <(docker compose logs "$container" -f 2>&1)
+timeout 240s grep -q -e 'Initialization complete' -e "Migrations complete" <(docker compose logs "$container" -f 2>&1)
```
jonathangreen, line 22: I needed to change this, because there is a race condition now that we test both webapp and scripts in the same container, previously we would have always hit initialization, now the first container to start does it.

tests/manager/api/util/test_xray.py
```diff
@@ -1,17 +1,21 @@
-from unittest.mock import MagicMock, call
+from unittest.mock import MagicMock, call, patch
+
+import pytest
```
jonathangreen, line 3: These test changes could be broken out to separate PR. They are necessary because we now run the tests against a container that has a version set. The tests previously assumed this was not the cause, causing some of these tests to fail. 

The changes just mock the `__version__` variable so it doesn't matter if its set or not, the tests will always behave correctly.

docker/Dockerfile.ci
tdilauro: Yeah, it's funny how persnickety Docker is about this. 😂

PR #2283, opened 2025-02-12
Playback time tracking failures return 207 multi-response, instead of 400/404. (PP-2129)

tdilauro:
## Description

For playback time tracking requests, return a 207 multi-status response instead of a 400 or 404 when a collection, library, or identifier is missing or not properly associated.

## Motivation and Context

When audiobook requests have unrecoverable errors, they should not be retried. However, existing versions of client apps do not handle the 400 and 404 responses; so, we send them the 207 multi-status response that they do understand to invalidate each entry individually.

[Jira [PP-2129](https://ebce-lyrasis.atlassian.net/browse/PP-2129)]

## How Has This Been Tested?

- Updated tests and added new ones.
- Tests pass locally.
- CI tests for local branch pass.

## Checklist

[PP-2129]: https://ebce-lyrasis.atlassian.net/browse/PP-2129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palace/manager/api/controller/playtime_entries.py
+++ b/src/palace/manager/api/controller/playtime_entries.py
@@ -51,29 +51,34 @@ class PlaytimeEntriesController(CirculationManagerController):
         except ValidationError as ex:
             return INVALID_INPUT.detailed(ex.json())
 
+        # TODO: For the time being, we need to return a 207 multi-response instead
+        #  of a 404 or 400 problem detail for missing or incorrectly associated
+        #  libraries, collections, and identifiers. We can switch back to the problem
+        #  detail responses once most instances of the client apps support them.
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
This looks good to me! 🚀 

One small suggestion: It might be helpful to our future selves to throw a comment in the code about why we are doing this (mobile apps can't currently handle 401 responses) and maybe a link to the tickets to implement 401 on mobile if we have them. So that eventually once we get is sorted on the mobile side this code in the CM can get dropped.

PR #2285, opened 2025-02-14
Refactor RedisLock (PP-1481)

jonathangreen:
## Description

Small refactor to the `RedisLock` class.

- Add a new parameter to the `lock()` function called `raise_when_not_acquired` that raises an exception if the lock isn't acquired.
- Make the `TaskLock` class only optionally take a `redis_client` parameter.

## Motivation and Context

This is a small refactor to attempt to DRY some uses of the `RedisLock` classes.

Almost all the uses of `RedisLock` have the same pattern where they try to acquire the lock, then they raise (or exit) if the lock wasn't acquired. Since we are doing this everywhere anyway, this should be the default behavior. I added `raise_when_not_acquired` to do this, and refactored existing code to use it.

Note: This changes the behavior of some tasks slightly. They used to silently exit if they couldn't acquire a lock, now they will exit with an error. Since we don't expect tasks to fail to get the lock, I think this is the correct thing to do, but it is a slight change.

The `TaskLock` class used to take a required `redis_client` parameter. Since it also takes a `Task` parameter, which has access to the redis client, we make it optional, and fall back to the redis client on the task.

## How Has This Been Tested?

- Unit tests

## Checklist

### DIFF
--- a/tests/manager/service/redis/models/test_lock.py
+++ b/tests/manager/service/redis/models/test_lock.py
@@ -4,7 +4,12 @@ from unittest.mock import create_autospec
 import pytest
 
 from palace.manager.celery.task import Task
-from palace.manager.service.redis.models.lock import LockError, RedisLock, TaskLock
+from palace.manager.service.redis.models.lock import (
+    LockNotAcquired,
+    LockValueError,
+    RedisLock,
+    TaskLock,
+)
 from tests.fixtures.redis import RedisFixture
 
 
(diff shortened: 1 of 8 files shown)


tests/manager/service/redis/models/test_lock.py
```diff
@@ -114,11 +119,22 @@ def test_lock(self, redis_lock_fixture: RedisLockFixture):
...
+        # raise_when_not_acquired parameter is set to False
         with redis_lock_fixture.no_timeout_lock.lock():
-            with redis_lock_fixture.lock.lock() as acquired:
+            with redis_lock_fixture.lock.lock(
+                raise_when_not_acquired=False
+            ) as acquired:
                 assert not acquired
 
+        # If the raise_when_not_acquired parameter is set, the context manager raises
+        # an exception is the lock is not acquired.
```
tdilauro, line 131: Minor typo here: "...[if] the lock..."

tdilauro (approved):
Looks good. 🎸  

Just one minor typo in a comment.

PR #2287, opened 2025-02-18
Adds collection subscriptions. (PP-1875)

tdilauro:
## Description

Adds implementation for collection subscriptions.

## Motivation and Context

Allow collections to become active or inactive on a schedule. (This is an addition to the existing ability to add/delete collection in real-time via the Admin UI.)

[Jira [PP-1875](https://ebce-lyrasis.atlassian.net/browse/PP-1875)]

## How Has This Been Tested?

- Tests pass locally.
- [CI tests for the associated branch](https://github.com/ThePalaceProject/circulation/actions/runs/13464239089) pass.

## Checklist

[PP-1875]: https://ebce-lyrasis.atlassian.net/browse/PP-1875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/fixtures/database.py
```diff
@@ -616,8 +643,23 @@ def collection(
...
+    def make_collection_inactive(collection: Collection):
+        """Make a collection inactive using some settings that will make it so."""
+        collection.integration_configuration.settings_dict.update(
+            {
+                "subscription_activation_date": datetime.date(2000, 12, 31),
+                "subscription_expiration_date": datetime.date(1999, 1, 1),
+            }
+        )
+        flag_modified(collection.integration_configuration, "settings_dict")
+        assert not collection.is_active
```
jonathangreen, line 661: Minor: Might be nice to use the methods on the integration to make this settings change, since it does a little more validation.

```suggestion
    def make_collection_inactive(self, collection: Collection) -> None:
        """Make a collection inactive using some settings that will make it so."""
        protocol_cls = self._goal_registry_mapping[Goals.LICENSE_GOAL][collection.protocol]
        protocol_cls.settings_update(
            collection.integration_configuration,
            {
                "subscription_activation_date": datetime.date(2000, 12, 31),
                "subscription_expiration_date": datetime.date(1999, 1, 1),
            },
            merge=True
        )
        assert not collection.is_active
```

tests/manager/api/admin/controller/test_collections.py
```diff
@@ -108,13 +109,24 @@ def test_collections_get_with_no_collections(
...
+    )
     def test_collections_get_collections_with_multiple_collections(
         self,
         controller: CollectionSettingsController,
         flask_app_fixture: FlaskAppFixture,
         db: DatabaseTransactionFixture,
+        is_inactive: bool,
     ) -> None:
-        [c1] = db.default_library().associated_collections
+        default_library = db.library(short_name="default", name="Default Library")
```
jonathangreen, line 126: Minor: This could be kind of confusing. I tend to assume the "default library" is what comes back from `db.default_library()`. Might be helpful to either:
- Just use `db.default_library()`
- use `db.library()` but rename the variable so its not called `default_library`

tests/manager/api/controller/test_loan.py
```diff
@@ -1417,14 +1417,14 @@ def test_active_loans(
...
-        assert sync_task.apply_async.call_count == len(
-            patron.library.associated_collections
-        )
-        for library in patron.library.associated_collections:
+        patron_collections = patron.library.active_collections
+        assert sync_task.apply_async.call_count == len(patron_collections)
+        for collection in patron_collections:
             sync_task.apply_async.assert_any_call(
-                (library.id, patron.id, loan_fixture.valid_credentials["password"]),
+                (collection.id, patron.id, loan_fixture.valid_credentials["password"]),
```
jonathangreen, line 1427: 😅

jonathangreen (approved):
I made a couple minor comments on the tests to think about, but overall looks good to me! 🚀

tests/manager/api/admin/controller/test_collections.py
tdilauro: I don't think I can use `db.default_library()` here because of its side effects. I'll change the variable name.

PR #2290, opened 2025-02-20
Simplify getting logger for function outside of a class

jonathangreen:
## Description

Just use the module name for a logger used by a function that doesn't live in a class. 

## Motivation and Context

In https://github.com/ThePalaceProject/circulation/pull/2288 I was looking at how we get a logger for a python function that lives outside of a class, because this is a somewhat common case in our celery tasks. 

I over-thought this by 100x when I initially wrote this function. We don't need the function name in the logger. It would be nice to have if it were easy to do, but the overhead of having to inspect the stack to get a logger this way doesn't seem worth it.

The `logger_for_function` function was only called in a couple places, and I replaced it with `log = logging.getLogger(__name__)` in those places, so the function will just use a logger with its module name.

## How Has This Been Tested?

- Running tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/api/lanes.py
+++ b/src/palace/manager/api/lanes.py
@@ -1,3 +1,5 @@
+import logging
+
 from sqlalchemy.orm import Session
 
 import palace.manager.core.classifier as genres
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
Looks good!

PR #2293, opened 2025-02-24
Update OPDS for distributors to respect the time tracking link (PP-1884)

jonathangreen:
## Description

Previously we overrode the normal time tracking flag parsing. Now we respect the flag set in the feed.

This is mostly test changes. I added an additional entry to the test feed, with the flag set, and added this to the test case to make sure it imports correctly.

## Motivation and Context

See PP-1884.

## How Has This Been Tested?

Running unit tests

## Checklist

### DIFF
--- a/tests/manager/api/test_opds_for_distributors.py
+++ b/tests/manager/api/test_opds_for_distributors.py
@@ -3,6 +3,7 @@ from collections.abc import Callable
 from unittest.mock import MagicMock, patch
 
 import pytest
+from freezegun import freeze_time
 
 from palace.manager.api.circulation_exceptions import (
     CannotFulfill,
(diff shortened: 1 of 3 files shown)


tests/manager/api/test_opds_for_distributors.py
```diff
@@ -526,40 +528,54 @@ def test_import(self, opds_dist_api_fixture: OPDSForDistributorsAPIFixture):
         # it adds delivery mechanisms for books with epub acquisition links
         # and sets pools' licenses_owned and licenses_available.
 
-        # Both works were created, since we can use their acquisition links
+        # All four works in the feed were created, since we can use their acquisition links
```
tdilauro, line 531: Going from "Both" to "All four" by adding one entry confused me for a minute.  🤦🏽‍♂️

tdilauro (approved):
This looks good! ✂️

PR #2294, opened 2025-02-24
Fix Axis360 import import all task.

dbernstein:
## Description
A follow on PR to fix a problem the celery axis.import_all_collections task execution.

## Checklist


dbernstein:
@jonathangreen :  I tested this one locally to make sure it is working.

alembic/versions/20250224_704dd5322783_add_task_type_to_timestamp_type_enum.py
```diff
@@ -0,0 +1,73 @@
...
+        " USING service_type::text::_service_type"
+    )
+    new_type.drop(op.get_bind(), checkfirst=False)
+    # Create and convert to the "old" service_type type
+    old_type.create(op.get_bind(), checkfirst=False)
+    op.execute(
+        "ALTER TABLE timestamps ALTER COLUMN service_type TYPE service_type"
+        " USING service_type::text::service_type"
+    )
+    tmp_type.drop(op.get_bind(), checkfirst=False)
```
jonathangreen, line 73: Minor: What you have here looks like it will work, but since enum updates are so painful, we do have a helper function that might be worth using `pg_update_enum`.

I think it would look something like this (I haven't actually tested this code, just out of memory):

```suggestion
def upgrade():
    # Add task type to the enum
    pg_update_enum(op, "timestamps", "service_type", "service_type", old_options, new_options)


def downgrade():
    # Convert 'task' service_type into 'script'
    op.execute("UPDATE timestamps SET service_type='script' where service_type='task'")
   
    # Update the enum
    pg_update_enum(op, "timestamps", "service_type", "service_type", new_options, old_options)
```

jonathangreen (approved):
🚀

alembic/versions/20250224_704dd5322783_add_task_type_to_timestamp_type_enum.py
dbernstein: I'll try it.

PR #2300, opened 2025-02-25
Update the exceptions returned from overdrive API (PP-2154)

jonathangreen:
## Description

Some minor updates to the Overdrive API based on the investigation I'm doing in PP-2154. 

- If a book is already on hold return the `AlreadyOnHold` exception instead of returning a hold info object.
- If a book is checked out when we go to make a loan, don't do an availability sync, just raise `NoAvailableCopies`. 

In both these cases the Circulation API that calls the Overdrive API was doing the same work that was being done in the Overdrive API, so we should just do the simple thing, and let the caller handle the outcome.

I'll add some context in comments on the PR.

## Motivation and Context

These don't fix the holds issue in PP-2154, but I think they are good changes to make anyway, having followed through some full OD requests in the debugger. 

## How Has This Been Tested?

- Running unit tests
- Locally testing with OD loans and holds

## Checklist

### DIFF
--- a/src/palace/manager/api/overdrive.py
+++ b/src/palace/manager/api/overdrive.py
@@ -1654,11 +1654,11 @@ class OverdriveAPI(
             # The book was successfully placed on hold. Return an
             # appropriate HoldInfo.
             data = response.json()
-            position1, date = self.extract_data_from_hold_response(data)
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
Looks good!

PR #2303, opened 2025-02-26
[PP-2109] Remove loan expiration events.

dbernstein:
## Description

This PR removes the loan expiration event, fixes the test, and does a minor refactor to move the event collection logic to the HoldReaper from the LoanLikeReapMonitor.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2109

## How Has This Been Tested?
Unit tests updated.

## Checklist

### DIFF
--- a/tests/manager/api/test_monitor.py
+++ b/tests/manager/api/test_monitor.py
@@ -189,7 +189,7 @@ class TestLoanlikeReaperMonitor:
         assert [sot_hold] == inactive_patron.holds
         assert 2 == len(current_patron.holds)
 
-        # verify expected circ event count and order for two monitor operations.
+        # verify expected circ event count for hold reaper run
         call_args_list = (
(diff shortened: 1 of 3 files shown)


tests/manager/api/test_monitor.py
```diff
@@ -193,11 +193,9 @@ def test_reaping(
         call_args_list = (
             services_fixture.analytics_fixture.analytics_mock.collect_event.call_args_list
         )
-        assert len(call_args_list) == 4
+        assert len(call_args_list) == 2
```
tdilauro, line 196: I don't think the comment on l. 192 makes sense anymore, in that you can no longer establish the order of anything. Might be worth changing that comment to note that we don't expect events from the reaped loans.

tdilauro (approved):
This looks good! 🎸 

One small requested change to a comment below.

PR #2307, opened 2025-02-27
Add some better logging when we have issues parsing an identifier (PP-2157)

jonathangreen:
## Description

Output the exception and a stack trace into our logs when an exception prevents us from parsing an identifier.

## Motivation and Context

Currently when an exception is triggered parsing an identifier, we don't include any details about the exception, or a stack trace we just output:
```
An unexpected exception occurred during parsing identifier
```

This makes it hard to know what caused the identifier to fail to parse. This gives us some additional logging info.

## How Has This Been Tested?

- Tested locally

## Checklist

### DIFF
--- a/src/palace/manager/core/opds_import.py
+++ b/src/palace/manager/core/opds_import.py
@@ -427,7 +427,7 @@ class BaseOPDSImporter(
                 parsed_identifier, _ = result
         except Exception as e:
             self.log.exception(
-                f"An unexpected exception occurred during parsing identifier {identifier}: {e}"
+                f"An unexpected exception occurred during parsing identifier '{identifier}': {e}"
             )
 
         return parsed_identifier
--- a/src/palace/manager/core/opds_import.py
+++ b/src/palace/manager/core/opds_import.py
@@ -425,9 +425,9 @@ class BaseOPDSImporter(
             result = Identifier.parse_urn(self._db, identifier)
             if result is not None:
                 parsed_identifier, _ = result
-        except Exception:
-            self.log.error(
-                f"An unexpected exception occurred during parsing identifier {identifier}"
+        except Exception as e:
+            self.log.exception(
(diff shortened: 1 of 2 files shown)


tests/manager/core/test_opds_import.py
```diff
@@ -1476,6 +1476,33 @@ def test_wayfless_url_no_saml_entity_id(self, wayfless_circulation_api):
...
+        test_identifier = "test"
+        value = importer.parse_identifier(test_identifier)
+        assert value == "returned value"
+        mock_parse_urn.assert_called_once_with(
+            opds_importer_fixture.db.session, test_identifier
+        )
+
+        # In the case of an exception, we log the relevant info.
+        mock_parse_urn.reset_mock()
+        mock_parse_urn.side_effect = ValueError("My god, it's full of stars")
```
tdilauro, line 1498: Hahaaaa. That is so "2001..." of you!

src/palace/manager/core/opds_import.py
```diff
@@ -425,9 +425,9 @@ def parse_identifier(self, identifier: str | None) -> Identifier | None:
             result = Identifier.parse_urn(self._db, identifier)
             if result is not None:
                 parsed_identifier, _ = result
-        except Exception:
-            self.log.error(
-                f"An unexpected exception occurred during parsing identifier {identifier}"
+        except Exception as e:
+            self.log.exception(
+                f"An unexpected exception occurred during parsing identifier {identifier}: {e}"
```
tdilauro, line 430: Very minor: Since we're touching this, could we quote the identifier, so that it doesn't blend in with the rest of the message?

```
f"An unexpected exception occurred during parsing identifier '{identifier}': {e}"
```

tdilauro (approved):
Looks good! 🪵 

One very minor suggestion. Feel free to ignore it.

PR #2308, opened 2025-02-27
Update simple auth to allow no password auth

jonathangreen:
## Description

Updates the configuration for the Simple Auth integration to allow auth with no password.

## Motivation and Context

I was using this integration to do some testing with a CT library that doesn't require passwords. Instead of setting up port forwarding or anything, i just setup simple auth as a fake ILS while I was testing.

I expected that I could just enter no password, and it would act as if the integration didn't need a password, but this wasn't the case. Since this integration is mostly just used in testing it seems reasonable to support this case.

## How Has This Been Tested?

- Tested locally
- Added test in CI

## Checklist


tdilauro (approved):
🎉

PR #2315, opened 2025-03-04
Type hint datasource, identifier and classification models (PP-1476)

jonathangreen:
## Description

- Adds the datasource, identifier and classification sqlalchemy models to strict mypy type checking
- Adds a DB migration that updates the database schema so some more columns are not nullable 
  - Makes some of the type hinting easier
  - Querying all our production DB, none of them actually had nulls in these columns

## Motivation and Context

This is broken out from other work to type hint the metadata layer done in https://github.com/ThePalaceProject/circulation/pull/2316. Doing this to set up some work to convert the opds importers to use celery queues.

## How Has This Been Tested?

- CI tests
- Production DB queries

## Checklist


alembic/versions/20250304_63825d889633_set_model_nullable_constraints.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    op.alter_column("datasources", "name", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("genres", "name", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column(
+        "identifiers", "type", existing_type=sa.VARCHAR(length=64), nullable=False
+    )
+    op.alter_column(
+        "identifiers", "identifier", existing_type=sa.VARCHAR(), nullable=False
+    )
+    op.alter_column("subjects", "type", existing_type=sa.VARCHAR(), nullable=False)
```
jonathangreen, line 40: It made sense when setting the type hints to make all of these columns non-nullable. Looking at our production databases we didn't have any nulls in any of these columns.

src/palace/manager/sqlalchemy/model/classification.py
```diff
@@ -204,86 +205,72 @@ def describes_format(self):
...
             new = False
-        if name and not subject.name:
+        if name and subject and not subject.name:
             # We just discovered the name of a subject that previously
             # had only an ID.
             subject.name = name
         return subject, new
 
-    @classmethod
-    def common_but_not_assigned_to_genre(
```
jonathangreen, line 244: These two functions were unused, so I removed them rather then type hint them.

src/palace/manager/sqlalchemy/model/datasource.py
```diff
@@ -160,25 +185,9 @@ def lookup_hook():
...
-        name = uri[len(cls.URI_PREFIX) :]
-        return unquote(name)
-
-    @classmethod
-    def from_uri(cls, _db, uri):
-        return cls.lookup(_db, cls.name_from_uri(uri))
-
-    @property
-    def uri(self):
-        return self.URI_PREFIX + quote(self.name)
```
jonathangreen, line 178: These also weren't used

src/palace/manager/sqlalchemy/model/identifier.py
```diff
@@ -69,11 +74,8 @@ def parse(self, identifier_string: str) -> tuple[str, str] | None:
 class ISBNURNIdentifierParser(IdentifierParser):
     """Parser for ISBN URN IDs."""
 
-    def __init__(self):
-        self._logger = logging.getLogger(__name__)
-
     def parse(self, identifier_string: str) -> tuple[str, str] | None:
-        self._logger.debug(f'Started parsing identifier string "{identifier_string}"')
+        self.log.debug(f'Started parsing identifier string "{identifier_string}"')
```
jonathangreen, line 78: While I was in here, I updated all the loggers to use `LoggerMixin`

src/palace/manager/sqlalchemy/model/identifier.py
```diff
@@ -319,48 +306,58 @@ def __repr__(self):
...
-        be the value of `asin`.
-        """
-        asin = asin.strip().replace("-", "")
-        if isbnlib.is_isbn10(asin):
-            asin = isbnlib.to_isbn13(asin)
-        if isbnlib.is_isbn13(asin):
-            type = cls.ISBN
-        else:
-            type = cls.ASIN
-        return cls.for_foreign_id(_db, type, asin, autocreate)
```
jonathangreen, line 336: Not used

src/palace/manager/sqlalchemy/model/identifier.py
```diff
@@ -811,9 +853,10 @@ def add_link(
         elif media_type and not resource.representation:
             # We know the type of the resource, so make a
             # Representation for it.
-            resource.representation, is_new = get_one_or_create(
+            rep, is_new = get_one_or_create(
                 _db, Representation, url=resource.url, media_type=media_type
             )
+            resource.representation = rep
```
jonathangreen, line 859: Mypy didn't like putting this into a unpacked tuple, putting it on its own line was fine. Just some mypy weirdness, but not a big change so I think its fine.

tests/manager/core/classifiers/test_classifier.py
```diff
@@ -429,7 +429,7 @@ def test_no_children_or_ya_signal_from_distributor_implies_book_is_for_adults(
         # YA. classifier.audience_weights[AUDIENCE_ADULT] will be set
         # to 500.
         i = data.identifier
-        source = DataSource.lookup(session, DataSource.OVERDRIVE)
+        source = DataSource.lookup(session, DataSource.OVERDRIVE, autocreate=True)
```
jonathangreen, line 432: This is just done for typing benefit, since setting autocreate make sure we always get a `Datasource` back rather then `Datasource | None`

src/palace/manager/sqlalchemy/model/identifier.py
tdilauro: Even though I must've seen it in the code at some point, I had no conscious idea that `resource.representation, is_new = get_one_or_create(...` was even allowed in Python. The more you know.

src/palace/manager/sqlalchemy/model/identifier.py
```diff
@@ -932,26 +985,32 @@ def classify(
...
                 Classification.data_source == data_source,
             )
-            all_classifications = all_classifications.all()
+            all_classifications = all_classifications_query.all()
             classification = all_classifications[0]
             for i in all_classifications[1:]:
                 _db.delete(i)
 
-        classification.weight = weight
+        if classification.weight != weight:
```
tdilauro, line 1002: Why can't we go straight to the assignment here? Is this an optimization for the database? Something to appease `mypy`? Might be worth a comment (or just get smarter code reviewers 😆).

tdilauro (approved):
Looks good! 🚀 

A few comments/questions below.

src/palace/manager/sqlalchemy/model/classification.py
```diff
@@ -541,28 +537,30 @@ def create():
...
+                genres.append(genre)
         return genres
 
     @property
-    def default_fiction(self):
+    def default_fiction(self) -> bool | None:
         if self.name not in classifier.genres:
             return None
-        return classifier.genres[self.name].is_fiction
+        return classifier.genres[self.name].is_fiction  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
jonathangreen: I decided I needed to cut my losses and stop type hinting at this file 😆

src/palace/manager/sqlalchemy/model/identifier.py
jonathangreen: An assignment would work fine here. I added this if statement because `weight` should already be set most of the time, and I didn't want to have SQLAlchemy flush it out to the database if its already set. I'm not sure if its a necessary optimization or not.

jonathangreen: did some reading of sqlalchemy docs, and apparently this isn't necessary, the ORM will handle internally, so i reverted this change

PR #2322, opened 2025-03-06
Move analytics provider into service.analytics package

jonathangreen:
## Description

Move our analytics provider into the `palace.manager.service.analytics` package.

When these files were moved, it started causing some type errors, because everything in `service` has strict type checking enabled, so I also added some type hints that were missing in:
- Local analytics
- S3 analytics
- `CirculationEvent` model

## Motivation and Context

Doing this as part of some refactoring work in PP-1470.

This PR sets up some more analytics refactoring done in https://github.com/ThePalaceProject/circulation/pull/2323.

## How Has This Been Tested?

- Ran tests

## Checklist


tdilauro (approved):
Looks good! 👍🏽

PR #2323, opened 2025-03-06
Refactor our analytics providers

jonathangreen:
## Description

This PR refactors our analytics providers to:
- Create a common abstract base class `AnalyticsProvider`
- Create a pydantic model for the json data that we output to s3 `AnalyticsEventData`
  - This model contains all the information needed to publish an analytics event, disconnected from the DB session, so we are able to save it across transaction boundaries without worrying about stale objects.
- Update our analytics providers to consume `AnalyticsEventData`
  - The local analytics provider and the s3 analytics provider both did overlapping serialization of the analytics data, they both now rely on the data provided by `AnalyticsEventData`
- Update opds_odl celery tasks to use the new `AnalyticsEventData` structure, instead of reattaching DB objects across the session.

## Motivation and Context

Did this refactoring as part of the work I was doing in PP-1470, as some of the reapers need to be able to persist event data across sessions similar to the opds_odl celery tasks

## How Has This Been Tested?

- Running tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/service/analytics/eventdata.py
+++ b/src/palace/manager/service/analytics/eventdata.py
@@ -1,0 +1,197 @@
+from __future__ import annotations
+
+from datetime import datetime
+from functools import cached_property
+from uuid import UUID
+
+import flask
+from pydantic import (
+    AwareDatetime,
+    BaseModel,
+    ConfigDict,
+    computed_field,
+    field_serializer,
+)
+from pydantic_core.core_schema import FieldSerializationInfo
+from typing_extensions import Self
+
+from palace.manager.sqlalchemy.model.library import Library
+from palace.manager.sqlalchemy.model.licensing import LicensePool
+from palace.manager.sqlalchemy.model.patron import Patron
+from palace.manager.util.datetime_helpers import utc_now
+from palace.manager.util.log import LoggerMixin
+
+
+class AnalyticsEventData(BaseModel, LoggerMixin):
+    """
+    This class represents the data that is sent to the analytics provider.
+
+    It is a Pydantic model that is used to validate the data before it is sent. It stores
+    all the data without references to any database objects, so it can be serialized, stored
+    and sent outside a database transaction.
+    """
+
+    type: str
+    start: AwareDatetime
+
+    # TODO: We include the 'end' field as a copy of 'start' because that is
+    #   what the pre-pydantic implementation did. Since this is just duplicated
+    #   we should remove it in a future release.
+    @computed_field  # type: ignore[prop-decorator]
+    @cached_property
+    def end(self) -> AwareDatetime:
+        return self.start
+
+    library_id: int
+    library_name: str
+    library_short_name: str
+    old_value: int | None
+    new_value: int | None
+
+    @computed_field  # type: ignore[prop-decorator]
+    @cached_property
+    def delta(self) -> int | None:
+        if self.new_value is None or self.old_value is None:
+            return None
+        return self.new_value - self.old_value
+
+    location: str | None
+    license_pool_id: int | None
+    publisher: str | None
+    imprint: str | None
+    issued: datetime | None
+    published: datetime | None
+    medium: str | None
+    collection: str | None
+    collection_id: int | None
+    identifier_type: str | None
+    identifier: str | None
+    data_source: str | None
+    distributor: str | None
+    audience: str | None
+    fiction: bool | None
+    summary_text: str | None
+    quality: float | None
+    rating: int | None
+    popularity: int | None
+    genre: str | None
+    availability_time: AwareDatetime | None
+    licenses_owned: int | None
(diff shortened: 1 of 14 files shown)


src/palace/manager/service/analytics/s3.py
```diff
@@ -1,202 +1,38 @@
...
 
 
-class S3AnalyticsProvider(LocalAnalyticsProvider):
+class S3AnalyticsProvider(AnalyticsProvider):
     """Analytics provider storing data in a S3 bucket."""
 
     def __init__(self, s3_service: S3Service | None):
         self.s3_service = s3_service
 
-    @staticmethod
```
jonathangreen, line 26: This function has been largely moved to `AnalyticsEventData.create`.

src/palace/manager/service/analytics/s3.py
```diff
@@ -1,202 +1,38 @@
...
-        )
-        content = json.dumps(
-            event,
-            default=str,
-            ensure_ascii=True,
-        )
+        # We exclude the collection_id from the json because it wasn't included
+        # in our pre-pydantic implementation, it was only used in the file key.
+        # TODO: See if adding collection_id will cause any issue with the ingest process.
+        content = event.model_dump_json(exclude={"collection_id"})
```
jonathangreen, line 32: @dbernstein I'm not sure how the file ingestion process from s3 to redshift works, if the json contains an unexpected field will that cause any issues? If not, we can probably just include `collection_id` instead of excluding it here.

src/palace/manager/service/analytics/eventdata.py
```diff
@@ -0,0 +1,197 @@
...
+    )
+
+    # We serialize the datetime fields as strings in the JSON output, to match what
+    # the output looked like before we switched to Pydantic.
+    # TODO: It would be nice to be able to drop this and just use the default
+    #   which is to serialize datetimes as ISO8601 strings. Need to see if the
+    #   analytics consumers can handle that.
+    @field_serializer("start", "end", "availability_time", when_used="json")
+    def serialize_dt(self, value: datetime, _info: FieldSerializationInfo) -> str:
+        return str(value)
```
jonathangreen, line 108: @dbernstein what sort of logic are we using to parse this timestamp when we are ingesting this data into redshift? I did this for compatibility with the previous implementation, but depending on that parsing, we might be able to drop this custom serializer.

src/palace/manager/sqlalchemy/model/circulationevent.py
```diff
@@ -126,46 +122,3 @@ class CirculationEvent(Base):
 
     # The time format used when exporting to JSON.
     TIME_FORMAT = "%Y-%m-%dT%H:%M:%S+00:00"
-
-    @classmethod
```
jonathangreen, line 130: Since the local analytics provider now uses `AnalyticsEventData` to create its `CirculationEvent` objects, this method didn't really make sense. So I removed it, so it doesn't cause confusion in the future.

tests/manager/service/analytics/test_analytics.py
```diff
@@ -48,53 +31,3 @@ def test_init_analytics(self):
         assert len(analytics.providers) == 2
         assert type(analytics.providers[0]) == LocalAnalyticsProvider
         assert type(analytics.providers[1]) == S3AnalyticsProvider
-
-    def test_user_agent_capture(
```
jonathangreen, line 52: The user agent logic was moved into `AnalyticsEventData.create`, so this test was moved into `test_eventdata.py`.

tests/manager/sqlalchemy/model/test_circulationevent.py
```diff
@@ -107,84 +107,6 @@ def test_new_title(self, db: DatabaseTransactionFixture):
         # updating the dataset.
         assert 0 == event.license_pool.licenses_owned
 
-    def test_log(self, db: DatabaseTransactionFixture):
```
jonathangreen, line 110: This test was moved to `test_local.py` in the `test_collect_end_to_end` test since that is where most of the logic this was testing moved to.

jonathangreen:
@dbernstein I tagged you in a couple comments with questions, since you are more familiar with how we get data from s3 into redshift then I am.

src/palace/manager/service/analytics/s3.py
dbernstein: Adding unexpected fields should not be a problem.

src/palace/manager/service/analytics/eventdata.py
dbernstein: We are not doing anything special.  We pull the dates as strings (presumably iso8601)  into a temporary database and then they are automatically converted to `timestamp` type in redshift.

jonathangreen: Okay. I'll remove this as well!

PR #2324, opened 2025-03-06
Convert database_reaper to celery (PP-1470)

jonathangreen:
## Checklist


src/palace/manager/celery/tasks/reaper.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    Remove seemingly abandoned holds from the database.
+    """
+    cutoff = utc_now() - timedelta(days=365)
+    analytics_service = task.services.analytics.analytics()
+    query = (
+        select(Hold)
+        .where(Hold.start < cutoff, or_(Hold.end == None, Hold.end < utc_now()))
+        .order_by(Hold.id)
+        .limit(batch_size)
```
jonathangreen, line 191: This query has been simplified in two ways:
- We aren't filtering out unlimited access and open access license pools, because you can't place holds on these items anyway. 
- We drop the concept of "source of truth" integrations.
  - The only source of truth integration we had was OPDS for distributors, which only supports unlimited access items, so you couldn't place a hold anyway. 

So this query should be effectively equivalent, even though its dropped some of the conditions.

src/palace/manager/celery/tasks/reaper.py
```diff
@@ -0,0 +1,239 @@
...
+
+    now = utc_now()
+    deletion_query = delete(Loan).where(
+        Loan.license_pool_id == LicensePool.id,
+        LicensePool.unlimited_access == False,  # type: ignore[comparison-overlap]
+        LicensePool.open_access == False,
+        or_(
+            Loan.end < now,
+            and_(Loan.start < now - timedelta(days=90), Loan.end == None),
+        ),
```
jonathangreen, line 233: This query drops the concept of "source of truth" integrations.

Previously the only source of truth integration we had was OPDS for distributors, which only supports unlimited access items, so these items are filtered out by the `LicensePool.unlimited_access == False` check, so no need to complicate the query by filtering on the collection protocol.

So this query should be effectively equivalent to what we had before.

src/palace/manager/celery/tasks/reaper.py
```diff
@@ -0,0 +1,239 @@
...
+@shared_task(queue=QueueNames.default, bind=True)
+def credential_reaper(task: Task) -> None:
+    """
+    Remove Credentials that expired more than a day ago.
+    """
+    cutoff = utc_now() - timedelta(days=1)
+    deletion_query = delete(Credential).where(Credential.expires < cutoff)
+    with task.transaction() as session:
+        rows_removed = _execute_delete(session, deletion_query)
+    task.log.info(f"Deleted {pluralize(rows_removed, 'expired credential')}.")
```
jonathangreen, line 49: For as many of the reapers as I could, i updated them to use a direct SQL delete statement, rather then loading each record into the ORM, to just delete it. Should allow us to perform the deletes much faster.

src/palace/manager/core/monitor.py
```diff
@@ -813,271 +809,3 @@ class CustomListEntryWorkUpdateMonitor(CustomListEntrySweepMonitor):
...
-    def run_once(self, *args, **kwargs):
-        rows_deleted = self.query().delete()
-        self._db.commit()
-        return TimestampData(achievements="Items deleted: %d" % rows_deleted)
-
-
-ReaperMonitor.REGISTRY.append(MeasurementReaper)
-
-
-class ScrubberMonitor(ReaperMonitor):
```
jonathangreen, line 1006: I didn't bother migrating either of the srubber monitors, since we are not using the patron location feature. Its pretty NYPL specific. I'm going to instead just drop whats left of that code in a follow up PR to this one.

src/palace/manager/service/celery/celery.py
```diff
@@ -92,6 +92,62 @@ def beat_schedule() -> dict[str, Any]:
...
+                hour="2",
+            ),  # Once a day at 2:35 AM
+        },
+        "loan_reaper": {
+            "task": "reaper.loan_reaper",
+            "schedule": crontab(
+                minute="40",
+                hour="2",
+            ),  # Once a day at 2:40 AM
+        },
```
jonathangreen, line 150: These all used to run at 2am every day. Since they are all separate task now, I decided to space them out a little bit.

src/palace/manager/sqlalchemy/model/patron.py
```diff
@@ -168,30 +168,36 @@ class Patron(Base, RedisKeyMixin):
     neighborhood: str | None = None
 
     loans: Mapped[list[Loan]] = relationship(
-        "Loan", back_populates="patron", cascade="delete", uselist=True
+        "Loan",
+        back_populates="patron",
+        cascade="delete",
+        uselist=True,
+        passive_deletes=True,
```
jonathangreen, line 175: See SQLAlchemy docs: https://docs.sqlalchemy.org/en/20/orm/cascades.html#passive-deletes

I updated these relationships to set the `passive_deletes` parameter since I added a DB level foreign key cascade.

tests/manager/api/test_monitor.py
```diff
@@ -1,260 +0,0 @@
...
-from palace.manager.sqlalchemy.model.circulationevent import CirculationEvent
-from palace.manager.sqlalchemy.model.datasource import DataSource
-from palace.manager.sqlalchemy.model.patron import Annotation
-from palace.manager.sqlalchemy.util import get_one_or_create
-from palace.manager.util.datetime_helpers import utc_now
-from tests.fixtures.database import DatabaseTransactionFixture
-from tests.fixtures.services import ServicesFixture
-
-
-class TestLoanlikeReaperMonitor:
```
jonathangreen, line 20: I moved the bulk of these tests into `test_reaper.py`

tests/manager/core/test_monitor.py
```diff
@@ -1031,298 +1018,3 @@ def test_set_item(self, db: DatabaseTransactionFixture):
         monitor = CustomListEntryWorkUpdateMonitor(db.session)
         monitor.process_item(entry)
         assert old_work == entry.work
-
-
-class MockReaperMonitor(ReaperMonitor):
```
jonathangreen, line 1036: I moved the bulk of these tests into `test_reaper.py`

src/palace/manager/core/monitor.py
jonathangreen: Here is the follow up PR: https://github.com/ThePalaceProject/circulation/pull/2326 (still in progress)

tests/manager/celery/tasks/test_reaper.py
tdilauro, line 1: I'm assuming that most of these just moved over, so I did't really give this much review. If there are specific things you want me to look at, just let me know.

tdilauro (approved):
LGTM! 🎸🤘🏽

PR #2325, opened 2025-03-06
[PP-2095] upload playtime reports to google drive

dbernstein:
## Description
This PR does several things:

1.  It establishes the GoogleDrive service which wraps the google drive client and provides  some useful methods for creating, deleting, and sharing files and folders.
2. For the playtime reports,  I have removed the s3 upload  + email and replaced it with a google drive upload using the email address as the user for sharing. 
3. It now separates the playtime reports into separate CSVs by collection name.
4. I have added a Github Repository secret to store the test credentials for  the google_drive module tests.  There is no google drive service endpoint emulator like we have for S3.  I thought it was important to use the live drive service using test credentials in order to have confidence that it will work in production rather than trying to mock the HTTP responses.
5. The google drive tests will be skipped if the `PALACE_TEST_GOOGLE_DRIVE_SERVICE_ACCOUNT_KEY_FILE_PATH` is not present.
 
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2095

## How Has This Been Tested?
I have unit tests that run against a test service_account in google drive.
I also stepped through the unit tests to confirm that the file creation, deletion  and sharing are visible to the user who is the target of the sharing. 

I will need to set up dev, staging,  and production credentials in `thepalaceproject` google cloud account, store those in the hosting-playbook secrets and ensure that the  `PALACE_GOOGLE_DRIVE_SERVICE_ACCOUNT_KEY_FILE_PATH` is being written as a secret and passed to the container when docker compose starts up.

## Checklist


.github/workflows/build.yml
```diff
@@ -264,6 +264,12 @@ jobs:
           path: ${{ runner.temp }}
           pattern: digests-${{ matrix.arch.name }}
 
+      - name: Create Google Drive Credentials File
+        run: |
+          mkdir ./secrets
+          google_service_account_json_secret_file=./secrets/test_google_service_account.json
```
dbernstein, line 270: Not sure how to make this DRYier - but there is duplicate config for the different targets.

jonathangreen: I'm not sure this is necessary. These images don't actually run tests, they just run the scripts:
- ./docker/ci/test_webapp.sh webapp
- ./docker/ci/test_scripts.sh scripts

and i don't think it will make much difference if the keys are configured or not to run these scripts

src/palace/manager/scripts/playtime_entries.py
```diff
@@ -179,76 +188,114 @@ def do_run(self):
...
+        leaf_folder = folder_results[-1]
+
+        # share root folder if not already shared
+        if recipient:
+            google_drive.share(
+                file_or_folder_id=root_folder["id"], email_addresses=[recipient]
+            )
+            self.log.info(
+                f"shared root folder ({'/'.join(root_folders)}) with {recipient} in Google Drive"
+            )
```
jonathangreen, line 220: minor: Is this sharing actually the behavior we want? Seems like we would just want this to upload to a google drive, and we can share the folder however is needed offline.

jonathangreen:
It looks like codecov is failing because we are skipping the google drive tests in the PR, which is where we report usage. If we don't have keys configured, can we fall back to local mocks or something? It would be nice to not have codecov think that we don't cover this at all.

It also looks like the docker build is failing here, because google was returning 500 errors for some reason. Depending how often that happens, its a pretty good reason not to have our CI calling out to an external service. It means that google drive blips can cause our CI tests to fail.

jonathangreen:
I re-ran the docker build to see if it will succeed this time

src/palace/manager/scripts/playtime_entries.py
```diff
@@ -179,76 +188,114 @@ def do_run(self):
...
-            _produce_report(
-                writer,
-                date_label=report_date_label,
-                records=self._fetch_report_records(start=start, until=until),
+        google_drive: GoogleDriveService = self.services.google_drive.palace_internal()
+
+        # create directory hierarchy
+        root_folders = os.environ.get(
+            Configuration.PALACE_GOOGLE_DRIVE_ROOT_ENVIRONMENT_VARIABLE,
+            "palace/no-env-specified",
```
jonathangreen, line 199: minor: This does not seem like a reasonable default. Wouldn't we just want the files to be placed in the root?

tests/fixtures/google_drive.py
jonathangreen, line 1: Why an empty file here?

jonathangreen (commented):
A couple things needed before this gets merged:
- README needs updates for new environment variables and re-purposing of REPORTING_EMAIL_ENVIRONMENT_VARIABLE.
- We need to figure out a way to get coverage for this stuff on or main tests, not just the docker builds. If anything, I'd say take the keys out of the docker builds and just use them in the test workflow.
- CI is failing with 500 errors from google. I tried re-running it, but its still failing for me.

poetry.lock
```diff
@@ -23,14 +23,14 @@ tz = ["tzdata"]
 
 [[package]]
 name = "amqp"
-version = "5.2.0"
+version = "5.3.1"
```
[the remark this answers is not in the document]
dbernstein: I'll get the latest poetry.lock from main and do a `poetry lock --no-update`

src/palace/manager/service/google_drive/container.py
```diff
@@ -0,0 +1,16 @@
...
+from dependency_injector.containers import DeclarativeContainer
+from dependency_injector.providers import Provider
+
+from palace.manager.service.google_drive.google_drive import GoogleDriveService
+
+
+class GoogleDrive(DeclarativeContainer):
+    config = providers.Configuration()
+
+    palace_internal: Provider[GoogleDriveService] = providers.Singleton(
```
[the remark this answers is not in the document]
dbernstein: Fair point.

.github/workflows/build.yml
dbernstein: Without this configuration,  the docker compose command fails because the google service config file is missing when the docker tries to bind the secrets.

docker-compose.yml
```diff
@@ -28,6 +28,10 @@ x-cm-variables: &cm
     PALACE_TEST_MINIO_USER: "palace"
     PALACE_TEST_MINIO_PASSWORD: "test123456789"
     PALACE_TEST_REDIS_URL: "redis://redis:6379/3"
+    PALACE_TEST_GOOGLE_DRIVE_SERVICE_ACCOUNT_KEY_FILE_PATH: /run/secrets/test_google_service_account_json
+  secrets:
+    - test_google_service_account_json
+
```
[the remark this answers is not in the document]
dbernstein: The contents of the Google Driver Service Account key file is json.  When I was trying to pass the config as a string representing the contents of a json to the application via an environment variable the json was unreadable.  It seemed that no matter what I tried - escaping the quotes, using single quotes, escaping single quotes, etc I couldn't seem to deliver valid json via the environment variable.

dbernstein:
@jonathangreen :  I reworked this PR so that we're passing the google drive creds via an env variable rather than creating a file and using the docker secrets mechanism to mount it in the container.   I think I addressed all the issues except for the one around injecting the configuration.  I didn't fully understand what you were suggesting there.

src/palace/manager/service/google_drive/google_drive.py
```diff
@@ -0,0 +1,169 @@
...
+if TYPE_CHECKING:
+    from googleapiclient._apis.drive.v3 import File, Permission
+
+
+class GoogleDriveService(LoggerMixin):
+    def __init__(
+        self,
+        service_account_key_file_path: str,
+    ) -> None:
+        credentials = service_account.Credentials.from_service_account_file(
```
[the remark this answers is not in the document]
dbernstein: I'm not sure if I addressed the issue here.  I'm not sure I understand what you're suggesting.

jonathangreen (approved):
🚀

PR #2326, opened 2025-03-07
Remove storing patron neighborhood / location 🔥

jonathangreen:
## Checklist


src/palace/manager/api/millenium_patron.py
```diff
@@ -187,8 +153,6 @@ def library_settings_class(cls) -> type[MilleniumPatronLibrarySettings]:
     RECORD_NUMBER_FIELD = "p81"  # e.g., "RECORD #[p81]"
     PATRON_TYPE_FIELD = "p47"  # e.g., "P TYPE[p47]"
     EXPIRATION_FIELD = "p43"  # e.g., "EXP DATE[p43]"
-    HOME_BRANCH_FIELD = "p53"  # e.g., "HOME LIBR[p53]"
```
tdilauro, line 190: I was a little worried about `HOME_BRANCH_FIELD` here, since the "p53" can be used in patron auth restrictions, but it doesn't use this constant and eliminating this constant won't affect how the ILS properties are modeled. So, this looks fine.

tdilauro (approved):
🔥

src/palace/manager/service/analytics/eventdata.py
```diff
@@ -0,0 +1,198 @@
...
+    @cached_property
+    def delta(self) -> int | None:
+        if self.new_value is None or self.old_value is None:
+            return None
+        return self.new_value - self.old_value
+
+    # TODO: This field has been removed from the database model, but is still included
+    #   in the response for backwards compatibility. It should be removed in a future
+    #   release.
+    location: None = None
```
jonathangreen, line 61: @dbernstein Do we need the "location" field in json for redshift? Or can this just be dropped?

PR #2327, opened 2025-03-07
[PP-2217] Fix Axis360 Celery DetachedInstanceError and downstream Ope…

dbernstein:
…rationalError (deadlock) error.

## Description

This small fix should address the task errors and the resulting back up they are causing.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2217

## Checklist


jonathangreen (approved):
Looks good. 

Couple thoughts though. Should you be updating to use `collection_name` in these places as well:
https://github.com/ThePalaceProject/circulation/blob/4447d7e9a5406720f6e6d2db50fe58cd17f56283/src/palace/manager/celery/tasks/axis.py#L231
https://github.com/ThePalaceProject/circulation/blob/4447d7e9a5406720f6e6d2db50fe58cd17f56283/src/palace/manager/celery/tasks/axis.py#L237
https://github.com/ThePalaceProject/circulation/blob/4447d7e9a5406720f6e6d2db50fe58cd17f56283/src/palace/manager/celery/tasks/axis.py#L241

And should you be using `collection_id` here:
https://github.com/ThePalaceProject/circulation/blob/4447d7e9a5406720f6e6d2db50fe58cd17f56283/src/palace/manager/celery/tasks/axis.py#L271

PR #2330, opened 2025-03-10
[PP-2233] Prevent deadlocks in axis.import_identifiers and axis.reap_…

dbernstein:
…collection

## Description

This update does the same two things for the reap collections  and import collections methods to try to address the deadlock issue. 

The deadlocks seem to be arising because all the CM's axis collections are being reaped and imported in parallel.  As a result when the data they are updating overlap - such us updates to an edition's publish - it is easy for one of many collections to get into a database deadlock conflict with another. 

To try to address this problem I've done 3 things: 

1.  I've introduced a time delay of 10 seconds between firings of import_collection and reap_collection tasks in order to prevent them from executing too close to the same moment.  
2. I've reduced the batch size from 25 to 10 items per task.  So both types of task will process at most 10 items and then requeue the task.  
3. I also added an OperationalError to ensure when there is a deadlock that the task is retried.

Also added some logging to see the elapsed time and the number of items processed in one message.

We may need to fine tune the `subtask_execution_interval_in_secs` and/or the batch size in order to find the right balance.

## Motivation and Context

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/axis.py
+++ b/src/palace/manager/celery/tasks/axis.py
@@ -1,8 +1,9 @@
 import time
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from celery import shared_task
 from sqlalchemy import select
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import Session
 from sqlalchemy.orm.exc import ObjectDeletedError, StaleDataError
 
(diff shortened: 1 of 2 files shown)


src/palace/manager/celery/tasks/axis.py
```diff
@@ -214,53 +224,34 @@ def log_run_end_message() -> None:
...
+        batch = identifiers[:batch_size]
+        batch_length = len(batch)
+        if batch_length > 0:
             try:
                 for metadata, circulation in api.availability_by_title_ids(
                     title_ids=batch
                 ):
                     process_book(task, session, api, metadata, circulation)
-            except (ObjectDeletedError, StaleDataError) as e:
+            except (ObjectDeletedError, StaleDataError, OperationalError) as e:
```
jonathangreen, line 237: Ideally, I'd like to not do this. I think adding this here papers over what is a bigger problem.

We should be able to sort out a way not to have deadlocks happen, rather then retry when they do happen. Deadlocks are slow, they stop all the processing while the deadlock is happening, and they take awhile to time out. Its also possible to have deadlocks that the PG deadlock detector does not catch, that just freeze until the session times out.

dbernstein:
I'm going to close this PR since the other approach seems to have brought the deadlocks down to a minimum. c.f. https://github.com/ThePalaceProject/circulation/pull/2333

PR #2332, opened 2025-03-11
Add type hints to Overdrive API (PP-2221)

jonathangreen:
## Description

Add type hints to the Overdrive API

## Motivation and Context

JIRA: PP-2221

## How Has This Been Tested?

Running tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/api/circulation.py
+++ b/src/palace/manager/api/circulation.py
@@ -636,19 +636,15 @@ class BaseCirculationAPI(
     def collection(self) -> Collection | None:
         return Collection.by_id(self._db, id=self.collection_id)
 
-    @classmethod
     def default_notification_email_address(
-        self, library_or_patron: Library | Patron, pin: str | None
-    ) -> str:
+        self, patron: Patron, pin: str | None
+    ) -> str | None:
         """What email address should be used to notify this library's
         patrons of changes?
 
-        :param library_or_patron: A Library or a Patron.
+        :param patron: a Patron.
         """
-        if isinstance(library_or_patron, Patron):
-            library = library_or_patron.library
-        else:
-            library = library_or_patron
+        library = patron.library
         return library.settings.default_notification_email_address
 
     def integration_configuration(self) -> IntegrationConfiguration:
--- a/src/palace/manager/api/overdrive.py
+++ b/src/palace/manager/api/overdrive.py
@@ -493,7 +493,8 @@ class OverdriveAPI(
         As a side effect, this will verify that the Overdrive
         credentials are working.
         """
-        if not self._collection_token:
+        collection_token = self._collection_token
+        if not collection_token:
             library = self.get_library()
             error = library.get("errorCode")
             if error:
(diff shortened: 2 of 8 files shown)


src/palace/manager/api/circulation.py
```diff
@@ -636,19 +636,15 @@ def __init__(self, _db: Session, collection: Collection):
     def collection(self) -> Collection | None:
         return Collection.by_id(self._db, id=self.collection_id)
 
-    @classmethod
     def default_notification_email_address(
-        self, library_or_patron: Library | Patron, pin: str | None
-    ) -> str:
+        self, patron: Patron, pin: str | None
```
jonathangreen, line 640: There was a test to make sure that this could work with a library, but all the places its actually called pass a `patron` and the overdrive implementation only allows a `patron`. So I updated the hint on the base class and removed the test where its called with a library.

src/palace/manager/api/overdrive.py
```diff
@@ -740,28 +756,15 @@ def recently_changed_ids(self, start, cutoff):
...
             self.METADATA_ENDPOINT,
             collection_token=self.collection_token,
             item_id=identifier.identifier,
         )
-        status_code, headers, content = self.get(url, {})
-        if isinstance(content, (bytes, str)):
-            content = json.loads(content)
-        return content
-
-    def metadata_lookup_obj(self, identifier):
```
jonathangreen, line 755: `metadata_lookup_obj` was unused, so I removed it

src/palace/manager/api/overdrive.py
```diff
@@ -1674,18 +1744,20 @@ def release_hold(self, patron, pin, licensepool):
...
+            return
         if not response.content:
             raise CannotReleaseHold()
         data = response.json()
         if not "errorCode" in data:
             raise CannotReleaseHold()
         if data["errorCode"] == "PatronDoesntHaveTitleOnHold":
             # There was never a hold to begin with, so we're fine.
-            return True
+            return
```
jonathangreen, line 1761: Base class only returned None, so I removed the bool return values here that were never used anyway.

src/palace/manager/api/overdrive.py
```diff
@@ -2155,18 +2268,10 @@ def process_item(self, identifier):
             break
 
 
-class OverdriveData:
-    overdrive_client_key: str
-    overdrive_client_secret: str
-    overdrive_website_id: str
-    overdrive_server_nickname: str = OverdriveConstants.PRODUCTION_SERVERS
-    max_retry_count: int = 0
```
jonathangreen, line 2163: This was unused.

tests/manager/core/test_metadata.py
```diff
@@ -398,6 +398,12 @@ def test_coverage_record(self, db: DatabaseTransactionFixture):
 
 
 class TestContributorData:
+    def test__init__(self):
+        assert ContributorData().roles == [Contributor.Role.AUTHOR]
+        assert ContributorData(roles="foo").roles == ["foo"]
+        assert ContributorData(roles=("x", "y")).roles == ["x", "y"]
+        assert ContributorData(roles=["x", "y"]).roles == ["x", "y"]
```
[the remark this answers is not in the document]
jonathangreen: Second last assert, roles is a tuple, and the last one roles is a list. I'll throw in a comment, since its easy to miss.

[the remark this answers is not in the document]
tdilauro: So it is! Well, all good, then.

PR #2333, opened 2025-03-11
[PP-2233] Fix deadlocks by ensuring only one book is updated per tran…

dbernstein:
…saction.

## Description

The PR attempts to solve the problem of deadlocks occurring between the same type of tasks.  The deadlocks appear to occur because  some CMs have many Axis collections which are pulling the same data.  As a result they were frequently updating the same batches of IDs at the same time.  Since the simultaneous updates can touch the same database objects - such as editions, this can result in a deadlock in the following way:

1.  task 1 opens transaction
2. task 2 opens transaction
3. task 1 updates editions A, B, C
4. task 2 updates editions A, B, C
5. task  1  tries to commit but fails due to dead lock

## Motivation and Context

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/axis.py
+++ b/src/palace/manager/celery/tasks/axis.py
@@ -380,26 +380,25 @@ def reap_collection(
             .all()
         )
 
-    with task.transaction() as session:
-        collection = Collection.by_id(session, collection_id)
-        api = create_api(session=session, collection=collection)
-        identifier_count = len(identifiers)
-        for identifier in identifiers:
+    for identifier in identifiers:
+        with task.transaction() as session:
+            collection = Collection.by_id(session, collection_id)
+            api = create_api(session=session, collection=collection)
             api.update_licensepools_for_identifiers(identifiers=[identifier])
 
     task.log.info(
-        f'Reaper updated {identifier_count} books in collection (name="{collection_name}", id={collection_id}.'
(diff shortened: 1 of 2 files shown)


src/palace/manager/celery/tasks/axis.py
```diff
@@ -216,16 +218,21 @@ def log_run_end_message() -> None:
             return
 
         api = create_api(session=session, collection=collection)
-        start_seconds = time.perf_counter()
-        total_imported_in_current_task = 0
-        while len(identifiers) > 0:
-            batch = identifiers[:batch_size]
 
+        identifier_batch = identifiers[:batch_size]
+        circ_data = [
```
dbernstein, line 223: Get the book metadata and circulation  info in one db session.

src/palace/manager/celery/tasks/axis.py
```diff
@@ -216,16 +218,21 @@ def log_run_end_message() -> None:
...
+            (metadata, circulation)
+            for metadata, circulation in api.availability_by_title_ids(
+                title_ids=identifier_batch
+            )
+        ]
+
+    for metadata, circulation in circ_data:
```
dbernstein, line 230: Then update each metadata/circulation data pair in a separate transaction

jonathangreen (approved):
Looks good!

PR #2335, opened 2025-03-12
Break overdrive.py file up into palace.manager.api.overdrive package (PP-2221)

jonathangreen:
## Description

The `api/overdrive.py` file has gotten very large, coming in at ~3000 lines currently. This PR breaks it out into a number of files in `palace.manager.api.overdrive` with as minimal a changes as possible. 

`api/overdrive/api.py` is still large, coming in at ~1800 lines now. So we will probably want to do some work to further break it up, but this PR is a good start.

## Motivation and Context

Make it a bit easier to work on the Overdrive intergration.

## How Has This Been Tested?

- Running tests in CI

## Checklist


src/palace/manager/api/overdrive/api.py
```diff
@@ -386,7 +231,7 @@ def child_settings_class(cls) -> type[OverdriveChildSettings]:
 
     @classmethod
     def label(cls) -> str:
-        return DataSourceConstants.OVERDRIVE
+        return OVERDRIVE_LABEL
```
jonathangreen, line 234: This had to get moved into the `OVERDRIVE_LABEL` constant outside of the class in order to break some import cycles.

src/palace/manager/api/overdrive/api.py
```diff
@@ -772,31 +617,6 @@ def metadata_lookup(self, identifier: Identifier) -> dict[str, Any]:
         status_code, headers, content = self.get(url)
         return json.loads(content)  # type: ignore[no-any-return]
 
-    @classmethod
```
jonathangreen, line 775: `make_link_safe` had to get moved to a function outside of `OverdriveAPI` in order to break an import cycle.

src/palace/manager/api/overdrive/constants.py
```diff
@@ -0,0 +1,43 @@
...
+    # the metadata wrangler) don't need to set this value.
+    ILS_NAME_KEY = "ils_name"
+    ILS_NAME_DEFAULT = "default"
+
+
+# An OverDrive defined constant indicating the "main" or parent account
+# associated with an OverDrive collection.
+OVERDRIVE_MAIN_ACCOUNT_ID = -1
+
+OVERDRIVE_LABEL = DataSourceConstants.OVERDRIVE
```
jonathangreen, line 43: `OVERDRIVE_LABEL` and `OVERDRIVE_MAIN_ACCOUNT_ID` were both moved here to break import cycles.

src/palace/manager/api/overdrive/constants.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+
+from palace.manager.sqlalchemy.constants import DataSourceConstants
+
+
+class OverdriveConstants:
```
jonathangreen, line 6: I think it would make sense to move all these constants outside of this class at some point, but I was trying to only make the minimal necessary changes in this PR, so I left them in here for now.

src/palace/manager/api/overdrive/coverage.py
```diff
@@ -0,0 +1,79 @@
...
+
+    SERVICE_NAME = "Overdrive Bibliographic Coverage Provider"
+    DATA_SOURCE_NAME = DataSource.OVERDRIVE
+    PROTOCOL = OVERDRIVE_LABEL
+    INPUT_IDENTIFIER_TYPES = Identifier.OVERDRIVE_ID
+
+    def __init__(
+        self,
+        collection: Collection,
+        api: OverdriveAPI | None = None,
```
jonathangreen, line 35: This used to be:
```python
api_class: type[OverdriveAPI] | OverdriveAPI = OverdriveAPI,
```

I updated (and renamed the parameter) to break an import cycle with `OverdriveAPI`. This is probably the biggest functional change I made in this PR, but I didn't see a nicer way to make the imports work. 

It looks like this parameter is only used in tests and one place in `OverdriveAPI`, so I think its safe to update.

tdilauro (approved):
Looks good! 💯

PR #2337, opened 2025-03-12
[PP-2231] Catch and log integration exceptions in Axis tasks.

dbernstein:
## Description
This PR catches Integration exceptions and logs them to prevent incomplete configurations from triggering alerts due to uncaught task exceptions.
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2231

## Checklist


jonathangreen:
@dbernstein I'd rather scope this much more narrowly, so we don't mask real issues in the task. Here is my suggestion for how we handle this: https://github.com/ThePalaceProject/circulation/pull/2338

dbernstein:
Closing in favor of @jonathangreen 's approach.

PR #2341, opened 2025-03-13
Don't include non-distributor data sources in library item reports. (PP-2250)

tdilauro:
## Description

- Update queries so that key key tables are joined from the license pool table.
- Use PG LATERAL JOIN to simplify some of the joins.
- Use the `sort_title` and `sort_author`, rather that `title` and `author` columns for sorting.

## Motivation and Context

- Some Palace Manager-internal data sources were appearing in the reports.
- There was some joining over identifiers/editions that was leading to some other unexpected combinations of collections and data sources.

[Jira PP-2250]

## How Has This Been Tested?

- Some local manual testing.
- Tests pass locally.
- CI tests.

## Checklist


dbernstein (approved):
Looks good to me.  Nice use of the `LATERAL` feature.  I've not seen that before.

PR #2342, opened 2025-03-14
Fix OPDS2 empty publication list validation (PP-2230)

jonathangreen:
## Description

On the last page of a feed sometimes we receive a feed that has no publications in it. This is causing a validation failure currently.

## Motivation and Context

In some cases we get an empty list for publications on the last page of the feed. We are failing validation in this case, but an empty list is allowed by the spec. I don't remember why I added this validation, perhaps I was just being overzealous in my validation. 

## How Has This Been Tested?

- Added test case
- Ran tests in CI

## Checklist


tdilauro (approved):
Here's to being more accepting! 🥂 

But not too accepting. 💣

PR #2343, opened 2025-03-14
Fix detached object error in axis reaper task.

dbernstein:
## Description
A follow on fix for PP-2233.  The identifier list, having been retrieved in one session, needed to be reattached (via the sqlalchemy session.merge method) in downstream session where the update operations occur.

## Motivation and Context

## Checklist


tdilauro (approved):
👍🏽

PR #2345, opened 2025-03-14
Add some type hints to the licensing model (PP-2218)

jonathangreen:
## Description

Add some type hints to the licensing model. This required a few changes to the hints in the metadata layer.

## Motivation and Context

Broken out from the work on PP-2218.

## How Has This Been Tested?

- Running mypy
- Running tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/core/coverage.py
+++ b/src/palace/manager/core/coverage.py
@@ -1334,7 +1334,10 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
         """
         error = None
         if circulationdata:
-            primary_identifier = circulationdata.primary_identifier(self._db)
+            try:
+                primary_identifier = circulationdata.primary_identifier(self._db)
+            except ValueError:
+                primary_identifier = None
             if identifier != primary_identifier:
                 error = "Identifier did not match CirculationData's primary identifier."
         else:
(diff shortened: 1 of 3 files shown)


src/palace/manager/core/metadata_layer.py
```diff
@@ -876,25 +876,22 @@ def __repr__(self) -> str:
...
 
-    def primary_identifier(self, _db: Session) -> Identifier | None:
+    def primary_identifier(self, _db: Session) -> Identifier:
         """Find the Identifier associated with this circulation information."""
         if not self.primary_identifier_obj:
             if self._primary_identifier:
                 obj, ignore = self._primary_identifier.load(_db)
             else:
-                obj = None
+                raise ValueError("No primary identifier provided!")
```
jonathangreen, line 894: In the calls `metadata_layer` makes to licensing, we are assuming these are not `None`. Since they don't seem to be `None` in our actual code, I modified the type hints and raise an `Exception` if they are None.

src/palace/manager/core/coverage.py
```diff
@@ -1334,7 +1334,10 @@ def _set_circulationdata(self, identifier, circulationdata):
         """
         error = None
         if circulationdata:
-            primary_identifier = circulationdata.primary_identifier(self._db)
+            try:
```
tdilauro, line 1337: Do you want to keep this exception in here long-term? Or is it just to make sure that it's behaving as you think it is in the short term? If the latter, would be handy to put a "TODO:" in here to note that.

tdilauro (approved):
Looks good! 🥇

One minor comment below.

PR #2348, opened 2025-03-18
Fix borrow method type hint (PP-2218)

jonathangreen:
## Description

Remove a type ignore comment from our circulation API implementation, and update the type hints to match in all the classes inheriting from `BaseCirculationApi`.

## Motivation and Context

Ran into this incorrect type hint when working on the Overdrive API, I had assumed it was set because of the type hint, but it was `None`.

## How Has This Been Tested?

- Running mypy

## Checklist


tdilauro (approved):
☑️

PR #2352, opened 2025-03-19
[PP-1391] remove title from patron shelf if removed from catalog

dbernstein:
## Description
This PR adds new scheduled and unscheduled tasks for reaping loans and holds meeting the criteria laid out in  PP-1391.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-1391

## How Has This Been Tested?
Unit tests cover the functionality but this will need to be tested more extensively in minotaur before being released.

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/integration_settings.py
+++ b/src/palace/manager/api/admin/controller/integration_settings.py
@@ -398,7 +398,7 @@ class IntegrationSettingsController(ABC, Generic[T], LoggerMixin):
         service: IntegrationConfiguration,
         libraries_data: str,
         settings_class: type[BaseSettings],
-    ) -> ChangedLibrariesTuple:
+    ) -> None:
         """
         Process the library settings for a service. This will create new
         IntegrationLibraryConfigurations for any libraries that don't have one,
--- a/src/palace/manager/api/admin/controller/integration_settings.py
+++ b/src/palace/manager/api/admin/controller/integration_settings.py
@@ -20,10 +20,6 @@ from palace.manager.api.admin.problem_details import (
     NO_SUCH_LIBRARY,
     UNKNOWN_PROTOCOL,
 )
-from palace.manager.celery.tasks.reaper import (
-    reap_unassociated_holds,
-    reap_unassociated_loans,
-)
 from palace.manager.core.problem_details import INTERNAL_SERVER_ERROR, INVALID_INPUT
 from palace.manager.core.selftest import HasSelfTests
 from palace.manager.integration.base import (
@@ -385,12 +381,6 @@ class IntegrationSettingsController(ABC, Generic[T], LoggerMixin):
         for library_integration in removed:
             self._db.delete(library_integration)
 
-        if removed:
-            # ensure that all loans and holds related
-            # with the deleted library integrations are purged.
-            reap_unassociated_loans.delay()
-            reap_unassociated_holds.delay()
-
     def process_updated_libraries(
         self,
         libraries: list[UpdatedLibrarySettingsTuple],
@@ -408,7 +398,7 @@ class IntegrationSettingsController(ABC, Generic[T], LoggerMixin):
         service: IntegrationConfiguration,
         libraries_data: str,
         settings_class: type[BaseSettings],
-    ) -> None:
+    ) -> ChangedLibrariesTuple:
         """
         Process the library settings for a service. This will create new
         IntegrationLibraryConfigurations for any libraries that don't have one,
@@ -419,6 +409,8 @@ class IntegrationSettingsController(ABC, Generic[T], LoggerMixin):
         self.process_deleted_libraries(removed)
         self.process_updated_libraries(new, settings_class)
         self.process_updated_libraries(updated, settings_class)
+
+        return ChangedLibrariesTuple(new=new, updated=updated, removed=removed)
 
     def delete_service(self, service_id: int) -> Response:
         """
--- a/src/palace/manager/api/admin/controller/integration_settings.py
+++ b/src/palace/manager/api/admin/controller/integration_settings.py
@@ -410,8 +410,6 @@ class IntegrationSettingsController(ABC, Generic[T], LoggerMixin):
         self.process_updated_libraries(new, settings_class)
         self.process_updated_libraries(updated, settings_class)
 
-        return ChangedLibrariesTuple(new=new, updated=updated, removed=removed)
-
     def delete_service(self, service_id: int) -> Response:
         """
         Delete a service.
--- a/src/palace/manager/api/admin/controller/integration_settings.py
+++ b/src/palace/manager/api/admin/controller/integration_settings.py
@@ -20,6 +20,10 @@ from palace.manager.api.admin.problem_details import (
     NO_SUCH_LIBRARY,
     UNKNOWN_PROTOCOL,
 )
+from palace.manager.celery.tasks.reaper import (
+    reap_unassociated_holds,
+    reap_unassociated_loans,
+)
 from palace.manager.core.problem_details import INTERNAL_SERVER_ERROR, INVALID_INPUT
 from palace.manager.core.selftest import HasSelfTests
 from palace.manager.integration.base import (
@@ -380,6 +384,12 @@ class IntegrationSettingsController(ABC, Generic[T], LoggerMixin):
         """
         for library_integration in removed:
             self._db.delete(library_integration)
+
+        if removed:
+            # ensure that all loans and holds related
+            # with the deleted library integrations are purged.
+            reap_unassociated_loans.delay()
+            reap_unassociated_holds.delay()
 
     def process_updated_libraries(
         self,
--- a/src/palace/manager/api/circulation.py
+++ b/src/palace/manager/api/circulation.py
@@ -57,11 +57,8 @@ from palace.manager.sqlalchemy.model.licensing import (
     License,
     LicensePool,
     LicensePoolDeliveryMechanism,
-    LicenseStatus,
-    RightsStatus,
 )
-from palace.manager.sqlalchemy.model.patron import Hold, Loan, LoanAndHoldMixin, Patron
-from palace.manager.sqlalchemy.model.resource import Resource
+from palace.manager.sqlalchemy.model.patron import Hold, Loan, Patron
 from palace.manager.sqlalchemy.util import get_one
 from palace.manager.util.datetime_helpers import utc_now
 from palace.manager.util.http import HTTP, BadResponseException, ResponseCodesT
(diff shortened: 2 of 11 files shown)


src/palace/manager/api/circulation.py
```diff
@@ -869,6 +869,62 @@ def sync_patron_activity(self, patron: Patron, pin: str | None) -> None:
...
+            if not collection.is_active:
+                log_message += (
+                    f'\n  * the collection("{collection.name}") is not currently active'
+                )
+                remove = True
+
+            # the associated license pool is (unlimited or open access) AND unfulfillable
+            if (
+                license_pool.open_access or license_pool.unlimited_access
+            ) and not license_pool.has_fulfillable_delivery_mechanism:
```
dbernstein, line 918: I wonder if it is necessary for us to check open_access and unlimited_access here.   If the title is neither open_access nor unlimited_access but it has no fulfillable delivery mechanism, shouldn't it be removed?  Or are fulfillment mechanisms updated with the availability update?

src/palace/manager/api/circulation.py
```diff
@@ -869,6 +869,62 @@ def sync_patron_activity(self, patron: Patron, pin: str | None) -> None:
 
         self.sync_loans(patron, remote_loans, local_loans)
         self.sync_holds(patron, remote_holds, local_holds)
+        self.delete_loans_and_holds_no_longer_available(patron)
+
+    def delete_loans_and_holds_no_longer_available(self, patron: Patron) -> None:
+        """
+        Delete the loans an holds associated with the patron that are no longer available to the patron
```
jonathangreen, line 876: Minor: Typo in comment (missing d)
> Delete the loans an***d*** holds...

src/palace/manager/api/circulation.py
jonathangreen: I think this check should come out. From my read of the ticket, this isn't necessary. I'm don't think `has_fulfillable_delivery_mechanism` is reliable (for example it says LCP content isn't deliverable), and if you look at where its set in the code, its not used for anything right now other then an informational script.

I've been wondering if that column in the database should just come out entirely.

src/palace/manager/api/circulation.py
```diff
@@ -869,6 +869,62 @@ def sync_patron_activity(self, patron: Patron, pin: str | None) -> None:
 
         self.sync_loans(patron, remote_loans, local_loans)
         self.sync_holds(patron, remote_holds, local_holds)
+        self.delete_loans_and_holds_no_longer_available(patron)
```
jonathangreen, line 872: Minor: This is fine, and it can go in as is, but I think it would be a better user experience, if this ran as a scheduled reaper task, rather then tacking it on to sync patron activity.

Patron activity sync is triggered, _after_ a patron requests their loans feed, since its slow to pull information from every distributor, and we assume our system is mostly up to date. So the way this is implemented, patrons will still see their outdated loans / holds when they first look at them, but upon refresh they will be gone. 

Since we have the information for all this locally, instead of waiting for the patron to request the feed, we could occasionally be doing this in the background.

src/palace/manager/api/circulation.py
jonathangreen: The cases from the ticket you are trying to accomplish here I believe it this one:
> An item has been removed from a feed
> An item has been marked as unavailable (using the availability collection in the feed entry)

I don't think this will handle those cases. 

For an item being marked as unavailable, here is what we do during import:
https://github.com/ThePalaceProject/circulation/blob/49da9ff0b3d4ef7053bf3b9bc9160d03b47af329/src/palace/manager/api/odl/importer.py#L178-L198

For an item being removed from the feed, here is what we do during import:
https://github.com/ThePalaceProject/circulation/blob/71ef9d48feb8d7cd0c86000ad3ae288dd0755531/bin/opds2_reaper_monitor#L167-L176

I don't think this will handle either of those cases.

src/palace/manager/api/circulation.py
dbernstein: @jonathangreen :  a couple of questions here
1) Frequency for the the periodic background update?  I'm imagining that every 12 hours would probably be more sufficiently frequent.  What do you think? 
2) If I add a scheduled background task, we want to move `delete_loans_and_holds_no_longer_available` rather than invoke it both places (ie in the background task as well as the patron sync)

dbernstein:
@jonathangreen :  I am thinking about reworking the implementation as a set of  tasks which I'll put with the other reaper tasks:   one task for each reason for deletion.  Instead of iterating through a list of items and checking each for the deletion conditions, I'll just execute a single delete query.  I think that will be way more efficient.

src/palace/manager/api/circulation.py
jonathangreen: Yes, I think removing it from patron sync makes sense.

jonathangreen:
>  I am thinking about reworking the implementation as a set of tasks which I'll put with the other reaper tasks: one task for each reason for deletion. Instead of iterating through a list of items and checking each for the deletion conditions, I'll just execute a single delete query. I think that will be way more efficient.

That seems fine to me. The one for *A collection has been deleted* may be better off as part of the `collection.delete()` that gets called by the collection delete task, rather then as a distinct task, since we already have a task that does cleanup when a collection gets deleted.

dbernstein:
@jonathangreen :  I've got the changes mostly done here. But  there is one question in my mind:  The sync patron activity routine will create new loans and holds if they don't exist.  So if the reaper tasks remove a loan or hold for whatever reason and the loan or hold still exists in the remote system, the loan will reappear if we're not removing it immediately after the synchronization.   I'm not even 100% sure if that scenario is possible but I thought I would raise it just in case.

src/palace/manager/api/admin/controller/integration_settings.py
```diff
@@ -381,6 +385,12 @@ def process_deleted_libraries(
         for library_integration in removed:
             self._db.delete(library_integration)
 
+        if removed:
+            # ensure that all loans and holds related
```
dbernstein, line 389: @jonathangreen : These task calls are breaking the tests.  Two questions here:  1) Do you agree that these should go here.  2) If so,  is there something special I need to do to patch these methods when testing in the flask context?  For example: in [this collection settings controller test](https://github.com/ThePalaceProject/circulation/blob/main/tests/manager/api/admin/controller/test_collections.py#L696) I tread using a `with patch("palace.manager.celery.tasks.reaper.reap_unassociated_loans") as reap_loans:` context manager, but when I actually debug the test I notice that the reap_unassociated_loans function is not being mocked.

dbernstein:
@jonathangreen :  this one is ready for re-review.  There are 4 broken tests  due to invocations of async reaper tasks in the collection settings controller.   I'm not sure if I need to be invoking the those tasks there.  If so I'm running into a problem monkey patching those functions.  See my comment for details.

src/palace/manager/sqlalchemy/model/licensing.py
```diff
@@ -511,7 +511,13 @@ def with_no_delivery_mechanisms(cls, _db):
     @property
     def deliverable(self):
         """This LicensePool can actually be delivered to patrons."""
-        return (self.open_access or self.licenses_owned > 0) and any(
+        return (
+            self.open_access or self.unlimited_access or self.licenses_owned > 0
+        ) and self.has_fulfillable_delivery_mechanism
+
+    @property
+    def has_fulfillable_delivery_mechanism(self) -> bool:
```
dbernstein, line 519: This method can probably be reverted since it is no longer doing anything.

src/palace/manager/api/admin/controller/integration_settings.py
jonathangreen: 1. This shouldn't go here. This class is a super-class for a lot of our settings controllers. It is not just for the collection settings. This method is called by all of them. That would mean we would trigger these celery tasks anytime a library is removed from any integration. I don't think we need to do that for all the different integration types like patron auth, or metadata, marc exporter, etc. You could either remove it entirely here, or target it at the specific controller you want.
2. The documentation for `unittest.mock` has a good section on where to patch https://docs.python.org/3/library/unittest.mock.html#where-to-patch that should help you.

src/palace/manager/sqlalchemy/model/licensing.py
jonathangreen: Yes. I'd revert this 👍🏻 

I think these should just be removed entirely, but lets do that in a separate PR.

src/palace/manager/api/admin/controller/integration_settings.py
dbernstein: Thanks for that reference.  I will move it to the right controller.

src/palace/manager/celery/tasks/reaper.py
```diff
@@ -237,3 +242,140 @@ def loan_reaper(task: Task) -> None:
...
+        f"because the associated collection is inactive."
+    )
+
+
+def reap_loans_or_holds_with_unavailable_license_pools(
+    task: Task, deletion_class: type[Loan | Hold]
+) -> None:
+    """
+    Delete the loans or holds where the associated licensepool has no licenses with status == "available" or
+    or the license_pool.licenses_owned == 0 and license_pool.licenses_available == 0
```
jonathangreen, line 354: I think you should break the "unavailable" reaper out into a separate ticket, because I think this needs more work and discussion before it goes in. I actually don't think we even keep enough information to know properly what is unavailable and what isn't, so we need to track that information before its possible to write a reaper. 

Turning the query here into SQL I get: 
```sql
SELECT loans.id
FROM loans JOIN licensepools ON licensepools.id = loans.license_pool_id 
WHERE licensepools.licenses_owned = 0 AND licensepools.licenses_available = 0 OR (licensepools.id NOT IN (SELECT available_license_pools.anon_1 
FROM (SELECT DISTINCT licenses.license_pool_id AS anon_1 
FROM licenses 
WHERE licenses.status = 'available') AS available_license_pools))
```

If I run that as a count query against one of our production CM:

```
=> SELECT count(*)
FROM loans JOIN licensepools ON licensepools.id = loans.license_pool_id
WHERE licensepools.licenses_owned = 0 AND licensepools.licenses_available = 0 OR (licensepools.id NOT IN (SELECT available_license_pools.anon_1
FROM (SELECT DISTINCT licenses.license_pool_id AS anon_1
FROM licenses
WHERE licenses.status = 'available') AS available_license_pools));
 count
-------
 12789
(1 row)

=> select count(*) from loans;
 count
-------
 23203
(1 row)
```

Doing the same exercise for holds: 
```
=> SELECT count(*)
FROM holds JOIN licensepools ON licensepools.id = holds.license_pool_id
WHERE licensepools.licenses_owned = 0 AND licensepools.licenses_available = 0 OR (licensepools.id NOT IN (SELECT available_license_pools.anon_1
FROM (SELECT DISTINCT licenses.license_pool_id AS anon_1
FROM licenses
WHERE licenses.status = 'available') AS available_license_pools));
 count
-------
 22147
(1 row)

=> select count(*) from holds;
 count
-------
 27632
(1 row)
```

So if this script ran it would delete 12789/23203 = 55% of all the outstanding loans and 22147/27632 = 80% of the outstanding holds. Those numbers seem wildly off, and this will very directly impact patrons, so we need to be really sure about these queries before they go in.

jonathangreen (changes requested):
This needs 2 changes before it can go in: 
- The unavailable license pools reaper needs to be re-thought or removed. 
- Need to test the other four reapers with some real data to make sure they aren't going to delete loans and holds that we are not expecting.

There are a couple minor comments there for consideration as well, but they aren't blockers to getting this merged.

jonathangreen (approved):
This looks good. I think it would be good to check the queries to make sure the data that comes back seems reasonable against a production DB, but reading the code this all looks good to me.

dbernstein:
Okay - confirmed:  I check against Georgia, three of the four deletion queries are currently removing no loans or holds, while one will remove a single loan.

PR #2358, opened 2025-03-20
Always use palace context for overdrive patron tokens (PP-2218)

jonathangreen:
## Description

Always use the Palace context when requesting patron credentials for the Overdrive integration.

## Motivation and Context

We already use the palace context almost everywhere that we perform actions on a patrons behalf. In the work on PP-2218, I needed to switch a number of other calls to `patron_request` to use `palace_context`. 

After those changes, the only calls left not using `palace_context` were:
 - `checkin`
 - `release_hold`
 - `get_patron_information`

All the other calls need the additional information returned when using the `palace_context` credentials. I've tested and all the calls left work with the `palace_context` credentials. It doesn't seem worth it to keep a second token around just for those three calls, so I was thinking about switching all the calls to use the `palace_context` credentials, as I've done in this PR.

## How Has This Been Tested?

- Tested with real API calls
- Tested running tests in CI

## Checklist

### DIFF
--- a/tests/manager/api/overdrive/test_api.py
+++ b/tests/manager/api/overdrive/test_api.py
@@ -1369,7 +1369,6 @@ class TestOverdriveAPI:
         # We will get the loan, try to lock in the format, and fail.
         overdrive_api_fixture.queue_access_token_response()
         http.queue_response(200, content=loan)
-        overdrive_api_fixture.queue_access_token_response()
         http.queue_response(400, content=lock_in_format_not_available)
 
         # Trying to get a fulfillment link raises an exception.
@@ -1395,7 +1394,6 @@ class TestOverdriveAPI:
         # format, fail, and then update the bibliographic information.
         overdrive_api_fixture.queue_access_token_response()
         http.queue_response(200, content=loan)
-        overdrive_api_fixture.queue_access_token_response()
         http.queue_response(400, content=lock_in_format_not_available)
         http.queue_response(200, content=bibliographic)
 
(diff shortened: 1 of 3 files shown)


jonathangreen:
@tdilauro tagging you in this one to take a look, since your most familiar with these tokens.

jonathangreen:
The reason we need to use the palace context for most of these calls is probably the same reason as this PR https://github.com/ThePalaceProject/circulation/pull/358. We get more format information in the responses when using our context for these calls.

tests/manager/api/overdrive/test_api.py
```diff
@@ -2028,45 +2028,6 @@ def test__refresh_patron_oauth_token_failure(
                 credential, patron, "a pin"
             )
 
-    def test__refresh_patron_oauth_token_palace_context(
-        self, overdrive_api_fixture: OverdriveAPIFixture
-    ):
```
tdilauro, line 2033: Is this already being tested elsewhere? I would've expected to see a test related to the collection context being removed.

jonathangreen: >  I would've expected to see a test related to the collection context being removed.
@tdilauro I'm not sure I understand 

This test, I think, covers the case you are talking about. If not can you expand on your question a bit?
https://github.com/ThePalaceProject/circulation/blob/e644bbbb82cbdb1ff0994e4bb3c756c1060b4a13/tests/manager/api/overdrive/test_api.py#L1957-L1995

tdilauro: Yes, that exactly answers my question. Thanks!

tdilauro (approved):
🔥

jonathangreen:
Since these changes overlap with https://github.com/ThePalaceProject/circulation/pull/2363 and this PR is much smaller. I'm going to merge that one first, then rebase this, cleanup the conflicts, then merge this one second.

PR #2360, opened 2025-03-21
Inventory reports exclude inactive collections. (PP-2237)

tdilauro:
## Description

Excludes otherwise eligible collections that are not active (e.g., subscription expired) from inclusion in the inventory reports.

## Motivation and Context

Books in inactive collections are not available for lending, so reporting them as available would be misleading.

[Jira PP-2237]

## How Has This Been Tested?

- Tests pass locally.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/13982806834/job/39151248655) pass.

NB: CI tests pass except for one Google Drive integration test with a known problem:
- `tests/manager/service/google_drive/test_google_drive.py::TestGoogleDriveService::test_share_document`

## Checklist


jonathangreen (approved):
Looks good!

PR #2364, opened 2025-03-21
Stagger the execution of the collection import tasks to minimize dead…

dbernstein:
…locks.

## Description
This update introduces a 5 second delay between the execution of each axis collection import task in order to reduce the probability of a simultaneous update of a particular book on CMs where there are multiple axis collections with overlapping contents.

## Motivation and Context

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/axis.py
+++ b/src/palace/manager/celery/tasks/axis.py
@@ -1,5 +1,5 @@
 import time
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from celery import shared_task
 from sqlalchemy import select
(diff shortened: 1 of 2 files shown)


src/palace/manager/celery/tasks/axis.py
```diff
@@ -49,6 +49,13 @@ def import_all_collections(
             )
             list_identifiers_for_import.apply_async(
                 kwargs={"collection_id": collection.id},
+                eta=utc_now()
+                + timedelta(
+                    seconds=count * 5
+                ),  # stagger the execution of the collection import
+                # in order to minimize chance of deadlocks caused by
+                # simultaneous updates to the metadata when CM has multiple
+                # axis collections configured with overlapping content
```
jonathangreen, line 58: Minor: Instead of having to work with a `timedelta` here, you could just use the [`countdown` parameter](https://docs.celeryq.dev/en/stable/userguide/calling.html#eta-and-countdown).
```suggestion
                delay=count * 5,  # stagger the execution of the collection import
                # in order to minimize chance of deadlocks caused by
                # simultaneous updates to the metadata when CM has multiple
                # axis collections configured with overlapping content
```

jonathangreen (approved):
🚀

src/palace/manager/celery/tasks/axis.py
dbernstein: I'll give it a shot - but it looks like we should be using `countdown` not `delay` unless I'm missing something.

dbernstein:
@jonathangreen : I updated this one to use countdown instead of ETA, employed a similar changed for the reap_all_collections task and updated the tests to reflect these changes in parameters.

PR #2365, opened 2025-03-21
Remove share/unshare methods from the google drive service since we a…

dbernstein:
…re not using them.

## Motivation and Context
We are not using these features and they are encountering rate limit issues with the google service api.

## Checklist


jonathangreen (approved):
🚀

PR #2366, opened 2025-03-22
Type hints for HTTP request kwargs

jonathangreen:
## Description

  This PR adds proper type hints to HTTP-related keyword arguments across the application. It introduces typed dictionaries for HTTP request keyword arguments, ensuring type safety when making HTTP requests.

  The changes include:

  - Added `GetRequestKwargs` and `RequestKwargs` typed dictionaries in http.py
  - Updated method signatures to use `Unpack[RequestKwargs]` for proper typing of kwargs
  - Created protocol classes for request callables to improve type checking
  - Updated tests to maintain compatibility with the new typing

## Motivation and Context

  This change improves type safety by properly typing HTTP request keyword arguments. Previously, these were untyped, which could lead to subtle bugs where incorrect parameter types were passed to HTTP requests. With these changes:

  - Static type checkers can catch errors at development time
  - IDE autocompletion provides better suggestions for available parameters
  - Code becomes more self-documenting through explicit parameter types

I am always forgetting what the param names for the http methods are, especially the palace specific ones, so the IDE autocomplete is really helpful here.

## How Has This Been Tested?

  The changes have been tested through:

  - Running the existing test suite to ensure all tests still pass
  - Manual verification of type checking using mypy
  - Ensuring IDE autocompletion correctly identifies parameter types

## Checklist

### DIFF
--- a/src/palace/manager/util/http.py
+++ b/src/palace/manager/util/http.py
@@ -2,9 +2,10 @@ from __future__ import annotations
 
 import logging
 import time
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, Iterable, Mapping, Sequence
+from io import BytesIO, StringIO
 from json import JSONDecodeError
-from typing import Any
+from typing import Any, Literal, Protocol, TypedDict
 from urllib.parse import urlparse
 
 import requests
@@ -12,7 +13,7 @@ from flask_babel import lazy_gettext as _
 from requests import PreparedRequest, sessions
 from requests.adapters import HTTPAdapter, Response
 from requests.auth import AuthBase
-from typing_extensions import Self
+from typing_extensions import Self, Unpack
 from urllib3 import Retry
 
 from palace import manager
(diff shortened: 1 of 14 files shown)


src/palace/manager/util/http.py
```diff
@@ -162,35 +216,37 @@ def set_quick_failure_settings(cls) -> None:
...
         """Make a GET request with timeout handling."""
-        return cls.request_with_timeout("GET", url, *args, **kwargs)
+        return cls.request_with_timeout("GET", url, **kwargs)
 
     @classmethod
     def post_with_timeout(
-        cls, url: str, payload: str | Mapping[str, Any], *args: Any, **kwargs: Any
+        cls,
+        url: str,
+        **kwargs: Unpack[RequestKwargs],
```
jonathangreen, line 227: Functional change here, I dropped the `payload` arg / kwarg, and instead use the `data` kwarg. This brings our implementation into line with requests. I followed this change through everywhere we call this method.

src/palace/manager/util/http.py
```diff
@@ -162,35 +216,37 @@ def set_quick_failure_settings(cls) -> None:
...
-        return cls.request_with_timeout("POST", url, *args, **kwargs)
+        return cls.request_with_timeout("POST", url, **kwargs)
 
     @classmethod
     def put_with_timeout(
-        cls, url: str, payload: str | Mapping[str, Any], *args: Any, **kwargs: Any
+        cls,
+        url: str,
+        **kwargs: Unpack[RequestKwargs],
     ) -> Response:
```
jonathangreen, line 237: Same as above, dropped `payload` and updated callers.

src/palace/manager/util/http.py
```diff
@@ -199,10 +255,11 @@ def request_with_timeout(
...
     def _request_with_timeout(
         cls,
+        http_method: str,
         url: str,
         make_request_with: Callable[..., Response],
-        *args: Any,
-        **kwargs: Any,
+        process_response_with: _ProcessResponseCallable | None = None,
+        **kwargs: Unpack[RequestKwargs],
     ) -> Response:
```
jonathangreen, line 263: This doesn't get called directly much, its is mostly used within this class, but I changed around the parameter order to be consistent with the other methods. This required some changes in tests.

src/palace/manager/util/http.py
```diff
@@ -230,39 +286,24 @@ def _request_with_timeout(
         )
         backoff_factor: float = float(kwargs.pop("backoff_factor", 1.0))
 
-        # Unicode data can't be sent over the wire. Convert it to UTF-8.
-        if "data" in kwargs and isinstance(kwargs["data"], str):
-            kwdata: str = kwargs["data"]
-            kwargs["data"] = kwdata.encode("utf8")
-
```
jonathangreen, line 237: This may have been true at one point in python 2, but its not in python 3. Requests is happy to accept `str`.

src/palace/manager/util/http.py
```diff
@@ -230,39 +286,24 @@ def _request_with_timeout(
...
-        ua_header = {"User-Agent": f"Palace Manager/{version}"}
-        headers = ua_header | (kwargs.get("headers") or {})
-
-        # Make sure headers are encoded as utf-8
-        kwargs["headers"] = {
-            k.encode() if isinstance(k, str) else k: (
-                v.encode() if isinstance(v, str) else v
-            )
-            for k, v in headers.items()
-        }
```
jonathangreen, line 253: Again, requests is happy to accept headers as `str`, so no need to encode them to `bytes` ourselves.

src/palace/manager/util/http.py
```diff
@@ -230,39 +286,24 @@ def _request_with_timeout(
...
+                    f"Sending {http_method} request to {url}: kwargs {kwargs!r}",
+                    url,
+                    kwargs,
                 )
-            if len(args) == 1:
-                # requests.request takes two positional arguments,
-                # an HTTP method and a URL. In most cases, the URL
-                # gets added on here. But if you do pass in both
-                # arguments, it will still work.
-                args = args + (url,)
```
jonathangreen, line 265: This ugly `arg` mangling isn't necessary, now that the functions parameter order is changed.

jonathangreen:
The tests failing here are the unrelated google drive tests that are failing elsewhere due to rate limits, so I think this is ready to go.

tdilauro (approved):
Looks good! 🔐🚀

PR #2372, opened 2025-03-26
Address Integration and Deadlock errors in axis.import_identifiers and axis.reap_collection

dbernstein:
This is a follow on PR to the many PRs associated with PP-2233.   I think this should bring the number of errors down very close to zero. 

* Add support for sharing a bearer token among multiple axis api instances:  
    We are creating a new instance of the axis api for each transaction. Since we're updating a single book in each transaction, both in the reaper and the import_identifiers, we are also calling out to the axis token endpoint man times per task.  Multiply that by 200+ (vt and ga libraries for starters) and it's no surprise that we're getting 502s back from axis.  This change ensures that we reducing our calls to the axis api to a minimum.

* Add retry support for non-401 IntegrationException in the axis reaper and import_identifiers operations: 
   In the case when we are getting integration exceptions, we should retry a few times before failing.
* Add retry support for DeadlockDetected errors in the axis reaper and import_identifiers operations
   Deadlocks are infrequent, but should not result in task failure.  This change will hopefully remedy this issue since consecutive dealocks on retries for a  single task should never occur (or at least extremely rarely).
* Add new tests for the various retry cases.

## Description

## Motivation and Context

## Checklist

### DIFF
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -212,7 +212,7 @@ class Axis360API(
         return ""
 
     def __init__(
-        self, _db: Session, collection: Collection, bearer_token: str = None
+        self, _db: Session, collection: Collection, bearer_token: str | None = None
     ) -> None:
         super().__init__(_db, collection)
         settings = self.settings
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -211,7 +211,9 @@ class Axis360API(
     def description(cls) -> str:
         return ""
 
-    def __init__(self, _db: Session, collection: Collection) -> None:
+    def __init__(
+        self, _db: Session, collection: Collection, bearer_token: str = None
+    ) -> None:
         super().__init__(_db, collection)
         settings = self.settings
         self.library_id = settings.external_account_id
@@ -229,7 +231,7 @@ class Axis360API(
         if not self.library_id or not self.username or not self.password:
             raise CannotLoadConfiguration("Axis 360 configuration is incomplete.")
 
-        self._cached_bearer_token: str | None = None
+        self._cached_bearer_token: str | None = bearer_token
         self.verify_certificate: bool = (
             settings.verify_certificate
             if settings.verify_certificate is not None
--- a/src/palace/manager/celery/tasks/axis.py
+++ b/src/palace/manager/celery/tasks/axis.py
@@ -2,7 +2,9 @@ import time
 from datetime import datetime
 
 from celery import shared_task
+from psycopg2.errors import DeadlockDetected
 from sqlalchemy import select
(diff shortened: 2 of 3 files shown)


src/palace/manager/api/axis.py
```diff
@@ -211,7 +211,9 @@ def label(cls) -> str:
     def description(cls) -> str:
         return ""
 
-    def __init__(self, _db: Session, collection: Collection) -> None:
+    def __init__(
```
dbernstein, line 214: Allow a bearer token to be reused for cases when user is making multiple calls to the same collection but in different transactions in rapid succession.  This reuse dramatically reduces the number of calls that need to be made to the axis token endpoint, especially when there are many api calls being processed in parallel

src/palace/manager/celery/tasks/axis.py
```diff
@@ -241,23 +252,34 @@ def log_run_end_message() -> None:
...
         with task.transaction() as session:
             collection = Collection.by_id(session, id=collection_id)
-            api = create_api(session=session, collection=collection)  # type: ignore[arg-type]
+            api = create_api(session=session, collection=collection, bearer_token=bearer_token)  # type: ignore[arg-type]
             try:
                 process_book(task, session, api, metadata, circulation)
                 total_imported_in_current_task += 1
-            except (ObjectDeletedError, StaleDataError) as e:
+            except (ObjectDeletedError, StaleDataError, OperationalError) as e:
+                _check_if_deadlock(e)
```
dbernstein, line 281: raises error if e is a deadlock based operational error.

jonathangreen (approved):
This looks good.

I'm a bit concerned about retrying on deadlocks. I think deadlocks in the database are a real problem, and I'd like to try to stop them from happening, rather then just retrying on them. 

So my preference here would be to remove the deadlock retry part of this PR, I think that will get us down to an exception or two a day. Then put some additional tickets in future sprints to work out the deadlock problem. However I'll defer to you on this, if you think retries here are the way to go, I'm fine with it going in as is.

dbernstein:
@jonathangreen :  I agree that it is a real problem, but it is a problem that arises in the specific circumstance where there are lots of overlapping axis collections on a CM.  One of the results of not retrying is that we may be missing updates to some books since the failures effective end the updates to that collection.  Since there is a separate and completely independent operation for resolving the list of ids to update,  failures in the import_identifiers task will lead to a premature cessation of processing identifiers.

Once we remove the coverage table I think the problem will become even more scarce since the number of database writes within a transaction will be even lower which should lower the likelihood of a deadlock.  If we're seeing dead locks after 4 retries then we will still know about it. 

As long as we are logging the occurrence of deadlocks then we can at least keep track of how frequently they are occurring (assuming they are not occurring after multiple retries).  I will verify that we are logging a warning and if not add one before merging.

PR #2376, opened 2025-03-28
Mitigate axis reaper task buildup

dbernstein:
## Description
Here are some changes that I'd like to try rolling out to the  Vermont and Georgia scripts instances to see if we can bring the default queues back to zero.

The first change is to reduce the frequency of the axis.reap_all_collections  task this should stop the growth of the queue. 
The second is to increase the maximum number of celery workers to 8 from 4. 
## Motivation and Context

## Checklist


dbernstein:
closing this one since #2377  is put up against main.

PR #2377, opened 2025-03-31
Reduce frequency of axis.reap_all_collections to once a month (on the…

dbernstein:
… 30th)

## Description
This is a temporary measure to reduce the backup of axis reap colleciton tasks: 

## Motivation and Context

## Checklist

### DIFF
--- a/src/palace/manager/service/celery/celery.py
+++ b/src/palace/manager/service/celery/celery.py
@@ -88,9 +88,10 @@ def beat_schedule() -> dict[str, Any]:
         "axis_reap_all_collections": {
             "task": "axis.reap_all_collections",
             "schedule": crontab(
+                day_of_month=30,
                 minute="0",
                 hour="4",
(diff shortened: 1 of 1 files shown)


jonathangreen (approved):

jonathangreen:
I'm not sure what this reaper does? But are you concerned about the user experience from running this only once a month?

dbernstein:
The reaper runs through each item in the database and updates the availability.  If the identifier is no longer part of the collection it deletes it.  Since we have the import (which also updates availability and creates new items) running every 15 minutes (which pulls in anything that has changed since the last run) I think this is okay until we get the bottleneck sorted out.  I'll add a follow up task to bring it back to once a day.

dbernstein:
I'm going to hold off merging this one as https://github.com/ThePalaceProject/circulation/pull/2380 will likely obviate the need for it.

dbernstein:
This is obsolete.

PR #2379, opened 2025-03-31
Fixes for Overdrive Integration

jonathangreen:
## Description

Fixes two issues I found while testing the changes in https://github.com/ThePalaceProject/circulation/pull/2359 with the mobile apps:
- Early returns were causing an exception because they return a blank response, not JSON
- Some checkout responses return links that are `dict[str, list[Link]]` rather then `dict[str, Link]`

## Motivation and Context

- Updated type hint to cover both cases of links, and added a test (and fixture file) based on the response I was seeing
- Updated the `patron_request` function to return a `Response` if the request isn't validated
- Add some more error handling when we can't validate the response we get from Overdrive

## How Has This Been Tested?

- Tested locally
- Added tests to CI

I'll need to do another round of testing with this using the mobile apps once this is merged.

## Checklist

### DIFF
--- a/src/palace/manager/api/overdrive/api.py
+++ b/src/palace/manager/api/overdrive/api.py
@@ -785,9 +785,8 @@ class OverdriveAPI(
                 # We were unable to validate the response as the expected type. Log some relevant details and
                 # raise a BadResponseException.
                 self.log.exception(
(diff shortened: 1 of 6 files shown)


dbernstein (approved):
Looks good.

PR #2380, opened 2025-04-01
Optimize the axis.reap_collection task for speed while also minimizin…

dbernstein:
…g the possibility of dead locks.

## Description
After running a series of performance tests it became clear that running individual identifiers through the api.update_license_pools_for_identifiers method was hurting us badly.   Looking back, dead locks on the axis collection reaper tasks were relatively rare  when they were batched:  2 in 24 hours.   So I brought back the batching for the reaper and added a longer delay (3 minutes)  between parallel collection reap tasks.  This should be more than enough to prevent all the but the extremely freak deadlock.

## Motivation and Context

## Checklist


src/palace/manager/celery/tasks/axis.py
```diff
@@ -378,9 +378,24 @@ def get_collections_by_protocol(
...
     """
     A shared task that  kicks off a reap collection task for each Axis 360 collection.
+    Palace Managers can be configured with many Axis collections which may have identical or nearly identical
+    lists of identifiers.  As a result reaping Axis collections in parallel can lead to frequent
+    deadlocks when launched in parallel.
+
+    The default 60 second delay between launching of reap_collection subtasks is based on
+    observations of performance in a local environment.  I observed that reaping in batches 100
+    can take from 10-20 seconds in my development environment. The production environment was
+    observed to be about 2 to 3 times slower.  Therefore,  a minute delay between task initiation provides a reasonable
```
jonathangreen, line 393: > Therefore,  a minute delay
Comment seems out of date. It looks like the delay is 3 minutes?

src/palace/manager/celery/tasks/axis.py
```diff
@@ -390,7 +405,8 @@ def reap_all_collections(task: Task) -> None:
             )
             reap_collection.apply_async(
                 kwargs={"collection_id": collection.id},
-                countdown=count * 5,  # stagger the execution of the collection import
+                countdown=count
+                * delay_in_seconds_between_reap_collection_tasks,  # stagger the execution subtasks
```
jonathangreen, line 409: The warning in [the manual](https://docs.celeryq.dev/en/stable/userguide/calling.html#eta-and-countdown) seems relevant here:
> Tasks with eta or countdown are immediately fetched by the worker and until the scheduled time passes, they reside in the worker’s memory. When using those options to schedule lots of tasks for a distant future, those tasks may accumulate in the worker and make a significant impact on the RAM usage.   
>...
> Therefore, using eta and countdown **is not recommended** for scheduling tasks for a distant future. Ideally, use values no longer than several minutes.

Unless I'm misunderstanding here, it looks like `delay_in_seconds_between_reap_collection_tasks` is 3 minutes, and there can be many collections, we may be scheduling these tasks fairly far into the future, which they recommend against.

jonathangreen (approved):
Added a couple comments for consideration

src/palace/manager/celery/tasks/axis.py
dbernstein: That's a valid point.  I guess the key question is what do they mean by "lots of tasks".    I'm thinking that this should be okay as we'll have at most on the order of 100 tasks _per day_ scheduled in this way.   With 5 worker threads (4 + 1 beat scheduler thread which I believe doubles as a worker), each thread shouldn't have many more than 20 tasks in memory.  Within a few hours, all the tasks will be consumed.  Also according to the docs, any task with count or ETA in the future will not be lost if the worker that has fetched it exits for any reason so we don't need to worry about losing tasks between restarts.

Another option here is to gather a list of collection ids, launch a single reap collection task (based on the first collection id in the list) and then pass the list of collection ids to a single task with an countdown 3 minutes in the future, repeating until the list is consumed.

I say let's give it a shot as is and adjust if we need to.

PR #2381, opened 2025-04-01
Return exception when an Overdrive book has a format that cannot be fulfilled

jonathangreen:
## Checklist


dbernstein (approved):
Looks good.

PR #2384, opened 2025-04-02
Bugfix for http verbose logging (PP-2319)

jonathangreen:
## Description

Fix bug in our HTTP class that was causing an exception when making a request with `verbose=True`. It was causing the failures in PP-2319.

It was causing this exception to occur:
```
Log message could not be formatted. Exception: TypeError('not all arguments converted during string formatting')
```

## Motivation and Context

Was caused when I converted the logging statement to use a f-string in https://github.com/ThePalaceProject/circulation/pull/2366. After converting it, I forgot to remove the args to the log function 🤦🏻 

## How Has This Been Tested?

- Added a test for this case, since the verbose option wasn't being tested at all. This would have caught the issue. 

## Checklist

### DIFF
--- a/src/palace/manager/util/http.py
+++ b/src/palace/manager/util/http.py
@@ -300,9 +300,7 @@ class HTTP(LoggerMixin):
         try:
             if verbose:
                 logging.info(
-                    f"Sending {http_method} request to {url}: kwargs {kwargs!r}",
-                    url,
-                    kwargs,
+                    f"Sending {http_method} request to {url}: kwargs {kwargs!r}"
                 )
 
             request_start_time = time.time()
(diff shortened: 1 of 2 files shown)


src/palace/manager/util/http.py
```diff
@@ -300,9 +300,7 @@ def _request_with_timeout(
         try:
             if verbose:
                 logging.info(
-                    f"Sending {http_method} request to {url}: kwargs {kwargs!r}",
-                    url,
-                    kwargs,
+                    f"Sending {http_method} request to {url}: kwargs {kwargs!r}"
```
dbernstein, line 303: Curious: what does `kwargs!r` do?

dbernstein (approved):
Looks good!

PR #2392, opened 2025-04-04
Adjust delay between launching of axis.reap_collection tasks to 20 se…

dbernstein:
…conds

Fire reap_all_collections on Saturday nights.

## Description
Celery default queues are filling up again because axis.reap_collection tasks are being scheduled past the redis visibility timeout of 1 hour.   This PR ensures that all tasks will be scheduled well within the one hour window.  It also changes the weekly schedule to Saturday mornings so that I can observe the behavior over the weekend to make sure things are fixed.

 

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/axis.py
+++ b/src/palace/manager/celery/tasks/axis.py
@@ -32,6 +32,7 @@ from palace.manager.util.log import pluralize
 
 DEFAULT_BATCH_SIZE: int = 10
 DEFAULT_START_TIME = datetime_utc(1970, 1, 1)
+MAX_DELAY_IN_SECONDS_BETWEEN_REAP_TASKS = 20
 
 
 @shared_task(queue=QueueNames.default, bind=True)
@@ -390,7 +391,8 @@ def get_collections_by_protocol(
 
 @shared_task(queue=QueueNames.default, bind=True)
 def reap_all_collections(
-    task: Task, delay_in_seconds_between_reap_collection_tasks: int = 20
+    task: Task,
+    delay_in_seconds_between_reap_collection_tasks: int = MAX_DELAY_IN_SECONDS_BETWEEN_REAP_TASKS,
 ) -> None:
     """
     A shared task that  kicks off a reap_collection task for each Axis 360 collection.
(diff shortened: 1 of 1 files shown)


src/palace/manager/celery/tasks/axis.py
```diff
@@ -390,18 +390,18 @@ def get_collections_by_protocol(
...
-    The default delay of 3 minutes between launching of reap_collection sub-tasks is based on
-    observations of performance in a local environment.  I observed that reaping in batches 100
-    can take from 10-20 seconds in my development environment. The production environment was
-    observed to be about 2 to 3 times slower.  Therefore,  a minute delay between task initiation provides a reasonable
+    The default delay of 20 seconds between launching of reap_collection sub-tasks is based on
+    observations of performance in a local environment.  I observed that reaping in batches 10 seconds
+    can take from 2-4 seconds in my development environment. The production environment was
+    observed to be about 2 to 3 times slower.  Therefore,  a 20 second delay between task initiation provides a reasonable
     empirically based time buffer such that batches of axis identifiers are not reaped simultaneously and won't likely
     cause a pile-up in the case of occasional deadlocks.
```
jonathangreen, line 406: https://github.com/ThePalaceProject/circulation/blob/8471111d638ffbf7a9809c0024a3e567ff3ad1fb/src/palace/manager/celery/tasks/axis.py#L417-L424

Should you perhaps have a maximum delay time here? So that adding additional collections doesn't cause this to go beyond the visibility timeout in the future? Seems like this presents a failure mode that will be unexpected if we start adding more collections.

jonathangreen (approved):
Added a comment for consideration

src/palace/manager/celery/tasks/axis.py
dbernstein: @jonathangreen : That's a good idea.  Maybe also I should increase the visibility timeout by 30 minutes or so.  What do you think?

dbernstein: Actually 20 seconds gives us 180 axis collections.  I think we can leave that.

PR #2393, opened 2025-04-06
Better parsing for BCP 47 language tags

jonathangreen:
## Description

OPDS technically allows BCP-47 language tags. Previously we were parsing these as ISO language codes. This is a subset of language tags, but doesn't account for any subtags. 

We now allow subtags, although we don't actually parse them, just validate the primary language as a language code.

## Motivation and Context

There were some technically valid language codes being used in the Palace Bookshelf feed ("en-US", and "en_UK"). I'm not sure these tags are adding a lot of value, but they are valid in a feed, so we should parse them. 

## How Has This Been Tested?

- Updated unit tests

## Checklist


dbernstein (approved):
🚀

PR #2394, opened 2025-04-07
Move the LicenseStatus enum into OPDS package

jonathangreen:
## Description

Move the `LicenseStatus` enum into OPDS package

## Motivation and Context

This enum is really part of our OPDS model, so it should be in the OPDS package. Having it there also means that importing the OPDS models, doesn't mean importing all of our sqlalchemy models. This is helpful for this PR: https://github.com/ThePalaceProject/palace-tools/pull/107

## How Has This Been Tested?

- Ran unit tests in CI

## Checklist


dbernstein (approved):
💯

PR #2398, opened 2025-04-07
Add support for specifying a shared google drive as root folder.

dbernstein:
## Description

This PR does two things:

1.  add the data_source_name column to the playtime summaries, entries, and reports.
2. break out reports by datasource name rather than collection name.
3. change the directory hierarchy to /{google root folder (if present)}/{data source name}/Usage Report/{cm name}/{year}

## Motivation and Context
These changes attempt to align the code with Tara's workflow based on feedback in the ticket.

## How Has This Been Tested?
Unit tests.  
Tested locally.

## Checklist


alembic/versions/20250407_f36442df213d_add_data_source_name_to_playtime_.py
```diff
@@ -0,0 +1,64 @@
...
+Create Date: 2025-04-07 21:00:49.422754+00:00
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from palace.manager.sqlalchemy.model.time_tracking import PlaytimeEntry, PlaytimeSummary
```
jonathangreen, line 14: Alembic models aren't supposed to import sqlalchemy models, since these can change over time, but the migration should always stay the same.

alembic/versions/20250407_f36442df213d_add_data_source_name_to_playtime_.py
```diff
@@ -0,0 +1,64 @@
...
+depends_on = None
+
+
+def upgrade() -> None:
+    session: Session = Session(bind=op.get_bind())
+    conn = session.connection()
+
+    op.add_column(
+        "playtime_entries",
+        sa.Column("data_source_name", sa.String(), nullable=False, default="Unknown"),
```
jonathangreen, line 29: This fails when you run the migration on a DB with data in it with the error: 
```
column "data_source_name" of relation "playtime_entries" contains null values
```

alembic/versions/20250407_f36442df213d_add_data_source_name_to_playtime_.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    for p in session.scalars(
+        select(PlaytimeEntry).where(PlaytimeEntry.collection_id is not None)
+    ).all():
+        p.data_source_name = p.collection.data_source.name
```
jonathangreen, line 34: If you update the nullable constraint above, then this line fails with: 
```
AttributeError: 'NoneType' object has no attribute 'data_source'
```

src/palace/manager/service/google_drive/configuration.py
```diff
@@ -7,8 +7,13 @@
...
 )
 
-# Environment variable specifying the google drive root folder for this environment
-PALACE_GOOGLE_DRIVE_ROOT_ENVIRONMENT_VARIABLE = "PALACE_GOOGLE_DRIVE_ROOT"
+# Environment variable specifying the google drive parent folder ID for this environment.
+# This value may be a folder ID or a shared drive ID. For example, the google drive folder URI will look like
+# https://drive.google.com/drive/u/1/folders/0AGtlKYStJaC3Uk9PVZ .   "0AGtlKYStJaC3Uk9PVZ" (not a real
+# folder ID) is the value that should be assigned environment variable.
+PALACE_GOOGLE_DRIVE_PARENT_FOLDER_ID_ENVIRONMENT_VARIABLE = (
+    "PALACE_GOOGLE_DRIVE_PARENT_FOLDER_ID"
```
jonathangreen, line 15: Need to update the `README` with the new environment variable, and the information you put here in a comment.

jonathangreen (commented):
I approved this, but it needs some changes before it can go in: 
- the migration needs some work
  - before you merge it, try running the migration on a DB with some playtime data already in it.
- the documentation needs updates. 
  - I notice you never actually check the check-boxes in the pull request template, but the "I have updated the documentation accordingly" one is there to remind us to actually update the readme. So it might be worth going through that checklist when you make a PR.

PR #2402, opened 2025-04-09
Bugfix detection of Overdrive loans that cannot be viewed in mobile clients (PP-2372)

jonathangreen:
## Description

This PR does two related things. They are in separate commits to make code review a bit easier.

- Handle the case where an overdrive ebook is available in multiple formats, none of which we support. We've seen this case in our real CMs where the formats available after a checkout are: `{'ebook-overdrive', 'ebook-kindle'}`. Before this change, we were failing to detect that as a case that the mobile apps couldn't view.
- Rename the `supported_formats` property on the `Checkout` model to `available_formats`, since the previous name was a bit misleading, and contributed to this bug.

## Motivation and Context

Issue came in via support. See ticket: PP-2372. 
Certain OverDrive titles on iOS staying in constant state of "Requesting" after clicking Download. Currently the mobile clients cannot handle a book that has no way to fulfill after checkout. We are handling this in the backend, but were failing to detect this case.

## How Has This Been Tested?

- Added new test case
- Tested locally with problematic title

## Checklist


tests/manager/api/overdrive/test_api.py
```diff
@@ -856,17 +880,21 @@ def test_checkout(
...
-            (
-                "post",
-                "https://integration-patron.api.overdrive.com/v1/patrons/me/checkouts",
-            ),
-            (
-                "delete",
-                "https://patron.api.overdrive.com/v1/patrons/me/checkouts/4f40589f-03e7-41b6-9d9a-5e6fd77fdf98",
-            ),
-        ]
+        # We made requests to checkout the book and return it. The first request is to get the patrons
```
tdilauro, line 883: Minor typo: "... patrons" -> "... patron's"

tdilauro (approved):
Looks good! 🏁

PR #2403, opened 2025-04-09
Bump poetry version to match dependabot

jonathangreen:
## Description

Bump up the Poetry version we are using to match what dependabot uses.

## Motivation and Context

I've noticed some messages in our CI lately that the lock file version may not be supported, since dependabot has moved to a 2.x version of Poetry.

## How Has This Been Tested?

- Build in CI

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -375,7 +375,7 @@ jobs:
           bash -c "
             cat src/palace/manager/_version.py &&
             source env/bin/activate &&
-            poetry install --without ci --no-root --sync &&
+            poetry sync --without ci --no-root &&
             pytest --no-cov tests
           "
         env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,13 @@ jobs:
       exec-repo: ${{ steps.ghcr-repo.outputs.exec }}
       exec-meta: ${{ steps.meta-exec.outputs.json }}
 
+    # https://docs.docker.com/build/ci/github-actions/local-registry/
+    services:
+      registry:
+        image: registry:2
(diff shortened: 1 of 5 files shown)


jonathangreen:
Moving this to draft, until I can sort out whats going on in the docker build

.github/workflows/build.yml
```diff
@@ -31,6 +31,13 @@ jobs:
       exec-repo: ${{ steps.ghcr-repo.outputs.exec }}
       exec-meta: ${{ steps.meta-exec.outputs.json }}
 
+    # https://docs.docker.com/build/ci/github-actions/local-registry/
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+
```
jonathangreen, line 40: Our builds when we change the baseimage within the PR were broken. We don't change the baseimage very often, and it looks like a change to docker buildx means that images are built within a container now, so you can't load a local image. You have to run a local registry and load the image from that registry. Those changes could be a separate PR, but since they are hard to test without a base image, I'm including them here with the poetry changes, since they change the base image and test this.

.github/workflows/build.yml
```diff
@@ -366,7 +375,7 @@ jobs:
           bash -c "
             cat src/palace/manager/_version.py &&
             source env/bin/activate &&
-            poetry install --without ci --no-root --sync &&
+            poetry sync --without ci --no-root &&
```
jonathangreen, line 378: Latest poetry has deprecated the `poetry install --sync` command in favor of `poetry sync`, so I switched to using it here.

tdilauro (approved):
Looks good! 🚀

PR #2404, opened 2025-04-09
Holds report includes only books with holds. (PP-2317)

tdilauro:
## Description

Fixes holds reports to include only titles with at least one hold.

NB: The overall associated inventory report should include all books.

## Motivation and Context

[Jira PP-2317]

## How Has This Been Tested?

- Updates a test to ensure that books with and without holds continue to be included in inventory reports; and that books without holds are excluded from the holds report.
- Tests pass locally.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/14365660091) pass.

## Checklist


jonathangreen (approved):
🎉🎸

PR #2408, opened 2025-04-11
Mock google drive api for tests (PP-2280)

jonathangreen:
## Description

Updates google drive integration to:
- Use more dependency injection
- Load all environment variables via pydantic
- Mock google drive API when no testing key is available

## Motivation and Context

Tests are failing both locally and in CI for dependabot PRs. Need to update how we are testing the google drive integration, so our tests aren’t failing.

## How Has This Been Tested?

- Running tests locally
- Running tests in CI

## Checklist


tests/manager/scripts/test_playtime_entries.py
```diff
@@ -681,37 +680,46 @@ def test_do_run(
...
-            patch(
-                "palace.manager.scripts.playtime_entries.os.environ",
-                new={
-                    Configuration.REPORTING_NAME_ENVIRONMENT_VARIABLE: reporting_name,
-                    PALACE_GOOGLE_DRIVE_PARENT_FOLDER_ID_ENVIRONMENT_VARIABLE: parent_folder_id,
-                },
-            ),
-        ):
-            # Act
-            PlaytimeEntriesReportsScript(db.session).run()
```
jonathangreen, line 703: I updated this so that we're not patching the csv writer. That way we can test that the data makes it all the way into the stream.

dbernstein (approved):
I like what you've done here.  It's much tidier than it was before.

PR #2411, opened 2025-04-14
Update our pre-commit linters

jonathangreen:
## Description

Updates our pre-commit linters. Figured I would do this while updating poetry.

## Motivation and Context

This PR was created by running:
```
pre-commit autoupdate
pre-commit run -a
```

So all the changes are automated from our linters. There are a few changes, but none that look problematic to me. 

## How Has This Been Tested?

- Running CI

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/library_settings.py
+++ b/src/palace/manager/api/admin/controller/library_settings.py
@@ -32,7 +32,9 @@ from palace.manager.sqlalchemy.listeners import site_configuration_has_changed
 from palace.manager.sqlalchemy.model.announcements import (
     SETTING_NAME as ANNOUNCEMENT_SETTING_NAME,
 )
-from palace.manager.sqlalchemy.model.announcements import Announcement
+from palace.manager.sqlalchemy.model.announcements import (
+    Announcement,
+)
 from palace.manager.sqlalchemy.model.library import Library, LibraryLogo
 from palace.manager.sqlalchemy.model.resource import Representation
 from palace.manager.sqlalchemy.session import json_serializer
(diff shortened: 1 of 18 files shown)


src/palace/manager/api/admin/controller/library_settings.py
```diff
@@ -32,7 +32,9 @@
 from palace.manager.sqlalchemy.model.announcements import (
     SETTING_NAME as ANNOUNCEMENT_SETTING_NAME,
 )
-from palace.manager.sqlalchemy.model.announcements import Announcement
+from palace.manager.sqlalchemy.model.announcements import (
+    Announcement,
+)
```
jonathangreen, line 37: I couple of these import changes are a little weird, but it appears to be related to a change isort made in handling `as` imports on the previous line. Even though is a bit weird, I don't think its a problem.

jonathangreen: Because of these kind of weird imports, I tried enabling the `combine_as_imports` isort option in our config. 

The automatic results are shown in https://github.com/ThePalaceProject/circulation/pull/2411/commits/87602eb9a5606ad9f6220b709d4cd4fafd1462db. I think I like how things are sorted better with this option enabled.

jonathangreen:
It creates a few more changes, but I added the `combine_as_imports` isort option, and I like the results, as I mentioned in https://github.com/ThePalaceProject/circulation/pull/2411#discussion_r2042290065. I'll leave that in as part of this PR for now, but I'm happy to break it out our drop adding that option all together. Let me know what you think.

src/palace/manager/api/admin/controller/library_settings.py
tdilauro: I prefer the `combine_as_imports` behavior.

tdilauro (approved):
🏅

PR #2422, opened 2025-04-22
Add additional sort column to playtime report

jonathangreen:
## Description

Add an additional `order by` to the playtime report. 

## Motivation and Context

This shouldn't come into play in real reports very often, but it makes sure we have a consistent order of results in our tests. Should stop this flakey test failure: https://github.com/ThePalaceProject/circulation/actions/runs/14595662530/job/40940999621

## How Has This Been Tested?

- Running tests in CI

## Checklist


dbernstein (approved):
LGTM

PR #2423, opened 2025-04-22
Additional information in SIP2 self test (PP-2418)

jonathangreen:
## Description

Run all the SIP2 self tests, even if there is a failure, as long as we have a test barcode configured.

## Motivation and Context

Running all the tests outputs helpful debug information that can help to figure out a failure such as the raw patron information. This also makes it more clear when requests are timing out.

See PP-2418

## How Has This Been Tested?

- Tested locally

## Checklist


tdilauro (approved):
Looks good! 🚀

PR #2424, opened 2025-04-22
Configurable SIP2 timeout (PP-2418)

jonathangreen:
## Description

Allow configuration of SIP2 connection timeout.

## Motivation and Context

We set the SIP2 connection timeout fairly low, because it has impacts on all the performance for all libraries on a CM. A single slow library, can slow the rest of them down. Libraries with slow authentication should cause less of a performance impact now that we have token auth rolled out everywhere, as we don't have to call out to the ILS as often, so I think adding a configuration that allows us to increase the timeout makes sense.

## How Has This Been Tested?

- Added tests
- Tested locally

## Checklist

### DIFF
--- a/src/palace/manager/api/sip/client.py
+++ b/src/palace/manager/api/sip/client.py
@@ -231,8 +231,6 @@ class Constants:
 class SIPClient(Constants, LoggerMixin):
     # Maximum retries of a SIP message before failing.
     MAXIMUM_RETRIES = 5
-    # Timeout in seconds
-    TIMEOUT = 3
 
     # These are the subfield names associated with the 'patron status'
     # field as specified in the SIP2 spec.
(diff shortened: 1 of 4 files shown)


src/palace/manager/api/sip/client.py
```diff
@@ -231,8 +231,6 @@ class Constants:
 class SIPClient(Constants, LoggerMixin):
     # Maximum retries of a SIP message before failing.
     MAXIMUM_RETRIES = 5
-    # Timeout in seconds
-    TIMEOUT = 3
```
tdilauro, line 235: Minor suggestion: Instead of removing this, could we just rename to DEFAULT_TIMEOUT and use it below, instead of hardcoding `3` down there?
```python
    # Default timeout in seconds, used when no timeout explicitly configured.
    DEFAULT_TIMEOUT =  3
```

tdilauro (approved):
Looks good! ⏲️ 

One small suggestion below.

PR #2428, opened 2025-04-24
[PP-2405] This update optimizes the axis identifiers import and reape…

dbernstein:
…r tasks by reducing the number of unnecessary calls to both CirculationData.apply and LicensePool.set_open_access_status. This call is particularly expensive when a CM has many licensepools associated with each identifier due to having many collections configured with the same set of books. Since the performance is significantly improved I have upped the default batch size for axis tasks from 10 to 50 to reduce the number of calls for token renewal. Finally this commit includes a new convenience script for manually kicking off an axis import.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2405

## How Has This Been Tested?
I tested this locally with the Georgia data.   The change reduced the average time per book from 1 second to .20 seconds.

## Checklist

### DIFF
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -620,9 +620,7 @@ class Axis360API(
         """
         remainder = set(identifiers)
         for bibliographic, availability in self._fetch_remote_availability(identifiers):
-            edition, ignore1, license_pool, ignore2 = self.update_book(
-                bibliographic, availability
-            )
+            edition, ignore1, license_pool, ignore2 = self.update_book(bibliographic)
             identifier = license_pool.identifier
             if identifier in remainder:
                 remainder.remove(identifier)
@@ -637,20 +635,21 @@ class Axis360API(
     def update_book(
         self,
         bibliographic: Metadata,
-        availability: CirculationData,
         analytics: Analytics = Provide[Services.analytics.analytics],
     ) -> tuple[Edition, bool, LicensePool, bool]:
         """Create or update a single book based on bibliographic
         and availability data from the Axis 360 API.
 
         :param bibliographic: A Metadata object containing
-            bibliographic data about this title.
-        :param availability: A CirculationData object containing
-            availability data about this title.
+            bibliographic and circulation (ie availability) data about this title
         """
-        license_pool, new_license_pool = availability.license_pool(
+        # The axis the bibliographic metadata always includes the circulation data
+        assert bibliographic.circulation
+
+        license_pool, new_license_pool = bibliographic.circulation.license_pool(
             self._db, self.collection
         )
+
         edition, new_edition = bibliographic.edition(self._db)
         license_pool.presentation_edition = edition
         policy = ReplacementPolicy(
@@ -662,9 +661,6 @@ class Axis360API(
             analytics=analytics,
         )
 
-        # NOTE: availability is bibliographic.circulation, so it is
-        # unnecessary to call availability.apply() -- it's taken
-        # care of inside bibliographic.apply().
         bibliographic.apply(edition, self.collection, replace=policy, db=self._db)
         return edition, new_edition, license_pool, new_license_pool
 
--- a/src/palace/manager/api/axis.py
+++ b/src/palace/manager/api/axis.py
@@ -662,7 +662,7 @@ class Axis360API(
             analytics=analytics,
         )
 
-        # NOTE: availability is bibliographic.circulation, so it's a
+        # NOTE: availability is bibliographic.circulation, so it is
         # unnecessary to call availability.apply() -- it's taken
         # care of inside bibliographic.apply().
         bibliographic.apply(edition, self.collection, replace=policy, db=self._db)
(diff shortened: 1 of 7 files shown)


src/palace/manager/sqlalchemy/model/licensing.py
```diff
@@ -1669,7 +1669,10 @@ def set(
         if not dirty and update_available and available is not None:
             lpdm.available = available
 
-        if not lpdm.rights_status or rights_status.uri != RightsStatus.UNKNOWN:
+        if not lpdm.rights_status or (
+            lpdm.rights_status != rights_status
```
dbernstein, line 1673: Without this check for non-equality, this test was returning true in almost every case leading to returning a dirty value for the delivery mechanism in every case which meant that 60 odd license pools were being updated unnecessarily for every book update.  So that means that the set_open_access_status was being called 3600  the georgia collections (60 license pools * 60 import tasks )  unnecessarily for every single book.  If were updating 28000 books in each of 12  collections (ie complete reimport for those 12 only) we'd see 1.2 billion unnecessary calls.

dbernstein: When you consider this method is also called by the reaper, we're taking 28000*60*3600 = 6 billion unnecessary calls every run of the reaper.

src/palace/manager/api/axis.py
```diff
@@ -662,11 +662,10 @@ def update_book(
             analytics=analytics,
         )
 
-        # NOTE: availability is bibliographic.circulation, so it's a
-        # little redundant to call availability.apply() -- it's taken
+        # NOTE: availability is bibliographic.circulation, so it is
+        # unnecessary to call availability.apply() -- it's taken
         # care of inside bibliographic.apply().
         bibliographic.apply(edition, self.collection, replace=policy, db=self._db)
-        availability.apply(self._db, self.collection, replace=policy)
```
jonathangreen, line 669: I'd suggest removing the `availability` parameter to this function and updating the functions doc comment. The parameter is unused after this change, and its misleading, since passing an availability to this function won't do anything with it, unless its already referenced by `bibliographic`.

tests/manager/api/test_axis.py
```diff
@@ -134,6 +134,7 @@ def __init__(self, db: DatabaseTransactionFixture, files: AxisFilesFixture):
             db.session, db.default_library()
         )
         self.api = MockAxis360API(db.session, self.collection)
+        self.BIBLIOGRAPHIC_DATA.circulation = self.AVAILABILITY_DATA
```
jonathangreen, line 137: Minor: I don't like updating a class constant in the `__init__` function like this. If you ever access it before `__init__` it won't be set. I'd suggest just removing `AVAILABILITY_DATA` and updating `BIBLIOGRAPHIC_DATA` to something like this: 
```python
    BIBLIOGRAPHIC_DATA = Metadata(
        DataSource.AXIS_360,
        publisher="Random House Inc",
        language="eng",
        title="Faith of My Fathers : A Family Memoir",
        imprint="Random House Inc2",
        published=datetime_utc(2000, 3, 7, 0, 0),
        primary_identifier=IdentifierData(
            type=Identifier.AXIS_360_ID, identifier="0003642860"
        ),
        identifiers=[IdentifierData(type=Identifier.ISBN, identifier="9780375504587")],
        contributors=[
            ContributorData(
                sort_name="McCain, John", roles=[Contributor.Role.PRIMARY_AUTHOR]
            ),
            ContributorData(sort_name="Salter, Mark", roles=[Contributor.Role.AUTHOR]),
        ],
        subjects=[
            SubjectData(
                type=Subject.BISAC, identifier="BIOGRAPHY & AUTOBIOGRAPHY / Political"
            ),
            SubjectData(type=Subject.FREEFORM_AUDIENCE, identifier="Adult"),
        ],
        circulation=CirculationData(
            data_source=DataSource.AXIS_360,
            primary_identifier=BIBLIOGRAPHIC_DATA.primary_identifier,
            licenses_owned=9,
            licenses_available=8,
            licenses_reserved=0,
            patrons_in_hold_queue=0,
            last_checked=datetime_utc(2015, 5, 20, 2, 9, 8),
        ),
    )
```

tests/manager/api/test_axis.py
```diff
@@ -774,6 +777,8 @@ def test_update_book(self, axis360: Axis360Fixture):
             licenses_available=7,
         )
 
+        axis360.BIBLIOGRAPHIC_DATA.circulation = new_circulation
+
         e2, e_new, lp2, lp_new = api.update_book(
             axis360.BIBLIOGRAPHIC_DATA,
             new_circulation,
```
jonathangreen, line 784: Shouldn't update a constant in a test like this. It will persist across tests, potentially causing issues for tests run after it.

I'd suggest something like: 
```suggestion
        biblographic = copy.deepcopy(axis360.BIBLIOGRAPHIC_DATA)
        bibliographic.circulation = new_circulation

        e2, e_new, lp2, lp_new = api.update_book(
            biblographic,
            new_circulation,
```

jonathangreen (approved):
This is a great improvement! Good hunting finding what was happening here.

I approved this one, but I added a couple minor comments for your consideration and there are a couple non-minor ones that would be nice to address before this goes in.

PR #2429, opened 2025-04-25
Mark LCP content as deliverable (PP-2334)

jonathangreen:
## Description

Add LCP to the list of deliverable formats in the mobile clients.

## Motivation and Context

Eventually, I think this property should be removed, as we have different clients, and they can deliver different formats. For example the web catalog can only deliver some formats, and the mobile apps might differ for some time in their ability to deliver a format. 

However we need bigger changes for that, and this property is still used in one of the importers and the custom list code. So I'm going to leave it for now, but update it so that LCP is marked as supported.

## How Has This Been Tested?

- Running tests in CI

## Checklist


tdilauro (approved):
👍🏽

jonathangreen:
It looks like this change either triggered or exposed from flakeyness in the OPDS2 tests. I was getting repeated test failures there, but only on the arm platform due to the order of the delivery mechanisms in the tests. This order doesn't matter so I updated the tests to use a set for comparison.

PR #2434, opened 2025-04-26
Add SIP2 dialect that sends SC Status message. (PP-2419)

tdilauro:
## Description

Adds a SIP2 dialect (SIP_V2 / "SIP v2.00 Compliant") that sends the SC Status (99) message before sending the Patron Information (63) message.

NB: Our other dialects, which do not send this message, are not compliant with the SIP v2.00 specification.

## Motivation and Context

Some SIP2 services (e.g., OCLC Wise) will not work properly if the SC Status (99) message is not sent (or, presumably, not sent in the proper sequence).

According to the SIP2 spec[1]:

> **SC Status**
> The SC status message sends SC status to the ACS. It requires an ACS Status Response message reply from the ACS. This message will be the first message sent by the SC to the ACS once a connection has been established (exception: the Login Message may be sent first to login to an ACS server program).
> The ACS will respond with a message that establishes some of the rules to be followed by the SC and establishes some parameters needed for further communication.
> `99<status code><max print width><protocol version>`

[1] https://developers.exlibrisgroup.com/wp-content/uploads/2020/01/3M-Standard-Interchange-Protocol-Version-2.00.pdf  Accessed 2025-04-24.

[Jira PP-2419]

## How Has This Been Tested?

- Manually tested in local development environment.
- All tests pass locally.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/14685037953) pass.

## Checklist

### DIFF
--- a/src/palace/manager/api/sip/__init__.py
+++ b/src/palace/manager/api/sip/__init__.py
@@ -114,17 +114,13 @@ class SIP2Settings(BasicAuthProviderSettings):
     ils: Sip2Dialect = FormField(
         Sip2Dialect.GENERIC_ILS,
         form=ConfigurationFormItem(
-            label="ILS",
+            label="ILS Dialect",
             description=(
                 "Some ILS require specific SIP2 settings. If the ILS you are using "
-                "is in the list please pick it otherwise select 'Generic ILS'."
+                f"is in the list, please pick it. Otherwise, select '{Sip2Dialect.preferred()}'."
             ),
             type=ConfigurationFormItemType.SELECT,
-            options={
-                Sip2Dialect.GENERIC_ILS: "Generic ILS",
-                Sip2Dialect.AG_VERSO: "Auto-Graphics VERSO",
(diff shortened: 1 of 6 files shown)


jonathangreen (approved):
Looks good!

This is a bit of an aside, this PR looks good as is and I don't want to delay getting it in. But it would be nice to know if many of our current ILS are SIP2 compliant. Its possible we could change the default to just be SIP2 compliant.

tdilauro:
> This is a bit of an aside, this PR looks good as is and I don't want to delay getting it in. But it would be nice to know if many of our current ILS are SIP2 compliant. Its possible we could change the default to just be SIP2 compliant.

@jonathangreen 💯 I had this in mind. I had suggested a similar idea when we added the "Folio" dialect, but there was concern that we might break some auth integrations. I'm going to add a ticket to do some testing; but, at least we're changing the default recommendation and we have a dialect with which we can do some of that testing.

PR #2444, opened 2025-05-01
Remove CSV metadata importer 🔥 (PP-1476)

jonathangreen:
## Description

Remove the unused `CSVMetadataImporter` class. 

This PR show a lot of changes, but its just from moving tests around. I'll make some comments to indicate this.

## Motivation and Context

I'm doing some refactoring of the metadata_layer and I found that this class is unused.

## How Has This Been Tested?

- Running tests in CI

## Checklist


tests/manager/core/test_metadata.py
```diff
@@ -49,675 +46,613 @@ def test_constructor(self):
...
-    return CSVFilesFixture()
+    def test_from_contribution(self, db: DatabaseTransactionFixture):
+        # Makes sure ContributorData.from_contribution copies all the fields over.
 
+        # make author with that name, add author to list and pass to edition
+        contributors = ["PrimaryAuthor"]
+        edition, pool = db.edition(with_license_pool=True, authors=contributors)
 
-class TestMetadataImporter:
-    def test_parse(self, csv_files_fixture: CSVFilesFixture):
```
jonathangreen, line 66: The diff here is confusing, but `test_parse` is the only test that we removed here. The other tests in `TestMetadataImporter` were still relevant, so I moved them down in the file into `TestMetadata` since that is what they are actually testing. This made for a kind of messy diff, but it preserves almost all the tests.

jonathangreen:
This one needs to go in before https://github.com/ThePalaceProject/circulation/pull/2445 since it touches the same files. 

I took this out mostly because it was unused and I didn't know where to put it when moving things around in https://github.com/ThePalaceProject/circulation/pull/2445

tdilauro (approved):
🔥

PR #2445, opened 2025-05-01
Split up core/metadata_layer.py (PP-1476)

jonathangreen:
## Description

Split up `core/metadata_layer.py`

## Motivation and Context

`manager/core/metadata_layer.py` was a 2000+ line file so before doing some refactoring work, I broke it out into a number of different files in `manager/metadata_layer`. I moved around the tests to match.

There are no logic changes here, just moving the existing classes around.

The only new code is a couple type hints that I had to add, because the file was moved into a directory with strict type hinting. I'll call these out in the review I put on this PR.

## How Has This Been Tested?

- Running tests in CI

## Checklist


src/palace/manager/metadata_layer/policy/presentation.py
```diff
@@ -13,18 +15,18 @@ class PresentationCalculationPolicy:
...
-        choose_summary=True,
-        calculate_quality=True,
-        choose_cover=True,
-        update_search_index=False,
-        verbose=True,
-        equivalent_identifier_levels=DEFAULT_LEVELS,
-        equivalent_identifier_threshold=DEFAULT_THRESHOLD,
-        equivalent_identifier_cutoff=DEFAULT_CUTOFF,
-    ):
+        choose_edition: bool = True,
```
jonathangreen, line 18: The few type hints added in this file are the only actual code changes.

src/palace/manager/core/monitor.py
```diff
@@ -810,3 +811,118 @@ class CustomListEntryWorkUpdateMonitor(CustomListEntrySweepMonitor):
 
     def process_item(self, item):
         item.set_work()
+
+
+class TimestampData:
```
jonathangreen, line 816: I moved this here from `metadata_layer` because its only used as part of the monitors, not really the metadata layer, so this seemed like a more appropriate home for it.

src/palace/manager/service/container.py
```diff
@@ -91,8 +91,8 @@ def wire_container(container: Services) -> None:
             "palace.manager.api.overdrive",
             "palace.manager.feed.annotator.circulation",
             "palace.manager.feed.acquisition",
+            "palace.manager.metadata_layer.policy.replacement",
```
jonathangreen, line 94: Had to update the path in the container where we are doing a dependency injection.

tdilauro (approved):
This looks good. 🔀 

I didn't do a deep review, since this was mostly moving code around.

PR #2447, opened 2025-05-01
Ensures that failures to retrieve an auth token are retried.

dbernstein:
## Description
I noticed that some of the Axis collections in Georgia that  we wanted to reimport from scratch were not completing the full run.  After some investigation I discovered the bug.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2464

## Checklist


tdilauro (approved):
This looks good.

It would be good to add more information to Motivation and Context, since this is an open source project and not everyone has access to our Jira tickets.

PR #2449, opened 2025-05-05
Optimize LibraryAnnotator NoveList config check. (PP-2491)

tdilauro:
## Description

- Lazy load and cache the value of `NoveList.is_configured` in the `LibraryAnnotator`. The previous behavior was to compute this value multiple times, once for every feed entry.
- Added test to ensure that `NoveList.is_configured` is called only once per feed generation, regardless of how many items are in the feed. Without the change this test was failing. After the change, it passes.

## Motivation and Context

Computing the value of `NoveList.is_configured` was contributing to slowness (1/5 to 1/4 of response time in my testing) when building our OPDS feeds because a database query is being run every time we check to see if NoveList is configured for a given library. However, the value should not change within a single feed build, so computing once and caching the result makes sense here.

## How Has This Been Tested?

- Added new test.
- All tests pass locally.
- Tested with profiling in local development environment.
  - After this change, calls to `NoveList.is_configured` were no longer contributing significantly to the response time and the response time improved.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/14845548918) pass.

## Checklist

### DIFF
--- a/src/palace/manager/feed/annotator/circulation.py
+++ b/src/palace/manager/feed/annotator/circulation.py
@@ -7,6 +7,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 from collections import defaultdict
+from functools import cached_property
 from typing import Any
 
 from dependency_injector.wiring import Provide, inject
@@ -740,18 +741,15 @@ class LibraryAnnotator(CirculationManagerAnnotator):
         self._top_level_title = top_level_title
         self.identifies_patrons = library_identifies_patrons
         self.facets = facets or None
-        self._is_novelist_configured: bool | None = None
 
-    @property
+    @cached_property
     def is_novelist_configured(self) -> bool:
         """Lazy load and cache NoveList's `is_configured` flag.
 
         This is an optimization to avoid a SQL query to check this
-        flag for every book in the feed.
+        flag for every entry in the feed.
         """
-        if self._is_novelist_configured is None:
-            self._is_novelist_configured = NoveListAPI.is_configured(self.library)
-        return self._is_novelist_configured
+        return NoveListAPI.is_configured(self.library)
 
     def top_level_title(self) -> str:
         return self._top_level_title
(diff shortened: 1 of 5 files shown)


src/palace/manager/feed/annotator/circulation.py
```diff
@@ -740,6 +740,18 @@ def __init__(
         self._top_level_title = top_level_title
         self.identifies_patrons = library_identifies_patrons
         self.facets = facets or None
+        self._is_novelist_configured: bool | None = None
+
+    @property
```
jonathangreen, line 745: Minor: could use `cached_property` if you don't need to control the caching here

src/palace/manager/feed/annotator/circulation.py
```diff
@@ -740,6 +740,18 @@ def __init__(
...
+    def is_novelist_configured(self) -> bool:
+        """Lazy load and cache NoveList's `is_configured` flag.
+
+        This is an optimization to avoid a SQL query to check this
+        flag for every book in the feed.
+        """
+        if self._is_novelist_configured is None:
+            self._is_novelist_configured = NoveListAPI.is_configured(self.library)
```
jonathangreen, line 753: Minor: Should `is_configured` maybe be private to make it clear the property should be the preferred way to access?

dbernstein (approved):
lgtm

jonathangreen (approved):
Looks good! Made a couple minor comments for consideration but, as always, feel free to merge as is

src/palace/manager/feed/annotator/circulation.py
tdilauro: Good point! I'll do that.

src/palace/manager/feed/annotator/circulation.py
tdilauro: I didn't do that because there was a class method that used `is_configured`, but could not access the instance property. I'll take another look to see if I can clean that up, too.

tdilauro: I decided not to make it private, since it's part of the public api of NoveListAPI. I added a docstring to indicate that there can be performance implications to running it. I also renamed it to hint in that direction, as well.

PR #2454, opened 2025-05-07
[PP-2503] Ensure google drive folders are not duplicated in a shared …

dbernstein:
…google drive.

## Description

This update just adds a setting to ensure that the shared drives are included in the google drive search for existing folders.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2503

## How Has This Been Tested?
Manually tested against the EBCE shared drive.

## Checklist


tdilauro (approved):
👍🏽

PR #2455, opened 2025-05-07
Refactor novelist integration to decouple from metadata layer (PP-2466)

jonathangreen:
## Description

Refactor `NoveListAPI` to decouple it from the `Metadata` object from the metadata layer.

## Motivation and Context

The `NoveListAPI` integration used to capture all its data into a `Metadata` object. However this data was never stored in the database by calling the `apply` method on it. The `Metadata` object was just used to pass book recommendations into the `RecommendationLane`, which isn't a very good fit for the `Metadata` object. 

This refactor decouples `NoveListAPI` from the metadata layer, allowing it to return a list of recommendations directly. This removes some of the parsing of the `NoveListAPI` data that was parsed but never used. If we decide in the future that we want to capture this data and store it in the database we can bring back in some of this code. (I'm not sure if we are even allowed to store this data).

## How Has This Been Tested?

- Running unit tests.

## Checklist

### DIFF
--- a/src/palace/manager/api/lanes.py
+++ b/src/palace/manager/api/lanes.py
@@ -1,4 +1,5 @@
 import logging
+from collections import defaultdict
 
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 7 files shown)


dbernstein (approved):
🚀

tdilauro (approved):
Looks good! 📚

PR #2458, opened 2025-05-07
Use pydantic for Metadata and CirculationData (PP-2466)

jonathangreen:
## Description

Update `Metadata` and `CirculationData` to use pydantic for parsing and serialization. 

Note: Unlike the conversions done in https://github.com/ThePalaceProject/circulation/pull/2448 the top level "containers" are not immutable and hashable, so they can be built up and modified in the importers.

## Motivation and Context

Follow up to: https://github.com/ThePalaceProject/circulation/pull/2448

Allow us to serialize `Metadata` and `CirculationData` objects, so that we can pass them into Celery tasks.

The biggest change that was necessary to convert these classes was to standardize how we are passing the `data_source` and `primary_identifier`. Previously we were not very strict about what went into these, and the constructor sorted it out. This updates the name of these fields to make it clear what they are, and adds some methods to load them from the database if necessary.

See: PP-2466

## How Has This Been Tested?

- Running unit tests

## Checklist


src/palace/manager/metadata_layer/container.py
```diff
@@ -0,0 +1,41 @@
...
+from pydantic import BaseModel, ConfigDict
+from sqlalchemy.orm import Session
+
+from palace.manager.metadata_layer.identifier import IdentifierData
+from palace.manager.sqlalchemy.model.datasource import DataSource
+from palace.manager.sqlalchemy.model.identifier import Identifier
+from palace.manager.util.log import LoggerMixin
+
+
+class BaseDataContainer(BaseModel, LoggerMixin):
```
jonathangreen, line 12: This is the mutable base model for `CirculationData` and `Metadata`. I had a bit of trouble coming up with a name for it, and I'm not 100% happy with `BaseDataContainer`, but its the best I could come up with.

Happy to change it if anyone has suggestions.

src/palace/manager/metadata_layer/container.py
```diff
@@ -0,0 +1,41 @@
...
+    def load_primary_identifier(self, _db: Session) -> Identifier:
+        """Find the Identifier associated with this circulation information."""
+        if self._primary_identifier is None:
+            if self.primary_identifier_data:
+                obj, ignore = self.primary_identifier_data.load(_db)
+                self._primary_identifier = obj
+                return obj
+            else:
+                raise ValueError("No primary identifier provided!")
+        return self._primary_identifier
```
jonathangreen, line 41: Renaming `data_source_name` and `primary_identifier_data` and making them consistent across `CirculationData` and `Metadata` is the bulk of the changes in this PR.

src/palace/manager/metadata_layer/container.py
```diff
@@ -0,0 +1,41 @@
...
+from palace.manager.metadata_layer.identifier import IdentifierData
+from palace.manager.sqlalchemy.model.datasource import DataSource
+from palace.manager.sqlalchemy.model.identifier import Identifier
+from palace.manager.util.log import LoggerMixin
+
+
+class BaseDataContainer(BaseModel, LoggerMixin):
+    model_config = ConfigDict(
+        frozen=False,
+        validate_assignment=True,
```
jonathangreen, line 15: We may eventually want to disable this, if its having negative performance impacts. Since it causes the validators to run every time the data in the container changes. 

This makes sure that some of the data manipulation that we do in the validators is always performed. So far the performance seems okay, but I wanted to note this.

src/palace/manager/metadata_layer/metadata.py
```diff
@@ -27,124 +30,76 @@
...
+    published: datetime.date | None = None
+    identifiers: list[IdentifierData] = Field(default_factory=list)
+    subjects: list[SubjectData] = Field(default_factory=list)
+    contributors: list[ContributorData] = Field(default_factory=list)
+    measurements: list[MeasurementData] = Field(default_factory=list)
+    links: list[LinkData] = Field(default_factory=list)
+    data_source_last_updated: datetime.datetime | None = None
+    duration: float | None = None
+    permanent_work_id: str | None = None
+    # Note: brought back to keep callers of bibliographic extraction process_one() methods simple.
```
dbernstein, line 79: This is what I was talking about above:  if there were a way to decouple CirculationData from the Metadata I _think_ (maybe?) it would be  easier to understand the model.

dbernstein (approved):
Nice surgery, Dr. Green. The patient's prognosis is much improved.

src/palace/manager/metadata_layer/container.py
jonathangreen: Those are all good suggestions @dbernstein. 

I agree untangling `CirculationData` from `Metadata`, would make things easier to reason about. I actually took a stab at doing that as part of this PR, but it will require some reasonably deep changes, so I ended up rolling it back and decided to defer it to another PR. Once this goes in, I'll see what I can do for a follow up to separate them.

I'll push a commit to this PR to rename `BaseDataContainer` -> `BaseMutableData` and `Metadata` -> `BibliographicData`.

src/palace/manager/metadata_layer/container.py
tdilauro: Should we put a note to that effect in the code itself?

src/palace/manager/metadata_layer/container.py
jonathangreen: I spent some time on this, and I pushed a commit to rename `BaseDataContainer` -> `BaseMutableData` but the `Metadata` rename is going to be much more work. I'm going to merge this one as is, and make a follow up PR with all the changes that are required for the metadata rename.

src/palace/manager/metadata_layer/container.py
jonathangreen: Done

src/palace/manager/metadata_layer/container.py
jonathangreen: Here's the follow up PR with the rest of the renaming if you want to take a look @dbernstein https://github.com/ThePalaceProject/circulation/pull/2463

PR #214, opened 2022-03-17
Feature/audit events update

RishiDiwanTT:
## Description

From ticket https://www.notion.so/lyrasis/Audit-Palace-Manager-events-ef18e581c66143a6b8326142c514309c
Removed Distributor events

## Checklist

- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

PR #2459, opened 2025-05-07
Clean up unnecessary analytics provider injections

jonathangreen:
## Description

Clean up some leftover dependency injections that are never actually used.

## Motivation and Context

I noticed this while working on: https://github.com/ThePalaceProject/circulation/pull/2458. Since these are unnecessary, we might as well remove them as it will speed up dependency injection.

Looks like this was leftover after https://github.com/ThePalaceProject/circulation/pull/214 went in.

## How Has This Been Tested?

- Running unit tests

## Checklist


tdilauro (approved):
🔥

PR #2461, opened 2025-05-08
Add pydantic serialization / deserialization to celery (kombu) (PP-2468)

jonathangreen:
## Description

This PR allows you to pass pydantic models into celery function calls, and have them automatically serialized and deserialized, so they are transparently passed into the function.

This PR also updates the marc exporter task to use this new functionality. Previously it was manually dumping the loading the pydantic models it was using.

## Motivation and Context

As part of PP-2468 we are going to be passing more pydantic models into celery tasks, so this is an effort to make that easier.

## How Has This Been Tested?

- Added unit test
- Converted marc exporter to use new functionality

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/marc.py
+++ b/src/palace/manager/celery/tasks/marc.py
@@ -237,6 +237,10 @@ def marc_export_collection(
             collection_name=collection_name,
             start_time=start_time,
             libraries=libraries,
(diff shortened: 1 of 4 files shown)


dbernstein (approved):
You are on a roll!  And thank you: this update will make my days with celery easier. : )

PR #2464, opened 2025-05-08
Remove celery `Job` class 🔥 

jonathangreen:
## Description

Remove the `Job` helper class from celery 🔥🔥 

## Motivation and Context

I added `Job` when doing the first celery tasks. I thought it would be a helpful abstraction, but as we've converted more tasks, we have moved away from using it. Of all our tasks only two were using `Job` and neither of them really benefited much from it.

So I removed it, and refactored those jobs to be task based, like the rest of our jobs are.

## How Has This Been Tested?

- Running unit tests locally

## Checklist


tdilauro (approved):
Looks good! 🔥

PR #2463, opened 2025-05-08
Rename metadata_layer.Metadata to data_layer.BibliographicData (PP-2468)

jonathangreen:
## Description

Rename the `metadata_layer` package to `data_layer` and rename `Metadata` -> `BibliographicData`.

## Motivation and Context

This PR is based on the code review discussion in this PR https://github.com/ThePalaceProject/circulation/pull/2458. 

This one is a big one. `Metadata` is referenced all over the code base, especially in some untyped parts, which makes renaming difficult. It is also referenced in function and variable names, and in comments pretty widely. I did my best to find all the references and clean them up as part of this PR. 

Some of this work is hard because we have `metadata` integrations, which do still need metadata references in their names. So its not as easy as doing a find and replace.

## How Has This Been Tested?

- Running unit tests

## Checklist

- [X] I have updated the documentation accordingly.
- [X] All new and existing tests passed.

PR #2465, opened 2025-05-09
Celery task to call BibliographicData.apply (PP-2468)

jonathangreen:
## Description

- Add a new queue `apply`
- Add a new worker to consume from that queue
- Add a new task `bibliographic_apply` that calls `BibliographicData.apply`

Note: Nothing is currently pushing data into this queue for the task, this is just getting the groundwork setup. I will modify one of our importers to push data into this queue in a follow up PR.

## Motivation and Context

JIRA: PP-2468

Getting tasks setup, so I can convert the OPDS importers to use Celery. 

This PR is based on https://github.com/ThePalaceProject/circulation/pull/2463, so that one should go in first.

## How Has This Been Tested?

- Added new tests for this case.

## Checklist


src/palace/manager/service/redis/models/lock.py
```diff
@@ -139,7 +140,7 @@ class RedisLock(BaseRedisLock):
     def __init__(
         self,
         redis_client: Redis,
-        lock_name: str | Sequence[str],
+        lock_name: str | Sequence[str | int | SupportsRedisKey],
```
jonathangreen, line 143: This was always supported, our type hint was just incomplete.

tests/fixtures/celery.py
```diff
@@ -77,6 +93,20 @@ class CeleryFixture:
     worker: WorkController
     session_maker: MockSessionMaker
 
+    @contextmanager
+    def patch_retry_backoff(self) -> Generator[CeleryRetriesMock, None, None]:
```
jonathangreen, line 97: I needed a way to test a celery task that used `autoretry_for` and `retry_backoff` without actually waiting for the backoffs to occur in the task. It seems like this is useful behaviour generally that we will want again, so I added it to the celery fixture, instead of just doing it in the test.

src/palace/manager/service/celery/celery.py
```diff
@@ -210,9 +226,9 @@ def task_queue_config() -> dict[str, Any]:
     """
     Configure the task queues for our Celery app.
 
-    Currently we have two queues, `high` and `default`. The `high` queue is for tasks that are
-    short running and should be processed quickly because they are time-sensitive. The `default`
-    queue is for tasks that are longer running and can be processed when the worker has capacity.
+    Currently we have three queues, `high`, `default` and `apply`.
+
```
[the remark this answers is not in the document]
jonathangreen: Good suggestion, done

PR #2468, opened 2025-05-13
Add celery tasks to mark unavailable identifiers (PP-2469)

jonathangreen:
## Description

Adds a series of tasks meant to provide the foundation for provider specific celery tasks that mark any identifiers not found in the providers feed as unavailable. The intention is to compose these tasks using celery workflows.

It also adds a new redis data structure to store sets of items, so we are able to store sets of identifiers that persist over mutliple celery task invocations.

Note: these tasks are not used anywhere yet. I'm planning to use them in a follow up to convert the OPDS importer tasks.

## Motivation and Context

See PP-2469. This is building up some of the tools needed to convert the OPDS importer tasks to Celery.

## How Has This Been Tested?

- New unit tests added to CI

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/identifiers.py
+++ b/src/palace/manager/celery/tasks/identifiers.py
@@ -23,13 +23,13 @@ from palace.manager.sqlalchemy.model.licensing import LicensePool
 @shared_task(queue=QueueNames.default, bind=True)
 def existing_available_identifiers(task: Task, collection_id: int) -> IdentifierSet:
     """
-    Get all the identifiers that have licensepools that are available (licenses_available > 0
-    and licenses_owned > 0) in the given collection and return them as a Redis IdentifierSet.
+    Retrieves all identifiers that have available licensepools (where licenses_available > 0 and
+    licenses_owned > 0) in the specified collection and returns them as a Redis IdentifierSet.
 
-    This is meant to be used as part of a chord that marks any identifiers not found in
-    a distributors feed as unavailable.
+    This function is designed to be used as part of a chord operation that identifies and
+    marks identifiers not present in a distributor's feed as unavailable.
 
-    See: mark_unavailable_chord.
+    See: `create_mark_unavailable_chord`.
     """
 
     redis_client = task.services.redis().client()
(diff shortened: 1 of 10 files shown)


src/palace/manager/data_layer/identifier.py
```diff
@@ -27,7 +28,14 @@ def from_identifier(cls, identifier: Identifier | IdentifierData) -> Self:
...
+        String representation of the IdentifierData object suitable for use
+        as a redis key.
+        """
         return (
             f"{self.__class__.__name__}{RedisKeyGenerator.SEPERATOR}"
             f"{self.type}{RedisKeyGenerator.SEPERATOR}{self.identifier}"
         )
+
+    def __str__(self) -> str:
+        return f"{self.type}/{self.identifier}"
```
jonathangreen, line 41: All the changes in this file are not necessary for this PR. I can pull them out of desired. They were just helpful to me when debugging, and I thought they were changes worth bringing into the codebase.

src/palace/manager/celery/tasks/identifiers.py
```diff
@@ -0,0 +1,159 @@
...
+
+@shared_task(queue=QueueNames.default, bind=True)
+def mark_identifiers_unavailable(
+    task: Task,
+    identifiers: list[RedisSetKwargs],
+    *,
+    collection_id: int,
+) -> None:
+    """
+    Takes a list of two RedisSetKwargs elements as the first positional argument. These are used to create
```
tdilauro, line 70: Just a question here: Is there any reason not to pass them separate sets? Seems like it would make it more clear at the call site with less chance of getting the order wrong.

src/palace/manager/celery/tasks/identifiers.py
```diff
@@ -0,0 +1,159 @@
...
+        identifier_set.delete()
+        raise
+
+    return identifier_set
+
+
+@shared_task(queue=QueueNames.default, bind=True)
+def mark_identifiers_unavailable(
+    task: Task,
+    identifiers: list[RedisSetKwargs],
```
tdilauro, line 65: Using "identifiers" here could be a little confusing, especially because we use `identifier` (singular) in this task.

Suggestion: Rename to make it more obvious that these aren't the identifier thingies that we're used to (`Identifier`s or `Identifier.id`s). You'll probably have a better idea of what to name this than I would.

src/palace/manager/celery/tasks/identifiers.py
```diff
@@ -0,0 +1,159 @@
...
+) -> Signature:
+    """
+    Creates a Celery chord that identifies and marks as unavailable any identifiers that were not
+    found in the distributor's feed for a given collection.
+
+    This chord performs the following sequence:
+    1. Calls the existing_available_identifiers task to retrieve all identifiers that are currently
+       available in the collection.
+    2. In parallel, executes the provided active_identifiers_sig task to retrieve identifiers that
+       are present in the distributor's feed.
```
tdilauro, line 142: Minor suggestion: Considering "numbering" these as "1.a." and "1.b", since they are meant to be running in parallel. Something like:

```
    1. Performs the following in parallel:
      a. Calls the existing_available_identifiers task to retrieve all identifiers that are currently
       available in the collection.
      b. Executes the provided active_identifiers_sig task to retrieve identifiers that
       are present in the distributor's feed.
    2. Finally, ...
```

tdilauro (approved):
Looks good!

A few questions/suggestions below.

src/palace/manager/celery/tasks/identifiers.py
jonathangreen: Since this is meant to be called via a Celery chord, we don't have control over how these args get passed to this task. Celery always passes the results of the chord header tasks as the first positional arg to the function in a list. It would be nice if we had more control over these, because I agree it would be nice to pass them separately.

src/palace/manager/celery/tasks/identifiers.py
jonathangreen: I renamed them to `existing_and_active_identifier_sets`

src/palace/manager/celery/tasks/identifiers.py
jonathangreen: Updated

PR #2469, opened 2025-05-13
Potential optimization for `Library.active_collections`. (PP-2492)

tdilauro:
## Description

Caches the result of `Library.active_collections` and adds event listeners to conservatively clear the cache.

## Motivation and Context

Needed to improve the performance of `LibraryAnnotator.active_licensepool_for`, which, according profiling, accounted for about a fifth of the response time. This turned out to be mostly tied to the performance of `Library.active_collections` and its underlying query.

[Jira PP-2492]

## How Has This Been Tested?

- All existing tests pass.
- Performance profiling to demonstrate that neither `LibraryAnnotator.active_licensepool_for` nor `Library.active_collections` accounts for 
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/15000149186).

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/model/__init__.py
+++ b/src/palace/manager/sqlalchemy/model/__init__.py
@@ -6,12 +6,6 @@ This is necessary to make sure that all of our models are properly reflected in
 the database when we run migrations or create a new database.
 """
 
-from typing import Any
-
-from sqlalchemy import event
-from sqlalchemy.engine import Connection
-from sqlalchemy.orm import Mapper
-
 import palace.manager.sqlalchemy.model.admin
 import palace.manager.sqlalchemy.model.announcements
 import palace.manager.sqlalchemy.model.base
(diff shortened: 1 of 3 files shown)


src/palace/manager/sqlalchemy/model/__init__.py
```diff
@@ -33,3 +39,29 @@
...
+
+
+@event.listens_for(IntegrationConfiguration, "after_insert")
+@event.listens_for(IntegrationConfiguration, "after_delete")
+@event.listens_for(IntegrationConfiguration, "after_update")
+@event.listens_for(IntegrationLibraryConfiguration, "after_insert")
+@event.listens_for(IntegrationLibraryConfiguration, "after_delete")
+@event.listens_for(IntegrationLibraryConfiguration, "after_update")
+def handle_collection_change(_: Mapper, _connection: Connection, target: Base) -> None:
+    Library.clear_active_collections_cache(target)
```
jonathangreen, line 60: Minor: Is it possible to move these to `listeners.py` with the rest of the event listeners?

src/palace/manager/sqlalchemy/model/library.py
```diff
@@ -206,9 +206,26 @@ def associated_collections(self) -> Sequence[Collection]:
 
     @property
     def active_collections(self) -> Sequence[Collection]:
```
jonathangreen, line 208: Minor: It might make sense to try to use a relationship instead of a property for this. I think that would let sqlalchemy handle the caching for us. 

Something like this: 
https://github.com/ThePalaceProject/circulation/blob/81967eba87b897287fee14e87e7b4f474f625a91/src/palace/manager/sqlalchemy/model/licensing.py#L318-L333

jonathangreen (approved):
Looks good @tdilauro! Thanks for tracking this down! 

Made a couple minor comments with some questions / ideas but also feel free to merge as is.

src/palace/manager/sqlalchemy/model/library.py
tdilauro: It doesn't look like we can do this right now because of how the collection subscriptions filters are currently structured. I do think it could be done if we switch to db date functions. I took a quick stab at it, but ran into an import cycle. I'm going to back off on this for now, since I want to continue to focus on the performance issues. I'll make a ticket to try this out in the near future.

src/palace/manager/sqlalchemy/model/__init__.py
tdilauro: Good suggestion. Thanks! Done.

PR #2472, opened 2025-05-14
Update our calls to Hold.collect_event_and_delete (PP-2527)

jonathangreen:
## Description

Update `Hold.collect_event_and_delete` `analytics` parameter to be an optional keyword only parameter and update all calls to it to reflect this.

## Motivation and Context

When you use dependency injector `Provide`, the parameter is always supplied as a keyword argument. So supplying it as a positional argument will result in:
> TypeError: Hold.collect_event_and_delete() got multiple values for argument 'analytics'

We were seeing this error happening in production due to the call in `palace/manager/api/circulation.py`

## How Has This Been Tested?

- Updated tests
- Running unit tests

## Checklist


dbernstein (approved):
🚀

PR #2473, opened 2025-05-14
Mark all dependency injected arguments as keyword only

jonathangreen:
## Description

Update all the arguments that we use dependency injection for, to make it clear that they should only be specified as keyword arguments.

## Motivation and Context

This is very similar to https://github.com/ThePalaceProject/circulation/pull/2472, but it updates all the arguments that we use dependency injection for so we don't run into the same bug we did in PP-2527.

## How Has This Been Tested?

- Running unit tests

## Checklist


tdilauro (approved):
🔧 ✅

PR #2474, opened 2025-05-14
Pytest pyinstrument fixture

jonathangreen:
## Description

A small fixture useful when profiling test code.

## Motivation and Context

I was trying to figure out why a test was running so slowly and wrote this fixture based on the example given in the pyinstrument documentation https://pyinstrument.readthedocs.io/en/latest/guide.html#profile-pytest-tests

Seems like it is a useful fixture to have in our toolbox, so I figured I'd make a PR for it.

## How Has This Been Tested?

- Used the fixture locally when profiling tests

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ nosetests.xml
 coverage.xml
 coverage.json
 
+# Testing profile data
+tests/profiles/
+
 # Translations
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
🎸

PR #2475, opened 2025-05-14
[PP-1792] add support for collection depriorization

dbernstein:
## Description

What I've done here is the following:
1. I added a configuration parameter for collections that allows the user to specify the "lane_priority_level" which is on a scale of 1 to 10. 
2. I updated the search index to include the licensepools.lane_priority_level field.
3. I added `field_value_factor` to the featurability_scoring_functions which simply adds the lane_priority_level to the score.
4. I added testing.

Note:  the `field_value_factor` allows for both a `modifier` property as well as a `factor`.  The modifier is one of a number of math functions you can apply to the value to manipulate the magnitude of the score.  So  I opted for no modifier and no factor.  To the extent that  a user  deviates from the middle of the range (ie 5), so she will give or take weight from the items in the collection.  The far end of the ranges the user can effectively trump the quality scores.  When the differences between colleciton priority levels are lower, the priority won't necessarily trump the quality scores or the effects of the randomizing function.

## Motivation and Context
Sometimes content from one collection can swamp a lane despite their being other collections matching the lane criteria. This feature enables users to specify a priority for the collection that applies to works that are featured in lanes.   For example a librarian may find that they the lane is only displaying books from an open access collection but s/he wants to give priority to the licensed content that is featured in the lane.  In this case, the user would increase or reduce collection lane priorities according to the desired outcome.

https://ebce-lyrasis.atlassian.net/browse/PP-1792

## How Has This Been Tested?
Manually and unit tested.

For manual testing you must do the following:
Starting conditions:
Have at least two collections. 
For a given lane, verify that works from both collections appear in the lane.

1. change the collection lane priority levels according to your tests.
2. reindex the collections
3. validate that the ordering of the lane is what you would expect (ie the collections with the highest priority appear first in the list and those with the lowest appear at the end.  If the spread between lowest and highest is minimal - say you have a 6 and 4, you may not see all the 6's coming before the 4's.   Quality may be a factor.   The wider the spread, the less quality will factor in.

## Checklist

### DIFF
--- a/src/palace/manager/api/circulation.py
+++ b/src/palace/manager/api/circulation.py
@@ -530,8 +530,8 @@ class BaseCirculationApiSettings(BaseSettings):
                 "other collections will push the contents of this collection to the bottom of any lane in which this "
                 "collection's contents appear. The default value "
                 f"({IntegrationConfigurationConstants.DEFAULT_LANE_PRIORITY_LEVEL}) "
-                "will be used, if no value is provided here. If two or more collections contain overlapping sets of "
-                "books, the highest of the lane priorities when ordering a book's position in a lane."
+                "will be used if no value is provided here. If two or more collections contain overlapping sets of "
+                "books, the highest of the lane priorities will be used when ordering a book's position in a lane."
             ),
             required=False,
         ),


src/palace/manager/search/external_search.py
```diff
@@ -2104,7 +2105,20 @@ def featurability_scoring_functions(self, random_seed):
...
+            field="licensepools.lane_priority_level",
+            factor=10,
+            modifier="none",
+            missing=IntegrationConfigurationConstants.DEFAULT_LANE_PRIORITY_LEVEL,  # assume default if missing
+        )
+
+        function_scores = [
+            quality_field,
+            available_now,
+            lane_priority_level,
```
dbernstein, line 2120: If I remove the lane_priority_level function from the function_scores list, I get the same results.

src/palace/manager/search/v5.py
```diff
@@ -266,6 +266,7 @@ def __init__(self) -> None:
         licensepools.add_property("suppressed", BOOLEAN)
         licensepools.add_property("licensed", BOOLEAN)
         licensepools.add_property("medium", keyword())
+        licensepools.add_property("lane_priority_level", INTEGER)
```
dbernstein, line 269: For my final draft,  what is the recommended way to add a new revision of the search index schema?

src/palace/manager/sqlalchemy/model/work.py
```diff
@@ -1707,6 +1710,13 @@ def _set_value(parent, key, target):
                     lc["medium"] = doc.presentation_edition.medium
                 lc["licensepool_id"] = license_pool.id
                 lc["quality"] = doc.quality
+                collection_settings = (
+                    license_pool.collection.integration_configuration.settings_dict
+                )
+                lc["lane_priority_level"] = collection_settings.get(
```
dbernstein, line 1716: I verified that the correct value is definitely coming in from the collection settings.

src/palace/manager/search/external_search.py
```diff
@@ -2104,7 +2105,20 @@ def featurability_scoring_functions(self, random_seed):
         nested = Nested(path="licensepools", query=available)
         available_now = dict(filter=nested, weight=5)
 
-        function_scores = [quality_field, available_now]
+        # Works with higher lane priority scores are more featurable
+        lane_priority_level = SF(
+            "field_value_factor",
+            field="licensepools.lane_priority_level",
+            factor=10,
```
dbernstein, line 2112: I add the factor of 10 here to ensure that the lane priority is the first consideration.

src/palace/manager/search/external_search.py
dbernstein: Even removing the quality_field and available_now functions, I still get the same results.

src/palace/manager/search/v5.py
dbernstein: Should I copy v5.py -> v6.py,  update the revision number and make my changes?

io7m: Yeah, that was the intention. Adding a new `.py` file for each new version means that you can go back to older schema versions in the case of some sort of disastrous design failure. :slightly_smiling_face:

tdilauro:
@dbernstein Will these changes affect search, as well? Or just browse?

dbernstein:
@tdilauro :  I believe it will affect browse only.

src/palace/manager/search/external_search.py
tdilauro: Based on our discussion the other day, I'm assuming that this conversation is resolved.

tests/manager/search/test_external_search.py
```diff
@@ -1559,6 +1560,78 @@ def assert_order(sort_field, order, **filter_kwargs):
...
+        assert data.e.license_pools[0].collection
+        # collection 1 has the highest priority
+        data.collection1._set_settings(lane_priority_level=10)
+        data.collection3._set_settings(lane_priority_level=1)
+
+        fixture.populate_search_index()
+        facets = FeaturedFacets(minimum_featured_quality=0, entrypoint_is_default=True)
+
+        # use deterministic ordering
+        facets.random_seed = Filter.DETERMINISTIC
```
tdilauro, line 1604: Do we need to use deterministic ordering here? I thought the lane priority level is supposed to overwhelm any randomness (although, I do note that the factor is down from the original 10 to just 1.

Suggestion: It seems like we want to demonstrate that the randomness will be overwhelmed.

src/palace/manager/sqlalchemy/model/work.py
```diff
@@ -1707,7 +1710,18 @@ def _set_value(parent, key, target):
...
+                    "lane_priority_level",
+                    IntegrationConfigurationConstants.DEFAULT_LANE_PRIORITY_LEVEL,
+                )
                 result["licensepools"].append(lc)
+            # use the maximum lane priority level associated with the work.
+            result["lane_priority_level"] = max(
+                [lc["lane_priority_level"] for lc in result["licensepools"]]
+            )
```
tdilauro, line 1724: Change requested: I might've missed something, but I didn't see anything testing this max lane priority level part. We could test that by adding a book to `test_lane_priority_level_ordering` that is in both collection 1 and collection 3 and making sure that the book appears in the priority collection in both scenarios.

tdilauro (approved):
This looks pretty good. 🎉 

I'm approving this now, in case you want to get this finished up before next week; but, I've left a few suggestions/change requests below.

tdilauro:
@dbernstein One other thing I forgot to mention: It would be good to put more information than just the Jira ticket link in the Motivation and Context section, since this is an open source project and not everyone will have access to our Jira instance. Also, it makes it easier to understand the reason for a change when looking back through close PRs.

tests/manager/search/test_external_search.py
dbernstein: I suppose that you're right.  I believe I can remove the determinism without a problem.

src/palace/manager/sqlalchemy/model/work.py
dbernstein: Excellent catch - I'll add a test for that.

dbernstein:
@tdilauro :  I made those changes and also updated the lane priority field description to include the use of the max lane priority level when books appear in more than one collection.

src/palace/manager/api/circulation.py
```diff
@@ -530,7 +530,8 @@ class BaseCirculationApiSettings(BaseSettings):
                 "other collections will push the contents of this collection to the bottom of any lane in which this "
                 "collection's contents appear. The default value "
                 f"({IntegrationConfigurationConstants.DEFAULT_LANE_PRIORITY_LEVEL}) "
-                "will be used, if no value is provided here."
+                "will be used, if no value is provided here. If two or more collections contain overlapping sets of "
+                "books, the highest of the lane priorities when ordering a book's position in a lane."
```
tdilauro, line 534: Looks like you might've omitted a word or two in the last sentence here. Maybe you wanted something like "...the highest of the lane priorities [will be used] when ordering..."

PR #2492, opened 2025-05-21
Create session fixture for services container (PP-2569)

jonathangreen:
## Description

- Create a session fixture that contains the autospec mocks for our services container, so we don't have to run `create_autospec` on every test run.
- Refactor our services fixture, so it uses the new session fixture. 
- Remove a number of fixtures that were created for the sub-containers that weren't very useful, and replace them with calls to the main services fixture.

## Motivation and Context

`create_autospec` is surprisingly expensive to run, and with our current services container, we have to run several times for almost every test run. This ends up making out tests take longer.

I think the new structure where we don't have a bunch of different services fixtures is a better design and it shaves a minute or so off our tests runs.

## How Has This Been Tested?

- Tested locally
- Tested in CI

## Checklist

### DIFF
--- a/tests/fixtures/celery.py
+++ b/tests/fixtures/celery.py
@@ -119,7 +119,7 @@ def celery_fixture(
     """Fixture to provide a Celery app and worker for testing."""
 
     # Make sure our services container has the correct celery app setup
-    container = services_fixture.celery_fixture.celery_container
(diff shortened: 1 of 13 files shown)


tdilauro (approved):
Smaller, simpler, and faster FTW! 🥇

PR #2493, opened 2025-05-22
Update get_one to not require a second trip to database for on_multiple=interchangeable

jonathangreen:
## Description

When we call `get_one` with `on_multiple="interchangeable"` we have to make a second call to the database if we get multiple row. Update the call to set a limit any time we have `interchangeable` results, to save an extra trip to the database.

## Motivation and Context

I noticed this while working on the other test performance improvements. I don't think this will provide any measurable gain in speed, but it does impact production since its in `get_one` and its nice to save the database roundtrip when we can.

## How Has This Been Tested?

- Running unit tests.

## Checklist


tdilauro (approved):
Great catch! And a very elegant solution. 🚀

PR #2519, opened 2025-06-12
[PP-1469] convert work classify unchecked to celery

dbernstein:
## Description
This PR refactors the work classification for unchecked subjects script's implementation into the celery framework, making use of the redis set for storing work_id and calculation policy tuples for later processing.   The celery task scheduler kicks off the unchecked subject processing once a day as before.   I've also preserved the ability to manually kick off the task via a script. 

## Motivation and Context
This PR converts the work classification for unchecked subjects script to use the celery task framework.

https://ebce-lyrasis.atlassian.net/browse/PP-1469

## Checklist


jonathangreen:
> @tdilauro or @jonathangreen : I'm not sure how to fix this mypy issue. Per the comment, just adding the Work type fixes mypy but causes the test to fail. It's not clear to me why it doesn't work. Any suggestions would be appreciated.

Type annotations with sqlalchemy are tricky since we are on version 1.4, and type support was much more experimental then. I think it'll be easier when we move to 2.x eventually, but to do that we need to address all our use of deprecated sqlalchemy features. So you may need the type ignores. 

That said, you will get further if you add:
```python
from __future__ import annotations
```

To the top of the file you are having trouble with. That will setup python so it doesn't evaluate type annotations at runtime for that file. That was supposed to become the default in newer versions of Python, and we import it in a number of our files already. 

After you add that `Query[Work]` will work as an annotation for by mypy and at runtime. Not sure if that will solve all the type issues here or not, but its worth trying.

dbernstein:
The reason I put DB Migration on there is because there is a migration of data from the workcoveragerecords table to the celery queue.  It is just happening outside of the alembic system.  It's true that there is no structural change to the database.  Nevertheless I think a major version change is warranted.  I'll defer to your judgement on the final call.

PR #2525, opened 2025-06-16
Refactor MockHttpClient to be pytest fixture

jonathangreen:
## Description

Refactor MockHttpClient to be pytest fixture. Previously it was a standalone class, but it makes more sense for it to be a fixture that other test functions can call in.

## Motivation and Context

This change was part of the work I'm doing in PP-2133. Breaking it out to its own PR, since PP-2133 is getting rather large.

## How Has This Been Tested?

- Running unit tests

## Checklist


dbernstein (approved):
Looks good!

PR #2530, opened 2025-06-16
Refactor axis 360 API response parsers to use Pydantic (PP-2133)

jonathangreen:
## Description

Refactor Axis 360 (Boundless) importers to use pydantic models to parse the API responses and refactor the code so that these models are used.

## Motivation and Context

I was having issues modifying the importer in a reasonable way to implement the new fulfillment workflow, because the API response parsers also made API requests. 

The fulfillment workflow previously had a lot of indirection. The response went into the `AvailabilityResponseParser` which returned a `AxisLoanInfo` with a `Axis360Fulfillment` attached. When this fulfillment was used, it made an additional API request, which was parsed via `Axis360FulfillmentInfoResponseParser` which then depending on the type of fulfillment made an additional API request, parsed that, and then turned it into a manifest to return.

Now all the fulfillment logic lives in a function, so its easier to reason about and modify. This means we don't need several classes that existed to override the default fulfillment logic like `AxisLoanInfo` and `Axis360Fulfillment`. 

## How Has This Been Tested?

- [X] Running unit tests in CI
- [X] Local testing with API

~~Note: I still need to do a bit of local testing with the real API to make sure everything looks good. I'm going to work on this today, but I don't expect much to change, so I think its fine to code review while I'm doing the local testing. I'll update this once that testing is complete.~~ This is complete.

## Checklist


src/palace/manager/api/axis/parser.py
```diff
@@ -52,155 +18,42 @@
...
-        available_copies = self.int_of_subtag(availability, "axis:availableCopies", ns)
-        size_of_hold_queue = self.int_of_subtag(availability, "axis:holdsQueueSize", ns)
 
-        circulation_data.licenses_owned = total_copies
-        circulation_data.licenses_available = available_copies
-        circulation_data.licenses_reserved = 0
-        circulation_data.patrons_in_hold_queue = size_of_hold_queue
-
-        return circulation_data
+class BibliographicParser(LoggerMixin):
```
jonathangreen, line 26: This takes the parsed Availability response and turns it into `BibliographicData` and `CirculationData` tuples used by the CM.

src/palace/manager/celery/tasks/axis.py
```diff
@@ -231,7 +229,6 @@ def import_identifiers(
     transaction.  If it has not finished processing the list of identifiers, it will requeue the task with the
     remaining unprocessed identifiers.
     """
-    count = 0
```
jonathangreen, line 234: Unused variable, removed it while I was here.

tests/files/axis/availability_with_checkouts_and_holds.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jonathangreen, line 1: None of our existing fixture files had checkouts and holds in them, so I added one from my testing.

tests/files/axis/availability_with_ebook_fulfillment.xml
```diff
@@ -1 +1 @@
...
...ies>1</availableCopies><totalCopies>2</totalCopies><Checkouts/><Holds/><holdsQueueSize>0</holdsQueueSize><isInHoldQueue>false</isInHoldQueue><estimatedHoldPeriod>0</estimatedHoldPeriod><isReserved>false</isReserved><isCheckedout>true</isCheckedout><checkoutFormat>ePub</checkoutFormat><transactionID>6670197A-D264-447A-86C7-E4CB829C0236</transactionID><downloadUrl>http://adobe.acsm/?src=library&amp;transactionId=2a34598b-12af-41e4-a926-af5e42da7fe5&amp;isbn=9780763654573&amp;format=F2</downloadUrl><checkoutStartDate>6/24/2020 2:34:33 PM +00:00</checkoutStartDate><checkoutEndDate>7/15/2020 2:34:00 PM +00:00</checkoutEndDate><updateDate>6/24/2020 2:40:03 PM +00:00</updateDate></availability><minLoanPeriod>3</minLoanPeriod><maxLoanPeriod>21</maxLoanPeriod><defaultLoanPeriod>21</defaultLoanPeriod><authors><author>Koehler, Jeff</author></authors><narrators/><formatType>EBT</formatType><imageUrl>http://contentcafecloud.baker-taylor.com/Jacket.svc/D65D0665-050A-487B-9908-16E6D8FF5C3E/9781620405147/Medium/Logo</imageUrl><status>OK</status><RTV>true</RTV></title><status><code>0000</code><statusMessage>Availability Data is Successfully retrieved.</statusMessage></status></availabilityResponse>
```
jonathangreen, line 1: I had to update this fixture file slightly. Either the API changed over time, or someone build this example themselves. The download URL here is an ACSM file, but the `checkoutFormat` indicated this is a `Blio` checkout.

This would cause our new code to raise an exception, as a `Blio` checkout format, isn't what we request for Adobe. I validated this against the real API. So the change here (its hard to see because this is all one line) is to update this to `<checkoutFormat>ePub</checkoutFormat>` which is what the API actually responds with.

tests/files/axis/availability_with_loan_and_hold.xml
```diff
@@ -1 +1 @@
...
...sCheckedout><checkoutFormat>ePub</checkoutFormat><checkoutStartDate>7/22/2015 5:40:27 PM +00:00</checkoutStartDate><checkoutEndDate>8/12/2015 5:40:27 PM +00:00</checkoutEndDate><downloadUrl>http://fulfillment/</downloadUrl><updateDate>7/8/2015 6:16:45 AM</updateDate></availability><minLoanPeriod>3</minLoanPeriod><maxLoanPeriod>21</maxLoanPeriod><defaultLoanPeriod>21</defaultLoanPeriod><RTV>true</RTV></title><title xsi:type="Axis360AvailCirculation"><titleId>1111111111</titleId><productTitle>Test</productTitle><isbn>12345</isbn><language>FRENCH</language><availability><availableCopies>0</availableCopies><totalCopies>2</totalCopies><Checkouts/><Holds/><holdsQueueSize>1</holdsQueueSize><holdsQueuePosition>1</holdsQueuePosition><isInHoldQueue>false</isInHoldQueue><isReserved>true</isReserved><reservedEndDate>1/1/2015 1:11:11 PM +00:00</reservedEndDate><isCheckedout>false</isCheckedout><updateDate>7/22/2015 4:05:11 PM</updateDate></availability><minLoanPeriod>3</minLoanPeriod><maxLoanPeriod>21</maxLoanPeriod><defaultLoanPeriod>21</defaultLoanPeriod></title><status><code>0000</code><statusMessage>Availability Data is Successfully retrieved.</statusMessage></status></availabilityResponse>
```
jonathangreen, line 1: This is another out of date or contrived example. The availability information for titleId 1111111111 was missing. This made the file fail validation. Looking at real API responses and the docs, this information should not be missing. I so I modified the example to add in the required availability info.

tests/manager/api/axis/test_api.py
```diff
@@ -362,118 +318,320 @@ def test_fulfill(self, axis360: Axis360Fixture):
 
         # If Axis 360 says a patron does not have a title checked out,
         # an attempt to fulfill that title will fail with NoActiveLoan.
-        data = axis360.sample_data("availability_with_audiobook_fulfillment.xml")
-        axis360.api.queue_response(200, content=data)
-        pytest.raises(NoActiveLoan, fulfill)
+        data = axis360.files.sample_data("availability_with_ebook_fulfillment.xml")
```
jonathangreen, line 321: Added better coverage for our fulfillment operations. Some of this comes from our old tests for parsers and fulfillment, and some of it is new tests.

PR #2543, opened 2025-06-23
Cleanup migration Axis 360 settings (PP-2133)

jonathangreen:
## Description

Removes the temporary validator in the `Axis360Settings` class and adds an alembic migration to set the new `server_nickname` key and remove the old `url` key.

## Motivation and Context

This is ready to go in now, but will need to be sequenced with the other migration waiting to go in: 
https://github.com/ThePalaceProject/circulation/pull/2513

Follow up migration for this PR: https://github.com/ThePalaceProject/circulation/pull/2540

## How Has This Been Tested?

- New unit test
- Ran some queries against our databases to make sure it covers all the cases I see there

## Checklist


dbernstein (approved):
Looks good!

PR #2549, opened 2025-06-23
[PP-2639] Add a new report to the inventory and holds reports zip tha…

dbernstein:
…t contains a list of holds with no remaining licenses.

## Description
This PR adds a a report to the inventory and holds report zip that is currently being generated.  The report contains holds that are associated with license pools without any owned copies and therefore will never be convertible to loans.

## Motivation and Context
We need to do the backend work to generate a report for libraries that lists all titles with holds but no remaining licenses. This report can be modelled after the current holds report.
https://ebce-lyrasis.atlassian.net/browse/PP-2639

## How Has This Been Tested?

I updated the unit tests which now cover the new functionality.  I will validate on minotaur once it is approved and merged.
## Checklist


src/palace/manager/celery/tasks/generate_inventory_and_hold_reports.py
```diff
@@ -104,10 +104,12 @@ def generate_report(
         )
     ]
 
+    MAX_LICENSE_COUNT = 1000 * 1000 * 1000 * 1000
```
dbernstein, line 107: An arbitrarily large number to represent the max license count.

tdilauro (approved):
I added a few suggestions and comments, but no showstoppers. So, this could go in as-is, if you don't want to make any changes.

PR #2553, opened 2025-06-25
Deprecate Axis 360 Datasource (PP-1911)

jonathangreen:
## Description

Update how deprecated data sources are handled:
- If a datasource using the old name exists, use it
- If it doesn't, create one using the new name

This will let us deprecate a datasource by:
1. Update `DEPRECATED_NAMES` to include the deprecated name
2. Put this into a release and roll it out everywhere
3. In a follow up, once we are running a version of the code the case deal with both the new and old names everywhere, do a DB update that updates the old name to the new name. 
4. Remove the deprecated name from `DEPRECATED_NAMES`.

This also updates the user facing references to the datasource name to display the new name, rather than the deprecated name.

This PR implements step 1 above. I will make a follow up PR to this one that implements step 3 & 4, which can go in once this PR is in a release.

## Motivation and Context

See: PP-1911. We want to display `Boundless` instead of `Axis 360` in the apps.

## How Has This Been Tested?

- New unit tests
- Running old unit tests
- Light local testing
- Grepping for places that we are directly creating / loading `DataSource` instead of using the `lookup` method.

## Checklist

### DIFF
--- a/tests/manager/sqlalchemy/model/test_lane.py
+++ b/tests/manager/sqlalchemy/model/test_lane.py
@@ -3,9 +3,10 @@ import logging
 import random
 from dataclasses import dataclass, fields
 from typing import cast
-from unittest.mock import MagicMock, call
+from unittest.mock import MagicMock, call, patch
 
 import pytest
+from bidict import frozenbidict
 from opensearchpy.exceptions import OpenSearchException
 from sqlalchemy import and_, text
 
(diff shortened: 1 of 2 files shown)


src/palace/manager/sqlalchemy/util.py
```diff
@@ -138,7 +140,16 @@ def get_one_or_create(
...
+            one = get_one(db, model, **kwargs)
+            if one is None:
+                # If we still don't have an object, raise the error.
+                log.error(
+                    "Unable to retrieve with get_one after IntegrityError with %r %r",
+                    model,
+                    kwargs,
+                )
+                raise
+            return one, False
```
jonathangreen, line 152: These changes are not directly related, but they fix a bug in the existing code that the changes in this PR surfaced. In the case where you are passing `constraint` to `get_one_or_create`, if there is an `IntegrityError` previously we would ignore the `constraint` when doing the second query. 

We likely wouldn't hit this often / ever in the real world, but good to handle that case, since i noticed it as part of these changes.

PR #2556, opened 2025-06-25
Fix issue with docker/ci/test_webapp.sh intermittantly failing

jonathangreen:
## Description

Recently in CI I've been seeing a lot of failures due to `docker/ci/test_webapp.sh`. It looks like this is because the tests are running before the container has fully started. Add a test to make sure the file exists before running the test. This will timeout after 15s, so if something is stopping the container from starting fully it shouldn't freeze.

## Motivation and Context

Try to reduce test failures.

Here is an example of a failure I hope this will prevent: https://github.com/ThePalaceProject/circulation/actions/runs/15880645569/job/44779939474

## How Has This Been Tested?

- Running locally

## Checklist

### DIFF
--- a/docker/ci/check_service_status.sh
+++ b/docker/ci/check_service_status.sh
@@ -16,6 +16,14 @@ function check_service_status()
   # The location of the runit service should be passed.
   service="$2"
 
+  # Wait up to 15 seconds for the service file to exist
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
⏲️

PR #2557, opened 2025-06-25
Rename Axis 360 references in our code base to Boundless

jonathangreen:
## Checklist


tdilauro (approved):
This looks good.

I couldn't find replacements for `bin/informational/axis-raw-bibliographic` or `bin/informational/axis-raw-patron-status` and didn't see anything about them in the main PR comment, so just want to confirm that you intended to remove them. If so, it be good to mention that they are being removed and why.

jonathangreen:
@tdilauro yes, I meant to remove them, thanks for the reminder. I forgot about that. 

I removed them because they are broken in `main` right now, and they don't have any tests, so no one noticed they were broken. We have similar functionality in palace-tools, so I figured I'd just remove them here.

PR #2558, opened 2025-06-25
[PP-2676] Eliminate deadlocks during presentation calculation by limi…

dbernstein:
…ting work updates to one per transaction.

## Description
This update includes a new test to verify that unprocessed work_id/policy objects are put back in the redis set in the case of failures.

## Motivation and Context
We're seeing deadlocks in the calculate_presentation task.
https://ebce-lyrasis.atlassian.net/browse/PP-2676

## How Has This Been Tested?
A new test was added to reflect changes. 
We likely won't know if this update solves the problem until we deploy to production.

## Checklist


src/palace/manager/celery/tasks/work.py
```diff
@@ -32,10 +32,9 @@ def calculate_work_presentations(
         work_policies = waiting.pop(batch_size)
 
         if work_policies:
-
             try:
+                remaining_work_policies = work_policies.copy()
```
jonathangreen, line 36: Minor: Given that we are processing one per session now, we lose the advantage of batching. I'm not sure there is much reason to batching these in a Redis set. Doesn't have to be part of this PR, but seems like we might want to just process them in a task one at a time.

src/palace/manager/celery/tasks/work.py
```diff
@@ -44,16 +43,24 @@ def calculate_work_presentations(
...
             except Exception as e:
                 # if a failure occurs requeue the items so that can be recalculated in the next round
-                waiting.add(*work_policies)
+                task.log.error(
+                    f"Failed to calculate_presentation for {wp.work_id}. "
+                    f"Re-queuing remaining {len(remaining_work_policies)} of {len(work_policies)} "
+                    f"for future processing.",
+                    exc_info=e,
+                )
+                waiting.add(*remaining_work_policies)
```
jonathangreen, line 63: Minor: You may want to change this block to look like this:

```python
  except:
      # if a failure occurs requeue the items so that can be recalculated in the next round
      task.log.exception(
          f"Failed to calculate_presentation for {wp.work_id}. "
          f"Re-queuing remaining {len(remaining_work_policies)} of {len(work_policies)} "
          f"for future processing.",
      )
      waiting.add(*remaining_work_policies)
      raise
```

By using `raise` instead of `raise e` the exception will be linked to the line where it originally occurred, instead of being raised from this line.

jonathangreen, line 63: Don't have to deal with this in this PR, but I am a bit worried with the structure here that an exception that isn't easily handled by a retry will just cause an infinite loop.

Like if something in `calculate_presentation` raised a `ValueError`. That same work would get added back into the set, and just raise a `ValueError` again next time the task runs. If we're going to re-add tasks to the set like this, we should probably fail them after a certain number of retries.

With the structure here, a retry count is a bit hard to implement, but if we process one work per task, it would be fairly easy to use the celery max retry functionality.

jonathangreen (approved):

dbernstein:
@jonathangreen : I'll fix the issue with `raise` as you suggested and merge it in the interest in trying to resolve the deadlocks asap. Then I'll create a follow-on ticket to remove the batching.

PR #2560, opened 2025-06-26
Update ability to deprecate identifiers (PP-1911)

jonathangreen:
## Description

This PR updates how we deprecate `Identifier` objects to match how we deprecate `DataSource` objects in https://github.com/ThePalaceProject/circulation/pull/2553.

## Motivation and Context

Did this as part of the work on PP-1911. 

This PR puts everything in place, but DOES NOT deprecate `AXIS_360_ID` because I'm concerned it will cause some issues in the apps with identifiers changing. I'm going to bring this up in the tech strategy meeting today. 

Even if we don't immediately end up using this code, I think its still worth merging in, since it:
- Aligns the deprecation strategy with `DataSource`
- Centralizes how we load identifiers passed into controllers

## How Has This Been Tested?

- Testing locally
- New unit tests

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/model/identifier.py
+++ b/src/palace/manager/sqlalchemy/model/identifier.py
@@ -9,7 +9,7 @@ from abc import ABC, abstractmethod
 from collections import defaultdict
 from collections.abc import Iterable, Mapping
 from functools import total_ordering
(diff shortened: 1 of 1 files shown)


dbernstein (approved):
Bravo 👏

PR #2526, opened 2025-06-16
Remove superseded license pool concept. (PP-2590)

tdilauro:
## Description

Removes the concept of a superseded ("superceded" [sic] in code) license pool.

NB: The `LicensePool.superceded` column will be removed in a future update, but all references to it have been removed.

## Motivation and Context

Before this change, Open Access license pools could be superseded by another license pool. This caused problems if the _superseding_ license pool was not associated with and active in all libraries that were associated with the _superseded_ pool.

## How Has This Been Tested?

- Removed or reworked tests that used superseded.
- All tests pass locally.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/15682544711).

## Checklist

- N/A - I have updated the documentation accordingly.
- [X] All new and existing tests passed.

PR #2563, opened 2025-06-26
Clean up "superceded" column from license pools table. (PP-2669)

tdilauro:
## Description

Removes "superceded" column from `LicensePool` and adds an alembic migration following deployment of #2526.

## Motivation and Context

[Jira PP-2669]

## How Has This Been Tested?

CI tests.

## Checklist

### DIFF
--- a/alembic/versions/20250626_bc471d8a83fb_remove_superseded_concept_from_license_pools.py
+++ b/alembic/versions/20250626_bc471d8a83fb_remove_superseded_concept_from_license_pools.py
@@ -21,6 +21,8 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
+    # Initially create the column with a default of False, to
+    # populate the value efficiently.
     op.add_column(
         "licensepools",
         sa.Column(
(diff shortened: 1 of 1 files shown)


alembic/versions/20250626_bc471d8a83fb_remove_superseded_concept_from_license_pools.py
```diff
@@ -0,0 +1,33 @@
...
+
+def downgrade() -> None:
+    op.add_column(
+        "licensepools",
+        sa.Column(
+            "superceded",
+            sa.BOOLEAN(),
+            autoincrement=False,
+            nullable=False,
+            server_default=sa.false(),
```
jonathangreen, line 31: @tdilauro, I believe the migration tests are failing because `server_default` is being treated differently from `default`. Since `server_default` is a database-side configuration and isn't set in the existing code, the migration is detecting a discrepancy in the database configuration.

It seems you're using `server_default` to bypass the nullable constraint. To ensure consistency in the configuration, you can either:  
1. Create the column as nullable, populate it, and then set it to non-nullable, or  
2. Create the column with `server_default` and later modify it to drop the `server_default`.

tdilauro: Thanks, @jonathangreen!

PR #2576, opened 2025-07-04
Include download link in Boundless DRM fulfillment (PP-2702)

jonathangreen:
## Description

Include a link to the encrypted epub along with the boundless license document when doing a boundless DRM fulfillment.

The new document looks like this:
```json
{
  "licenseDocument": {},
  "links": [
    {
      "rel": "publication",
      "type": "application/epub+zip",
      "href": "http://blahblah"
    }
  ]
}
```

## Motivation and Context

Currently the apps don't have a way to get from the fulfillment document to the encrypted epub, since the license document does not provide a link. So we encapsulate the document, to provide the apps with all the information they need to complete the fulfillment.

## How Has This Been Tested?

- Tested locally

## Checklist

### DIFF
--- a/src/palace/manager/api/boundless/api.py
+++ b/src/palace/manager/api/boundless/api.py
@@ -24,6 +24,7 @@ from palace.manager.api.boundless.models.json import (
     AxisNowFulfillmentInfoResponse,
     FindawayFulfillmentInfoResponse,
 )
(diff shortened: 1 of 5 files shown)


tdilauro (approved):
Looks good!

PR #2581, opened 2025-07-07
M pp 2705 remove coverage from calculate presentation

dbernstein:
## Description
This PR removes the updates to the CoverageRecord table when an edition's cover or metadata is recalculated.

## Motivation and Context
Since we are using celery now to recalculate presentation information, we no longer need to support Edition related coverage records.  
https://ebce-lyrasis.atlassian.net/browse/PP-2705

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/model/coverage.py
+++ b/src/palace/manager/sqlalchemy/model/coverage.py
@@ -320,7 +320,6 @@ class CoverageRecord(Base, BaseCoverageRecord):
 
     __tablename__ = "coveragerecords"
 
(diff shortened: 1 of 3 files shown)


dbernstein:
@jonathangreen :  I was thinking the same thing.  I didn't see anything that caught my eye, but I'll give a second look before I merge.

dbernstein:
@jonathangreen : the only side effect I can see is that we will update the metadata for an edition associated with a licensepool every time a book appears in the feed even if only because it's availability information has changed. Before this change we would only update the metadata if the book's coverage record's timestamp was older than the data sources last update or the coverage record did not exist.

PR #2583, opened 2025-07-08
Add disable_async_calculation flags in order to prevent infinite loop.

dbernstein:
## Checklist


src/palace/manager/sqlalchemy/model/work.py
```diff
@@ -1188,7 +1195,6 @@ def queue_indexing(
             waiting.add(work_id)
 
     @staticmethod
-    @inject
```
dbernstein, line 1191: This removal is not related to the PR but in testing I noticed it was producing a warning.  It was left over from a previous version of the method that had an injected parameter.

jonathangreen (approved):

PR #2591, opened 2025-07-10
Add IntegrationRegistry.configurations_query (PP-2470)

jonathangreen:
## Description

Add two new functions:
- `IntegrationRegistry.configurations_query`
  - This gives a sqlalchemy select statement to find all the `IntegrationConfiguration` for a particular protocol in the registry.
- `Collection.select_by_protocol`
  - Leverages `IntegrationRegistry.configurations_query` to provide a select statement targeted at collections with a particular protocol.

And refactored some older code to use these new functions. Note: Not all code has been updated, only code that was already doing the same thing. Eventually it would be good to move all code using `Collection.by_protocol` to `Collection.select_by_protocol`, but for now I just added a comment saying that `select_by_protocol` should be used for new code.

## Motivation and Context

In PP-2470 I need to query collections by protocol. We already do this in the opds_odl celery task, and the boundless celery task, so I decided to pull this out to a function, so its not being duplicated.

## How Has This Been Tested?

- New unit tests
- Running existing tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/marc.py
+++ b/src/palace/manager/api/controller/marc.py
@@ -9,7 +9,6 @@ from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from palace.manager.api.util.flask import get_request_library
-from palace.manager.integration.goals import Goals
 from palace.manager.marc.exporter import MarcExporter
 from palace.manager.service.integration_registry.catalog_services import (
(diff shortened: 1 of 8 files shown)


tdilauro (approved):
Looks good! 🚀

PR #2593, opened 2025-07-10
Refactor OPDS http methods (PP-2470)

jonathangreen:
## Description

- Move the boundless `Token` model into a common location, so it can be used by both Boundless and Opds.
  - As part of the move I renamed it from `Token` to `OAuthTokenResponse`
  - This is a common token structure, so its nice to have a more general model to deal with it.
- Refactor the `OdlAuthenticatedRequest` mixin class to out to be a set of objects that we can delegate to in order to make requests.

## Motivation and Context

In PP-2470 I'm turning the OPDS importers into celery tasks, and I need a way to make requests within these tasks. It didn't really make sense to implement `OdlAuthenticatedRequest`, but having a request object I can call on to make my http requests made sense. 

## How Has This Been Tested?

- Added some new tests
- Moved over old tests

## Checklist


src/palace/manager/api/boundless/models/json.py
```diff
@@ -149,38 +146,6 @@ class AudiobookMetadataResponse(BaseBoundlessJsonResponse):
     )
 
 
-class Token(BaseBoundlessJsonModel):
```
jonathangreen, line 152: This is just moved around

PR #2598, opened 2025-07-16
Add type hints to bibliotheca integration (PP-2720)

jonathangreen:
## Description

Add type hints to the Bibliotheca integration.

Note: As well as adding type hints, this does some light refactoring to make it easier to type hint. I'll add some comments with reasoning for these changes.

## Motivation and Context

- Trying to slowly make sure all our code is type hinted
- Blocks doing some code organization

## How Has This Been Tested?

- Running in CI

## Checklist

### DIFF
--- a/src/palace/manager/api/bibliotheca.py
+++ b/src/palace/manager/api/bibliotheca.py
@@ -10,15 +10,18 @@ import re
 import time
 import urllib.parse
 from abc import ABC
-from collections.abc import Generator
+from argparse import ArgumentParser, Namespace
+from collections.abc import Collection as CollectionT, Generator, Iterable
 from datetime import datetime, timedelta
 from io import BytesIO
-from typing import Optional, TypeVar
+from typing import Any, Literal, Optional, TypeVar, overload
 
 import dateutil.parser
 from flask_babel import lazy_gettext as _
+from frozendict import frozendict
 from lxml.etree import Error, _Element
-from pymarc import parse_xml_to_array
+from pymarc import Record, parse_xml_to_array
+from requests import Response
 from sqlalchemy.orm import Session
 from typing_extensions import Unpack
 
@@ -53,7 +56,7 @@ from palace.manager.core.config import (
     CannotLoadConfiguration,
     ConfigurationAttributeValue,
 )
-from palace.manager.core.coverage import BibliographicCoverageProvider
+from palace.manager.core.coverage import BibliographicCoverageProvider, CoverageFailure
 from palace.manager.core.monitor import (
     CollectionMonitor,
     IdentifierSweepMonitor,
@@ -100,7 +103,8 @@ from palace.manager.util.datetime_helpers import (
     to_utc,
     utc_now,
 )
-from palace.manager.util.http import HTTP, RemoteIntegrationException
+from palace.manager.util.http import HTTP, RemoteIntegrationException, RequestKwargs
+from palace.manager.util.log import LoggerMixin
 from palace.manager.util.problem_detail import BaseProblemDetailException
 from palace.manager.util.xmlparser import XMLParser, XMLProcessor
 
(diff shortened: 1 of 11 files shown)


src/palace/manager/api/bibliotheca.py
```diff
@@ -234,60 +234,43 @@ def signature(self, method, path):
...
+        method: str = "GET",
+    ) -> Response:
         path = self.full_path(path)
         url = self.full_url(path)
         if method == "GET":
             headers = {"Accept": "application/xml"}
         else:
             headers = {"Content-Type": "application/xml"}
         self.sign(method, headers, path)
-        # print headers
```
jonathangreen, line 257: I dropped the `max_age` parameter and the logic to handle it here, because it wasn't used, and because it makes the response type of this function complicated since `Representation.get` has a different return type from `self._request_with_timeout`. This seemed needlessly complicated, when it wasn't actually used.

src/palace/manager/api/bibliotheca.py
```diff
@@ -234,60 +234,43 @@ def signature(self, method, path):
...
-            return self._request_with_timeout(
-                method,
-                url,
-                data=body,
-                headers=headers,
-                allow_redirects=False,
-                timeout=60,
-            )
-
-    def get_bibliographic_info_for(self, editions, max_age=None):
```
jonathangreen, line 281: Removed unused function

src/palace/manager/api/bibliotheca.py
```diff
@@ -378,18 +364,14 @@ def _count_activity():
...
 
-    def get_events_between(self, start, end, cache_result=False, no_events_error=False):
+    def get_events_between(
+        self, start: datetime, end: datetime, no_events_error: bool = False
+    ) -> Generator[tuple[str, str, str | None, datetime, datetime | None, str]]:
         """Return event objects for events between the given times."""
-        start = start.strftime(self.ARGUMENT_TIME_FORMAT)
-        end = end.strftime(self.ARGUMENT_TIME_FORMAT)
-        url = f"data/cloudevents?startdate={start}&enddate={end}"
-        if cache_result:
```
jonathangreen, line 386: Remove the unused `cache_result` parameter.

src/palace/manager/api/bibliotheca.py
```diff
@@ -646,14 +645,7 @@ def findaway_license_to_webpub_manifest(cls, license_pool, findaway_license):
         return DeliveryMechanism.FINDAWAY_DRM, str(manifest)
 
 
-class DummyBibliothecaAPIResponse:
```
jonathangreen, line 649: Unused

src/palace/manager/api/bibliotheca.py
```diff
@@ -768,7 +762,7 @@ def value(bibliotheca_key):
 
         # Presume all images from Bibliotheca are JPEG.
         media_type = Representation.JPEG_MEDIA_TYPE
-        cover_url = value("CoverLinkURL").replace("&amp;", "&")
+        cover_url = self.text_of_subtag(tag, "CoverLinkURL").replace("&amp;", "&")
```
jonathangreen, line 765: This is the same as `value`, but it raises if the element doesn't exist. The element must exist, since we were calling `replace` on it. Otherwise we would be calling `replace` on `None`. So the logic here is equivalent in the normal case, and gives a more descriptive error in the error case.

src/palace/manager/api/circulation/base.py
```diff
@@ -98,8 +99,13 @@ def __init__(self, _db: Session, collection: Collection):
...
     @property
-    def collection(self) -> Collection | None:
-        return Collection.by_id(self._db, id=self.collection_id)
+    def collection(self) -> Collection:
+        collection = Collection.by_id(self._db, id=self.collection_id)
+        if collection is None:
+            raise PalaceValueError(
+                f"Collection with id {self.collection_id} not found for {self.__class__.__name__}"
+            )
+        return collection
```
jonathangreen, line 108: I updated this type hint in the base class since all the implementations check if its none, then raise if it is. Instead of doing this in each implementation, I just moved this up to the base class, and removed the checks in each implementation.

PR #2602, opened 2025-07-17
Update booleans in configuration forms to use bool values

jonathangreen:
## Description

Update type hint, and update configuration form booleans to use boolean values in options.

## Motivation and Context

In one of our config forms we were using "True" instead of "true" in `options` which was causing the form not to work. Instead of mixing string and boolean, just update them all use to use boolean values for options to make it clear what we should be using.

## How Has This Been Tested?

- Tested locally

## Checklist


tdilauro (approved):
Looks good!

PR #2606, opened 2025-07-17
Allow circulation api to sort delivery mechanisms (PP-2744)

jonathangreen:
## Description

Allow `BaseCirculationAPI` implementations to sort the list of delivery mechanisms to be included in feeds.

This includes some refactoring to move the sorting that was being done for `OPDS` integration's out to the appropriate API implementations, instead of being tightly coupled to the `Collection`. 

## Motivation and Context

In PP-2744 I need a setting to prioritize Boundless DRM. This provides the hooks I need for that. That implementation will come in a future PR.

## How Has This Been Tested?

- New unit tests

## Checklist


src/palace/manager/api/boundless/settings.py
```diff
@@ -7,15 +7,14 @@
...
-from palace.manager.integration.configuration.formats import FormatPrioritiesSettings
 from palace.manager.integration.settings import (
     ConfigurationFormItem,
     ConfigurationFormItemType,
     FormField,
 )
 
 
-class BoundlessSettings(BaseCirculationApiSettings, FormatPrioritiesSettings):
+class BoundlessSettings(BaseCirculationApiSettings):
```
jonathangreen, line 17: This was my first attempt at allowing us to set the priority, but it didn't make much sense, since the settings here reference LCP. Removing this, and I'll do a followup PR with a setting specifically for Boundless.

src/palace/manager/core/opds_format_priority.py
```diff
@@ -12,106 +14,74 @@
...
-    _prioritized_drm_schemes: Mapping[str, int]
-    _prioritized_content_types: Mapping[str, int]
-    _hidden_content_types: list[str]
-    _deprioritize_lcp_non_epubs: bool
-
     def __init__(
         self,
         prioritized_drm_schemes: list[str],
         prioritized_content_types: list[str],
-        hidden_content_types: list[str],
```
jonathangreen, line 36: This now gets filtered before we call `FormatPriorities`.

tests/manager/api/controller/test_urn_lookup.py
```diff
@@ -12,7 +12,10 @@ def test_work_lookup(self, controller_fixture: ControllerFixture):
         work = controller_fixture.db.work(with_open_access_download=True)
         [pool] = work.license_pools
         urn = pool.identifier.urn
-        with controller_fixture.request_context_with_library("/?urn=%s" % urn):
+        with (
+            controller_fixture.request_context_with_library("/?urn=%s" % urn),
+            controller_fixture.wired_container(),
```
jonathangreen, line 17: This is needed because the containers must be setup now in order to get the registry.

tests/manager/api/controller/test_work.py
```diff
@@ -68,7 +68,6 @@ def work_fixture(
 class TestWorkController:
     def test_contributor(self, work_fixture: WorkFixture):
         m = work_fixture.manager.work_controller.contributor
-        work_fixture.collection.data_source = None
```
jonathangreen, line 71: This is actually invalid, because datasource cannot be null. I think we set this previously in order to reduce the number of facets coming back. Doing this here causes the configuration to error because datasource is required. 

So I removed this and adjusted for the extra facets having a datasource generates.

tests/manager/api/controller/test_work.py
```diff
@@ -812,7 +811,6 @@ def test_related_no_works(self, work_fixture: WorkFixture):
         assert result == NOT_FOUND_ON_REMOTE
 
     def test_series(self, work_fixture: WorkFixture):
-        work_fixture.collection.data_source = None
```
jonathangreen, line 815: Same reason for change as above.

tests/manager/core/test_opds_format_priority.py
```diff
@@ -0,0 +1,356 @@
...
+    FormatPriorities,
+    FormatPrioritiesSettings,
+)
+from palace.manager.sqlalchemy.model.licensing import (
+    DeliveryMechanism,
+    LicensePoolDeliveryMechanism,
+)
+
+
+class TestFormatPriorities:
```
jonathangreen, line 19: This is moved from `test_licensing.py`, since it didn't seem like it belonged there in the first place.

Its mostly copied, there are a couple minor changes to the test because `FormatPriorities` no longer filters out mechanisms.

PR #2608, opened 2025-07-17
Sort known delivery mechanisms (PP-2744)

jonathangreen:
## Description

Provide a baseline reasonable sorting of Delivery Mechanisms to get returned in our feeds.

## Motivation and Context

Right now the CM returns our delivery mechanisms in the feed in the order that they are returned from the database, which is generally insertion order. This is kind of random, and isn't ideal. It mostly doesn't matter too much, since mostly things only have one or two delivery mechanisms, but we are getting more cases where items have multiple delivery mechanisms, so its nice to provide a reasonable order that can be adjusted over time as we add new DRM methods.

## How Has This Been Tested?

- New unit tests

## Checklist

### DIFF
--- a/tests/manager/sqlalchemy/model/test_licensing.py
+++ b/tests/manager/sqlalchemy/model/test_licensing.py
@@ -240,6 +240,47 @@ class TestDeliveryMechanism:
         pytest.raises(IntegrityError, create, session, dm, **without_drm_args)
         session.rollback()
 
+    def test_sort(self) -> None:
+        def create_lpdm(
+            content_type: str | None, drm_scheme: str | None
+        ) -> LicensePoolDeliveryMechanism:
+            return LicensePoolDeliveryMechanism(
+                delivery_mechanism=DeliveryMechanism(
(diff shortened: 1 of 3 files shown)


tests/manager/sqlalchemy/model/test_licensing.py
```diff
@@ -240,6 +240,47 @@ def test_uniqueness_constraint(
...
+        assert DeliveryMechanism.sort([no_drm, adobe_drm, bearer_token]) == [
+            no_drm,
+            bearer_token,
+            adobe_drm,
+        ]
+
+        # If we have unknown DRM schemes, they are sorted last, but maintain their relative order.
+        assert DeliveryMechanism.sort(
+            [adobe_drm, unknown_drm_1, unknown_drm_2, no_drm]
+        ) == [no_drm, adobe_drm, unknown_drm_1, unknown_drm_2]
```
tdilauro, line 282: Minor suggestion: Repeat this assertion with the unknowns in a different relative order.

tdilauro (approved):
Looks good! 🔢 

One small test suggestion.

PR #2610, opened 2025-07-18
Add a setting so we can manually set the timeout on boundless requests (PP-2755)

jonathangreen:
## Description

Add a configuration option to override the default timeout for B&T requests.

## Motivation and Context

While testing Boundless DRM on Minotaur we are seeing timeouts. This seems like its worse in their dev environment, although looking at the logs we are seeing some timeouts in production as well. Hopefully we won't have to leave this config setting in forever, but for now I think its helpful to be able to tune the timeout for B&T via config.

## How Has This Been Tested?

- New unit tests

## Checklist


tdilauro (approved):
Looks good!

One question about a test assertion below.

PR #2616, opened 2025-07-21
Fix issue with BibliographicData validator (PP-2722)

jonathangreen:
## Description

Fix an issue where the `BibliographicData` published validator was causing `BibliographicData` not to be able to be round tripped from JSON.

To fix this I rolled back the changes from https://github.com/ThePalaceProject/circulation/pull/2501 and took another approach to solve the problem that PR was attempting to solve.

## Motivation and Context

In PP-2722, I need to be able to round trip `BibliographicData` to / from JSON. When I tried to do this, I hit the issue I mentioned in this comment: https://github.com/ThePalaceProject/circulation/pull/2501#discussion_r2126673379

It looks like the `BibliographicData` published validator was added in https://github.com/ThePalaceProject/circulation/pull/2501 to resolve PP-2610.

Instead of solving this via a validator, I updated the OPDS importer code to make sure what it passes into `BibliographicData` is the type that is expected.

## How Has This Been Tested?

- Tested locally
- Once its deployed, I'll check the imports on minotaur and make sure they are succeeding

## Checklist

### DIFF
--- a/src/palace/manager/core/opds2_import.py
+++ b/src/palace/manager/core/opds2_import.py
@@ -656,9 +656,7 @@ class OPDS2Importer(BaseOPDSImporter[OPDS2ImporterSettings]):
 
         first_imprint = first_or_default(publication.metadata.imprints)
         imprint = str(first_imprint.name) if first_imprint else None
-
         published = self._extract_published_date(publication.metadata.published)
-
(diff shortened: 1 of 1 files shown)


tdilauro:
@jonathangreen It looks like @dbernstein has some deeper background here; however, let me know if you want me to review it anyway.

PR #2617, opened 2025-07-21
Fix size update for large custom lists. (PP-2711)

tdilauro:
## Description

Use a database row count, rather than a Python `len` to calculate the current size of a custom list.

## Motivation and Context

Updates to auto update custom lists were loading the list's entries into memory and counting them. When lists were large enough, the list update script would fail. We didn't need most of those records for anything else, so it's a lot less expensive (and dangerous) to just count them at the database and return the count.

[Jira PP-2711]

## How Has This Been Tested?

- Existing custom list tests pass (specifically `tests/manager/scripts/test_customlist.py::TestCustomListUpdateEntriesScript`'s `test_process_custom_list` and `test_repopulate_state`).
- All tests pass locally.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/16429590778).

## Checklist

### DIFF
--- a/src/palace/manager/core/query/customlist.py
+++ b/src/palace/manager/core/query/customlist.py
@@ -4,6 +4,8 @@ import datetime
 import json
 from typing import TYPE_CHECKING
 
+from sqlalchemy import func, select
+
 from palace.manager.api.admin.problem_details import (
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
🎸

PR #2623, opened 2025-07-22
Avoid performing updates when no changes have occurred during present…

dbernstein:
…ation calculation to reduce probability of deadlocks.

## Description
On CMs with multiple overlapping Boundless collections we are seeing long database lock acquisition times  and Deadlocks when attempting to  update the editions table.   This problem becomes more acute as the number of overlapping collections increases. 

This PR attempts address this problem by being smarter about not trying to update the cover url and thumbnail information if it hasn't changed.  

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2725

## How Has This Been Tested?
Unit tests are passing.  There is no change in behavior expected at the application level.  

I'm not sure how to effectively test that non-data changing sql updates are not being propagated to the database without having to dig into the sqlalchemy layer.   @jonathangreen or @tdilauro: if you think this level of coverage is necessary,  I'll spend some time looking into it.  

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/model/edition.py
+++ b/src/palace/manager/sqlalchemy/model/edition.py
@@ -422,9 +422,10 @@ class Edition(Base, EditionConstants):
     def set_cover(self, resource):
         old_cover = self.cover
         old_cover_full_url = self.cover_full_url
-        self.cover = resource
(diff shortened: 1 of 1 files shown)


jonathangreen (approved):
👍🏻

PR #2625, opened 2025-07-23
Revert "Increase autoscale limit on default queue worker to 8 from 4 …

dbernstein:
…(#2592)"

This reverts commit 2e257432bd65df0052d64393094d9d25a8fbc837.

## Description
Now that #2623 is in place, it appears that this commit wasn't really helping after all.   I tested ga-georgia with 4 default workers and all was well so I think we can safely revert this.

## Checklist


jonathangreen (commented):
👍

PR #2629, opened 2025-07-24
Trigger collection task from collection controller (PP-2722)

jonathangreen:
## Description

Add a new method to trigger a collection indexing task when collection settings are changed.

## Motivation and Context

I need this functionality for the PR I'm making for PP-2722. Breaking this part out, because that PR is a beast, and I'd like to get some of the pieces of it landed.

This doesn't actually do anything right now. I'm going to make a couple PRs to implement the `import_task` method as a follow on to this.

## How Has This Been Tested?

- New tests

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/collection_settings.py
+++ b/src/palace/manager/api/admin/controller/collection_settings.py
@@ -147,7 +147,8 @@ class CollectionSettingsController(
             # as quickly as possible.
             try:
                 impl_cls.import_task(integration.collection.id).apply_async(
-                    countdown=10  # Delay the task to ensure the collection is created the task starts
+                    # Delay the task to ensure the collection has been created by the time the task starts
+                    countdown=10
                 )
             except NotImplementedError:
                 # If the protocol does not support import tasks, we just skip it.


src/palace/manager/api/admin/controller/collection_settings.py
```diff
@@ -142,6 +142,17 @@ def process_post(self) -> Response | ProblemDetail:
             # Trigger a site configuration change
             site_configuration_has_changed(self._db)
 
+            # If we have an importer task for this protocol, we start it
+            # in the background, so that the collection is ready to go
+            # as quickly as possible.
+            try:
+                impl_cls.import_task(integration.collection.id).apply_async(
+                    countdown=10  # Delay the task to ensure the collection is created the task starts
```
tdilauro, line 150: Do we want the import task to be run on every change (i.e., some changes -- e.g., name change -- may not require an import)? Although, I suppose that if we were to be selective, that selection probably needs to happen in the implementation class, since difference settings would require import in different implementations. Maybe food for thought for a future optimization, if it were to prove useful.

tdilauro (approved):
Looks good!

A minor comment issue and food for thought for the implementation import tasks.

PR #2634, opened 2025-07-25
Add helper functions to data_layer (PP-2722)

jonathangreen:
## Description

Add some helper functions for PP-2722 to the data_layer code.

- Add ability to call `bibliographic_apply` task without passing an edition
- Add `load_edition` method to `BibliographicData` to load a existing edition without autocreating it
- Add `parse_urn` method to `IdentifierData` to load an `IdentifierData` from a URN without creating the related `Identifier`
- Make `Identifier` raise `PalaceValueError` instead of a `ValueError`

## Motivation and Context

A smattering of changes I made to the datalayer to support the work on PP-2722. Since all these changes stand on their own, I think it makes sense to PR them separately, to make the final PP-2722 PR easier to review.

## How Has This Been Tested?

- New unit tests

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/apply.py
+++ b/src/palace/manager/celery/tasks/apply.py
@@ -75,8 +75,8 @@ def circulation_apply(
 def bibliographic_apply(
     task: Task,
     bibliographic: BibliographicData,
-    edition_id: int,
-    collection_id: int | None,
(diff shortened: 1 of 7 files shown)


tdilauro (approved):
🚀🚀🚀

PR #2638, opened 2025-07-28
[PP-2704] Ignore contributors with blank display name values.

dbernstein:
## Description
Before this PR,  opds2 imports were failing if a contributor's name was blank.  This condition can arise when metadata property that is mapped to a contributor role,  such as "translator",  appears in the json with a blank value.

As a result, after the contributors were parsed, we were seeing a ValueError being thrown [here](https://github.com/ThePalaceProject/circulation/blob/main/src/palace/manager/data_layer/contributor.py#L220) which was causing the book import to fail (just for that book).  But since all of the books in the PressBoooks feed contained translator="", all the books were failing.

This PR changes the behavior such that a WARNING is logged and the nameless contributor ignored rather than recording a  import failure for the book in question.  An exception would be justified if OPDS2 specified that metadata properties MUST be non-blank.  In fact the spec says that value SHOULD NOT be blank. 

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2704

## How Has This Been Tested?
Unit test added. 
Existing tests pass.
Tested manually in my local environment with with PressBooks OER collection.

## Checklist

### DIFF
--- a/src/palace/manager/data_layer/bibliographic.py
+++ b/src/palace/manager/data_layer/bibliographic.py
@@ -753,6 +753,14 @@ class BibliographicData(BaseMutableData):
             edition.contributions = []
 
         for contributor_data in self.contributors:
+
+            dn = contributor_data.display_name
+            if not dn or not dn.strip():
+                self.log.warning(
+                    f"contributor display name is blank for {contributor_data}: skipping..."
+                )
+                continue
+
             contributor_sort_name = contributor_data.find_sort_name(_db)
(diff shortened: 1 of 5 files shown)


jonathangreen:
👍 I do think this is a good approach. Just log a warning and ignore the empty contributor. My only nit is I think it should go up the stack a bit so it targets the data layer code instead of just OPDS2.

dbernstein:
@jonathangreen : this one is ready to go.

jonathangreen (approved):

PR #2639, opened 2025-07-29
Fix bug with B&T DRM Patron Sync (PP-2758)

jonathangreen:
## Description

We were ignoring the `Blio` format when doing a `patron_sync` operation for Boundless collections.

## Motivation and Context

This isn't a fix for PP-2758, but I found it while working on that fix, and it solves a related bug for Boundless `patron_sync`, so I wanted to get it fixed as well.

## How Has This Been Tested?

- New unit tests

## Checklist


tdilauro (approved):
Looks good (and cleaner). 🥇 

On minor typo below.

PR #2642, opened 2025-07-30
Refactor our ServicesFixture to make sure a wired services container is always available

jonathangreen:
## Description

This PR refactors our `ServicesContainer` fixture to make sure that tests always have a wired services container available.

## Motivation and Context

Previously, if a test required a wired `services_container`, it had to either wire it manually or use the `wired_services_container` fixture. This approach was chosen because wiring the container is a relatively expensive operation.

However, as we increasingly rely on the `services_container` and use it for more dependency injection, managing it manually has become cumbersome. We’re finding ourselves needing to request a wired container in more and more places, which adds overhead and friction.

Specifically in https://github.com/ThePalaceProject/circulation/pull/2643, because we call up `Collection.data_source` in many places, that PR required adding a lot of code to the tests, just to wire a services container in various places.

This PR introduces an alternative approach:

* A session-level `services_container` is created and wired once.
* A function-level fixture then ensures the container is reset between tests to avoid state leakage.

There is some risk of test data leaking if the container isn’t fully reset, but the dependency injection framework provides helper methods to assist with cleanup. If we encounter issues, we can enhance the session fixture to handle additional reset logic.

Overall, I believe the convenience of always having a pre-wired container outweighs the potential risk of state leakage between tests.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/tests/fixtures/api_controller.py
+++ b/tests/fixtures/api_controller.py
@@ -109,11 +109,6 @@ class ControllerFixture:
         # TestScopedSession to hang.
         app.manager = self.circulation_manager_setup()
 
-    @contextmanager
-    def wired_container(self):
-        with self.services_fixture.wired():
-            yield
-
     def circulation_manager_setup_with_session(
         self, session: Session, overrides: ControllerFixtureSetupOverrides | None = None
     ) -> CirculationManager:
--- a/tests/fixtures/search.py
+++ b/tests/fixtures/search.py
@@ -13,7 +13,7 @@ from palace.manager.celery.tasks.search import get_work_search_documents
 from palace.manager.search.external_search import ExternalSearchIndex
 from palace.manager.search.revision import SearchSchemaRevision
 from palace.manager.search.service import SearchServiceOpensearch1
-from palace.manager.service.container import Services, wire_container
+from palace.manager.service.container import Services
 from palace.manager.service.search.container import Search
 from palace.manager.sqlalchemy.model.work import Work
 from palace.manager.util.log import LoggerMixin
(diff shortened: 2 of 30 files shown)


tests/fixtures/api_controller.py
```diff
@@ -109,11 +109,6 @@ def __init__(
         # TestScopedSession to hang.
         app.manager = self.circulation_manager_setup()
 
-    @contextmanager
-    def wired_container(self):
-        with self.services_fixture.wired():
-            yield
-
```
jonathangreen, line 116: No longer need the various functions we have for providing a wired container.

tests/fixtures/search.py
```diff
@@ -44,34 +44,26 @@ def __init__(
...
-        # Make sure the services container is wired up with the newly created search container
-        wire_container(self.services_container)
-
     def close(self):
         # Delete our index prefix
         self.client.indices.delete(f"{self.index_prefix}*")
-
-        # Unwire the services container
-        self.services_container.unwire()
-        self.services_container.search.reset_override()
```
jonathangreen, line 74: The search fixture used to do a bunch of work to wire and unwire the container that is no longer necessary.

tests/fixtures/services.py
```diff
@@ -33,7 +34,11 @@ def mock_services_container(
         container._container_instance = None
 
 
-class MockServicesFixture:
+class ServicesFixture:
```
jonathangreen, line 37: This is where the bulk of the logic changes in this PR are. Most of the other changes are changes to tests to remove all the wiring we were doing.

dbernstein (approved):
🐰

PR #2649, opened 2025-07-31
[PP-2763] reduce database load caused by calculate work presentation

dbernstein:
## Description
This PR includes code changes that should prevent database update statements that do not change the state of the database from being generated in the sqlalchemy layer.  As a result we should see a drop in database loads caused by unnecessary row locking.

Like previous similar updates,  it is tricky to test whether the changes will produce the reduction in load without mocking the sqlalchemy layer.  Since we had good results in the past with similar updates, I think this one is good to go.   The coverage plugin is complaining a bit, but the cases it wants us to cover are trivial IMHO.
 
## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-2763

## How Has This Been Tested?
No functional changes:  unit tests should keep passing.  
The proof of success will be in seeing lower database load in production RDS performance insights.

## Checklist


jonathangreen (approved):
Looks good!

PR #2655, opened 2025-08-05
More lenient OPDS2 language parsing

jonathangreen:
## Description

Be more lenient about how we parse the OPDS2 language information.

## Motivation and Context

I was checking in on PP-2722 this morning, to make sure it looks good as it rolls out everywhere. It generally is looking good, but I'm seeing a lot of failures to parse the language information in some collections. 

This would have existed before the update to Celery, but since I'm noticing it, I thought it would make sense to make our language parsing a bit more lax. The two major categories of errors I'm seeing are: specifying the full language name like "english" or using the 3-letter bibliographic code, instead of the terminological code ("ger" instead of "deu"). 

This PR updates our language parsing to handle both those cases.

## How Has This Been Tested?

- New unit tests.

## Checklist

### DIFF
--- a/src/palace/manager/opds/types/language.py
+++ b/src/palace/manager/opds/types/language.py
@@ -30,6 +30,7 @@ class LanguageTag(str):
 
     _original: str
     _subtags: tuple[str, ...]
+    _name: str
 
     def __new__(
         cls,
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 🚀 

A couple of food for thought questions/suggestions below.

jonathangreen:
Good suggestions @tdilauro. I added a warning, and moved things into their own test.

PR #2659, opened 2025-08-05
Update MockHttpClientFixture to allow queuing responses out of order (PP-2724)

jonathangreen:
## Description

Update `MockHttpClientFixture` to:
- Add `index` parameter, so we can queue out of order
- While I was there, I updated the parameter for headers from `other_headers` to `headers` and updated the type hint.
- Change requests to use a `deque` instead of a `list`, since we are inserting and removing from both sides of the list.

## Motivation and Context

In PP-2724, I need to be able to queue up responses out of order, to fit the existing test structure. Adding this in a separate PR to make it a little easier to review.

## How Has This Been Tested?

- Running unit tests

## Checklist

### DIFF
--- a/tests/fixtures/http.py
+++ b/tests/fixtures/http.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from collections.abc import Callable, Generator
+from collections import deque
+from collections.abc import Callable, Generator, Mapping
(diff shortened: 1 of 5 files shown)


tdilauro (approved):
🛠️

PR #2666, opened 2025-08-08
Allow HTTP request_with_timeout to use a Session (PP-2722)

jonathangreen:
## Description

Allow us to pass a requests session to `request_with_timeout` to improve the performance of repeated requests.

## Motivation and Context

I took a look at the request times from our Celery ODL importer this morning, and the requests to fetch the license info documents take up the majority of time for that task. Its not too bad, 30 - 45 seconds per page, but it does take a lot of time when a feed as many many pages, as some of our production feeds do.

I was looking at ways to speed this up. I think the best solution here would be to use async to do this fetching, but that is much harder to implement. I'm going to make a ticket to do the work to make this possible. But an easier solution is to allow `request_with_timeout` to take a session, which allows it to use a collection pool and reuse connections. This speeds up repeated fetches by about 40%, which benefits the OPDS2+ODL importer.

## How Has This Been Tested?

- Added tests
- Tested locally

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/opds/odl/importer.py
+++ b/src/palace/manager/integration/license/opds/odl/importer.py
@@ -30,7 +30,7 @@ from palace.manager.service.integration_registry.license_providers import (
 )
 from palace.manager.service.redis.models.set import IdentifierSet
 from palace.manager.sqlalchemy.model.collection import Collection
-from palace.manager.util.http import BadResponseException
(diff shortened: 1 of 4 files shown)


jonathangreen:
I moved this back to draft, because a requests session isn't thread safe, and I think that is going to be a problem here. Need to do a bit of checking and possibly update this PR before it gets merge.

PR #2668, opened 2025-08-11
Remove reap_loans_with_unavailable_license_pools task (PP-2804)

jonathangreen:
## Description

Remove `reaper.reap_loans_with_unavailable_license_pools` from our beat schedule, since this task does not exist. 

Update other tasks to lookup their name via their function, so its easier to catch situations like this in the future.

## Motivation and Context

It looks like this task got added to the schedule in: https://github.com/ThePalaceProject/circulation/pull/2352 but was later removed from that PR before it was merged, but its entry was still in the celery conf. This was causing an error to occur in our celery workers: 
```
Received unregistered task of type 'reaper.reap_loans_with_unavailable_license_pools'
```

## How Has This Been Tested?

- Ran the scheduler locally to confirm tasks still ran correctly

## Checklist

### DIFF
--- a/src/palace/manager/service/celery/celery.py
+++ b/src/palace/manager/service/celery/celery.py
@@ -44,71 +44,87 @@ def beat_schedule() -> dict[str, Any]:
     be a timedelta object or a crontab string. The `task` key is required and should be the name
     of the task to run.
     """
+    from palace.manager.celery.tasks import (
+        boundless,
+        marc,
+        notifications,
+        novelist,
+        nyt,
+        opds2,
+        opds_odl,
+        playtime_entries,
+        reaper,
+        rotate_jwe_key,
+        saml,
+        search,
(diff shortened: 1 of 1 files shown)


src/palace/manager/service/celery/celery.py
```diff
@@ -44,204 +44,213 @@ def beat_schedule() -> dict[str, Any]:
     be a timedelta object or a crontab string. The `task` key is required and should be the name
     of the task to run.
     """
+    from palace.manager.celery.tasks import (
```
jonathangreen, line 47: We have to import the tasks within the function to avoid an import loop. I think it still makes sense to reference the tasks by their function, so its easier to notice when they are removed, or refactored, instead of just referring to them by string.

tdilauro (approved):
👍🏽

PR #2685, opened 2025-08-19
Abort and cleanup if an IdentifierSet is None (PP-2721)

jonathangreen:
## Description

Handle the case where the `mark_identifiers_unavailable` task receives `None` instead of an `IdentifierSet`.

## Motivation and Context

In some cases one of the tasks in the `mark_identifiers_unavailable` chord may want to abort the operation and log information, without raising an exception.In this case we wouldn't want the `mark_identifiers_unavailable` task to proceed, and we want to try to clean up any data. 

We handle this by returning None in the task in the chord, and handling it appropriately within `mark_identifiers_unavailable`.

## How Has This Been Tested?

- New unit test

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/identifiers.py
+++ b/src/palace/manager/celery/tasks/identifiers.py
@@ -71,6 +71,10 @@ def mark_identifiers_unavailable(
     two IdentifierSets: the first represents existing identifiers that are available in the collection, and
     the second contains active identifiers received from the distributor.
 
+    If the calling task wants to abort marking identifiers as unavailable, it can return None for either
+    of the sets. This will cause the function to log a warning, clean up any IdentifierSets it was
+    passed and exit without marking any identifiers as unavailable.
+
     Any identifiers present in the existing set but not in the active set will be marked as unavailable in
     the collection. This is done by sending a circulation_apply task that sets both licenses_available and
     licenses_owned to 0 for each identifier in the collection.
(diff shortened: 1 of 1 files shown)


src/palace/manager/celery/tasks/identifiers.py
```diff
@@ -82,6 +86,18 @@ def mark_identifiers_unavailable(
     existing_identifier_kwargs, active_identifier_kwargs = (
         existing_and_active_identifier_sets
     )
+
+    if existing_identifier_kwargs is None or active_identifier_kwargs is None:
+        # If one of the sets is None, one of the tasks in the chord failed.
+        task.log.warning(
+            "Received None instead of IdentifierSet. Aborting without marking any identifiers as unavailable."
+        )
```
tdilauro, line 94: Question: Is it useful to know which of the tasks aborted? Or will there already be enough information logged elsewhere to satisfy any need?

tdilauro (approved):
Looks good! 🪣

One question about more specific logging below.

src/palace/manager/celery/tasks/identifiers.py
jonathangreen: I think it'll be notable that the calling task wants to abort this operation, so I expect it will log some detailed information about what went wrong, so I don't think that we need more detail here.

PR #2690, opened 2025-08-20
Output extra information from log record in JSONFormatter

jonathangreen:
## Description

Allow extra context to be added to log records like this:
```
self.log.warning("Log message", extras={"palace_context": {"helpful": "context"})
```

## Motivation and Context

As part of PP-2723, I'd like to add some additional context to the log output when a publication fails to index.

## How Has This Been Tested?

- Tested locally

## Checklist


tdilauro (approved):
🪵 🚀

PR #2696, opened 2025-08-26
Convert boundless importer to use shared reaper and bibliographic tasks (PP-2805)

jonathangreen:
## Description

Update the Boundless importer to use a similar structure to the OPDS importer where we use:
- `identifiers.create_mark_unavailable_chord` for reaping
- `apply.circulation_apply` to actually apply the circulation data

## Motivation and Context

I hope this will make it easier to maintain the importers, since they are all using the same common components. I'm also hoping that this will solve the issue we are seeing in Georgia where multiple copies of the reaper task are running. Because the reaper shouldn't have to requeue it should complete fairly quickly and its protected by a lock, so only one copy can be running at at a time.

Once this one is merged, I'd like to cut a hotfix branch for it and give it a try on Georgia and see how it performs there. We may have to adjust depending how that test goes.

## How Has This Been Tested?

- New unit tests

## Checklist


src/palace/manager/celery/tasks/boundless.py
```diff
@@ -1,520 +1,112 @@
...
-        ):
-            task.log.info(
-                f'Queued collection("{collection.name}" [id={collection.id}] for importing...'
-            )
-            list_identifiers_for_import.apply_async(
-                kwargs={"collection_id": collection.id},
-                countdown=count * 5,  # stagger the execution of the collection import
-                # in order to minimize chance of deadlocks caused by
-                # simultaneous updates to the metadata when CM has multiple
-                # Boundless collections configured with overlapping content
```
jonathangreen, line 61: I'm hoping this won't be necessary now, since the apply task is protected by a lock, and retries if it can't get the lock. Overlapping tasks should just wait their turn in the queue.

src/palace/manager/celery/tasks/boundless.py
```diff
@@ -1,520 +1,112 @@
...
-    deadlocks when launched in parallel.
-
-    The default delay of 20 seconds between launching of reap_collection sub-tasks is based on
-    observations of performance in a local environment.  I observed that reaping in batches 10 seconds
-    can take from 2-4 seconds in my development environment. The production environment was
-    observed to be about 2 to 3 times slower.  Therefore,a 20 second delay between task initiation provides a reasonable
-    empirically based time buffer such that batches of boundless identifiers are not reaped simultaneously and won't likely
-    cause a pile-up in the case of occasional deadlocks.  With the redis visibility_timeout set to one hour, this will
-    allow for a maximum of 180 collections.  Due to the visibility_timeout constraint,  20 seconds will be the maximum
-    delay:  any value over 20 will be ignored and the default will be used in its place.
```
jonathangreen, line 411: I think this will also be handled by the collection level locks, but since its a concurrency thing, its hard to tell until it goes out to a busy server.

src/palace/manager/integration/license/boundless/api.py
```diff
@@ -670,16 +671,6 @@ def _reap(self, identifier: Identifier) -> None:
...
 
-    def recent_activity(
-        self, since: datetime.datetime
-    ) -> Generator[tuple[BibliographicData, CirculationData]]:
-        """Find books that have had recent activity.
-
-        :yield: A sequence of (BibliographicData, CirculationData) 2-tuples
-        """
-        availability_response = self.api_requests.availability(since=since)
-        yield from BibliographicParser.parse(availability_response)
```
jonathangreen, line 681: This was only used in the importer, so it was moved over there.

src/palace/manager/service/celery/celery.py
```diff
@@ -129,9 +129,9 @@ def beat_schedule() -> dict[str, Any]:
             "task": boundless.reap_all_collections.name,
             "schedule": crontab(
                 day_of_week="6",
-                minute="0",
+                minute="8",
                 hour="4",
-            ),  # Every Saturday at 4:00 AM
+            ),  # Every Saturday at 4:08 AM
```
jonathangreen, line 134: I offset this slightly, so the reaper and import don't run at the exact same time.

src/palace/manager/sqlalchemy/model/coverage.py
```diff
@@ -312,6 +315,48 @@ def to_data(self):
             counter=self.counter,
         )
 
+    @property
```
jonathangreen, line 318: These are just helpers to make working with `Timestamp` easier. They could come out to a separate PR if desired.

PR #2706, opened 2025-08-28
Move calculate_work_presentation task to apply queue (PP-2842)

jonathangreen:
## Description

Move `work.calculate_work_presentation` to the `apply` queue and protect calls to it with the same lock used for the other apply tasks.

## Motivation and Context

- I have a theory that a lot of the database load that `work.calculate_work_presentation` causes happens because we have two async tasks doing the same work at the same time. Both the apply task and `work.calculate_work_presentation`
- Protecting this with the same lock should make it clear if that is the case.

If that is the case, I think long term we need to refactor how the two of them work. I have a ticket in the backlog for this: PP-2726. This should help in the short term.

## How Has This Been Tested?

- Added new tests

## Checklist

### DIFF
--- a/tests/manager/celery/tasks/test_work.py
+++ b/tests/manager/celery/tasks/test_work.py
@@ -3,7 +3,7 @@ from unittest.mock import call, patch
 import pytest
 from sqlalchemy.exc import SQLAlchemyError
 
-from palace.manager.celery.tasks import work as work_tasks
+from palace.manager.celery.tasks import apply, work as work_tasks
 from palace.manager.data_layer.policy.presentation import PresentationCalculationPolicy
 from palace.manager.service.logging.configuration import LogLevel
 from palace.manager.sqlalchemy.model.classification import Subject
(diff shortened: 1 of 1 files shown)


src/palace/manager/celery/tasks/work.py
```diff
@@ -1,48 +1,64 @@
...
+        pool = first_or_default(work.license_pools)
+        lock_ctx_manager: AbstractContextManager[Any]
+        if pool is None:
+            task.log.warning(
+                f"Work {work.id} has no LicensePool. Continuing without lock."
+            )
+            lock_ctx_manager = nullcontext()
+        else:
+            identifier = IdentifierData.from_identifier(pool.identifier)
+            lock_ctx_manager = apply_task_lock(redis_client, identifier).lock()
```
dbernstein, line 52: Hopefully applying a task lock for these relatively small tasks won't introduce too much overhead.

dbernstein (approved):
Looking good.

src/palace/manager/celery/tasks/work.py
dbernstein: But if it solves the problem with the conflicting updates then it should improve the current performance.

PR #2719, opened 2025-09-04
[PP-2183] refactor overdrive import to minimize duplicative updates

dbernstein:
## Description
This update makes the following changes:
1.  The overdrive circulation monitors only import "parent" overdrive collections - ie any account that is configured with a parent will be ignored by the overdrive import.
2. During the import of the "parent" collections, the circulation data for all the child (ie Advantage) collections which are returned in the overdrive availability api with their respective owned and available counts as well as their shared status. This means that we are no longer querying circulation data twice in cases where an advantage collection's books are also found in the parent collection.   Additionally we are no longer importing the bibliographic data multiple times:  only once for every edition that appears in the parent of child collections, regardless of whether the books appears in more than one of the related collections.
3. Whenever a OverdriveCirculationMonitor runs on a parent collection, it will automatically set up OverDrive Advantage Collections that are listed in the parent collections "/advantageAccounts" endpoint.  I'm not sure if this is the desired behavior.  However, it does ensure that any available and owned copies that are shared with the consortial (parent) account will be included in the parent licensepool regardless of whether there are libraries associated with the child that is sharing the titles.

## Motivation and Context

PP-2183
## How Has This Been Tested?
Unit tests have been updated.  
Manually tested overdrive monitors locally.

## Checklist

### DIFF
--- a/tests/manager/integration/license/overdrive/test_monitor.py
+++ b/tests/manager/integration/license/overdrive/test_monitor.py
@@ -74,6 +74,8 @@ class TestOverdriveCirculationMonitor:
         #
         # The method stops when should_stop() -- called on every book
         # -- returns True.
+        advantage_account = MagicMock()
+
         class MockAPI:
             def __init__(self, *ignore, **kwignore):
                 self.licensepools = []
@@ -83,6 +85,9 @@ class TestOverdriveCirculationMonitor:
                 is_changed = self.licensepools.pop(0)
                 self.update_licensepool_calls.append(book_id)
                 return is_changed
+
+            def get_advantage_accounts(self):
+                return [advantage_account]
 
         class MockMonitor(OverdriveCirculationMonitor):
             recently_changed_ids_called_with = None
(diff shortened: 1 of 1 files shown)


tests/manager/integration/license/overdrive/test_api.py
```diff
@@ -1726,166 +1730,9 @@ def test_update_licensepool_provides_bibliographic_coverage(
         assert not pool.work
         http.queue_response(200, content=availability)
         http.queue_response(200, content=bibliographic)
-        pool, was_new, changed = overdrive_api_fixture.api.update_licensepool(
-            identifier.identifier
-        )
-        assert was_new is False
-        assert pool.work.presentation_ready is True
-
-    def test_update_new_licensepool(
```
dbernstein, line 1735: The function that these tests target is not being used anywhere in the application code so I removed them.

src/palace/manager/integration/license/overdrive/monitor.py
```diff
@@ -65,6 +65,15 @@ def catch_up_from(
         """
         overdrive_data_source = DataSource.lookup(self._db, DataSource.OVERDRIVE)
 
+        if self.collection.parent is None:
+            # If it is a parent account, generate child (Advantage) collections if they don't already exist.
```
dbernstein, line 69: I'm unsure if we want to do this.   As a note:  we have a CM with a main collection  and a single Advantage account.  However with this block of code included, we create 28 other advantage accounts in the course of an import.  I'm not sure if this is typical and/or a concern, but I wanted to call it out.  
@jonathangreen  - see the ticket for details.

PR #2728, opened 2025-09-11
Async HTTP Client (PP-2900)

jonathangreen:
## Description

Add a new async http client. The client is just a light wrapper around `httpx.AsyncClient`.

We wrap it to provide some of the same niceties that `HTTP.request_with_timeout` provides:
- Wrap httpx exceptions with our `RemoteIntegrationError` exceptions
- Provide error status checking via `allowed_response_codes` and `disallowed_response_codes`
- Provide basic retry functionality (no built into httpx).
- Set default parameters to meet our needs (such as setting user-agent).

This PR also includes some refactoring of the existing `HTTP` class:
- So that parts of it can be reused in the new async class
- To move it into a http package, so we have a home for the reusable components

Notes:
- Eventually I'd like to remove the use of requests in `HTTP.request_with_timeout` and replace it with `httpx`, so we aren't depending on two different http libraries. I think this PR gives a good framework to do that, but I'm going to make that effort to replace it in a different PR.
- `AsyncClient` isn't actually used anywhere yet, that PR just adds it. I'll make a follow up PR to update the OPDS2 + ODL importer to use it.

## Motivation and Context

I want to use async to speed up the requests in PP-2900, but I need an async HTTP client in order to be able to do that.

## How Has This Been Tested?

- New unit tests

## Checklist

### DIFF
--- a/src/palace/manager/util/http/async_http.py
+++ b/src/palace/manager/util/http/async_http.py
@@ -251,9 +251,11 @@ class AsyncClient(LoggerMixin):
         self._max_backoff = max_backoff
 
     @staticmethod
-    def _defaults(kwargs: ClientKwargs) -> ClientKwargs:
+    def _defaults(kwargs: ClientKwargs) -> None:
         """
         Sets our global defaults for httpx.AsyncClient parameters.
+
+        Modifies the passed in kwargs in place.
         """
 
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Looks great! 🚀 

One possible side-effect issue below.

PR #2730, opened 2025-09-12
OPDS importer asynchronous HTTP requests (PP-2900)

jonathangreen:
## Description

This PR updates the OPDS importer to use asynchronous HTTP requests when fetching license documents, improving performance by fetching multiple license documents concurrently rather than sequentially.

## Motivation and Context

The original OPDS importer would fetch license documents one by one in a synchronous manner, which could be slow when processing feeds with many publications. This was causing us problems with large feeds on our production CM instances, where we were not able to complete a feed harvest in one hour. 

Doing some local bench-marking on my system, it used to take about ~20 seconds per feed page to fetch the license documents, and it takes ~2 seconds now.

This change improves import performance by:

  - Fetching all license documents for a feed concurrently using async/await
  - Restructuring the import process into three clear phases: validation/filtering, concurrent license fetching, and bibliographic data extraction
  - Maintaining the same functional behavior while significantly reducing I/O wait time

This change builds on the async HTTP client infrastructure added in https://github.com/ThePalaceProject/circulation/pull/2728.

## How Has This Been Tested?

  - Updated existing unit tests in test_importer.py to work with the new async methods
  - Tests now use MockAsyncClientFixture instead of MockHttpClientFixture
  - Tests verify proper retry behavior with the async client (4 total requests: 1 initial + 3 retries)
  - Maintained test coverage for error handling scenarios (bad response codes, validation errors)
  - All existing test scenarios continue to pass, ensuring backward compatibility

  The changes maintain the same public interface while restructuring internal implementation for better performance.

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/opds/importer.py
+++ b/src/palace/manager/integration/license/opds/importer.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 from collections.abc import Iterable
 from dataclasses import dataclass
 from functools import cached_property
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good!

PR #2742, opened 2025-09-17
Add report query definition. (PP-1888)

tdilauro:
## Description

Adds a mechanism for associating an id and title with a query for tabular data for reports. 

## Motivation and Context

As we add more reports, it is handy to have a consistent way of defining and being able to lookup the queries used to produce the tabular data that will comprise those reports.

[Jira PP-1888]

## How Has This Been Tested?

- New tests.
- Local tests pass.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/17810446785) running.

## Checklist


src/palace/manager/reporting/model.py
```diff
@@ -0,0 +1,71 @@
...
+    title: str
+    statement: Select | TextualSelect = field(repr=False, hash=False)
+
+    @cached_property
+    def headings(self) -> tuple[str, ...]:
+        """Return the headings for the table's rows."""
+        try:
+            headings = tuple(c.name for c in self.statement.c)
+        except AttributeError as e:
+            raise ValueError(
```
jonathangreen, line 31: Minor: Might want to use `PalaceValueError`. I've been trying to use that instead of `ValueError`, so we have a common base class for any Palace exceptions.

jonathangreen (approved):
Looks good!

PR #2750, opened 2025-09-19
Library collection reports support (PP-1888)

tdilauro:
## Description

Adds support for collection data for a specified library. By default, this data is restricted to the active collections of a library, but that can be overridden by specifying a list of any or all collections associated with a library, whether those collections have an active subscription or not.

## Motivation and Context

This provides support for the library title-level report requested in PP-1888.

## How Has This Been Tested?

- New tests.
- All tests pass in locally.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/17870762495).

## Checklist

### DIFF
--- a/src/palace/manager/reporting/reports/library_collection.py
+++ b/src/palace/manager/reporting/reports/library_collection.py
@@ -1,0 +1,302 @@
+from __future__ import annotations
+
+import tempfile
+import zipfile
+from collections.abc import Sequence
+from datetime import datetime
+from functools import cached_property, partial
+from pathlib import Path
+from typing import IO, ClassVar, TypedDict, TypeVar
+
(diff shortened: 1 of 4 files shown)


jonathangreen (approved):
This looks good @tdilauro!

The only minor comment I have about it is that it would be nice to be able to use context manager for resource cleanup and closing (temp files, file handles, etc). I followed the logic here though, and I believe everything is cleaned up right now. My concern is that over time as this code evolves the cleanup approach is going to be brittle, and we are going to end up leaking temp files or other resources.

Thats just a minor, future worry though so feel free to merge this as is.

tdilauro:
That's fair. I did try to do that, but some of the flow was really tricky to manage. I hope this will gel a little better after we have a couple more examples.

PR #2762, opened 2025-09-26
Update AsyncClient backoff implementation

jonathangreen:
## Description

This PR updates the `AsyncClient` backoff implementation to provide more flexibility and update the backoff calculation logic. The changes replace the hardcoded backoff configuration with a customizable function-based approach.

## Motivation and Context

The previous implementation had a few issues:
1. The exponential backoff calculation was off by one (using `retries + 1` instead of `retries`)
    - This happened as part of https://github.com/ThePalaceProject/circulation/pull/2754, I added the new `factor` parameter, but I didn't start using `retries`.
    - This change aligns the backoff calculation with the one done in urllib3 and celery, so our back off is more standard.
    - Using the default parameters, the backoff values haven't changed at all since:
      - `1 * (3 ^ (x+1)) == 3 * (3 ^ x)`
3. The backoff configuration was limited to two parameters (`factor` and `max_backoff`), which didn't allow for full customization
    - We could add additional parameters, but it seems like we would continue to need to add new ones, so just passing a backoff function seems easier. Especially while this is still relatively unused.

This change:
- Fixes the backoff calculation to correctly start at `factor` seconds for the 0th retry
- Replaces `backoff_factor` and `max_backoff` parameters with a single `backoff` function parameter that takes the retry count and returns the wait time
- Updates tests to use `backoff=None` to avoid delays during testing
- Provides sensible defaults for both web and worker clients using `functools.partial`
  - The defaults for the worker are a little more aggressive then previously, but they have been working well in my testing.

## How Has This Been Tested?

- All existing tests pass with the new implementation
- Updated tests to verify the correct backoff calculation behavior

## Checklist

### DIFF
--- a/src/palace/manager/util/backoff.py
+++ b/src/palace/manager/util/backoff.py
@@ -9,7 +9,7 @@ def exponential_backoff(
     retries: int,
     *,
     max_time: float | None = None,
-    factor: float = 1.0,
+    factor: float = 3.0,
     base: float = 3.0,
     jitter: float = 0.3,
 ) -> float:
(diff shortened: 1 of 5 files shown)


tdilauro (approved):
Looks good! 🔄

PR #2763, opened 2025-09-26
Library title-level report (PP-1888)

tdilauro:
## Description

- Adds a library title-level report based on the `LibraryAllTitleReportTable`.
- Adds a new `/admin/reports/<report_key>` endpoint through which a properly configured report may be requested by their keys.
- Adds a `generate_reports` celery task that will be used to run reports requested through the endpoint.

**Note**: This PR depends on #2750. It should not be merged before that one has made it in.

## Motivation and Context

Provides the report capability requested on PP-1888.

## How Has This Been Tested?

- **Still needs tests**

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/report.py
+++ b/src/palace/manager/api/admin/controller/report.py
@@ -1,4 +1,5 @@
 import json
+import uuid
 from http import HTTPStatus
 
 import flask
@@ -13,12 +14,20 @@ from palace.manager.api.admin.model.inventory_report import (
     InventoryReportCollectionInfo,
     InventoryReportInfo,
 )
-from palace.manager.api.admin.problem_details import ADMIN_NOT_AUTHORIZED
+from palace.manager.api.admin.problem_details import (
+    ADMIN_NOT_AUTHORIZED,
+    INVALID_REPORT_KEY,
+)
 from palace.manager.celery.tasks.generate_inventory_and_hold_reports import (
     generate_inventory_and_hold_reports,
     library_report_integrations,
 )
+from palace.manager.celery.tasks.reports import (
+    REPORT_KEY_MAPPING,
+    generate_report,
+)
 from palace.manager.core.problem_details import INTERNAL_SERVER_ERROR
+from palace.manager.reporting.reports.library_collection import LibraryCollectionReport
 from palace.manager.service.integration_registry.license_providers import (
     LicenseProvidersRegistry,
 )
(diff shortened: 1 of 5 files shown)


src/palace/manager/celery/tasks/reports.py
```diff
@@ -0,0 +1,36 @@
...
+
+from palace.manager.celery.task import Task
+from palace.manager.reporting.reports.library_collection import (
+    LibraryCollectionReport,
+    LibraryReportKwargs,
+    LibraryTitleLevelReport,
+)
+from palace.manager.service.celery.celery import QueueNames
+
+REPORT_KEY_MAPPING: dict[str, type[LibraryCollectionReport]] = {
```
jonathangreen, line 14: Minor: Might be nice to use a `frozendict` here. Its nice to have these global variables be immutable data structures.

src/palace/manager/api/admin/controller/report.py
```diff
@@ -46,10 +56,65 @@ def _authorize_from_request(
...
+            task = generate_report.delay(
+                key=report_key,
+                request_id=request_id,
+                library_id=library.id,
+                email_address=email_address,
+            )
+        except Exception as e:
+            msg = f"Failed to generate report '{report_title}' ({report_key}). (request ID: {request_id})"
+            self.log.error(msg=msg, exc_info=e)
+            self._db.rollback()
```
jonathangreen, line 97: Minor: Is this rollback actually necessary here? It doesn't look like we should have done anything that needs a rollback.

tests/manager/api/admin/controller/test_report.py
```diff
@@ -43,8 +49,12 @@ def report_fixture(
 
 
 class TestReportController:
+    @patch(
+        "palace.manager.api.admin.controller.report.generate_inventory_and_hold_reports"
```
jonathangreen, line 53: Minor: I've taken to trying to use `patch.object(report, "generate_inventory_and_hold_reports")` since its handled better when using pycharm refactor to do a rename.

src/palace/manager/reporting/reports/library_collection.py
```diff
@@ -368,8 +369,7 @@ def _run_report(self, *, session: Session) -> bool:
         # Send success notification.
         self.send_success_notification(access_url=location)
         self.log.info(
-            f"Emailed notification for report '{self.title}' ({self.key}) for "
-            f"library {library.name} ({library.short_name}) to {self.email_address}."
+            f"Creating report '{self.key}' for {library.name} ({library.short_name})."
```
jonathangreen, line 372: Minor: This log message seems misleading (and is a duplicate of line 347). I'd suggest changing "Creating" to "Completed".

jonathangreen (approved):
Looks good, just a couple (very) minor comments! 🎸

src/palace/manager/api/admin/controller/report.py
tdilauro: No, it is not. Good catch. 🎉

PR #2765, opened 2025-09-26
Add Retry-After header support to AsyncClient (PP-2950)

jonathangreen:
## Description

This PR adds support for the Retry-After HTTP header in `AsyncClient`, allowing the client to respect server-specified retry delays when encountering rate limits or temporary unavailability.

## Motivation and Context

Getting UL harvesting to work with our new AsyncClient in PP-2950 has been a bit of a journey. I ended up going down a few different paths that didn't work. After doing a bunch of testing, they are returning `Retry-After` headers, and when we respect these headers, I am finally able to successfully harvest the whole feed locally. 🎉 

Building on the backoff implementation in https://github.com/ThePalaceProject/circulation/pull/2762, this adds intelligent retry behaviour by parsing and respecting the Retry-After header sent by servers. 

The implementation:
- Supports both delay-seconds and HTTP-date formats for Retry-After headers specified in the RFC
- Uses the larger of the calculated backoff and Retry-After delay
- Includes a configurable maximum delay cap to prevent excessive waits
- Can be disabled if needed via the `respect_retry_after` parameter

## How Has This Been Tested?

Added new unit tests covering:
- Retry-After with delay-seconds format
- Retry-After with HTTP-date format  
- Comparison logic between backoff and Retry-After delays
- Maximum delay capping behavior
- Disabling Retry-After support
- Invalid header handling

All existing and new tests pass.

## Checklist

### DIFF
--- a/src/palace/manager/util/http/base.py
+++ b/src/palace/manager/util/http/base.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 from collections.abc import Collection
+from email.utils import parsedate_to_datetime
 from typing import Literal, TypeVar
 
 import httpx
 import requests
 
 from palace import manager
+from palace.manager.util.datetime_helpers import utc_now
 from palace.manager.util.http.exception import BadResponseException
 
 # In case an app version is not present, we can use this version as a fallback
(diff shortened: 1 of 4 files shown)


src/palace/manager/util/http/base.py
```diff
@@ -107,3 +109,34 @@ def raise_for_bad_response(
...
+        pass
+
+    # Try to parse as HTTP-date
+    try:
+        retry_date = parsedate_to_datetime(header_value)
+        delay = (retry_date - utc_now()).total_seconds()
+        # Return delay only if it's positive (future date)
+        return max(0, delay)
+    except (TypeError, ValueError, AttributeError):
+        # Invalid date format
```
tdilauro, line 141: Should we log these?

PR #2772, opened 2025-10-03
Refactor OPDS2 API tests to use MockHttpClientFixture

jonathangreen:
## Description

Refactored OPDS2 API tests to use the shared `MockHttpClientFixture` instead of directly mocking the HTTP object with `unittest.mock`. This improves test consistency and maintainability across the codebase.

## Motivation and Context

- Standardizes HTTP mocking patterns across the test suite
- Removes the need to patch internal API modules
- Makes test setup more explicit by requiring tests to queue expected responses
- Improves test readability by using the fixture's cleaner API

## How Has This Been Tested?

- All existing tests continue to pass with the new mocking approach
- Verified that tests properly queue responses before making assertions

## Checklist

### DIFF
--- a/tests/manager/integration/license/opds/opds2/test_api.py
+++ b/tests/manager/integration/license/opds/opds2/test_api.py
@@ -59,8 +59,7 @@ def opds2_api_fixture(
     db: DatabaseTransactionFixture,
     http_client: MockHttpClientFixture,
 ) -> Generator[Opds2ApiFixture, None, None]:
-    fixture = Opds2ApiFixture(db, http_client)
-    yield fixture
+    yield Opds2ApiFixture(db, http_client)
 
 
 class TestOpds2Api:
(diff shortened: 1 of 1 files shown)


tests/manager/integration/license/opds/opds2/test_api.py
```diff
@@ -51,31 +46,36 @@ def __init__(self, db: DatabaseTransactionFixture, mock_http: MagicMock):
...
 @pytest.fixture
 def opds2_api_fixture(
     db: DatabaseTransactionFixture,
+    http_client: MockHttpClientFixture,
 ) -> Generator[Opds2ApiFixture, None, None]:
-    with patch.object(api, "HTTP") as mock_http:
-        fixture = Opds2ApiFixture(db, mock_http)
-        yield fixture
+    fixture = Opds2ApiFixture(db, http_client)
+    yield fixture
```
tdilauro, line 63: Minor: We could skip the variable and go straight to the yield here:

```suggestion
    yield Opds2ApiFixture(db, http_client)
```

PR #2773, opened 2025-10-03
Add SAML template parameter support for OPDS2 token authentication (PP-2990)

jonathangreen:
## Description

- Added `TemplateVariable` enum defining supported template variables: `patron_id`, `saml_entity_id`, and `saml_person_scoped_affiliation`
- Refactored `get_authentication_token` from classmethod to instance method to support SAML credential lookups
- Implemented `_build_template_parameters` method that intelligently fetches only the data required by the template
- Enhanced error handling with detailed logging including structured extra fields for debugging
- Added support for multiple SAML affiliations (passed as comma-separated values)

## Motivation and Context

**Jira:** PP-2990

This PR adds support for SAML template parameters in OPDS2 token authentication URLs. Token authentication endpoints can now receive SAML identity information when authenticating patrons, enabling content providers to make authorization decisions based on the patron's identity provider attributes.

Needed for upcoming Taylor & Francis integration.

## How Has This Been Tested?

- New tests
- Running tests in CI

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/opds/opds2/api.py
+++ b/src/palace/manager/integration/license/opds/opds2/api.py
@@ -44,9 +44,7 @@ class TemplateVariable(StrEnum):
 SAML_TEMPLATE_VARIABLES: frozenset[str] = frozenset(
     var for var in TemplateVariable if var.startswith("saml_")
 )
-SUPPORTED_TEMPLATE_VARIABLES: frozenset[str] = frozenset(
-    var for var in TemplateVariable
-)
+SUPPORTED_TEMPLATE_VARIABLES: frozenset[str] = frozenset(TemplateVariable)
 
 
 class OPDS2API(BaseOPDSAPI):
--- a/src/palace/manager/integration/license/opds/opds2/api.py
+++ b/src/palace/manager/integration/license/opds/opds2/api.py
@@ -77,6 +77,10 @@ class OPDS2API(BaseOPDSAPI):
     ) -> tuple[str | None, list[str] | None]:
         """Get the SAML token arguments for a patron
 
+        This method extracts values for the SAML-related members of TemplateVariable
+        (SAML_ENTITY_ID and SAML_PERSON_SCOPED_AFFILIATION). If / when we need to
+        support additional SAML attributes, they can be added here.
+
         :param patron: The patron to get parameters for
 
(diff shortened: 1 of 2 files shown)


src/palace/manager/integration/license/opds/opds2/api.py
```diff
@@ -48,28 +74,155 @@ def __init__(self, _db: Session, collection: Collection):
...
+                and person_scoped_affiliation
+            ):
+                parameters[TemplateVariable.SAML_PERSON_SCOPED_AFFILIATION] = (
+                    person_scoped_affiliation
+                )
+
+            # Check if any requested SAML variables are still missing from parameters.
+            # This will be true if the patron lacks the required SAML credentials.
+            missing_saml_variables = requested_saml_variables - parameters.keys()
+            if missing_saml_variables:
```
tdilauro, line 159: Minor: Could use ye olde Walrus operator here, but the linting might make that harder to read.

```suggestion
            if (
                missing_saml_variables := requested_saml_variables
                - parameters.keys()
            ):
```

src/palace/manager/integration/license/opds/opds2/api.py
```diff
@@ -48,28 +74,155 @@ def __init__(self, _db: Session, collection: Collection):
...
+    def _get_saml_token_template_parameters(
+        self, patron: Patron
+    ) -> tuple[str | None, list[str] | None]:
+        """Get the SAML token arguments for a patron
+
+        :param patron: The patron to get parameters for
+
+        :return: A tuple containing the SAML entity ID and a list of person scoped affiliations,
+            or (None, None) if the patron has no SAML credentials, or they cannot be
+            extracted.
```
tdilauro, line 86: This seems very specific to the T&F integration, but that might be because that's all we need to support right now (and all the SAML token properties we can support with the current `TemplateVariable` enum).

It might be helpful to add a comment that ties this to the SAML-related members of `TemplateVariable` and/or vice versa.

src/palace/manager/integration/license/opds/opds2/api.py
jonathangreen: I initially had it using walrus, but it was really ugly after formatting, so I changed it back

PR #2774, opened 2025-10-06
Handle unknown collection protocols gracefully

jonathangreen:
## Description

This bugfix ensures that collections with unknown protocols (e.g., from removed integrations) are handled gracefully with warnings instead of causing application crashes.

## Motivation and Context

When an integration is removed from the codebase, any existing collections in the database with that protocol would cause the application to crash when trying to instantiate them. This creates a problematic deployment scenario where code removal would break production systems.

This change implements graceful degradation by catching `LookupException` and logging warnings instead of crashing, allowing for a safer two-step migration: (1) remove code, (2) clean up database.

## How Has This Been Tested?

- Added unit test `test_unknown_protocol` in `test_patron_activity.py` that verifies patron activity sync handles unknown protocols gracefully
- Added unit test `test_circulation_api_unknown_protocol` in `test_collection.py` that verifies the collection circulation API logs warnings when protocol is unknown
- Ran full test suite - all existing tests pass
- Manually verified warning messages are logged correctly

## Checklist

### DIFF
--- a/src/palace/manager/api/circulation_manager.py
+++ b/src/palace/manager/api/circulation_manager.py
@@ -44,6 +44,7 @@ from palace.manager.feed.annotator.circulation import (
 from palace.manager.integration.patron_auth.saml.controller import SAMLController
 from palace.manager.service.analytics.analytics import Analytics
 from palace.manager.service.container import Services
+from palace.manager.service.integration_registry.base import LookupException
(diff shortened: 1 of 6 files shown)


tdilauro (approved):
Looks good to me! 🛠️

PR #2780, opened 2025-10-06
Fix flaky tests with associated_collections ordering

jonathangreen:
## Description

Fixed multiple flaky tests that were occasionally failing in CI due to non-deterministic collection ordering.

Five test assertions were comparing `associated_collections` using list comparisons (order-dependent), but the database doesn't guarantee a specific order for these collections. This caused intermittent failures in CI when the collections were returned in a different order.

Changed all assertions to use set comparisons instead for consistency.

## Motivation and Context

These tests were failing intermittently in CI with assertion errors showing that collections were in a different order than expected. The fixes ensure all affected tests are stable regardless of the database query result ordering.

## How Has This Been Tested?

- Ran all affected test classes locally:
  - `TestDatabaseBackedWorkList` - 8 tests passed
  - `TestCrawlableFeed::test_crawlable_library_feed` - passed
  - `TestFilter::test_from_worklist` - passed
- Tests run in CI
- All changes are minimal and only affect test assertion logic, not application code
- Pattern is consistent with existing set comparisons already used in the same test files

## Checklist

### DIFF
--- a/tests/manager/api/controller/test_crawlfeed.py
+++ b/tests/manager/api/controller/test_crawlfeed.py
@@ -85,10 +85,10 @@ class TestCrawlableFeed:
         lane = kwargs.pop("worklist")
         assert isinstance(lane, CrawlableCollectionBasedLane)
         assert library.id == lane.library_id
-        assert library.associated_collections == [
+        assert set(library.associated_collections) == {
             active_collection,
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
Looks good!

PR #2781, opened 2025-10-07
Fix race condition in IntegrationConfiguration.context_update (PP-2787)

jonathangreen:
## Description

Fixed a race condition in `IntegrationConfiguration.context_update` that could cause lost updates when multiple processes modified the context simultaneously.

The method now uses PostgreSQL's JSONB concatenation operator (`||`) for atomic database-level merging instead of in-memory updates.

## Motivation and Context

The original implementation used a read-modify-write pattern:
```python
self.context.update(new_context)
flag_modified(self, "context")
```

This created a race condition where:
1. Process A reads context: `{"key1": "val1"}`
2. Process B reads context: `{"key1": "val1"}`
3. Process A updates: `{"key1": "val1", "key2": "val2"}` and commits
4. Process B updates: `{"key1": "val1", "key3": "val3"}` and commits
5. Result: `{"key1": "val1", "key3": "val3"}` ❌ **key2 is lost!**

This couldn't happen as we currently use the context, since only one Celery task can update it at a time. In PP-2787 I want to use `context` to store the last time a collection was reaped, and this expanded use means that we need to be more aware of data races. 

## How Has This Been Tested?

- ✅ Added `test_context_update_basic` - tests basic functionality
- ✅ Added `test_context_update_overwrite` - tests overwriting existing keys
- ✅ Added `test_context_update_no_session` - tests error handling
- ✅ Added `test_context_update_atomic` - **simulates concurrent updates from separate database sessions** to verify that both updates are preserved
- ✅ All existing tests pass, including `test_auth_token_feed` which uses this method
- ✅ Type checking passes with mypy

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/model/integration.py
+++ b/src/palace/manager/sqlalchemy/model/integration.py
@@ -10,13 +10,15 @@ from sqlalchemy import (
     Index,
     Integer,
     Unicode,
+    cast,
     select,
+    update,
 )
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.orm import Mapped, Query, Session, relationship
-from sqlalchemy.orm.attributes import flag_modified
 
+from palace.manager.core.exceptions import PalaceValueError
 from palace.manager.integration.goals import Goals
 from palace.manager.sqlalchemy.model.base import Base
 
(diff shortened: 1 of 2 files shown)


jonathangreen:
No. We wouldn't have seen it anywhere, based on how we are currently using `context`. It isn't used very widely, and only one collection import is happening at a time, so there wouldn't have been an opportunity for it to happen. 

I want to use `context` more widely in PP-2728 though, to track when a collection is reaped, so I wanted to make sure it couldn't happen.

PR #2788, opened 2025-10-07
Fix flaky test in LibraryAllTitleReportTable

jonathangreen:
## Description

Fixed a flaky test in `test_included_collections` that was failing intermittently due to non-deterministic ordering of integration IDs.

## Motivation and Context

The test was comparing `integration_ids` as ordered lists when verifying mock calls, but the order of IDs returned from the database query is not guaranteed. This caused random test failures when the IDs appeared in a different order.

The fix changes the assertion to compare sets instead of lists, making the test order-independent while still verifying the correct integration IDs are passed.

## How Has This Been Tested?

The existing test suite validates the fix. The test now passes consistently regardless of the order in which integration IDs are returned.

## Checklist


tdilauro (approved):
✅

PR #2791, opened 2025-10-08
Update project dependencies

jonathangreen:
## Description

Update Poetry lock file with latest dependency versions. This updates numerous packages to their latest compatible versions within the constraints defined in pyproject.toml.

## Motivation and Context

Dependabot should be keeping all our dependencies up to date, but it seems like it has some issue with transitive dependencies and poetry, where it doesn't always keep the transitive dependencies up to date.

This PR was made by running:
```
poetry lock --regenerate
```

This caused a couple dependencies whose most recent versions we have issues with to update:
- [lxml](https://github.com/ThePalaceProject/circulation/pull/2752)
- [pydantic](https://github.com/ThePalaceProject/circulation/pull/2783)
- [types-aws-xray-sdk](https://github.com/ThePalaceProject/circulation/pull/2582)

So I went back and pinned their versions in pyproject.toml and re-ran `poetry lock --regenerate`.

## How Has This Been Tested?

- Running mypy
- Running tests

## Checklist

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -24,14 +24,14 @@ tz = ["tzdata"]
 
 [[package]]
 name = "amqp"
-version = "5.2.0"
+version = "5.3.1"
 description = "Low-level AMQP client for Python (fork of amqplib)."
 optional = false
 python-versions = ">=3.6"
 groups = ["main", "dev"]
 files = [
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
👍🏽

PR #3199, opened 2026-04-02
Content-hash-based change tracking for data imports

jonathangreen:
## Description

Replaces the timestamp-only change-detection logic in the data import pipeline with a content-hash-based system. Previously, `BibliographicData` and `CirculationData` used only the data source's "last updated" timestamp to decide whether to re-apply incoming data to an `Edition` or `LicensePool`. This caused two problems:

- Re-publishing the same content with a newer timestamp triggered unnecessary writes and work creation.
- A real change arriving with the same (or missing) timestamp was silently skipped.

This branch reverts the original timestamp-throttling PR (#3198) and replaces it with a proper content-hash approach. A SHA-256 hash of the canonical, serialized form of the incoming data is stored on the database record after each import. Subsequent imports compare both the timestamp and the hash before deciding whether to apply an update.

**Key changes:**

- New `json_hash()` / `json_canonical()` utilities (`util/json.py`) produce a stable, order-independent SHA-256 fingerprint of any JSON-serializable structure.
- `BaseMutableData` gains `updated_at`, `created_at`, `as_of_timestamp`, `calculate_hash()`, and `should_apply_to()`. The `should_apply_to()` method is now the single decision point for both bibliographic and circulation data.
- `BibliographicData.has_changed()` and `CirculationData.has_changed()` are removed and replaced by the shared `should_apply_to()` logic.
- `Edition` and `LicensePool` each gain an `updated_at_data_hash` column. `LicensePool` also gains `created_at` and `updated_at` columns to track when its `CirculationData` was first and most recently imported.
- Individual-license pools (e.g. ODL) always re-apply availability even when the hash matches, because license availability can change as licenses expire independently of feed content.
- `LicensePool.created_at` is set to `as_of_timestamp` (the data source's own availability date) rather than the import-time timestamp, preserving the "Added to Collection" feed sort order.
- Database migration `f98e4049c87d` adds all four new columns.

**Bug fixes (addressed during review):**

- `_update_edition_timestamp` now updates `updated_at` and `updated_at_data_hash` together (or not at all), preserving the invariant that the hash always reflects the content as of `updated_at`. Previously, force-applying stale data via `even_if_not_apparently_updated=True` overwrote the hash without advancing the timestamp, leaving the two fields inconsistent.
- `_update_edition_timestamp` now uses `<=` instead of `<` when comparing timestamps. The previous strict less-than caused an infinite re-apply loop: when incoming data had the same timestamp as the stored record but different content, the update ran on every import but the new hash was never persisted.
- `PublicationImportResult.changed` is now `needs_apply or called_circulation_apply`. Previously, when bibliographic data was unchanged but an ODL circulation task was queued, `changed` was `False`, causing `found_unchanged_publication` to return `True` and halt feed pagination after the first page.
- OPDS import task now passes `apply_circulation` to `importer.import_feed`, restoring the fallback path for "bibliographic unchanged, circulation changed". Previously this path was dead code, causing circulation-only updates to be silently skipped.

## Motivation and Context

https://ebce-lyrasis.atlassian.net/browse/PP-3997

The original `has_changed()` implementation only compared timestamps, which is insufficient: a data source can re-publish identical content with a newer timestamp, or publish changed content with the same timestamp. Content hashing is the correct primitive for detecting genuine data changes and avoiding redundant imports.

## How Has This Been Tested?

- Updated unit tests for `BibliographicData` and `CirculationData` cover the new `should_apply_to()` logic, including the null-hash bootstrap case, the timestamp-is-older short-circuit, the hash-match skip, the equal-timestamp/changed-content case, and the stale force-apply invariant.
- New unit tests for `json_canonical()` and `json_hash()` verify ordering stability across dict keys, list items, and float precision.
- New test for the ODL pagination fix verifies that `found_unchanged_publication` stays `False` when only a circulation task is queued.
- All existing integration tests for Boundless, OPDS, ODL, and Overdrive importers pass with the updated field names (`updated_at` in place of `data_source_last_updated`).
- Full test suite run via `tox -e py312-docker -- --no-cov`.

Manually tested OverDrive locally by me (@dbernstein) with a non-advantage collection..  It seems to be working as advertised.  Before merging I will confirm that it is working with a parent + two child advantage accounts.

## Checklist


src/palace/manager/data_layer/base/mutable.py
```diff
@@ -72,3 +91,62 @@ def load_primary_identifier(
...
+        ):
+            # We don't have a db object, or the db object has never been updated.
+            # We should apply this data.
+            return True
+
+        if self.as_of_timestamp < db_object.updated_at:
+            # The data we have is strictly older than what is stored, no update needed.
+            return False
+
+        return self.calculate_hash() != db_object.updated_at_data_hash
```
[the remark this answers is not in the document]
dbernstein: This issue will be addressed: Fix by moving the hash update inside the same conditional as the timestamp
    update, so both fields are always kept in sync.

dbernstein (approved):
I think this one is good to go.  @jonathangreen  I thoroughly tested our combined changes.  It's working as expected.  Claude made a bunch of suggestions which I thought were useful so those went in.   I also was able to remove some effectively dead and confusing code (ie the removal of the early exit routine - last commit).  Since we both worked on it, you may want to take a look at it before I merge it?

jonathangreen:
Sounds good @dbernstein. I'll take a run through it today before you merge.

src/palace/manager/data_layer/base/mutable.py
```diff
@@ -24,11 +29,25 @@ class BaseMutableData(BaseModel, LoggerMixin):
         validate_assignment=True,
     )
 
-    _data_source: DataSource | None = None
-    _primary_identifier: Identifier | None = None
+    _data_source: DataSource | None = PrivateAttr(default=None)
+    _primary_identifier: Identifier | None = PrivateAttr(default=None)
+    # Lazily computed and cached for the lifetime of this object. Cleared by
+    # __setattr__ whenever a public field is mutated so the hash stays consistent.
+    _hash_cache: str | None = PrivateAttr(default=None)
```
jonathangreen, line 36: This cache is a nice addition.

src/palace/manager/data_layer/circulation.py
```diff
@@ -227,6 +220,20 @@ def apply(
         pool = None
         if collection:
             pool, ignore = self.license_pool(_db, collection)
+            # Skip circulation data update if the content hasn't changed, UNLESS we
+            # have individual license objects that may have expired since the last
+            # import (ODL-style pools). License expiry is time-dependent and cannot
+            # be detected by content hashing alone.
```
jonathangreen, line 226: I'd really like to see the expiration be handled in its own process instead of tangled up in change tracking. I think this carve out makes it kind of hard to reason able ODL-style pools, and means that they cannot participate in change tracking which I think would be beneficial for them.

I'm okay with this going in, but I'd like to make a follow up to have an async task to process license expiry, so its not tied together here.

jonathangreen (commented):
Looks good @dbernstein. I'm okay with this getting merged. I am concerned about the queue management we will have to do when this goes in though, as a full re-import of everything will probably fill all of our queues.

src/palace/manager/data_layer/circulation.py
dbernstein: That seems reasonable.  Here's the ticket: https://ebce-lyrasis.atlassian.net/browse/PP-4177

PR #3204, opened 2026-04-06
Refresh OverDrive collection token every 30 days (PP-3937)

dbernstein:
## Description
Add a two-layer cache for OverDrive collection tokens so they are
automatically refreshed as OverDrive rotates them, without requiring a
process restart or configuration change.
## Motivation and Context
OverDrive is retiring legacy collection tokens within the next 3–4 months
and moving to a model where tokens must be refreshed periodically.
`collectionToken` is embedded in the library account response
(`GET /v1/libraries/{id}`) and is required for all collection-scoped API
calls (search, product listings, availability, metadata).
Two problems existed in the prior implementation:
1. **DB cache had no TTL.** `get_library()` and `get_advantage_accounts()`
   called `Representation.get()` with no `max_age`, so the library document
   (and the `collectionToken` within it) could persist in the
   `representations` table indefinitely.
2. **In-memory cache was unbounded.** `_collection_token` was a plain
   `str | None` set once and never cleared. Flask workers hold one
   `OverdriveAPI` instance per collection for the entire process lifetime
   (`CirculationManager.load_settings`), so a rotated token would never be
   picked up without a full process restart or an admin config change.
### Solution: two-layer cache
| Layer | Constant | TTL | Purpose |
|---|---|---|---|
| In-memory (`_cached_collection_token`) | `COLLECTION_TOKEN_MAX_AGE` | 5 minutes | Avoids a DB hit on every request |
| DB (`representations` table) | `LIBRARY_MAX_AGE` | 30 days | Avoids a network hit on every 5-minute miss |
A long-lived Flask worker re-checks the DB every 5 minutes and the DB
re-fetches from OverDrive every 30 days. Rotated tokens are picked up
within 5 minutes.
## How Has This Been Tested?
- `test_collection_token` — verifies the in-memory cache is populated on
  first access and reused on subsequent calls within `COLLECTION_TOKEN_MAX_AGE`.
- `test_collection_token_cache_expires` — verifies that aging the
  in-memory cache past `COLLECTION_TOKEN_MAX_AGE` causes `get_library` to
  be called again and returns the new token.
- `test_collection_token_error` — verifies that an `errorCode` in the
  library response raises `CannotLoadConfiguration`.
- `test_get_library_passes_max_age` — verifies `LIBRARY_MAX_AGE` is
  forwarded to `Representation.get()` in `get_library()`.
- `test_get_library_refreshes_stale_token` — end-to-end DB staleness test:
  ages the `Representation` record past `LIBRARY_MAX_AGE` and verifies the
  next call re-fetches from the network.
- `test_get_advantage_accounts_passes_max_age` — verifies `LIBRARY_MAX_AGE`
  is also forwarded in the advantage accounts `Representation.get()` call.
## Checklist

### DIFF
--- a/src/palace/manager/integration/license/overdrive/api.py
+++ b/src/palace/manager/integration/license/overdrive/api.py
@@ -241,6 +241,11 @@ class OverdriveAPI(
 
     EVENT_DELAY = datetime.timedelta(minutes=120)
 
+    # Maximum age of the cached library document before it is re-fetched.
+    # OverDrive periodically rotates collection tokens, so this ensures we
+    # always have a current token without manual intervention.
+    LIBRARY_MAX_AGE: datetime.timedelta = datetime.timedelta(days=30)
+
     TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
     NEXT_REL = "next"
@@ -531,6 +536,11 @@ class OverdriveAPI(
     def get_library(self) -> dict[str, Any]:
         """Get basic information about the collection, including
         a link to the titles in the collection.
+
+        The response is cached in the ``representations`` table and refreshed
+        at most once every :attr:`LIBRARY_MAX_AGE`. This ensures the
+        ``collectionToken`` embedded in the response stays current, since
+        OverDrive periodically rotates collection tokens.
         """
         url = self._library_endpoint
         with self.lock:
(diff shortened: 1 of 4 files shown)


jonathangreen (approved):
I approved this @dbernstein, but I think there is some interaction between the two layers of caching here that are not ideal. I added some comments with my thoughts for consideration. 

I don't think we really need two layers of caching here. It makes it kind of hard to reason about what is going on. I think we might be better off just dropping the in-memory cache and caching via `Representation.get`.

PR #3213, opened 2026-04-08
Prevent Overdrive double-start by adding workflow lock guard to import_collection_group (PP-4025)

dbernstein:
## Description
Add a workflow lock guard at the `import_collection_group` level to prevent
redundant Overdrive import chains from being created while another workflow
is already in progress.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-4025

When Celery Beat fires every 15 minutes, `import_all_collections` →
`import_collection_group` would unconditionally create and enqueue a new
`import_collection` chain for each collection. The only duplicate-start
guard lived inside `import_collection` itself (first-page check against
the workflow lock).
If the workflow lock (2-hour TTL) expired between pages — for example due
to a queue backlog — a subsequent Beat tick could create a brand-new chain
that acquired a fresh lock and started a full-history import concurrently
with the still-queued continuation pages of the original import. Both would
log `Modified since: None` and process from epoch, causing duplicate work
and unexpected behavior.
The fix adds a `locked()` check in `import_collection_group` before chain
creation. If the workflow lock key already exists in Redis (regardless of
which UUID holds it), the task returns the import-skipped payload without
creating a chain. The `import_collection`-level guard remains the
definitive enforcement point; this is defense-in-depth.

## How Has This Been Tested?
Added `test_import_collection_group_skips_when_lock_held` to
`TestImportCollectionGroup`, which pre-acquires the workflow lock in Redis
and asserts that `import_collection_group` returns the skip payload without
calling `chain`. All existing `TestImportCollectionGroup` tests continue to
pass.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/overdrive.py
+++ b/src/palace/manager/celery/tasks/overdrive.py
@@ -240,7 +240,7 @@ def import_collection_group(
     import_all: bool = False,
     modified_since: datetime.datetime | None = None,
     start_time: datetime.datetime | None = None,
-) -> dict[str, Any]:
+) -> dict[str, Any] | ImportSkippedPayload:
     """Import an Overdrive collection and all its child (Advantage) collections.
 
     This task orchestrates the import of a parent Overdrive collection and chains
--- a/src/palace/manager/celery/tasks/overdrive.py
+++ b/src/palace/manager/celery/tasks/overdrive.py
@@ -261,8 +261,20 @@ def import_collection_group(
                           See import_collection docstring for detailed behavior.
     :param start_time: The datetime when this import began. Used to update the
                       collection's timestamp. If None, uses current time.
-    :return: Dictionary containing the chain_id for tracking the async import chain.
+    :return: Dictionary containing the chain_id for tracking the async import chain,
+             or an import-skipped payload if another workflow is already in progress.
     """
+    redis = task.services.redis().client()
+    # Defense-in-depth: skip chain creation if a workflow is already running.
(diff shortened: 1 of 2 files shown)


src/palace/manager/celery/tasks/overdrive.py
```diff
@@ -261,8 +261,20 @@ def import_collection_group(
...
-    :return: Dictionary containing the chain_id for tracking the async import chain.
+    :return: Dictionary containing the chain_id for tracking the async import chain,
+             or an import-skipped payload if another workflow is already in progress.
     """
+    redis = task.services.redis().client()
+    # Defense-in-depth: skip chain creation if a workflow is already running.
+    # This prevents redundant chains from being queued when the workflow lock
+    # expires between pages and a new Beat tick fires before import_collection
+    # can re-acquire and enforce the first-page guard itself.
+    if import_workflow_lock(redis, collection_id, str(uuid4())).locked():
```
jonathangreen, line 272: Nit: The `str(uuid4())` here  is a little confusing (for me anyway). Since its just thrown away, and since `RedisLock` defaults to a random uuid. You could reduce this down to:
 
```suggestion
    if import_workflow_lock(redis, collection_id).locked():
```

You will have to update the type hints on `import_workflow_lock` to:
 ```python
 # importer.py
 def import_workflow_lock(
     client: Redis, collection_id: int, random_value: str | None = None
 ) -> RedisLock:
 ```

jonathangreen (approved):
Looks good!

PR #3219, opened 2026-04-09
Reduce import log noise (PP-4036)

dbernstein:
## Description
Now that we've moved OverDrive imports to celery we're seeing a significant increase in parallel processing.  This is leading to a significant increase in log size.  As part of an effort to keep the logs from overwhelming disks on the scripts instance, this update reduces the log level of two particularly noisy lines.

## Motivation and Context

https://ebce-lyrasis.atlassian.net/browse/PP-4036

## How Has This Been Tested?
Unit tests  pass.

## Checklist


dbernstein:
I'm going to close this for now:  @jonathangreen  suggested it wait until we see the results of the logrotate adjustments.

PR #3221, opened 2026-04-10
Migrate from Poetry to uv for dependency management (PP-4076)

jonathangreen:
## Description

Migrate the entire project from Poetry to uv for dependency management, packaging, and virtualenv orchestration.

## Motivation and Context

Poetry lacks native workspace/monorepo support, which is needed for future project restructuring. uv also provides faster dependency resolution, built-in Python version management, and a simpler CI setup via `astral-sh/setup-uv`.

## How Has This Been Tested?

- `uv lock` resolves 213 packages successfully
- `uv build` produces a clean sdist + wheel (597 files, correct namespace)
- `uv sync --all-groups` installs all dependencies
- `uv run mypy` passes (1123 source files, no issues)
- `uv run pytest --no-cov tests/manager/util/test_log.py` passes (23 tests)
- All pre-commit hooks pass
- Tests pass in CI

## Checklist


docker/Dockerfile.baseimage
```diff
@@ -57,27 +59,22 @@ RUN useradd -ms /bin/bash -U palace && \
...
+COPY --chown=palace:palace uv.lock pyproject.toml /var/www/circulation/
 
 # Setup virtualenv and install our python dependencies.
-# What we install is based on the poetry.lock file in the repo at the time this
+# What we install is based on the uv.lock file in the repo at the time this
 # image is built. These may get out of date, but we always rerun this step when
 # building the final image, so it will be up to date then. This gives is a base
 # to work from which speeds up the final image build.
-RUN python3 -m venv env && \
+RUN uv sync --frozen --no-dev --no-install-project && \
```
tdilauro, line 69: Is `no-editable` desirable here, as well?

tdilauro (approved):
✔️

docker/Dockerfile.baseimage
jonathangreen: `--no-install-project` already skips installing the project package entirely, so `--no-editable` (which controls how the project is installed) has nothing to apply to. They're effectively mutually exclusive.

PR #3222, opened 2026-04-10
Fix revoke links missing from OPDS2 feeds (PP-4056)

jonathangreen:
## Description

Remove the unnecessary type requirement from `_serialize_acquisition_link` in the OPDS2 serializer that was silently dropping links without a media type. This was filtering out revoke links for holds and checkouts, preventing integration partners from being able to cancel holds or return titles early via OPDS2 feeds.

## Motivation and Context

An integration partner reported that revoke links were no longer appearing in OPDS2 loan feeds.

The regression was introduced in #3027, which added `_acquisition_link_type()` a gate in `_serialize_acquisition_link` that required all acquisition links to have a media type, returning `None` and dropping the entire link otherwise. Revoke links are intentionally created without a type (they aren't content delivery links), so they were being silently dropped with only a log error. 

At the time of #3027, the OPDS2 Pydantic models used `StrictLink` which required both `rel` and `type`. However, in #3137, we removed `StrictLink` and replaced it with `opds2.Link`, which has `type` as optional — matching the [OPDS 2.0 spec](https://drafts.opds.io/opds-2.0.html) and the [Readium WebPub Manifest link schema](https://readium.org/webpub-manifest/schema/link.schema.json) where only `href` is required. That PR also renamed `_strict_link()` to `_link()` in the serializer, but did not update `_serialize_acquisition_link` to remove its now-redundant type gate.

Since the underlying `opds2.Link` model already accepts `None` for `type`, the `_acquisition_link_type` check is no longer needed and can simply be removed.

JIRA: PP-3056

## How Has This Been Tested?

- Updated existing OPDS2 serializer tests to verify links without a type are now serialized rather than dropped
- All 40 tests in `test_opds2_serializer.py` pass
- mypy passes with no errors

## Checklist

### DIFF
--- a/src/palace/manager/feed/serializer/opds2.py
+++ b/src/palace/manager/feed/serializer/opds2.py
@@ -242,9 +242,7 @@ class OPDS2Serializer(SerializerInterface[dict[str, Any]], LoggerMixin):
             )
 
         for acquisition in data.acquisition_links:
-            acq_link = self._serialize_acquisition_link(acquisition)
-            if acq_link is not None:
-                links.append(acq_link)
+            links.append(self._serialize_acquisition_link(acquisition))
         return links
 
     def _serialize_link(self, link: Link) -> opds2.Link:
@@ -255,14 +253,11 @@ class OPDS2Serializer(SerializerInterface[dict[str, Any]], LoggerMixin):
             title=link.title,
         )
 
-    def _serialize_acquisition_link(self, link: Acquisition) -> opds2.Link | None:
-        link_type = self._acquisition_link_type(link)
-        if link_type is None:
-            return None
+    def _serialize_acquisition_link(self, link: Acquisition) -> opds2.Link:
         return self._link(
             href=link.href,
             rel=link.rel or opds2.AcquisitionLinkRelations.acquisition,
-            type=link_type,
+            type=self._resolve_type(link.type),
             title=link.title,
             properties=self._serialize_acquisition_properties(link),
             templated=link.templated,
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
@jonathangreen : looks good.  Below are a couple of things CLAUDE surfaced.  Neither one of them requires a change, but I offer it for your consideration before merging.


1. Removed indirect acquisition type fallback (potential regression)
The old _acquisition_link_type() had a fallback:

def _acquisition_link_type(self, link: Acquisition) -> str | None:
    if link.type:
        return self._resolve_type(link.type)
    for indirect in link.indirect_acquisitions:  # <-- fallback behavior
        indirect_type = self._resolve_type(indirect.type)
        if indirect_type:
            return indirect_type

The PR removes this by replacing it with a direct _resolve_type(link.type) call. This means acquisition links with no direct type but with indirect acquisitions will now serialize with type: None instead of the indirect type.

The deleted tests (test_acquisition_link_type_fallback_to_indirect, test_acquisition_link_type_fallback_to_semantic_indirect) confirm this is intentional — but it's worth explicitly calling out in the PR description whether this is a safe behavioral change. If any integrations relied on the indirect type appearing as the top-level link type, this could be a quiet regression.

2. Inconsistency in _serialize_publication_links
For other_links, the method still logs an error and skips links without a type (src/palace/manager/feed/serializer/opds2.py:231-233). This is now inconsistent with the new behavior for acquisition links. Not necessarily wrong, but worth a comment or issue for follow-up.

PR #3223, opened 2026-04-10
Fix KeyError when Overdrive API response missing 'products' key (PP-3379)

dbernstein:
## Description

Handle the case where the Overdrive API returns a response without a
`"products"` key in `fetch_book_info_list`. Previously, this caused an
unhandled `KeyError` that crashed the `overdrive.import_collection`
Celery task entirely. The fix uses `.get("products")` and logs a warning
with the endpoint URL and response body when the key is absent, then
raises a BadResponseException.  The assumption here is that the issue
is transient and will be resolved on a retry.  We've logged a warning with the 
response object to help us easily understand what's in the response.

## Motivation and Context
https://ebce-lyrasis.atlassian.net/browse/PP-3979

The `overdrive.import_collection` task was raising: KeyError: 'products'

at `api.py` line 674 whenever the Overdrive API returned a response
without the expected `"products"` key. This killed the entire task,
blocking collection imports for affected libraries.
## How Has This Been Tested?
Added a new unit test `test_fetch_book_info_list_missing_products_key`
in `TestSyncBookshelf`. All existing `test_fetch_book_info_list*` tests
continue to pass.
## Checklist

### DIFF
--- a/src/palace/manager/integration/license/overdrive/api.py
+++ b/src/palace/manager/integration/license/overdrive/api.py
@@ -711,6 +711,13 @@ class OverdriveAPI(
             async_task_list = list()
             response_products = data.get("products")
             if response_products is None:
+                self.log.warning(
+                    f"Overdrive response missing 'products' key for endpoint {endpoint.url}.",
+                    extra={
+                        "palace_response_data": data,
+                        "palace_response_status_code": response.status_code,
+                    },
+                )
                 raise BadResponseException(
                     endpoint.url,
                     f"Overdrive response missing 'products' key. Response data: {data}",
--- a/src/palace/manager/integration/license/overdrive/api.py
+++ b/src/palace/manager/integration/license/overdrive/api.py
@@ -712,8 +712,11 @@ class OverdriveAPI(
             response_products = data.get("products")
             if response_products is None:
                 self.log.warning(
-                    f"Overdrive response missing 'products' key for endpoint {endpoint.url}. "
-                    f"Response data: {data}"
+                    f"Overdrive response missing 'products' key for endpoint {endpoint.url}."
+                    extra={
+                        "palace_response_data": data,
+                        "palace_response_status_code": response.status_code,
+                    },
                 )
                 return [], next_endpoint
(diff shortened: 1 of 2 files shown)


src/palace/manager/integration/license/overdrive/api.py
```diff
@@ -709,7 +709,13 @@ async def fetch_book_info_list(
                 BookInfoEndpoint(next_url) if next_url else None
             )
             async_task_list = list()
-            response_products = data["products"]
+            response_products = data.get("products")
+            if response_products is None:
+                self.log.warning(
+                    f"Overdrive response missing 'products' key for endpoint {endpoint.url}. "
+                    f"Response data: {data}"
+                )
```
jonathangreen, line 720: Minor: When logging structured data like this, I like to use `extra`, so it ends up in our logs as json, rather then json encoded in a json string, which is hard to read and query on.

```suggestion
                self.log.warning(
                    f"Overdrive response missing 'products' key for endpoint {endpoint.url}."
                    extra={
                        "palace_response_data": data,
                        "palace_response_status_code": response.status_code,
                    },
                )
```

jonathangreen (approved):
A couple comments to consider before merging:

- The PR title says PP-3379 but the description links to PP-3979. Looks like PP-3379 is an Android ticket, so this should be fixed before merging, so the wrong ticket doesn't end up in git log / release notes.
- Do we know right now what causes this? If we think this is a transient issue with Overdrive, we should raise a retry-able error instead. As this PR is written, it will look like we successfully got this page. So a transient malformed page will either advance to `next_endpoint` or finish the import and update the timestamp, permanently skipping every title on that page. You are deeper in the Overdrive code then I am, so perhaps I'm misunderstanding here, but this seems like a concern to me.

dbernstein:
@jonathangreen :  that is a good point re the fact the possibility of bailing out of an import early due to a transient issue.  I think I'll need to study the response data before determining whether it is safe to return with an empty product list.  I'll raise a retryable error for now and circle back to see what's what.

PR #3229, opened 2026-04-10
Set explicit target Python versions for black

jonathangreen:
## Description

Explicitly set `-t py312`, `-t py313`, and `-t py314` target version flags for black in the pre-commit configuration, and apply the resulting formatting changes.

## Motivation and Context

In #3221, a change to `python-requires` in `pyproject.toml` allows black to auto-detect the supported Python versions. This auto-detection brings in unexpected formatting changes (parenthesized `with` statements, trailing commas on `**kwargs`, etc.). By explicitly setting the target versions in pre-commit and applying the formatting changes here, we can land #3221 without mixing unrelated formatting noise into that PR.

Reference: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#t-target-version

## How Has This Been Tested?

- `pre-commit run -a` passes cleanly with the updated configuration.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,6 +41,13 @@ repos:
     hooks:
       - id: black
         name: Run black
+        args:
+          - -t
+          - py312
+          - -t
+          - py313
+          - -t
+          - py314
 
   - repo: https://github.com/PyCQA/isort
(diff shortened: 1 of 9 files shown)


tdilauro (approved):
✔️

PR #3230, opened 2026-04-13
Extract palace.opds into a uv workspace package (PP-4074)

jonathangreen:
## Description

Move `palace.manager.opds` into a new workspace package `palace-opds` (namespace `palace.opds`) and retarget all callers to `palace.opds.*`. `palace-opds` depends on `palace-util` for the small set of cross-cutting symbols the OPDS models need (`PalaceValueError`, `utc_now` / `to_utc`, `LoggerMixin`); it depends on no other Palace package — same dependency-graph rule documented in #3231.

## Motivation and Context

The OPDS Pydantic model tree describes formats exchanged across system boundaries — other Palace services and ingestion tools could reuse it without dragging in the full `palace-manager` application. Extracting it as a workspace member lets us publish it independently in the future while keeping the monorepo as the source of truth.

JIRA: PP-4074

## How Has This Been Tested?

- `uv sync` resolves the workspace cleanly with all three members.
- `uv build --package palace-opds` produces a wheel containing only `palace/opds/**`.
- `mypy` passes (1126 source files).
- `pytest tests/palace_opds tests/palace_util` — 232 tests pass.
- Full DB-requiring suite (`tox -e py312-docker`) will run in CI.

## Conventions

This PR follows the workspace-member conventions established in #3231:

- Tests live under `tests/palace_opds/` (mirrors the `tests/palace_util/` layout) so they share root pytest fixtures and conftest plugins automatically.
- No re-export shims in `palace.manager` — callers import from `palace.opds.*` directly.
- Workspace-member `pyproject.toml` files are copied (manifest-only) into the Docker cache layer so `uv` can resolve workspace deps.
- `palace-opds/README.md` follows the same template as `palace-util/README.md` — purpose, scope, dependency-graph hard rule, stability, development.

## Checklist


tdilauro (approved):
This looks good to me, too. 🚀 

One overall comment: Since this is just for OPDS models, should it be named `palace-opds-models` to avoid confusion with other Palace OPDS stuff? If so, I'd say that now is the time to do it.

jonathangreen:
I think `palace-opds` works as a package name. Kind of like `pyopds2`.

PR #3238, opened 2026-04-14
Publish palace-util and palace-opds to PyPI on release (PP-4076)

jonathangreen:
## Description

Two things:

1. **Dynamic versioning** for both workspace packages. Each package's `pyproject.toml` now declares `dynamic = ["version"]` and sources the version from a committed `_version.py` stub via hatchling's built-in `regex` version source. The stub carries a dev placeholder (`0.0.0.dev0`, `__commit__ = None`) so `uv build --package …` works out-of-box for local and CI builds without an explicit version override. The release workflow overwrites `_version.py` with real dunamai-computed values right before `uv build`, so the wheel metadata picks up the real version.

2. **`publish-pypi.yml` workflow** triggered by published GitHub Releases. Matrix-builds both packages in parallel, uses PyPI's [Trusted Publishing (OIDC)](https://docs.pypi.org/trusted-publishers/) flow. The workflow also pins the intra-workspace dep, so `palace-opds` always depends on the most recent palace-util.

## Motivation and Context

With `palace-util` and `palace-opds` extracted as standalone packages, they should be installable from PyPI so downstream Palace services (and the wider OPDS-consuming community for `palace-opds`) can depend on them without pulling in the `palace-manager` monorepo. `palace-manager` itself is intentionally NOT published here — it needs additional work first (alembic and friends prevent a clean wheel build).

## One-time PyPI setup

~~I need to setup trusted publishing in pypi before we cut a release with this. This is a work in progress. 🚧~~
This setup is complete now.

## How Has This Been Tested?

I set this up in my own fork, and published test packages to test.pypi.org. Using the same workflow they published correctly for me:
- https://github.com/jonathangreen/circulation/actions/runs/24460001637
- https://test.pypi.org/project/palace-opds/
- https://test.pypi.org/project/palace-util/

## Checklist


tdilauro (approved):
🔨

PR #3246, opened 2026-04-15
Drop unused pydantic[email] extra from palace-opds

jonathangreen:
## Description

`palace-opds`'s `pyproject.toml` declared `pydantic[email]` but nothing in the package uses `EmailStr` / `email_validator` — the only email field (the LCP license owner email in `palace.opds.lcp.license`) is a plain `str`. Drop the extra so anyone installing `palace-opds` standalone doesn't pull in `email-validator` and its transitive deps.

## How Has This Been Tested?

- `uv lock` — only the palace-opds edge changed; no other resolution churn.
- `mypy` passes.
- `pytest tests/palace_opds` — 174 tests pass.

## Checklist


tdilauro (approved):
🪛

PR #3272, opened 2026-04-21
Rename "Filtered genres" setting to "Filtered categories" (PP-4074)

jonathangreen:
## Description

Renames the user-facing label for the library setting previously shown as "Filtered genres" to "Filtered categories". The description is updated to match, with "(genres)" kept as a parenthetical hint. The underlying Pydantic field name `filtered_genres` is intentionally unchanged — the values it filters against are referred to as genres throughout the codebase, so renaming the variable would create a wider divergence than it would resolve. A comment on the field documents this.

## Motivation and Context

PP-4074 asked for the admin-facing terminology to say "categories" instead of "genres". This PR makes the UI-only change.

## How Has This Been Tested?

No behavior changes — label/description strings only. Existing tests pass unchanged.

## Checklist

### DIFF
--- a/src/palace/manager/integration/configuration/library.py
+++ b/src/palace/manager/integration/configuration/library.py
@@ -405,11 +405,14 @@ class LibrarySettings(BaseSettings):
             level=Level.SYS_ADMIN_OR_MANAGER,
         ),
     ] = []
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
🪛

PR #3276, opened 2026-04-21
Default OPDS2+ODL skipped_license_formats to empty (PP-4044)

jonathangreen:
## Description

Change the default value of `OPDS2WithODLSettings.skipped_license_formats` from `["text/html"]` to `[]`. ODL 2.0 integrations will no longer suppress the `text/html` license format by default, which means the streaming delivery mechanism (`STREAMING_DRM`) will be emitted for those titles.

## Motivation and Context

PP-4044. We want streaming web-reader delivery to be active by default for ODL integrations.

A production audit of the `integration_configurations.settings` JSONB column across our environments confirmed that every existing ODL 2.0 row either omits `skipped_license_formats` or already stores `[]` — none store the previous `["text/html"]` default explicitly. Rows that omit the key will pick up the new default and start emitting streaming formats; that behavior change is the goal, and because nothing stores the old default explicitly we don't need a data migration to clear it out.

## How Has This Been Tested?

Existing extractor tests in `tests/manager/integration/license/opds/odl/test_extractor.py` continue to cover the explicit-skip path. The change is a single default-value flip; behavior under explicit configuration is unchanged.

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/opds/odl/settings.py
+++ b/src/palace/manager/integration/license/opds/odl/settings.py
@@ -99,7 +99,7 @@ class OPDS2WithODLSettings(OPDS2ImporterSettings):
             type=FormFieldType.LIST,
             required=False,
         ),
-    ] = ["text/html"]
+    ] = []
 
     loan_limit: Annotated[
         PositiveInt | None,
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🔧 🚀

PR #3284, opened 2026-04-23
Fix Palace Marketplace BISAC N-suffix codes misclassified as Adult

dbernstein:
## Description

Fix `BISACClassifier.scrub_identifier()` to correctly handle the non-standard BISAC codes sent by Palace Marketplace, and add a database migration to trigger reclassification of the affected subjects.

## Motivation and Context

Palace Marketplace sends Juvenile Fiction BISAC codes with an `FB` prefix and an `N` suffix (e.g. `FBJUV000000N`). The scrubber already stripped the `FB` prefix, but the resulting identifier (e.g. `JUV000000N`) was absent from `bisac.csv`. With no canonical name found, classification fell back to the bare stored name (`"Themes"`, `"Family"`, etc.) which carries no audience context, causing these subjects to hit the Adult catch-all rule and be misclassified as Adult.

### Root cause

Official BISAC codes always end with digits. A trailing `N` is a non-standard distributor extension. After FB-stripping, `FBJUV000000N` → `JUV000000N` → not in `bisac.csv` → keyword fallback on `"Themes"` → Adult.

### Fix

Two changes to `scrub_identifier()`:

1. **Strip trailing `N` suffix** — after FB-prefix stripping, any identifier ending in `N` has the `N` removed. This resolves e.g. `FBJUV000000N` → `JUV000000` → `"Juvenile Fiction / General"` → Children/Fiction.

2. **`NON_STANDARD_CODE_ALIASES` dict** — for codes that still don't resolve to a canonical entry after N-stripping, an explicit remap is applied. `FBJUV009001N` strips to `JUV009001`, which is not a valid BISAC code; it is remapped to `JUV013000` (`"Juvenile Fiction / Family / General"`). New entries can be added here when a distributor uses codes absent from the official BISAC list.

The five affected Palace Marketplace codes and their resolutions:

| Stored identifier | After scrubbing | Canonical name |
|---|---|---|
| `FBJUV000000N` | `JUV000000` | Juvenile Fiction / General |
| `FBJUV009000N` | `JUV009000` | Juvenile Fiction / Concepts / General |
| `FBJUV009001N` | `JUV009001` → alias → `JUV013000` | Juvenile Fiction / Family / General |
| `FBJUV022000N` | `JUV022000` | Juvenile Fiction / Legends, Myths, Fables / General |
| `FBJUV038000N` | `JUV038000` | Juvenile Fiction / Short Stories |

### Impact on existing database records

The `subjects.identifier` column is **not changed** — records continue to store the original distributor value (e.g. `FBJUV000000N`). The scrubber transforms the identifier in memory at classification time only. The migration resets `checked=False` on the 5 affected subjects; the `classify_unchecked_subjects` Celery task then calls `work.calculate_presentation()` on every work linked to those subjects, updating the subject's `audience`, `fiction`, `genre`, and `target_age` fields and recalculating each work's overall classification.

### Impact on future imports

Because the fix lives in the scrubber (which runs on every classification), reimported records with these identifiers will be classified correctly without any further intervention.

## How Has This Been Tested?

- Added `test_palace_marketplace_n_suffix_codes` verifying that all 5 FB+N codes classify as Children/Fiction.
- Extended `test_scrub_identifier` with assertions for N-suffix stripping (with and without the `FB` prefix) and for the `NON_STANDARD_CODE_ALIASES` remap.
- All 21 existing BISAC classifier tests continue to pass.

## Checklist


src/palace/manager/core/classifier/bisac.py
```diff
@@ -702,12 +702,31 @@ def genre(cls, identifier, name, fiction, audience):
...
             return identifier
         if identifier.startswith("FB"):
             identifier = identifier[2:]
+        # Some distributors (e.g. Palace Marketplace) append an "N" suffix to
+        # standard BISAC codes (e.g. "FBJUV000000N" becomes "JUV000000N" after
+        # FB-stripping). Official BISAC codes always end with digits, so a
+        # trailing "N" is always a non-standard extension; strip it so the code
+        # resolves to its canonical entry.
+        if identifier.endswith("N"):
+            identifier = identifier[:-1]
```
tdilauro: Minor - We're Python 3.12+, so we just use `removesuffix` here

```suggestion
        identifier = identifier.removesuffix("N")
```

src/palace/manager/core/classifier/bisac.py
```diff
@@ -702,12 +702,31 @@ def genre(cls, identifier, name, fiction, audience):
...
+        # ("Juvenile Fiction / Family / General").
+        "JUV009001": "JUV013000",
+    }
+
     @classmethod
     def scrub_identifier(cls, identifier):
         if not identifier:
             return identifier
         if identifier.startswith("FB"):
             identifier = identifier[2:]
```
tdilauro: Minor - We're Python 3.12+, so can just use `removeprefix` here:


```suggestion
        identifier = identifier.removeprefix("FB")
```

tests/manager/core/classifiers/test_bisac.py
```diff
@@ -344,6 +344,43 @@ def test_feedbooks_bisac(self):
...
+            ),  # → JUV009001 → alias → JUV013000 "Juvenile Fiction / Family / General"
+            (
+                "FBJUV022000N",
+                "Lifestyles",
+            ),  # → JUV022000 "Juvenile Fiction / Legends, Myths, Fables / General"
+            (
+                "FBJUV038000N",
+                "Social situations",
+            ),  # → JUV038000 "Juvenile Fiction / Short Stories"
+        ]
```
tdilauro: Would be better to use `pytest.mark.parametrize` with `ids=...` here. That way, the cases can fail individually without causing other cases not to be tested.

tests/manager/core/classifiers/test_bisac.py
```diff
@@ -359,6 +396,27 @@ def test_scrub_identifier(self):
             "FBFIC015000"
         )
 
+        # Trailing "N" suffixes are stripped. Official BISAC codes always end
```
tdilauro: Same suggestion here, re: `parametrize`. I know this test came in already with these individual assertions, so feel free to ignore, but parametrize would probably be better, if individual cases started failing.

tdilauro (approved):
Looks good to me. ↔️ 🚀

A few suggestions below.

jonathangreen:
@dbernstein I see you already made the switch from `MappingProxyType` -> `frozendict`. Good call, I was going to suggest the same. Might be worth putting a small update in `CLAUDE.md` to mention this as well. I've noticed it always wants to suggest `MappingProxyType`.

dbernstein:
@tdilauro : thanks for the feedback:  I pulled in your commit suggestions and parameterized the test.
@jonathangreen :  PR is up for the CLAUDE.md updates.

PR #3288, opened 2026-04-25
Update BISAC classification codes and subject headings (PP-4187)

dbernstein:
## Description

Merges the updated BISAC classification codes into the existing `bisac.csv` file. Where codes exist in both the old and new files, the updated subject heading is used. All existing codes are preserved.

- 4,226 existing codes updated with new subject headings
- 1,633 new codes added
- 126 old-only codes preserved
- 5,984 total codes (no duplicates, sorted)

## Motivation and Context

The BISAC classification system has been updated with new codes and revised subject headings. This keeps the classifier data current.  Preserving the codes that no longer exist in the updated list ensures that we don't lose any  context. If for example we allowed a JUV* code that is no longer in the updated BISAC list to be removed from our csv, then if that code were to appear in a feed and there was nothing in the feed's subject heading to indicate a juvenile classification, the work could end up being removed from lists with an audience filter of "Children" which is undesirable. 

Jira: PP-4187

## How Has This Been Tested?

Verified via script that no existing codes were lost, duplicates were resolved in favor of the updated headings, and the file is sorted with no duplicates.

## Checklist


tdilauro (approved):
👍🏽

PR #3289, opened 2026-04-27
Add frozendict convention to CLAUDE.md

dbernstein:
## Description

Add an instruction to CLAUDE.md specifying that `frozendict` should be used when creating immutable dictionaries, unless directed otherwise.

## Motivation and Context

This makes the immutable dictionary convention explicit and consistent with the existing guideline to use immutable datastructures for global and class constants.

## How Has This Been Tested?

Documentation-only change; no tests required.

## Checklist


jonathangreen (approved):
Looks good

PR #3295, opened 2026-04-27
Adjust Palace Marketplace juvenile book audience classification thresholds

dbernstein:
## Description
This is a follow on PR to #3284.  After doing some tests I realized that not all of the books identified in the original JIRA (PP-4128) ticket were fixed.  After some more investigation,  it became clear that the migration needed to touch a wider swath of subjects and the JV audience threshold needed a minor adjustment.

So this PR contains two fixes to correctly classify juvenile books from Palace Marketplace as Children rather than Adult:

1. **Migration**: Resets `checked=False` for all BISAC subjects whose identifier starts with `FBJUV`. The previous migration (45f74fdcec18) only targeted a specific set of N-suffix codes; this broader pass catches remaining subjects (e.g. `FBJUV000000`, `FBJUV037000`) that were classified before the FB-prefix scrubber fix landed.

2. **Audience threshold**: Changes the juvenile audience threshold from strictly greater than (`>`) to at least (`>=`) twice the adult weight. Palace Marketplace sends `FBFIC000000` (a generic "Fiction / General" parent category) alongside `FBJUV*` codes for juvenile books, which produced exactly 2× the adult weight in children votes — enough to tie but not enough to win under the old strict check. Applies consistently to the Children, YA, and combined-juvenile checks.

## Motivation and Context
Books from Palace Marketplace with `FBJUV*` BISAC codes and a single `FBFIC000000` code were being classified as Adult because:
- `FBJUV` subjects last classified before the FB-prefix scrubber fix still had `audience=Adult` stored
- Once those were corrected, `FBFIC000000` (1 adult vote) produced a threshold of 2, which the 2 children votes could not exceed under the strict `>` check

## How Has This Been Tested?

- Traced classification logic manually against real subject data from affected books
- Updated the boundary-case unit test in `TestWorkClassifier` to reflect the new `>=` semantics
- Verified mypy reports no issues on changed files

## Checklist

### DIFF
--- a/alembic/versions/20260427_05a95c828149_reset_checked_for_all_fbjuv_subjects.py
+++ b/alembic/versions/20260427_05a95c828149_reset_checked_for_all_fbjuv_subjects.py
@@ -1,0 +1,47 @@
+"""reset_checked_for_all_fbjuv_subjects
+
+Palace Marketplace sends BISAC codes with an "FBJUV" prefix (e.g. FBJUV000000,
+FBJUV009000N). The previous migration only targeted a specific set of known
+identifiers. This migration casts a wider net by resetting checked=False for
+every BISAC subject whose identifier starts with "FBJUV", ensuring the
+classify_unchecked_subjects task reclassifies any remaining affected subjects
+with the corrected scrubber.
+
+Revision ID: 05a95c828149
+Revises: 45f74fdcec18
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
Looks good

PR #3296, opened 2026-04-27
Fix ValueError in calculate_work with retryable InconsistentLicensePoolState (PP-3977)

dbernstein:
## Description

Replace the bare set-unpack `[self.work] = existing_works` in `LicensePool.calculate_work()` with an explicit guard. When `existing_works` is empty, raise a new `InconsistentLicensePoolState` exception (a `BasePalaceException` subclass) instead of crashing with an unhandled `ValueError`. Both Celery apply tasks now include `InconsistentLicensePoolState` in `autoretry_for`, so the task retries up to 5 times with backoff.

## Motivation and Context
JIRA: PP-3977

A widespread `ValueError: not enough values to unpack (expected 1, got 0)` was observed across many circulation manager instances, correlated with OverDrive-only imports. The crash occurs at:

```
licensing.py, calculate_work:
    [self.work] = existing_works   # ValueError when set is empty
```

`existing_works` is built from `self.identifier.licensed_through`. It is empty when the Celery session loads the Identifier (and its eagerly-joined `licensed_through` relationship) before a sibling import transaction commits, leaving the relationship snapshot empty. Silently continuing would create a duplicate Work; retrying allows the next attempt to see the committed pool.

## How Has This Been Tested?

- `test_calculate_work_raises_on_empty_licensed_through` — verifies `InconsistentLicensePoolState` is raised when `licensed_through` is empty.
- `test_inconsistent_license_pool_state_retried` — verifies the task retries 5 times and re-raises after exhausting retries.
- `mypy` — passes with no new errors.

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/model/licensing.py
+++ b/src/palace/manager/sqlalchemy/model/licensing.py
@@ -1235,9 +1235,20 @@ class LicensePool(Base):
                 lp.work = None
                 if lp.presentation_edition:
                     lp.presentation_edition.work = None
-        else:
+        elif existing_works:
             # There is a consensus Work for this Identifier.
             [self.work] = existing_works
+        else:
+            # licensed_through is empty — the session doesn't yet see any pools
+            # for this identifier. This is a transient race condition during
+            # parallel imports; raising here allows the Celery task to retry.
+            from palace.util.exceptions import InconsistentLicensePoolState
+
+            raise InconsistentLicensePoolState(
+                f"No LicensePools found for identifier {self.identifier!r} in "
+                "licensed_through — this is likely a transient session state "
+                "issue caused by a parallel import; the task should be retried."
+            )
 
         if self.work:
             # This pool is already associated with a Work. Use that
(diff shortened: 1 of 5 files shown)


src/palace/manager/sqlalchemy/model/licensing.py
```diff
@@ -1235,9 +1235,20 @@ def calculate_work(
...
                     lp.presentation_edition.work = None
-        else:
+        elif existing_works:
             # There is a consensus Work for this Identifier.
             [self.work] = existing_works
+        else:
+            # licensed_through is empty — the session doesn't yet see any pools
+            # for this identifier. This is a transient race condition during
+            # parallel imports; raising here allows the Celery task to retry.
+            from palace.util.exceptions import InconsistentLicensePoolState
```
dbernstein: Do not use inline imports

dbernstein: Fixed — moved the import to the module level.

dbernstein:
Closing to reopen with branch hosted directly on upstream.

PR #3298, opened 2026-04-28
Add country/state geographic configuration for circulation events (PP-4127)

dbernstein:
## Description

Adds first-class \`country\`, \`state\`, and \`palace_manager_name\` fields to the analytics pipeline, replacing the fragile convention of parsing geographic data from CM instance names.

**Geographic resolution (per-library):**
- \`country\` and \`state\` are optional fields on \`LibrarySettings\`, visible to system admins only
- \`collect_event()\` reads them directly: \`library.settings.country or "US"\` / \`library.settings.state or "All"\`
- No env-var tier, no sitewide GlobalSettings — just the library record

**CM identity field:**
- \`PALACE_REPORTING_NAME\` (already set per-CM in the hosting-playbook compose templates) is read in \`collect_event()\` and embedded as \`palace_manager_name\` on every \`AnalyticsEventData\` event written to S3 JSON

\`country\`, \`state\`, and \`palace_manager_name\` are top-level keys in every S3 JSON event, consumed by the AWS Glue ETL job into Redshift.

## Motivation and Context

JIRA: PP-4127

Canadian libraries use a \`can-<province>\` naming convention in the hosting-playbook, which is unreliable for Redshift geo data. This makes country and state explicit, per-library admin-configurable values propagated through the analytics pipeline. \`palace_manager_name\` makes the originating CM instance unambiguous in Redshift without parsing hostnames.

Companion hosting-playbook PR: ThePalaceProject/hosting-playbook#1012

## Modified files

- \`src/palace/manager/integration/configuration/library.py\` — \`country\`/\`state\` optional fields (sys-admin only)
- \`src/palace/manager/service/analytics/eventdata.py\` — \`country\`, \`state\`, and \`palace_manager_name\` fields on \`AnalyticsEventData\`
- \`src/palace/manager/service/analytics/analytics.py\` — reads \`country\`/\`state\` from library settings and \`PALACE_REPORTING_NAME\` into \`palace_manager_name\` in \`collect_event()\`

## How Has This Been Tested?

- Existing analytics tests extended: \`test_analytics.py\`, \`test_eventdata.py\`
- All affected tests pass against a live local database

## Checklist


jonathangreen (commented):
I'd like to push back on the overall shape here before we land it. I think this is heavier than it needs to be and points in a direction we're actively trying to move away from.

## Architectural concern
Our roadmap is to consolidate all libraries into a single CM instance. That makes any new sitewide/global state a liability: every "one value per CM" setting becomes a "needs to be split across N libraries" migration when we consolidate. This PR introduces three new sitewide surfaces in service of two fields:

- A new `Goals.SITEWIDE_SETTINGS` IntegrationConfiguration goal + migration `0422_a1b2c3d4e5f6`
- A `GlobalSettings` Pydantic class and admin controller (`global_settings.py` + route)
- `PALACE_DEFAULT_COUNTRY` / `PALACE_DEFAULT_STATE` env vars feeding both the DB row's defaults and a no-DB fallback

It also collides conceptually with the existing `service/sitewide.py::SitewideConfiguration` (env-var-driven, instance-level). Having two different "sitewide settings" mechanisms is going to confuse the next person who needs to add one.

## Suggested simplification

I think the actual feature, country/state on Redshift events, only needs the per-library half of this PR:

1. Keep `LibrarySettings.country` / `state` (with sane non-None defaults, e.g. `"US"` / `"All"`, or keep `None` and resolve at the call site).
2. Replace `resolve_geo` with a direct read in `collect_event`: `country = library.settings.country or "US"` etc. The 3-tier cascade collapses to one line, so the helper isn't pulling its weight.
3. Drop everything else: `GlobalSettings`, the admin controller and route, `Goals.SITEWIDE_SETTINGS`, the migration, the env vars, `geo.py`, `test_geo.py`, and the global-settings tests.

That keeps the feature intact, ships less code, doesn't introduce a new sitewide concept, and means every library already has correct values when we move to a single CM hosting all libraries.

## Related ticket

I also think the work here intersects with https://ebce-lyrasis.atlassian.net/browse/PP-4237, in that ticket we need even finer grained data about library locations. I think these two needs should be fulfilled by a single design.

dbernstein:
Hi @jonathangreen :  I took your advice (https://github.com/ThePalaceProject/circulation/pull/3298#pullrequestreview-4230308724).  This one is ready for a re-review.

src/palace/manager/service/analytics/analytics.py
```diff
@@ -55,7 +57,11 @@ def collect_event(
         new_value: int | None = None,
         patron: Patron | None = None,
     ) -> None:
-        session = Session.object_session(library)
+        country = library.settings.country or "US"
+        state = library.settings.state or "All"
```
jonathangreen: Minor: Instead of falling back here, would it just make sense for these to be nullable values, and return `None`. That would distinguish not configured from the default.

src/palace/manager/service/analytics/eventdata.py
```diff
@@ -107,6 +110,9 @@ def create(
         new_value: int | None = None,
         patron: Patron | None = None,
         user_agent: str | None = None,
+        country: str = "US",
+        state: str = "All",
```
jonathangreen: Minor: We are defining the defaults in two places that can get out of sync. Maybe just making them nullable would be reasonable.

jonathangreen (approved):
Approving this one. I added a couple minor comments to consider. 

I do wonder though, like I mentioned in my last comment if we want to automatically configure these, as we are looking at doing for locations so we can do registry lookups by location.

Code can go in as is thought.

dbernstein:
> Approving this one. I added a couple minor comments to consider.
> 
> I do wonder though, like I mentioned in my last comment if we want to automatically configure these, as we are looking at doing for locations so we can do registry lookups by location.
> 
> Code can go in as is thought.

Yes - we will definitely use the registry data for auto-configuration.  I have laid out a multi-step plan for how to accomplish this in https://ebce-lyrasis.atlassian.net/browse/PP-4308

PR #3308, opened 2026-04-29
Refine filtered categories description text

jonathangreen:
## Description

Refines the description for the "Filtered categories" library setting to clarify what the filter applies to. Follow-up to #3272.

Before:
> Content in these categories (genres) will be hidden from catalog browse and search results.

After:
> Content in these categories (fiction genres and non-fiction subjects) will be hidden from catalog browse and search results.

## Motivation and Context

Follow-up feedback on #3272 — the parenthetical "(genres)" understated the filter's scope. The new wording makes it clear to admins that both fiction genres and non-fiction subjects are covered.

## How Has This Been Tested?

No behavior changes — description string only. Existing tests pass unchanged.

## Checklist

### DIFF
--- a/src/palace/manager/integration/configuration/library.py
+++ b/src/palace/manager/integration/configuration/library.py
@@ -412,7 +412,7 @@ class LibrarySettings(BaseSettings):
         list[str],
         LibraryFormMetadata(
             label="Filtered categories",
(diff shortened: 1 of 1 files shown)


dbernstein (approved):
🤖

PR #3311, opened 2026-04-29
Handle license expiration in async Celery task, enable change tracking for ODL pools (PP-4177)

dbernstein:
## Description

Introduces a dedicated Celery task (`update_expired_licenses`) that processes ODL license expirations, and removes the carve-out that prevented AGGREGATED pools from participating in content-hash-based change tracking.

## Motivation and Context

ODL-style (`AGGREGATED`) license pools carry individual `License` rows with `expires` timestamps. Because expiry is time-dependent, a pool's availability can drop to zero without the vendor feed changing at all — making hash-based change detection blind to it.

The previous workaround forced every import pass to always re-process AGGREGATED pools, bypassing the hash check entirely (the `self.licenses is None` guard in `CirculationData.apply()` and an early `return True` in `needs_apply()`). This had two downsides:

1. ODL pools could never skip unchanged imports — every feed poll triggered a full re-apply.
2. Time-based expiry concerns were tangled into the import pipeline rather than being handled as a scheduled concern.

**Fix:** Remove the carve-out so ODL pools participate in normal hash-based change tracking, and add a dedicated Celery task that finds pools where a license expired after the pool was last checked, then calls `update_availability_from_licenses()` on each stale pool.

### New task: `update_expired_licenses`

Scheduled hourly at :30 (every hour at 30 minutes past the hour). Uses a targeted query to find only pools that need updating:

- License has `expires <= now` (it has expired)
- Pool's `last_checked < license.expires` (the expiry occurred after the last availability recalculation)

`update_availability_from_licenses(as_of=now)` advances `last_checked` to `now`, so processed pools are automatically excluded on the next run. The `updated_at` importer freshness sentinel is never touched by this task.

This is O(newly-stale pools) rather than O(all AGGREGATED pools), so it scales efficiently even as the ODL catalog grows.

A `bin/update_expired_licenses` script is also provided for manual invocation.

## How Has This Been Tested?

- `mypy` passes clean on all modified and new files.
- New test class `TestUpdateExpiredLicenses` (`tests/manager/celery/tasks/test_license_expiration.py`) covers:
  - No expired licenses: pool unchanged
  - Stale pool (license expired after last check): availability recalculated downward; `last_checked` advanced to `now`
  - Already-processed expiry: pool correctly skipped; `last_checked` unchanged
  - Mixed multi-pool scenario: only stale pool updated; current pool's `last_checked` unchanged
- New `TestUpdateExpiredLicensesScript` (`tests/manager/scripts/test_license_expiration.py`) verifies the script queues the task and logs the confirmation message.
- New test `test_needs_apply_aggregated_pool_uses_hash` (`tests/manager/data_layer/test_circulation.py`) confirms AGGREGATED pools now go through the normal hash check rather than always returning `True`.
- Full DB-dependent suite: `tox -e py312-docker -- --no-cov tests/manager/data_layer/test_circulation.py tests/manager/celery/tasks/test_license_expiration.py tests/manager/scripts/test_license_expiration.py`

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/license_expiration.py
+++ b/src/palace/manager/celery/tasks/license_expiration.py
@@ -13,7 +13,7 @@ from palace.manager.sqlalchemy.model.licensing import License, LicensePool
 
 
 @shared_task(queue=QueueNames.default, bind=True)
-def expire_licenses(task: Task) -> None:
+def update_expired_licenses(task: Task) -> None:
     """Find pools with newly-expired licenses and recalculate their availability.
 
     A pool is considered stale when any of its licenses has an ``expires`` timestamp at or
--- a/src/palace/manager/celery/tasks/license_expiration.py
+++ b/src/palace/manager/celery/tasks/license_expiration.py
@@ -42,11 +42,15 @@ def expire_licenses(task: Task) -> None:
             task.log.info("No pools with newly-expired licenses found.")
             return
 
-        pools = session.scalars(
-            select(LicensePool)
-            .where(LicensePool.id.in_(stale_pool_ids))
-            .options(selectinload(LicensePool.licenses))
-        ).all()
+        pools = (
+            session.scalars(
+                select(LicensePool)
+                .where(LicensePool.id.in_(stale_pool_ids))
+                .options(selectinload(LicensePool.licenses))
+            )
+            .unique()
+            .all()
+        )
 
         for pool in pools:
             pool.update_availability_from_licenses(as_of=now)
--- a/src/palace/manager/celery/tasks/license_expiration.py
+++ b/src/palace/manager/celery/tasks/license_expiration.py
@@ -17,9 +17,10 @@ def update_expired_licenses(task: Task) -> None:
     """Find pools with newly-expired licenses and recalculate their availability.
 
     A pool is considered stale when any of its licenses has an ``expires`` timestamp at or
-    before the current time AND the pool's ``updated_at`` predates that expiry — meaning
+    before the current time AND the pool's ``last_checked`` predates that expiry — meaning
     the expiry occurred after the last availability calculation and has not yet been
-    accounted for.
+    accounted for. ``update_availability_from_licenses(as_of=now)`` advances
+    ``last_checked`` to ``now``, so processed pools are automatically excluded on the next run.
     """
     now = utc_now()
 
@@ -31,8 +32,8 @@ def update_expired_licenses(task: Task) -> None:
             .where(License.expires <= now)
             .where(
                 or_(
-                    LicensePool.updated_at.is_(None),
-                    LicensePool.updated_at < License.expires,
+                    LicensePool.last_checked.is_(None),
+                    LicensePool.last_checked < License.expires,
                 )
             )
             .distinct()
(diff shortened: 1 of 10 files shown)


src/palace/manager/service/celery/celery.py
```diff
@@ -78,6 +79,13 @@ def beat_schedule() -> dict[str, Any]:
                 minute="0",
             ),  # Run every day at 1:00 AM
         },
+        "expire_odl_licenses": {
+            "task": license_expiration.expire_licenses.name,
+            "schedule": crontab(
+                minute="30",
+                hour="1",
+            ),  # Run every day at 1:30 AM
+        },
```
dbernstein, line 87: Since the query should be very efficient, pulling a minimal number of license pools it may make more sense to run this on an hourly schedule, rather than daily.

src/palace/manager/celery/tasks/license_expiration.py
```diff
@@ -0,0 +1,61 @@
...
+                .where(LicensePool.id.in_(stale_pool_ids))
+                .options(selectinload(LicensePool.licenses))
+            )
+            .unique()
+            .all()
+        )
+
+        for pool in pools:
+            pool.update_availability_from_licenses(as_of=now)
+            pool.updated_at = now
```
jonathangreen: `updated_at` is part of the importer freshness check, so I don't think it's safe to update it here. `BaseMutableData.should_apply_to()` returns early when the incoming feed timestamp is older than `LicensePool.updated_at`, before it reaches the hash comparison. If this task advances `updated_at` to the task run time, a later ODL import with an older feed updated value can be skipped even when the license content actually changed. That can hide real circulation or license updates until a forced reimport.

tests/manager/celery/tasks/test_license_expiration.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Pool already accounts for the expiry
+        pool.updated_at = last_updated
+        pool.licenses_available = 0
+
+        expire_licenses.delay().wait()
+
+        db.session.refresh(pool)
+        # Still 0 — the task correctly skipped this pool
+        assert pool.licenses_available == 0
```
jonathangreen: Minor: This test doesn't actually prove the pool was skipped. `pool.licenses_available` is set to 0 manually, and an expired license also yields 0 from `update_availability_from_licenses`. So if the task incorrectly re-processed the pool, the assertion here still passes. I think we want to be tested `updated_at` to prove that we were skipped.

jonathangreen (commented):
A couple things to look at here: 

1. The PR description still reads "Scheduled nightly at 1:30 AM. Uses a targeted query…". It should get updated before merge, so the commit message it correct. So it reflects that we no run this every hour.

2. I'm concerned about the `updated_at` timestamp update causing us to falsely skip processing.

dbernstein:
@jonathangreen :  I think this one is good to go now.

src/palace/manager/celery/tasks/license_expiration.py
```diff
@@ -0,0 +1,61 @@
...
+            .join(LicensePool, License.license_pool_id == LicensePool.id)
+            .where(License.expires.is_not(None))
+            .where(License.expires <= now)
+            .where(
+                or_(
+                    LicensePool.last_checked.is_(None),
+                    LicensePool.last_checked < License.expires,
+                )
+            )
+            .distinct()
```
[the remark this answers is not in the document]
dbernstein: We decided not to do this.

jonathangreen (approved):

PR #3312, opened 2026-05-01
Use shared github-actions reusable workflows

jonathangreen:
## Description

Point the `claude-review` and `jira-release-sync` workflows at the `v1` tag of [ThePalaceProject/github-actions](https://github.com/ThePalaceProject/github-actions) instead of inline definitions / a local composite action.

- `.github/workflows/claude-review.yml` — replaces the 42-line inline job with a call to the reusable `ThePalaceProject/github-actions/.github/workflows/claude-review.yml@v1` workflow.
- `.github/workflows/jira-release-sync.yml` — swaps `./.github/actions/jira-release-sync` for the shared composite at `ThePalaceProject/github-actions/jira-release-sync@v1`. Drops the preceding `actions/checkout` step since the action does not read the caller's repo.
- Deletes `.github/pr_review_prompt.md`. The reusable workflow embeds an identical prompt and exposes an `additional_prompt` input for any repo-specific guidance we want to layer on later.

The local `.github/actions/jira-release-sync/` composite is intentionally kept for now — it is referenced from outside this repo and will be removed in a follow-up once those callers are migrated.

## Motivation and Context

I've wanted to break our actions out to their own repo for a while. Before proliferating the Claude code review action across more repositories, it makes sense to move it (and the existing jira-release-sync action) into the shared [github-actions](https://github.com/ThePalaceProject/github-actions) repo so future updates land in one place.

## How Has This Been Tested?

- Pre-commit (including the GitHub workflow validator) passes.
- The shared workflow / action contract was diff-checked against the prior inline versions: input names match for `jira-release-sync`, and the embedded review prompt matches `pr_review_prompt.md` verbatim.
- End-to-end behaviour will verify on the first PR and the next release published after merge.

## Checklist


tdilauro (approved):
🎸🤘🏽

PR #3318, opened 2026-05-05
Include UNLIMITED pools in reaper's existing-identifiers query (PP-4267)

jonathangreen:
## Description

Fix `existing_available_identifiers` so it correctly considers `LicensePoolType.UNLIMITED` pools as available. The query previously filtered on `licenses_available != 0 AND licenses_owned != 0`, but UNLIMITED pools have both columns forced to `0` in `CirculationData.apply`, so they were always excluded. The query now matches UNLIMITED pools on `status == ACTIVE` while keeping the original license-count check for metered/aggregated pools.

Added `TestExistingAvailableIdentifiers.test_unlimited_pools_included` covering the new branch, including a REMOVED unlimited pool decoy.

## Motivation and Context

The reap chord (`create_mark_unavailable_chord`) runs `existing_available_identifiers` and the active-identifiers signature in parallel, then diffs the two sets in `mark_identifiers_unavailable`. Because the existing-set query never matched any UNLIMITED pool, the "before" set was always empty, the underlying Redis key was never created, and `mark_identifiers_unavailable` short-circuited with:

```
WARNING palace.manager.celery.tasks.identifiers.mark_identifiers_unavailable: Existing identifiers set does not exist in Redis. No identifiers to mark as unavailable.
```

Net effect: **reaping has been a silent no-op for OPDS for Distributors collections** (and any other importer that produces UNLIMITED pools — opds1/opds2/opds_odl etc.) — no withdrawn titles have been marked `REMOVED` by the reaper.

The bug was introduced when `LicensePoolType.UNLIMITED` started zeroing the count columns (PP-2992) without updating the older identifiers reaper query (PP-2469).

JIRA: PP-4267

## How Has This Been Tested?

- New regression test `test_unlimited_pools_included` builds three ACTIVE UNLIMITED pools plus one REMOVED UNLIMITED pool and asserts only the active three are returned. Fails on `main`, passes with the fix.
- Existing tests in `TestExistingAvailableIdentifiers` continue to pass (metered behavior unchanged).
- `mypy` clean.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/identifiers.py
+++ b/src/palace/manager/celery/tasks/identifiers.py
@@ -2,7 +2,7 @@ from functools import partial
 
 from celery import shared_task
 from celery.canvas import Signature, chord
-from sqlalchemy import select
+from sqlalchemy import and_, or_, select
 from sqlalchemy.orm import raiseload
 
 from palace.util.exceptions import PalaceValueError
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
🚀 LGTM & C

PR #3321, opened 2026-05-05
Coerce miscased a11y feature/hazard values with a warning (PP-4213)

jonathangreen:
## Description

When parsing OPDS publications, miscased accessibility `feature`/`hazard` values previously failed `StrEnum` validation and caused the whole publication to be rejected. The `Accessibility` model now case-insensitively coerces these list entries to their canonical spec form, and drops unknown values rather than failing. Both branches log a `palace.opds.a11y` WARNING so the OPDS validator (which captures that logger) can surface them for upstream reporting.

## Motivation and Context

Found while investigating PP-4213. DeMarque publications include miscased entries — for example, `"TaggedPDF"` instead of the spec's `"taggedPDF"`:

```json
"accessibility": {
  "summary": "This publication lacks alternative text for non-text elements, but it otherwise meets WCAG 2.2 Level A. Furthermore, the publication generally has been produced to meet WCAG Level AA, but it has not been checked for color contrast within images.",
  "conformsTo": "http://www.idpf.org/epub/a11y/accessibility-20170105.html#wcag-a",
  "feature": [
    "readingOrder",
    "structuralNavigation",
    "tableOfContents",
    "index",
    "TaggedPDF",
    "ttsMarkup"
  ],
  "accessModeSufficient": ["textual"]
}
```

These publications now import successfully with the canonical value persisted, while leaving a warning trail so we can ask DeMarque to fix the source data.

## How Has This Been Tested?

Added unit tests in `tests/palace_opds/test_a11y.py` covering: miscased values are coerced to canonical form and emit a warning; already-canonical values produce no warning; unknown `feature` and `hazard` values are dropped with a warning. `tox -e py312-docker -- --no-cov tests/palace_opds/test_a11y.py` passes.

## Checklist

### DIFF
--- a/packages/palace-opds/src/palace/opds/a11y.py
+++ b/packages/palace-opds/src/palace/opds/a11y.py
@@ -7,15 +7,57 @@ https://readium.org/webpub-manifest/schema/a11y.schema.json
 
 from __future__ import annotations
 
+import logging
 from collections.abc import Sequence
 from enum import StrEnum, auto
 from functools import cached_property
 from typing import Any, cast
 
-from pydantic import Field, SerializerFunctionWrapHandler, model_serializer
+from pydantic import (
+    Field,
+    SerializerFunctionWrapHandler,
+    field_validator,
+    model_serializer,
+)
 
 from palace.opds.base import BaseOpdsModel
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
🥇

PR #3322, opened 2026-05-05
Add library setting to allow borrowing with expired credentials (PP-4214)

dbernstein:
## Description

Adds an `allow_borrowing_with_expired_credentials` boolean setting to `LibrarySettings`. When enabled for a library, patrons whose library cards have expired are still permitted to borrow. The `authorization_expires` date is preserved in the database unchanged — only the enforcement of the expiry check is bypassed.

The setting is stored in the existing `Library.settings_dict` JSONB column, so no database migration is required. It appears in the library admin UI under a new **Patron Authentication** category with Yes/No options (default: No). The setting is restricted to **SYS_ADMIN_OR_MANAGER** level access, meaning only library managers and system admins can change it.

## Motivation and Context

A library requested that patrons with expired cards not be blocked from borrowing (PP-4214). The canonical expiry check lives in `PatronUtility.authorization_is_active()`, called at borrow time via `Dispatcher.borrow()`. This change adds a per-library opt-in to skip that check while keeping all existing expiry data intact.

Note: if a library's SIP2 server also sets charge privileges denied in the patron status field for expired patrons, the admin can additionally set `patron_status_block = False` on the SIP2 integration to suppress that separate block.

## How Has This Been Tested?

Added `test_authorization_is_active_allow_borrowing_with_expired_credentials` to `TestPatronUtility` covering:
- Default (False): expired patron is blocked as before
- Opt-in (True): expired patron passes the active check and `assert_borrowing_privileges` does not raise
- Non-expired patron: unaffected by the setting in both states
- Patron with no expiry: unaffected by the setting

Added `test_allow_borrowing_with_expired_credentials_level` to the library configuration test suite asserting the field is gated at `Level.SYS_ADMIN_OR_MANAGER`.

mypy passes cleanly on all changed source files.

## Checklist

### DIFF
--- a/src/palace/manager/api/util/patron.py
+++ b/src/palace/manager/api/util/patron.py
@@ -67,11 +67,17 @@ class PatronUtility:
 
         :raises AuthorizationExpired: If the patron's authorization has expired.
         :raises OutstandingFines: If the patron has too many outstanding fines.
+        :raises AuthorizationBlocked: If the patron is blocked for another reason.
 
         """
+        allow_expired = (
+            patron.library.settings.allow_borrowing_with_expired_authorization
+        )
+
         if not cls.authorization_is_active(patron):
-            # The patron's card has expired.
-            raise AuthorizationExpired()
+            # authorization_expires path (SIP2/Millennium): card is past its expiry date.
+            if not allow_expired:
+                raise AuthorizationExpired()
 
         if cls.has_excess_fines(patron):
             raise OutstandingFines(fines=patron.fines)
@@ -86,6 +92,12 @@ class PatronUtility:
             # the patron has outstanding fines, even if the circulation
             # manager is not configured to make that deduction.
             raise OutstandingFines(fines=patron.fines)
+
+        if patron.block_reason is PatronData.EXPIRED:
+            # Sirsi reports expiry via block_reason rather than authorization_expires.
+            if not allow_expired:
+                raise AuthorizationExpired()
+            return
 
         raise AuthorizationBlocked()
 
@@ -116,8 +128,7 @@ class PatronUtility:
         if patron.authorization_expires and cls._to_date(
             patron.authorization_expires
         ) < cls._to_date(now_local):
-            if not patron.library.settings.allow_borrowing_with_expired_credentials:
-                return False
+            return False
         return True
 
     @classmethod
--- a/src/palace/manager/api/util/patron.py
+++ b/src/palace/manager/api/util/patron.py
@@ -116,7 +116,8 @@ class PatronUtility:
         if patron.authorization_expires and cls._to_date(
             patron.authorization_expires
         ) < cls._to_date(now_local):
-            return False
+            if not patron.library.settings.allow_borrowing_with_expired_credentials:
+                return False
         return True
 
     @classmethod
--- a/src/palace/manager/api/util/patron.py
+++ b/src/palace/manager/api/util/patron.py
@@ -75,7 +75,8 @@ class PatronUtility:
         )
 
         if not cls.authorization_is_active(patron):
-            # authorization_expires path (SIP2/Millennium): card is past its expiry date.
+            # Covers authorization_expires (SIP2/Millennium) and
+            # block_reason=EXPIRED (SirsiDynix) in one place.
             if not allow_expired:
                 raise AuthorizationExpired()
 
@@ -93,10 +94,8 @@ class PatronUtility:
             # manager is not configured to make that deduction.
             raise OutstandingFines(fines=patron.fines)
 
-        if patron.block_reason is PatronData.EXPIRED:
-            # Sirsi reports expiry via block_reason rather than authorization_expires.
-            if not allow_expired:
-                raise AuthorizationExpired()
+        if patron.block_reason is PatronData.EXPIRED and allow_expired:
+            # Expiry was waived by the library; don't raise AuthorizationBlocked.
             return
 
         raise AuthorizationBlocked()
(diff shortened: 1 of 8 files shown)


src/palace/manager/integration/configuration/library.py
```diff
@@ -546,6 +546,19 @@ class LibrarySettings(BaseSettings):
...
+        LibraryFormMetadata(
+            label="Allow borrowing with expired credentials?",
+            description="When enabled, patrons whose library cards have expired will still be permitted to borrow.",
+            type=FormFieldType.SELECT,
+            options={
+                True: "Yes",
+                False: "No",
+            },
+            category="Patron Authentication",
+        ),
```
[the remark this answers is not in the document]
dbernstein: Level.SYS_ADMIN_OR_MANAGER seems appropriate to me.

src/palace/manager/api/util/patron.py
jonathangreen, line 120: `authorization_is_active`'s docstring says "Return True unless the patron's authorization has expired." With this change it can return `True` even when authorization is expired, depending on a library setting that has nothing to do with expiry status. The method now answers the question "is the patron allowed to borrow?" rather than what its name claims.

We already have a method that is supposed to answer that question: `assert_borrowing_privileges`. It seems like this override belongs there. Moving the override into `assert_borrowing_privileges` keeps `authorization_is_active` honest as a pure expiry check and confines the policy to the borrowing-privilege gate where it belongs.

src/palace/manager/api/util/patron.py
```diff
@@ -116,7 +116,8 @@ def authorization_is_active(cls, patron: Patron) -> bool:
         if patron.authorization_expires and cls._to_date(
             patron.authorization_expires
         ) < cls._to_date(now_local):
-            return False
+            if not patron.library.settings.allow_borrowing_with_expired_credentials:
```
jonathangreen: This global setting fails silently for sirsi-expired patrons. The override only intercepts the `authorization_expires` branch, but Sirsi reports expiry via `block_reason = SirsiBlockReasons.EXPIRED` and never sets `authorization_expires`. With the setting enabled, the Sirsi-expired patron skips this branch and falls through to `assert_borrowing_privileges`, which raises `AuthorizationBlocked`. As a result someone using sirsi auth will still get blocked based on expired credentials, even though that is what this setting is meant to prevent.

src/palace/manager/integration/configuration/library.py
```diff
@@ -546,6 +546,20 @@ class LibrarySettings(BaseSettings):
...
+        bool,
+        LibraryFormMetadata(
+            label="Allow borrowing with expired credentials?",
+            description="When enabled, patrons whose library cards have expired will still be permitted to borrow.",
+            type=FormFieldType.SELECT,
+            options={
+                True: "Yes",
+                False: "No",
+            },
+            category="Patron Authentication",
```
jonathangreen: Minor: It seems like this belongs in the `Loans, Holds, & Fines` category. Alongside other settings that impact borrowing like `max_outstanding_fines`, `loan_limit`, `hold_limit`. I don't think it makes sense to create a new category for just this setting when we have other related settings already.

src/palace/manager/integration/configuration/library.py
```diff
@@ -546,6 +546,20 @@ class LibrarySettings(BaseSettings):
             level=Level.ALL_ACCESS,
         ),
     ] = None
+    allow_borrowing_with_expired_credentials: Annotated[
```
jonathangreen: Nit: The model column is `authorization_expires`, the exception is `AuthorizationExpired`, but the setting calls it `expired_credentials`. `allow_borrowing_with_expired_authorization` would line up with the rest of the codebase better. The user facing text doesn't have to change, just the variable name in the model.

tests/manager/integration/configuration/test_library.py
```diff
@@ -97,6 +101,17 @@ def test_minimum_featured_quality_constraints(
     assert "less than or equal to 1" in excinfo.value.problem_detail.detail
 
 
+def test_allow_borrowing_with_expired_credentials_level() -> None:
```
jonathangreen: Minor: This test isn't really meaningful. It doesn't test any behaviour, just a constant that gets set. The level mechanism is enforced by the framework and tested generically at the controller layer. I think this test should just be removed.

jonathangreen (commented):
Added some comments for consideration. I think the issue with sirsi patron auth needs to be resolved before this one gets merged. That exposes a deeper issue that will need a bit of refactoring. 

Right now the "expiry" concept is split across two fields, `authorization_expires` for SIP2/Millenium, `block_reason = EXPIRED` for Sirsi. Any feature that wants to act on expiry has to know to check both.

I think we will need to move the expired block reason from Sirsi into `PatronData`, and update the code to enforce this consistently. So sirsi expired cards are treated the same as others.

dbernstein:
@jonathangreen :  thanks for the review:  I will revisit this after the release.

dbernstein:
@jonathangreen : your suggestions have all been implemented.

src/palace/manager/integration/patron_auth/sirsidynix_authentication_provider.py
```diff
@@ -302,7 +302,7 @@ def remote_patron_lookup(
             patrondata.fines = float(fines.get("amount", 0))
 
         if status_fields.get("expired"):
-            patrondata.block_reason = SirsiBlockReasons.EXPIRED
+            patrondata.block_reason = PatronData.EXPIRED
```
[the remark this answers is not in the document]
dbernstein: I just removed that.

src/palace/manager/api/util/patron.py
```diff
@@ -87,6 +93,12 @@ def assert_borrowing_privileges(cls, patron: Patron) -> None:
             # manager is not configured to make that deduction.
             raise OutstandingFines(fines=patron.fines)
 
+        if patron.block_reason is PatronData.EXPIRED:
```
jonathangreen: Minor: I think this check should live in `authorization_is_active`, as its part of that check. Then we don't need two code paths here that both raise `AuthorizationExpired`.

jonathangreen (approved):
Minor comment to consider

PR #3340, opened 2026-05-13
Run Overdrive import every 3 hours, offset from OPDS2+ODL (PP-4337)

jonathangreen:
## Description

Change `overdrive_import_all_collections` from `crontab(minute="*/15")` (every 15 minutes) to `crontab(minute="15", hour="2-23/3")` — every 3 hours at minute 15, offset 1.5h from the OPDS2+ODL beat tick (which runs at minute 45 every 3 hours).

## Motivation and Context

We've been seeing a steady trickle of `LockNotAcquired` errors raised by `opds_odl.import_collection` in production. The lock errors aren't a bug in ODL, they're the canary that ODL's 3-hour beat tick is firing while the previous chain is still walking pages. The underlying cause is that Overdrive's import is saturating the `default` Celery queue, so ODL pages can't get scheduler time between `task.replace()` calls and the chain stretches past its 3-hour window.

Data from CloudWatch Logs Insights for the prod accounts:

### Top consumers of the `default` queue (worker-hours, last 12h)

| Task | Hours | Runs |
|---|---|---|
| `overdrive.import_collection` | **16.54** | 12,060 |
| `search.index_works` | 5.46 | 10,611 |
| `work.classify_unchecked_subjects` | 3.95 | 23 |
| `patron_activity.sync_patron_activity` | 1.86 | 7,957 |
| `overdrive.reap_collection` | 1.64 | 5,284 |
| `opds_odl.import_collection` | **0.61** | 752 |

### Overdrive's share of the `default` queue, per CM

| CM | default total | overdrive | OD % |
|---|---|---|---|
| ct-connecticut | 5.66h | 4.99h | **88%** |
| vt-vermont | 3.77h | 3.77h | **100%** |
| nj-newjersey | 2.92h | 2.45h | **84%** |
| ri-rhodeisland | 0.67h | 0.67h | **100%** |
| il-illinois | 0.40h | 0.40h | **100%** |
| ca-california | 4.64h | 2.01h | 43% |
| wa-washington | 0.96h | 0.56h | 59% |
| ga-georgia | 2.32h | 0.76h | 33% |

On ct-connecticut, 2,884 Overdrive page-tasks executed in 12 hours vs. 25 ODL page-tasks. ODL barely gets a turn.

### Overdrive import duration is already on a ~3h cycle in practice (last 24h)

| CM | completions | sum_elapsed_h | median_min | p90_min | max_min | avg concurrent chains |
|---|---|---|---|---|---|---|
[...]

PR #3341, opened 2026-05-13
Run Overdrive reap once a day instead of every hour (PP-4337)

jonathangreen:
## Description

Change `overdrive_reap_all_collections` from `crontab(minute="0")` (every hour) to `crontab(minute="0", hour="23")` (once a day at 23:00 UTC).

## Motivation and Context

While investigating Celery `LockNotAcquired` errors on `opds_odl.import_collection` (see #3340), I traced the contention back to Overdrive saturating the `default` queue and discovered that my initial worker-time accounting was significantly under-reading reality.

### The measurement gap

CloudWatch Insights queries against `Task X[id] succeeded in Ys` log lines miss every `task.replace()`-driven page or batch — those are logged as `Task X[id] ignored` with no duration string. For tasks like Overdrive `reap_collection` that paginate over thousands of identifiers via `task.replace()`, only the *final* batch shows up in succeeded counts.

Fleet-wide counts over 12 hours:

| Task | succeeded (counted) | ignored (missed) | total runs |
|---|---|---|---|
| **overdrive.reap_collection** | 5,284 | **129,022** | 134,306 |
| overdrive.import_collection | 12,060 | 29,206 | 41,266 |
| search.search_reindex | 25 | 11,544 | 11,569 |
| opds_odl.import_collection | 752 | 6,655 | 7,407 |

`reap_collection` was under-counted by ~25×.

### What reap actually costs

Sampling per-event durations (received → finished, including ignored events) on `tpp-prod-zeta` over 33 minutes:

| Task | events | total sec | avg concurrent workers busy | per-event p75 / p90 |
|---|---|---|---|---|
| overdrive.reap_collection | 321 | 3,084 | **1.56** | 21s / 23s |
| overdrive.import_collection | 382 | 1,527 | 0.77 | 5s / 7s |

The distribution of reap batches is bimodal: ~50% complete in 0–1s (workflow-lock-skip early returns) and ~25% take 21–23s of real work — 50 sequential synchronous HTTP calls to OverDrive's availability endpoint per batch (`api.update_licensepool(identifier.identifier)` in a for-loop, src/palace/manager/celery/tasks/overdrive.py:594).

Extrapolated fleet-wide for a 12h window, real reap work is on the order of **~370 worker-hours** — substantially more than the `overdrive.import_collection` chain work I previously identified as the dominant consumer.

### Why hourly reap is overkill

Reap's purpose is to detect titles that have left the OverDrive collection. Title removals don't happen on an hourly cadence; detecting them ~24 hours after the fact is fine. Hourly reap also means a heavy reap cycle (which can take hours wall-clock for a collection with tens of thousands of titles) is essentially always running — the workflow lock skips the in-cycle beat ticks, so freshness is already much worse than "1 hour" in practice on heavy CMs.

### Choice of time

23:00 UTC sits in a relatively quiet block between `work.classify_unchecked_subjects` at 22:30 and `full_search_reindex` at 00:10, giving the reap cycle a clear lane to start before the early-morning reaper / import block kicks off at 02:00–05:00.

## How Has This Been Tested?

Going to have to deploy this out and monitor queues to see how it works in practice.

## Checklist

### DIFF
--- a/src/palace/manager/service/celery/celery.py
+++ b/src/palace/manager/service/celery/celery.py
@@ -273,7 +273,7 @@ def beat_schedule() -> dict[str, Any]:
         },
         "overdrive_reap_all_collections": {
             "task": overdrive.reap_all_collections.name,
-            "schedule": crontab(minute="0"),  # Run every hour
+            "schedule": crontab(minute="0", hour="23"),  # Once a day at 11:00 PM
         },
     }
 


tdilauro (approved):
🕚

dbernstein (approved):
Good idea.

PR #3342, opened 2026-05-13
Drop unused Celery per-task CloudWatch metrics (PP-4335)

jonathangreen:
## Description

Remove the `TaskSucceeded`, `TaskFailed`, and `TaskRuntime` custom CloudWatch metrics emitted by `Cloudwatch` in `src/palace/manager/celery/monitoring.py`, along with the `TaskStats` dataclass and the supporting `update_task_stats` / `reset_tasks` / `is_celery_task` / `task_info_str` machinery. `Cloudwatch.publish` now only takes `queues`.

`task_send_sent_event = True` in `service/celery/configuration.py` stays on — Celery still emits the events, we just stop translating them to CloudWatch. Worker overhead is negligible and turning the flag off is unrelated cleanup.

## Motivation and Context

These three metrics account for roughly 85% of the ~6,000 series in the Celery namespace (roughly $1,800/mo on the CloudWatch custom-metric tier). I thought they would be useful when I initially set up monitoring, but overtime as we have added more and more tasks, the number of metrics we monitor has exploded and we haven't been using them for any monitoring or alarms.

Audit findings: `TaskFailed` and `TaskRuntime` are referenced by **zero** alarms and **zero** dashboards; `TaskSucceeded` feeds a single widget on one manager. 

After deploy, no new datapoints are written to these series, which will immediately save us the custom metrics costs, additional savings on storage will phase in gradually as the abandoned series age out of the retention window.

## How Has This Been Tested?

- `tox -e py312-docker -- --no-cov tests/manager/celery/test_monitoring.py` — 6 tests pass.
- `mypy` clean.
- Pre-commit clean.

## Checklist

### DIFF
--- a/src/palace/manager/celery/monitoring.py
+++ b/src/palace/manager/celery/monitoring.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-from collections import defaultdict
 from collections.abc import Sequence
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from datetime import datetime
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
@@ -11,7 +10,7 @@ import boto3
 from boto3.exceptions import Boto3Error
 from botocore.exceptions import BotoCoreError
 from celery.events.snapshot import Polaroid
-from celery.events.state import State, Task
+from celery.events.state import State
 from kombu.transport.redis import PrefixedStrictRedis
 from redis import ConnectionPool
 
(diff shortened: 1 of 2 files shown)


dbernstein (commented):
@jonathangreen :  I have found these useful for doing custom dashboards showing throughput and error rates.  Are these metrics costing us a lot?

dbernstein:
I think it's costing about $80/month...so maybe not worth it.

~30 CMs * 3 queues * 3 metrics * $0.30/metric/month ~ $80

dbernstein:
I was thinking it might be useful to have Claude add some dashboards via the terraform definitions since that is so easy to do now.

dbernstein:
It is handy to be able to look across the production instances and see where the bottlenecks are in terms of execution time and task throughput especially when there's a problem.

jonathangreen:
@dbernstein The metrics are dimensioned per (`Manager`, `TaskName`), not per queue, so each task name across each manager is a separate series. The audit before this PR found ~5,900 active series across the three (`TaskFailed`: 2,112, `TaskRuntime`: 1,672, `TaskSucceeded`: 2,112) → ~$1,800/mo at $0.30/series. (I mentioned this in the PR description). That is a huge chunk of our total cloudwatch spend, so it didn't seem worth keeping this around.

jonathangreen:
That said, you're right that throughput / error rate / latency are real signals to lose. I think we can derive the same stats for dashboards from the celery worker logs if we want to, or we could publish CM-level or queue-level `Failed` / `Succeeded` metrics (at this granularity, I'm not sure `Runtime` makes sense).

dbernstein:
It totally makes sense to remove it.  I missed the bit in the PR description.  :)

dbernstein (approved):
🦞

PR #3343, opened 2026-05-13
Add QueueOldestAge CloudWatch metric for Celery queues (PP-4336)

jonathangreen:
## Description

Emit a new per-queue, per-manager `QueueOldestAge` CloudWatch metric (in seconds) measuring how long the oldest unconsumed task has been waiting. To compute it, outgoing task messages are stamped with an `enqueued_at` ISO-8601 header via a `before_task_publish` Celery signal handler in `src/palace/manager/service/celery/celery.py`. The `Cloudwatch` polaroid camera reads the tail of each redis queue (`LINDEX <queue> -1` — kombu LPUSH-on-publish / BRPOP-on-consume puts the oldest message there), parses the header, and emits the age alongside the existing `QueueWaiting` count.

Other notes:
- `setdefault` keeps the original timestamp across retries and chained re-publishes, so the metric reflects how long the work has actually been waiting rather than restarting on each republish.
- Messages without the header (pre-rollout, or published by an external system) return `None` silently — no metric emitted, no error logged.
- JSON-parse / shape-validation failures log at ERROR with a traceback so we know if something on the broker ever goes off the rails.
- `service.celery.*` was added to the existing `palace.manager.celery.*` mypy override block since both wrap untyped Celery features.

## Motivation and Context

Queue depth alone is a load-dependent signal — a 1M-item queue draining in 60s is fine, but 3 items waiting 6h is not. The user-impact signal is the **age** of the oldest waiting task. CloudWatch alarms on `RATE(min QueueWaiting) > 0` had been firing ~57 times per 2 weeks across four managers on noise; tuning that threshold was the short-term fix in the infra repo, but the durable signal needs to be age-based. Celery itself doesn't emit it and no broker header carries enqueue time (verified against a prod redis dump), so we inject it ourselves.

## How Has This Been Tested?

- `tox -e py312-docker -- --no-cov tests/manager/celery/test_monitoring.py tests/manager/service/celery/test_celery.py` — 13 tests pass.
- `mypy` clean across 1137 source files.
- Pre-commit clean.
- **Local smoke**: ran the webapp and Celery beat against a local redis, confirmed every message in `palacedefault` (33 beat-scheduled tasks + 1 webapp-queued task) had `headers.enqueued_at` set. Verified `LINDEX palacedefault -1` returns the oldest message (the 4-minute-older webapp task), not the freshly-published beat batch.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -325,6 +325,7 @@ disallow_subclassing_any = false
 disallow_untyped_decorators = false
 module = [
     "palace.manager.celery.*",
+    "palace.manager.service.celery.*",
 ]
 
 [[tool.mypy.overrides]]
--- a/src/palace/manager/celery/monitoring.py
+++ b/src/palace/manager/celery/monitoring.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import datetime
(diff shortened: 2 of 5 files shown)


dbernstein (approved):
🔢

PR #3361, opened 2026-05-19
500 on /search/ from Accept-Language header (PP-4368)

jonathangreen:
## Description

`parse_accept_language` in `src/palace/manager/util/accept_language.py` called `accept_lang_segment.split(";")` with no `maxsplit`, so any segment with more than one `;` (e.g. `en-US;q=0.8;something`) raised `ValueError: too many values to unpack`. The exception bubbled up from `SearchFacets.from_request` and returned **500** on `/<library>/search/<lane_id>` endpoints.

The parser is now lenient with malformed input rather than relying on the caller to catch exceptions:

- Scan all `;`-separated params for `q=` instead of assuming it's the first param. Real-world clients send extras before or after `q` (the RFC technically forbids it, but `meta-externalagent/1.1` and others ignore that).
- Invalid q-values (`q=`, `q=abc`) log a warning and fall back to the default quality, keeping the language code. Dropping a usable language because one param is garbage would be too aggressive.
- Over-length headers (>8192 chars) log a warning and return `[]` instead of raising. Every caller wanted the empty-list fallback anyway.

## Motivation and Context

I'm looking into alerts from over the weekend and it looks like `meta-externalagent/1.1` (Facebook crawler) hit `/FL0249/search/915`, `/VT0014/search/60583`, and other lane-search endpoints with a multi-`;` `Accept-Language` header, generating 500s. It crawled enough of them for us to get an alert on 500s coming from production.

Jira ticket: PP-4368.

## How Has This Been Tested?

- New test file `tests/manager/util/test_accept_language.py` with 11 cases covering: docstring example, sort ordering, `default_quality` argument, extras after `q`, extras before `q`, segments with no `q=` at all, empty segments, invalid q-values (`q=`, `q=abc`), invalid language codes, over-length header returning `[]` with a warning, and whitespace handling.
- Existing `tests/manager/feed/facets/test_search.py` (8 tests) continues to pass — no behavior change for the happy path.
- `mypy` clean on the changed files.

## Checklist

### DIFF
--- a/src/palace/manager/util/accept_language.py
+++ b/src/palace/manager/util/accept_language.py
@@ -54,15 +54,41 @@ def parse_accept_language(
         return []
 
     if len(accept_language_str) > MAX_HEADER_LEN:
-        raise ValueError("Accept-Language too long, max length is 8192")
+        # Reject pathological input rather than spending CPU on it. We log
+        # the size + a prefix so we can see what's hitting us in metrics,
+        # but callers get an empty list — a malformed header from an
+        # external client should not break the request.
+        logger.warning(
+            "parse_accept_language: header exceeds max length %d (got %d); "
+            "returning no languages. Prefix: %r",
+            MAX_HEADER_LEN,
+            len(accept_language_str),
+            accept_language_str[:200],
+        )
+        return []
 
     parsed_langs = []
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
Looks good.

PR #3363, opened 2026-05-19
BuildError on loans feed when Work language is null (PP-4368)

jonathangreen:
## Description

`CirculationAnnotator.add_author_links` and `add_series_link` (in `src/palace/manager/feed/annotator/circulation.py`) now build the `url_for` call without the `languages` / `audiences` kwargs when either is `None`, so Werkzeug selects the no-filter route variant (`/works/contributor/<contributor_name>` or `/works/series/<series_name>`) instead of trying — and failing — to fill `<languages>` with `None`.

## Motivation and Context

`/<library>/loans/` was returning a 500 for any patron whose loans included a `Work` with `language IS NULL`. The traceback ended in:

```
werkzeug.routing.exceptions.BuildError: Could not build url for endpoint 'contributor' with values ['audiences', 'contributor_name', 'library_short_name']. Did you forget to specify values ['languages']?
```

The root cause: `language_and_audience_key_from_work` returns `work.language` verbatim, which is `None` for affected Works. The annotator then passed `languages=None, audiences=None` to `url_for("contributor", ...)`. Werkzeug treats an explicit `None` kwarg as "present," so it never falls back to the rule whose `defaults=dict(languages=None, audiences=None)` would have covered the missing values — it iterates the longer-path rules instead, none of which accept a `None` path variable, and raises `BuildError`. The series link path (`add_series_link`) had the same shape and would have hit the same error.

Observed in production for patron `external_identifier=1526824` at `CA0081` (Oakland Public Library) — three failures in two minutes from the iOS app; the loans page was completely broken for that patron until the offending Work was fixed.

Found via CloudWatch alerts, looks like this was happening enough to occasionally push the 500 errors rate up enough to cause an error.

## How Has This Been Tested?

- Added `test_add_author_links_when_language_is_none` and `test_add_series_link_when_language_is_none` in `tests/manager/feed/test_library_annotator.py`. Both run inside `app.test_request_context("/")` so the real Flask routing is exercised (the default `_patched_url_for` test fixture short-circuits route matching, so a request context is required to catch this kind of bug).
- Extended `test_language_and_audience_key_from_work` with a `language=None` case to lock in the `(None, <audience_key>)` return shape.
- Verified regression coverage by reverting the source change locally — both new tests fail with `BuildError`, confirming they would have caught this in CI.
- `tox -e py312-docker -- --no-cov tests/manager/feed/test_library_annotator.py` — 41 passed.
- `mypy` — clean.

## Checklist

### DIFF
--- a/src/palace/manager/feed/annotator/circulation.py
+++ b/src/palace/manager/feed/annotator/circulation.py
@@ -963,6 +963,14 @@ class LibraryAnnotator(CirculationManagerAnnotator):
             return None
 
         languages, audiences = self.language_and_audience_key_from_work(entry.work)
+        # Werkzeug treats an explicit `None` kwarg as "present" and won't fall
+        # back to the route variant whose `defaults` cover the missing values,
+        # so we drop the keys entirely when either is missing.
+        extra_kwargs: dict[str, str] = {}
+        if languages and audiences:
+            extra_kwargs["languages"] = languages
+            extra_kwargs["audiences"] = audiences
+
         for author_entry in entry.computed.authors:
             if not (name := author_entry.name):
                 continue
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 👍🏽

dbernstein (approved):
LGTM :) 🥇

PR #3365, opened 2026-05-19
Migrate Bibliotheca event monitor to Celery (PP-4390) 

dbernstein:
## Description

Replaces the legacy `BibliothecaEventMonitor` cron script with two Celery tasks and a new `BibliothecaEventImporter` helper class:

- **`import_all_collections`**: dispatcher that queues one `import_collection` task per Bibliotheca collection once per hour (Celery beat).
- **`import_collection`**: processes one 5-minute slice of circulation events, stamps the `Timestamp`, then re-queues itself via `task.replace()` until the collection is caught up to `now - 5 min`.

A Redis workflow lock (keyed to `collection_id`, UUID carried across `task.replace()` calls and Celery retries) ensures at most one chain runs per collection at a time. The lock auto-expires after 2 hours to handle worker death mid-chain.

**New:**
- `BibliothecaEventImporter` class (`bibliotheca_importer.py`) encapsulates the per-slice logic: `get_start()`, `import_time_slice()`, `_handle_event()`. Slice duration defaults to 5 minutes but is overridable.
- `SliceImportResult` frozen dataclass carries `events_handled`, `slice_start`, `slice_end` back to the task.
- `tests/manager/integration/license/test_bibliotheca_importer.py` — unit tests for `BibliothecaEventImporter` directly.
- `tests/manager/celery/test_importer.py` — standalone unit tests for `import_workflow_key` / `import_workflow_lock`.

**Deleted:**
- `BibliothecaEventMonitor` class from `bibliotheca.py`
- `bin/bibliotheca_monitor` script
- `bibliotheca_monitor` cron entry from `docker/services/cron/cron.d/circulation`
- `TestBibliothecaEventMonitor` test class (replaced by new tests)

**Retained for PR 2:**
- `BibliothecaTimelineMonitor` — still inherited by `BibliothecaPurchaseMonitor`; will be removed when the purchase monitor is migrated.

This is PR 1 of 3. PR 2 will migrate `BibliothecaPurchaseMonitor`; PR 3 will migrate `BibliothecaCirculationSweep`.

## Motivation and Context

JIRA: PP-4390

Moves Bibliotheca background work onto Celery (consistent with Overdrive, Boundless, OPDS2), enabling centralized scheduling via Celery beat, per-collection parallelism, and observable retries — while deleting the external cron entry points.

## How Has This Been Tested?

`tests/manager/celery/tasks/test_bibliotheca.py` covers the Celery task layer:
- Dispatcher queues one task per Bibliotheca collection, ignoring non-Bibliotheca collections
- First-run timestamp creation and resume from stored timestamp (`finish - OVERLAP`)
- Explicit `start` forwarded directly on chained-slice path (no Timestamp consult)
- Up-to-date check (no API call when already caught up)
- `task.replace()` fired when more slices remain; not fired on the last slice
- `lock_value` forwarded unchanged through `task.replace()` chains
- Concurrent run blocked when lock already held (warning logged); expired-between-slices path logs warning and continues
- Workflow lock held across retries (not released on `BadResponseException` / `RequestTimedOut`)
- End-to-end smoke: events processed → `LicensePool` created, `Timestamp` updated

`tests/manager/integration/license/test_bibliotheca_importer.py` covers `BibliothecaEventImporter` directly:
- `get_start`: no prior timestamp, prior timestamp, null-finish timestamp
- `import_time_slice`: correct window bounds, cutoff capping, custom slice size, `Timestamp` stamped, event count
- `_handle_event`: `LicensePool` created, ISBN equivalency linked at strength 1, availability updated, `bibliographic_apply` queued / skipped based on hash

`tests/manager/celery/test_importer.py` covers `import_workflow_key` / `import_workflow_lock` in isolation (key structure, distinct-ID independence, acquire/release, 2-hour TTL, same-collection conflict, different-collection independence).

All tests pass. Type-checked with mypy.

## Checklist


src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,290 @@
...
+
+    Ported from ``BibliothecaEventMonitor.handle_event``.  Creates or updates
+    the ``LicensePool``, links the ISBN identifier, adjusts availability based
+    on the event delta, and queues a ``bibliographic_apply`` task when the
+    title's metadata has changed (hash-based deduplication).
+    """
+    session = api._db
+
+    license_pool, _ = LicensePool.for_foreign_id(
+        session,
```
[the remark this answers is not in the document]
dbernstein: This is intentional by design: the goal is hash-based deduplication rather than a 'new LicensePool only' guard. `bibliographic_lookup` is called on every event, but `needs_apply()` computes a SHA-256 hash of the returned `BibliographicData` and short-circuits without queuing a `bibliographic_apply` task when the hash matches what is already stored. This means metadata updates propagate the next time an event for that title fires, rather than only on first ingestion. The extra API call per event is a deliberate trade-off for fresher metadata.

tests/manager/celery/tasks/test_bibliotheca.py
```diff
@@ -0,0 +1,479 @@
...
+
+    def mock_api(self) -> MockBibliothecaAPI:
+        """Return a fresh MockBibliothecaAPI bound to this fixture's collection."""
+        return MockBibliothecaAPI(self.db.session, self.collection)
+
+    def stamp_event_monitor(
+        self, finish: datetime | None = None, collection: Collection | None = None
+    ) -> Timestamp:
+        """Create (or update) the event-monitor Timestamp for the collection."""
+        from palace.util.datetime_helpers import utc_now
```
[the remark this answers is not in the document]
dbernstein: Fixed — removed the local re-import.

tests/manager/celery/tasks/test_bibliotheca.py
```diff
@@ -0,0 +1,479 @@
...
+from palace.manager.sqlalchemy.model.coverage import Timestamp
+from tests.fixtures.celery import CeleryFixture
+from tests.fixtures.database import DatabaseTransactionFixture
+from tests.fixtures.files import BibliothecaFilesFixture
+from tests.fixtures.redis import RedisFixture
+from tests.mocks.bibliotheca import MockBibliothecaAPI
+from tests.mocks.mock import MockRequestsResponse
+
+if TYPE_CHECKING:
+    pass
```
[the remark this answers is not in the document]
dbernstein: Fixed — removed both the `TYPE_CHECKING` import and the empty guard block.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,286 @@
+"""Celery tasks for Bibliotheca (3M Cloud) collection management.
+
+Three tasks handle different aspects of keeping a Bibliotheca collection in sync:
+
+- ``import_all_collections`` / ``import_collection``:  Near-real-time circulation
+  events, processed in 5-minute slices.
+- ``purchase_monitor_all_collections`` / ``purchase_monitor_collection`` (PR 2):
```
dbernstein: coming soon (in the next two PRs)

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,286 @@
...
+  MARC-record-based purchase history, one day at a time.
+- ``circulation_sweep_all_collections`` / ``circulation_sweep_collection`` (PR 3):
```
dbernstein: coming soon

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,281 @@
+"""Celery tasks for Bibliotheca (3M Cloud) collection management.
+
+Three tasks handle different aspects of keeping a Bibliotheca collection in sync:
+
+- ``import_all_collections`` / ``import_collection``:  Near-real-time circulation
+  events, processed in 5-minute slices.
+- ``purchase_monitor_all_collections`` / ``purchase_monitor_collection`` (PR 2):
```
jonathangreen: Minor: The PR could reference these forthcoming PRs, but I don't think the in code comments should. Seems liek this comment should talk about the task that exists, and future PRs should add information about the tasks that they add.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,281 @@
...
+    isbn: str,
+    foreign_patron_id: str | None,  # present in the API payload but not acted on here
+    start_time: datetime,
+    end_time: datetime | None,
+    internal_event_type: str,
+    logger: LoggerType,
+) -> None:
+    """Process a single Bibliotheca circulation event.
+
+    Ported from ``BibliothecaEventMonitor.handle_event``.  Creates or updates the
```
jonathangreen: Does it make sense to have the comment here call out where it was ported from? I think we're going to remove `BibliothecaEventMonitor` right? So doesn't this comment just become stale and misleading. We don't care where the function came from. We care what it does right now.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,281 @@
...
+    bibliotheca_identifier = license_pool.identifier
+    isbn_identifier, _ = Identifier.for_foreign_id(session, Identifier.ISBN, isbn)
+
+    edition, _ = Edition.for_foreign_id(
+        session, api.data_source, Identifier.BIBLIOTHECA_ID, bibliotheca_id
+    )
+
+    bibliotheca_identifier.equivalent_to(api.data_source, isbn_identifier, strength=1)
+
+    license_pool.update_availability_from_delta(internal_event_type, start_time, 1)
```
jonathangreen: Shouldn't all this happen in the apply task? The general design here has been to just queue up the mutations and have them done in the apply tasks.

src/palace/manager/service/celery/celery.py
```diff
@@ -295,6 +296,10 @@ def beat_schedule() -> dict[str, Any]:
             "task": overdrive.reap_all_collections.name,
             "schedule": crontab(minute="0", hour="23"),  # Once a day at 11:00 PM
         },
+        "bibliotheca_import_all_collections": {
+            "task": bibliotheca.import_all_collections.name,
+            "schedule": crontab(minute="*/15"),  # Every 15 minutes
```
jonathangreen: Similar to the overdrive tasks, it seems like running this every 15 minutes is way too often. we might want to dial this back to once an hour, or once every couple hours. We don't want this to crowd out other tasks that also need time to run.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    return RedisLock(
+        client,
+        [
+            "BibliothecaEventImportWorkflow",
+            Collection.redis_key_from_id(collection_id),
+        ],
+        random_value=random_value,
+        lock_timeout=timedelta(hours=2),
+    )
```
jonathangreen: This is a near exact copy of `import_workflow_lock`, why not reuse that instead of copying it? The only difference I see is the redis lock key. If you want different locks for your future PRs, then why not extend the function instead of copying it.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,281 @@
...
+    if lock_value is None:
+        lock_value = str(uuid4())
+
+    workflow_lock = _event_import_workflow_lock(redis, collection_id, lock_value)
+
+    with workflow_lock.lock(
+        raise_when_not_acquired=False,
+        # task.replace() raises Ignore; listing it here keeps the lock alive
+        # across the re-queue so the chain holds the lock for all its slices.
+        ignored_exceptions=(Ignore,),
```
jonathangreen: Overdrive lists `ignored_exceptions=(Ignore, BadResponseException, RequestTimedOut)`, why the behavior change here? Your design seems the same between these, so we should make this align.

https://github.com/ThePalaceProject/circulation/blob/f2fd04ade373801b8a35a54621948dfcdfaa42d3/src/palace/manager/celery/tasks/overdrive.py#L117-L123

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,281 @@
...
+        [
+            "BibliothecaEventImportWorkflow",
+            Collection.redis_key_from_id(collection_id),
+        ],
+        random_value=random_value,
+        lock_timeout=timedelta(hours=2),
+    )
+
+
+def _handle_event(
```
jonathangreen: Minor: There is a bunch of buisness logic in this file, that seems like it belongs elsewhere. Somewhere it can be independently tested. Boundless has `BoundlessImporter`, Overdrive has `OverdriveImporter`, OPDS has `OpdsImporter`. It seems like some of this logic that lives in this file should go into an importer class.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -0,0 +1,281 @@
...
+from palace.manager.sqlalchemy.model.edition import Edition
+from palace.manager.sqlalchemy.model.identifier import Identifier
+from palace.manager.sqlalchemy.model.licensing import LicensePool
+from palace.manager.util.http.exception import (
+    BadResponseException,
+    RemoteIntegrationException,
+    RequestTimedOut,
+)
+
+EVENT_IMPORT_SERVICE_NAME = "Bibliotheca Event Monitor"
```
jonathangreen: This kind of conflicts with our new naming scheme right? Its not really a monitor anymore.

jonathangreen (approved):

src/palace/manager/celery/tasks/bibliotheca.py
dbernstein: Good catch.

PR #3371, opened 2026-05-20
Migrate OIDC validator from authlib.jose to joserfc

jonathangreen:
## Description

Migrates the OIDC ID token validator off the deprecated `authlib.jose` module and onto `joserfc`, the library the Authlib maintainers now recommend (and which Authlib already depended on internally). `authlib` is dropped as a direct dependency and replaced by `joserfc`; the `authlib.jose` / `authlib.jose.errors` entries in the mypy `ignore_missing_imports` list are removed because `joserfc` ships type hints (`py.typed`).

Mapping of the changes in `validator.py`:

- `from authlib.jose import JsonWebKey, JsonWebToken` / `from authlib.jose.errors import JoseError` → `from joserfc import jwt`, `from joserfc.errors import JoseError`, `from joserfc.jwk import KeySet, KeySetSerialization`. `JoseError` keeps the same name, so the existing error handling is unchanged.
- The `JsonWebToken(algorithms=[...])` instance is replaced by an `ALLOWED_ALGORITHMS` class constant passed to `jwt.decode(...)`, so the `__init__` is no longer needed.
- `JsonWebKey.import_key_set(jwks)` → `KeySet.import_key_set(...)`, and the decoded payload is read from `token.claims`.

Claims validation is done by the validator's own `validate_claims` method (not Authlib's `claims_options`), so that behavior is untouched.

## Motivation and Context

Running the test suite emitted an `AuthlibDeprecationWarning: authlib.jose module is deprecated, please use joserfc instead.` The validator was the only consumer of `authlib` in the codebase, so migrating it removes the warning and the now-unnecessary direct dependency.

## How Has This Been Tested?

- `tox -e py312-docker -- --no-cov tests/manager/integration/patron_auth/oidc/` — 371 passed.
- `mypy` clean on the validator.
- Confirmed importing the module under `-W error::DeprecationWarning` no longer raises.

## Checklist

### DIFF
--- a/src/palace/manager/integration/patron_auth/oidc/validator.py
+++ b/src/palace/manager/integration/patron_auth/oidc/validator.py
@@ -12,8 +12,9 @@ import time
 from re import Pattern
 from typing import Any, cast
 
-from authlib.jose import JsonWebKey, JsonWebToken
-from authlib.jose.errors import JoseError
+from joserfc import jwt
+from joserfc.errors import JoseError
+from joserfc.jwk import KeySet, KeySetSerialization
 
 from palace.util.exceptions import BasePalaceException
 from palace.util.log import LoggerMixin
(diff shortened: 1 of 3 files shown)


src/palace/manager/integration/patron_auth/oidc/validator.py
```diff
@@ -54,26 +54,25 @@ def validate_signature(self, id_token: str, jwks: dict[str, Any]) -> dict[str, A
...
+            # 1. Find the correct key from the set using the 'kid' header
+            # 2. Verify the signature using one of the allowed algorithms
+            # 3. Return a token whose 'claims' is the decoded payload
+            # We validate the claims separately for better error messages.
+            token = jwt.decode(
+                id_token,
+                key_set,
+                # This type ignore can go away if / when upstream merges
+                # https://github.com/authlib/joserfc/pull/98
+                algorithms=self.ALLOWED_ALGORITHMS,  # type: ignore[arg-type]
```
tdilauro, line 71: This was on my list for avoiding a potential "None" algorithm exploit. 🎉

tdilauro (approved):
Excellent! This change was on my list, so appreciate this! 🔧

PR #3376, opened 2026-05-21
Exclude trashed items from Google Drive folder lookup

dbernstein:
## Summary

- Adds `trashed = false` to the query in `GoogleDriveService.get_file` so that folders moved to the Drive trash are never treated as live parents.
- Without this filter, reports uploaded into a trashed parent folder are invisible to users — they appear to be missing but are actually sitting inside the trash hierarchy.
- Adds a test that inspects the raw HTTP request URI to assert the filter is present in every `files.list` call.

## Test plan

- [x] New test `test_get_file_excludes_trashed_items` verifies `trashed = false` appears in the Drive API query
- [x] All existing Google Drive service tests pass
- [ ] Manually verify on a staging environment that existing (non-trashed) folders are still found and reused correctly

### DIFF
--- a/src/palace/manager/service/google_drive/google_drive.py
+++ b/src/palace/manager/service/google_drive/google_drive.py
@@ -17,8 +17,19 @@ class GoogleDriveService(LoggerMixin):
         self.api_client = api_client
 
     def get_file(self, name: str, parent_folder_id: str | None = None) -> File | None:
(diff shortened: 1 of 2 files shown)


dbernstein:
Superseded by the race-condition fix PR which includes this change.

PR #3377, opened 2026-05-22
Reduce Google Drive folder race condition with Redis lock, jitter, and pick-oldest (PP-4434)

dbernstein:
## Description

Three independent layers guard against duplicate Google Drive folder creation in `generate_playtime_report`:

1. **Redis lock** — a `RedisLock` scoped per `(root_folder_id, data_source_name)` serialises workers within the same Redis instance (i.e. the same shard). If the lock cannot be acquired within `FOLDER_LOCK_ACQUIRE_TIMEOUT` (60 s), the worker logs a warning and proceeds without it (graceful degradation — no task stalling).

2. **Random jitter** — each task invocation sleeps `random.uniform(0, FOLDER_CREATION_JITTER_MAX)` (up to 10 s) once before the data-source loop. Independent shards share no Redis instance, so the lock alone cannot coordinate them; a single random offset per run spreads their arrival times across all folder levels without multiplying the delay by the number of data sources.

3. **Pick-oldest** — `get_file` passes `orderBy=createdTime` to `files.list`, so it always returns the oldest matching folder first. If two groups do race and each creates a copy, every subsequent worker converges on the same canonical (oldest) folder for all levels of the hierarchy and the file upload. Orphaned empty duplicates may exist but no files are misplaced.

**Residual risk:** The three layers together make the collision window very narrow but do not provide a 100% guarantee. A fully watertight fix would require a coordination mechanism shared across all shards (e.g. storing canonical folder IDs in a shared database). We have chosen to accept this low-stakes residual risk to keep the code simple: the task runs monthly, any orphaned folders are cosmetic (no data loss), and the previous behaviour created duplicates on every concurrent run.

## Motivation and Context

Multiple CMs grouped in a shard all write reports to the same Google Drive root. When `generate_playtime_report` fires at the same scheduled time, all workers check for the folder hierarchy simultaneously. Because Google Drive allows same-named folders and its `files.list` API is eventually consistent, every worker sees "folder not found" and creates its own copy — producing duplicate same-named directories (e.g. multiple "BiblioBoard" folders). Different CM groups share no Redis instance, so a lock alone is insufficient.

JIRA: PP-4434

## How Has This Been Tested?

- `redis_fixture` added to `test_generate_playtime_reports` so the task's Redis call resolves to a real client; `FOLDER_CREATION_JITTER_MAX` monkeypatched to 0 to keep the test fast.
- `test_generate_playtime_report_folder_lock_contention`: pre-acquires the folder lock, patches `FOLDER_LOCK_ACQUIRE_TIMEOUT` to 0.1 s and jitter to 0, asserts the warning is logged and the file upload still completes (graceful degradation).
- `test_get_file_orders_by_created_time`: verifies `orderBy=createdTime` is present in the Drive API request URI and that the first (oldest) result is returned when multiple exist.
- All 21 affected tests pass.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/playtime_entries.py
+++ b/src/palace/manager/celery/tasks/playtime_entries.py
@@ -149,15 +149,16 @@ class Writer(Protocol):
 REPORT_DATE_FORMAT = "%m-%d-%Y"
 
 # How long (seconds) to wait for the folder-creation lock before proceeding without it.
-# Exposed as a module-level constant so tests can monkeypatch a short value.
-FOLDER_LOCK_ACQUIRE_TIMEOUT: int = 60
+# Typed as float | int to match acquire_blocking's signature; tests monkeypatch this
+# to a small float (e.g. 0.1) to avoid waiting a full 60 seconds.
+FOLDER_LOCK_ACQUIRE_TIMEOUT: float | int = 60
 
-# Maximum seconds of random jitter added before acquiring the folder-creation lock.
-# Independent CM groups share no Redis instance and therefore cannot coordinate via
-# the lock alone; spreading their start times reduces the probability that two groups
-# race to create the same Google Drive folder at the same moment.
+# Maximum seconds of random jitter added once per task invocation before the
+# folder-creation loop. Independent CM groups share no Redis instance and therefore
+# cannot coordinate via the lock alone; a single random offset per run spreads their
+# arrival times across all folder levels without multiplying the delay by N data sources.
 # Exposed as a module-level constant so tests can set it to 0.
-FOLDER_CREATION_JITTER_MAX: int = 30
+FOLDER_CREATION_JITTER_MAX: int = 10
 
 
 @shared_task(queue=QueueNames.default, bind=True)
@@ -200,6 +201,12 @@ def generate_playtime_report(
 
     redis_client = task.services.redis.client()
 
+    # Jitter: a single random delay per task invocation spreads independent CM groups
+    # (which share no Redis instance) in time, reducing the probability that two groups
+    # race to create the same Google Drive folder. Taking it once before the loop keeps
+    # the per-data-source overhead constant regardless of how many sources are processed.
+    time.sleep(random.uniform(0, FOLDER_CREATION_JITTER_MAX))
+
     with task.session() as session:
         # get list of collections
         data_source_names = _fetch_distinct_eligible_data_source_names(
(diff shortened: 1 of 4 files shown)


dbernstein:
Superseded — the Redis lock approach does not work across independent CM groups that share different Redis instances. Replacing with a jitter + pick-oldest-folder strategy that requires no shared state.

dbernstein:
Reopening — the Redis lock is still needed for within-group serialization. Will layer pick-oldest and jitter on top to handle the cross-shard case.

src/palace/manager/celery/tasks/playtime_entries.py
```diff
@@ -145,6 +148,18 @@ def writerow(self, row: Iterable[Any]) -> Any: ...
 
 REPORT_DATE_FORMAT = "%m-%d-%Y"
 
+# How long (seconds) to wait for the folder-creation lock before proceeding without it.
+# Typed as float | int to match acquire_blocking's signature; tests monkeypatch this
+# to a small float (e.g. 0.1) to avoid waiting a full 60 seconds.
```
tdilauro: Minor - probably better to put the test comment in the test file, rather than here, since they might diverge.

src/palace/manager/celery/tasks/playtime_entries.py
```diff
@@ -223,12 +246,39 @@ def generate_playtime_report(
...
+                folder_lock = RedisLock(
+                    redis_client,
+                    lock_name=[
+                        "playtime_report_folder",
+                        root_folder_id or "default",
+                        data_source_name,
+                    ],
+                    lock_timeout=timedelta(minutes=5),
+                )
+                lock_acquired = folder_lock.acquire_blocking(
```
tdilauro: The PR description mentions that failure to acquire the lock should result in graceful degradation.

Question: Will `RedisLock.acquire` raise if, for example, Redis is down or not responding for some reason?

If an exception is thrown here, it won't be handled and we won't end up proceeding to the lock-less fallback below.

tdilauro (approved):
This looks good! 🎸 

One significant question and a minor suggestion below, but I'm approving it now, in case my question has a happier answer than I expected.

src/palace/manager/celery/tasks/playtime_entries.py
dbernstein: You're right that is a problem:  I will fix it.  If lock acquisition fails due to an issue with Redis, we should just carry on since the purpose of the lock is to reduce the probability of generating duplicate directories not eliminate it entirely.

PR #3383, opened 2026-05-22
Restrict playtime reports to Blackstone Unlimited and Unlimited Listens (PP-4346)

dbernstein:
## Description

Replaces the hardcoded data-source allowlist with an explicit per-collection opt-in flag that controls whether a collection's playtime data is included in the monthly Google Drive report.

### Changes

**New setting: `generate_playtime_report` (bool, default `False`)**

Added to `OPDS2ImporterSettings` and `OPDSForDistributorsSettings` as a sysadmin-facing SELECT field ("Generate playtime report for audio books"). Only OPDS 2.0 Import and OPDS for Distributors collections expose this setting; all other protocols are ineligible by design.

**Migration**

Sets `generate_playtime_report = True` for all existing OPDS 2.0 / OPDS for Distributors collections whose `data_source` name begins with `"Blackstone"` or `"Unlimited"`. All other collections default to `False` and require no update.

Downgrade removes the key entirely (absence is equivalent to the Pydantic default of `False`).

**`_fetch_distinct_eligible_data_source_names`**

Now returns only data source names for collections where `generate_playtime_report` is explicitly `True`. The `PlaytimeSummary` fallback (which was the root cause of folder proliferation — it included every historical data source regardless of intent) has been removed.

## Motivation and Context

The Partner Success team observed large numbers of Google Drive folders being generated for every data source in the system. The underlying cause was a combination of:

1. `_fetch_distinct_eligible_data_source_names` returning all data sources that ever appeared in `PlaytimeSummary` records, and
2. A prior change expanding the set of eligible protocols beyond the two Blackstone-platform integrations.

The flag-based approach gives sysadmins explicit control, makes the opt-in visible and auditable in the collection settings UI, and removes the unbounded fallback.

JIRA: PP-4433

## How Has This Been Tested?

- `test_fetch_distinct_eligible_data_source_names` (8 parametrised cases): verifies that only flagged OPDS 2.0/OFD collections are included; unflagged collections, ineligible protocols, and mixed scenarios all behave correctly.
- `test_fetch_distinct_eligible_data_source_names_deleted_collection`: verifies the new no-fallback behaviour — a deleted collection is excluded even if it had prior playtime data.
- `test_generate_playtime_reports`: updated to set `generate_playtime_report=True` on active collections and verify that the deleted `collection4` (ds_d) no longer produces a report (3 files instead of 4).
- All 13 affected tests pass.

## Checklist

### DIFF
--- a/src/palace/manager/service/logging/log.py
+++ b/src/palace/manager/service/logging/log.py
@@ -107,6 +107,9 @@ class JSONFormatter(logging.Formatter):
                 "method": flask_request.method,
                 "host": flask_request.host_url,
             }
+            # Stable per-route pattern for log aggregation; omitted when no route matched (e.g. 404).
+            if flask_request.url_rule:
+                data["request"]["rule"] = flask_request.url_rule.rule
             if flask_request.query_string:
                 data["request"]["query"] = flask_request.query_string.decode()
             if user_agent := flask_request.headers.get("User-Agent"):
--- a/tests/manager/service/logging/test_log.py
+++ b/tests/manager/service/logging/test_log.py
@@ -193,6 +193,18 @@ class TestJSONFormatter:
             assert "query" not in request
             assert "user_agent" not in request
             assert "forwarded_for" not in request
+            # No route matched this path, so there is no rule to include.
+            assert "rule" not in request
+
+        # When the request matches a registered route, its rule pattern is included.
+        flask_app_fixture.app.add_url_rule(
+            "/books/<identifier>", "book", lambda identifier: ""
+        )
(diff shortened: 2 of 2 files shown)


jonathangreen (changes requested):
@dbernstein I really don't like baking these strings into our codebase. I imagine we will want to add more or remove various collections over time. And I'm 100% sure our naming isn't consistent across CMs, so this won't always match. 

I think a better fix here would be to let collections opt in via a setting. We could write a little script or something to set the flag on existing collections. That way it can be configured over time.

PR #3385, opened 2026-05-22
Migrate BibliothecaPurchaseMonitor to Celery tasks

dbernstein:
## Description

Replaces the legacy script-driven `BibliothecaPurchaseMonitor` (and its helper `BibliothecaTimelineMonitor`) with a pair of Celery tasks that page through MARC purchase records one day at a time, chaining via `task.replace()` until the collection is fully caught up to `utc_now()`.

This is **PR 2 of 3** in the Bibliotheca → Celery migration series (PR 1 migrated the event monitor; PR 3 will migrate the circulation sweep).

FYI:  I create a follow-on [ticket](https://ebce-lyrasis.atlassian.net/browse/PP-4467) to remove dead monitor infrastructure code.

---

### Task architecture

**`import_purchase_records_for_all_collections`** (fan-out)
- Scheduled via Celery beat daily at 4:00 AM.
- Queries all Bibliotheca collections from the database and enqueues one `import_purchase_records_by_collection` task per collection.
- Accepts an optional `force_reimport=True` flag (see [Manual scripts](#manual-scripts) below).

**`import_purchase_records_by_collection`** (per-collection worker)
- Processes a single page of up to 50 MARC records for a one-day window `[current_day, day_end]`, where `day_end = min(current_day + 1 day, utc_now())`.
- After each page it re-queues itself via `task.replace()`:
  - **Full page (50 records):** more records remain for the same calendar day → re-queues with `offset` advanced by 50, `current_day` unchanged.
  - **Partial page (<50 records):** the day is fully processed → re-queues with `current_day` advanced to `day_end` and `offset` reset to 1.
- The loop terminates naturally when `current_day >= utc_now()` (collection is up to date).

---

### First run vs. subsequent runs

| | First run | Subsequent run |
|---|---|---|
| Start derived from | `DEFAULT_PURCHASE_RECORD_START_TIME` (2014-01-01) when no `Timestamp` exists; otherwise `Timestamp.finish` | `Timestamp.finish` (resume from last completed day) |
| Scope | Full historical backfill from 2014-01-01 to now | Only the delta since last run |

Mid-chain invocations receive `current_day` as an explicit parameter, bypassing `get_start()` entirely so the stored `Timestamp` is never consulted again until the chain finishes.

---

### Timestamp checkpointing

`Timestamp.finish` is updated after every page — not just at the end of the chain — to guard against crashes mid-backfill:

- **Day still in progress** (page was full): `finish = current_day`. A restart will re-process the current day from offset 1 rather than falling back to a previous day.
- **Day fully processed** (partial page): `finish = day_end`. The next beat trigger picks up from the next day.

---

### Failure modes

| Failure | Behaviour |
|---|---|
| `BadResponseException` / `RequestTimedOut` | Auto-retried up to 4 times with 60 s exponential back-off. The Redis workflow lock is held across retries (via `ignored_exceptions`) so a transient API error does not open a window for a second concurrent run. |
| `RemoteIntegrationException` | Declared in `throws=` — treated as a terminal, non-retriable error. The exception is logged and discarded; Celery does not mark the task as failed. This approach mirrors other celery importer routines. |
| Redis lock already held on first invocation | The task logs a warning and returns immediately, leaving the in-progress chain undisturbed. |
| Redis lock expires mid-chain (rare) | The task logs a warning ("workflow lock expired between invocations") and continues processing; a concurrent run may briefly exist for the same collection. |
| Collection deleted mid-chain | `ModelNotFoundError` is caught, a warning is logged, and the chain stops cleanly without raising. |
| Collection marked for deletion | `collection.marked_for_deletion` is checked after load; a warning is logged and the chain stops without making any API calls. |

---

### Redis workflow lock

- Key: `["PurchaseRecordCollectionWorkflow", <collection-redis-key>]`
- Lock timeout: 2 hours (generous ceiling for a large historical backfill page).
- A UUID `lock_value` is generated on the first invocation and forwarded unchanged to every subsequent re-queued task. Because `task.replace()` raises `celery.exceptions.Ignore`, `Ignore` is listed in `ignored_exceptions` so the lock is not released when the task replaces itself.
- The lock key is **distinct** from the event-import lock so both workflows can run concurrently for the same collection without blocking each other.

---

### MARC record processing

Each MARC record goes through `BibliothecaPurchaseRecordImporter._process_record()`:

1. Extracts the Bibliotheca ID from MARC field `001` (logs an error and skips if missing or ambiguous — i.e. zero or more than one `001` field).
2. Creates or finds the `LicensePool` via `LicensePool.for_foreign_id()`.
3. **For new titles only** (`is_new=True`): calls `api.bibliographic_lookup()`. If the lookup returns no results, a warning is logged (LicensePool exists but metadata may be incomplete). Otherwise, for each result, checks `bibliographic.needs_apply()` (hash-based deduplication). Only changed metadata queues a `bibliographic_apply` task; unchanged records produce no database writes. Existing titles skip the lookup entirely to avoid unnecessary outbound API calls during historical backfills.

---

### Manual scripts

**`bin/bibliotheca_purchase_record_import`** (CLI entry point) dispatches `ImportPurchaseRecordCollection`:

```
# Import a single named collection
bin/bibliotheca_purchase_record_import --collection "My Bibliotheca"

# Queue all collections
bin/bibliotheca_purchase_record_import --import-all

# Force a full reimport from 2014-01-01, ignoring the stored Timestamp
bin/bibliotheca_purchase_record_import --collection "My Bibliotheca" --force-reimport
bin/bibliotheca_purchase_record_import --import-all --force-reimport
```

`--force-reimport` passes `reset_timestamp=True` and `current_day=DEFAULT_PURCHASE_RECORD_START_TIME` to `import_purchase_records_by_collection.delay()`. On the first invocation, `reset_timestamp=True` clears `Timestamp.finish` before deriving the start date, so that a crash before any records are processed causes the next scheduled run to fall back to 2014-01-01 rather than the stale pre-reimport finish date. `reset_timestamp` is not forwarded to replacement tasks.

---

## Motivation and Context

JIRA: (PP-4391)

The three Bibliotheca monitors (event, purchase, circulation sweep) are driven by external cron jobs invoking one-shot scripts backed by a legacy `Monitor` base-class hierarchy. Migrating to Celery gives us:

- Centralised scheduling via beat (no external cron dependency)
- `task.replace()` to avoid tying up a worker thread for the full duration of a large historical backfill
- Standard retry/back-off handling via `autoretry_for` instead of bespoke try/except loops
- Deletion of several hundred lines of monitor scaffolding

## How Has This Been Tested?

~65 tests across five test files — all passing:

- **`test_bibliotheca.py`** — `TestImportPurchaseRecordsForAllCollections` and `TestImportPurchaseRecordsByCollection`: first-run default start, stored-timestamp resume, already-up-to-date early exit, full-page replace with advanced offset, partial-page replace with advanced day, `force_reimport` flag propagation (including `reset_timestamp`), force-reimport clears stale `Timestamp.finish`, `reset_timestamp` not forwarded to replacement tasks, lock-held skip/warning, lock-expired mid-chain warning, lock forwarded across replaces, lock not released on autoretry, lock independence from event-import lock, collection deleted mid-chain stops cleanly, collection marked for deletion stops cleanly, end-to-end chain integration test (real `task.replace()` across 3 days without mocking).
- **`test_bibliotheca_purchase_record_importer.py`** — `get_start()` with and without prior timestamp; `import_day()` window bounds, cutoff capping, timestamp stamping (mid-day vs. day-complete), record count, offset forwarding, next-offset logic; `_process_record()` happy path, missing control number, multiple control numbers, `bibliographic_apply` queued/skipped based on `needs_apply()`, `bibliographic_lookup` skipped for existing records, warning logged when lookup returns no results.
- **`test_bibliotheca_scripts.py`** — both `ImportEventCollection` and `ImportPurchaseRecordCollection`: `--import-all`, `--collection`, `--force-reimport` (with `reset_timestamp` propagation), unknown collection error, missing args error, conflicting args error.

## Checklist


src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +159,142 @@ def import_collection(
                     lock_value=lock_value,
                 )
             )
+
+
+# ---------------------------------------------------------------------------
+# Purchase monitor
+# ---------------------------------------------------------------------------
```
dbernstein: remove comment.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +159,142 @@ def import_collection(
...
+
+
+def _purchase_workflow_lock(
+    client: Redis, collection_id: int, random_value: str
+) -> RedisLock:
+    """Create a workflow-level lock for the purchase monitor.
```
dbernstein: monitor -> importer

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +159,142 @@ def import_collection(
...
+
+    Uses a key distinct from ``import_workflow_lock`` so that event-import
+    and purchase-import runs for the same collection do not block each other.
```
dbernstein: remove comment.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +159,142 @@ def import_collection(
...
+    return RedisLock(
+        client,
+        ["PurchaseCollectionWorkflow", Collection.redis_key_from_id(collection_id)],
+        random_value=random_value,
+        lock_timeout=timedelta(hours=2),
+    )
+
+
+@shared_task(queue=QueueNames.default, bind=True)
+def purchase_all_collections(task: Task) -> None:
```
dbernstein: change name according to first comment.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +159,142 @@ def import_collection(
...
+
+@shared_task(
+    queue=QueueNames.default,
+    bind=True,
+    max_retries=4,
+    autoretry_for=(BadResponseException, RequestTimedOut),
+    throws=(RemoteIntegrationException,),
+    retry_backoff=60,
+)
+def purchase_collection(
```
dbernstein: Same as above.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +161,180 @@ def import_collection(
                     lock_value=lock_value,
                 )
             )
+
+
+# ---------------------------------------------------------------------------
+# Purchase record importer
+# ---------------------------------------------------------------------------
```
dbernstein: Please remove this comment.

src/palace/manager/integration/license/bibliotheca_purchase_record_importer.py
```diff
@@ -0,0 +1,205 @@
...
+from palace.manager.integration.license.bibliotheca import BibliothecaAPI
+from palace.manager.sqlalchemy.model.collection import Collection
+from palace.manager.sqlalchemy.model.coverage import Timestamp
+from palace.manager.sqlalchemy.model.identifier import Identifier
+from palace.manager.sqlalchemy.model.licensing import LicensePool
+
+PURCHASE_RECORD_SERVICE_NAME = "Bibliotheca Purchase Record Importer"
+
+# The importer starts from this date when no prior Timestamp exists.
+# Bibliotheca collections typically go back to 2014-01-01.
```
dbernstein: Please remove "Bibliotheca collections typically go back to 2014-01-01." Add a note saying:  "Earlier versions of this importer set the default start date to 2014-01-01.  For the sake of consistency, we will preserve this default until there is a clear rationale for changing it.

src/palace/manager/integration/license/bibliotheca_purchase_record_importer.py
```diff
@@ -0,0 +1,207 @@
...
+        Falls back to :data:`DEFAULT_PURCHASE_RECORD_START_TIME` (2014-01-01) when no
+        prior run has been recorded, so the first run begins a full historical
+        backfill from the earliest possible purchase record date.
+
+        :returns: The start datetime for the next import day.
+        """
+        timestamp = Timestamp.lookup(
+            self._session,
+            PURCHASE_RECORD_SERVICE_NAME,
+            Timestamp.MONITOR_TYPE,
```
dbernstein: This is not a monitor type: it should be a TASK_TYPE.

src/palace/manager/integration/license/bibliotheca_purchase_record_importer.py
```diff
@@ -0,0 +1,207 @@
...
+        day_complete = records_handled < _MARC_PAGE_SIZE
+        next_offset = None if day_complete else offset + _MARC_PAGE_SIZE
+
+        # Always checkpoint: advance finish to day_end when the day is done,
+        # or to current_day while still in progress, so a restart after a
+        # crash resumes from this day rather than the previous one.
+        Timestamp.stamp(
+            self._session,
+            service=PURCHASE_RECORD_SERVICE_NAME,
+            service_type=Timestamp.MONITOR_TYPE,
```
dbernstein: Also TASK_TYPE , not  MONITOR_TYPE

.pre-commit-config.yaml
```diff
@@ -1,5 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+default_language_version:
+  python: python3.12
```
jonathangreen: This seems unrelated?

bin/bibliotheca_purchase_monitor
```diff
@@ -1,11 +0,0 @@
-#!/usr/bin/env python
-"""Ask the Bibliotheca API about license purchases, potentially purchases
-that happened many years in the past."""
-
-
-from palace.manager.integration.license.bibliotheca import (
```
jonathangreen, line 6: Need to remove this from the crontab as well as deleting it.

https://github.com/ThePalaceProject/circulation/blob/a7c83c9899f544f5bd8d6e650278ec830eed6193/docker/services/cron/cron.d/circulation#L23

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +164,208 @@ def import_collection(
...
+    task: Task, *, force_reimport: bool = False
+) -> None:
+    """Queue an ``import_purchase_records_by_collection`` task for every Bibliotheca collection.
+
+    :param force_reimport: When ``True``, each per-collection task ignores the stored
+        ``Timestamp`` and reimports from :data:`DEFAULT_PURCHASE_RECORD_START_TIME`
+        (2014-01-01) rather than resuming where the last run left off.  The stored
+        ``Timestamp.finish`` is also cleared so that a crash before the first record
+        is processed causes the next scheduled run to start from
+        :data:`DEFAULT_PURCHASE_RECORD_START_TIME` rather than the stale finish date.
```
jonathangreen: Minor: Is this claim actually true? It all happens in a transaction, so a crash will just roll back. Including the `timestamp.finish`. To be clear, I think the existing behavior is fine. But maybe we want to make the documentation here accurate.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -152,3 +164,208 @@ def import_collection(
...
+                return
+
+            importer = BibliothecaPurchaseRecordImporter(session, collection)
+
+            # On a force-reimport's first invocation, clear Timestamp.finish so
+            # that a crash before the first record is processed causes the next
+            # scheduled run to fall back to DEFAULT_PURCHASE_RECORD_START_TIME
+            # rather than the stale pre-reimport finish date.
+            if is_first_invocation and reset_timestamp:
+                ts = Timestamp.lookup(
```
jonathangreen, line 324: Minor: Do we want to migrate the old timestamps, so we don't have to do full reimports? Alternatively, if we don't want to migrate them. Do we want to delete the old timestamps, so they aren't hanging around in the database.

jonathangreen (approved):
Looks good. A couple minor comments to consider and a couple others I'd like to see addressed before merging.

src/palace/manager/celery/tasks/bibliotheca.py
dbernstein: I went back and forth on this.  I thought it would be good to do a full reimport to exercise the new tooling.   I'd like to remove the old timestamps  as a cleanup task in the future after all three of the bibliotheca monitors have been merged and released.  That way if we need to roll back for any reason, the old monitor states are preserved.

.pre-commit-config.yaml
dbernstein: I will revisit.

PR #3394, opened 2026-05-27
Generate Authentication For OPDS documents from pydantic models (PP-3842)

jonathangreen:
## Description

Builds the library Authentication For OPDS document from the `palace-opds` pydantic models instead of assembling a `dict` by hand and serializing it with `json.dumps`.

Standard fields follow the [Authentication for OPDS 1.0 spec](https://drafts.opds.io/authentication-for-opds-1.0.html) and its [JSON schema](https://drafts.opds.io/schema/authentication.schema.json). Palace-specific (non-spec) fields are kept in their own extension classes in `palace.opds.palace` and composed into the served document, mirroring how `opds2.py` composes `rwpm` + `palace` link properties.

Key changes:
- Add Palace auth extension models to `palace/opds/palace.py`: `LocalizedValue`, `AuthenticationInput`/`AuthenticationInputs`, `WebColorScheme`, `PublicKey`, `Features`, `AnnouncementDocument`, and `AuthenticationExtension` (the per-flow `description`/`inputs`).
- Add `AuthenticateLink`, `PalaceAuthentication`, and `PalaceAuthenticationDocument` to `palace/opds/authentication.py`; switch the spec models (`Authentication`, `AuthenticationLabels`, `AuthenticationDocument`) to `BaseOpdsModel`.
- Each auth provider (`basic`, `basic_token`, `oidc`, `saml`) now returns a `PalaceAuthentication` via a new `OPDSAuthenticationFlow` ABC in `api/authentication/opds.py`, replacing the legacy `util/authentication_for_opds.py` module (deleted).
- `LibraryAuthenticator.create_authentication_document` builds a `PalaceAuthenticationDocument` and returns `json.dumps(document.serialize())`.

The emitted JSON is unchanged, with one tiny intentional exception (see Motivation), so existing mobile/web clients are unaffected. This is effectively a pure refactor.

## Motivation and Context

The authentication document was assembled as a plain `dict` and then mutated to bolt on Palace-specific fields, with each provider hand-building its own `dict`. We already had pydantic spec models for parsing remote authentication documents but were not using them to build our own. This moves document generation onto those typed, validated models and keeps the Palace extensions cohesively defined alongside the other OPDS extensions. PP-3842.

Backward-compatibility details preserved: `features`/`announcements`/SAML+OIDC localized arrays stay present-when-empty; the logout link carries no localized keys or `properties`; `public_key.type` is always emitted; the library description is served under `service_description` (not the spec `description`); and the served `Content-Type` remains `application/vnd.opds.authentication.v1.0+json`.

The one deliberate deviation: a `help` link built from `help_email` (whose media type is `None`) previously serialized as `"type": null`; it now omits the `type` key entirely. This is semantically identical for any client (a missing media type and a null media type read the same) and is arguably more correct. It's covered by a test.

## How Has This Been Tested?

- Ran the affected test suite via `tox -e py312-docker` (authenticator, index controller, basic-token, OIDC, SAML providers, and the OPDS-license auth-document parsing tests) — all pass. The strict `==` assertions in the provider/document tests act as a byte-equivalence guard against JSON drift.
- Added `tests/palace_opds/test_authentication.py` covering the new composed models (always-present-empty arrays, dropped optional fields, plain vs authenticate links, parsing remote documents with extra fields).
- `mypy` passes across the repo; pre-commit is clean.

## Checklist


tdilauro (approved):
Looks good! 🚀

PR #3399, opened 2026-05-27
Add matched route rule to request log data

jonathangreen:
## Description

Include the matched Flask route rule in the `request` section of JSON log records. The rule is the route pattern string (e.g. `/<library_short_name>/works/<identifier_type>/<path:identifier>`) rather than the concrete request path.

The field is added under `request.rule` and is guarded: `flask_request.url_rule` is `None` when no route matched (e.g. a 404), in which case the field is omitted.

## Motivation and Context

We want to be able to collapse and aggregate logging information by route, and there wasn't a nice way to do that. The existing `request.path` varies per request (e.g. `/books/12345` vs `/books/67890`), so it can't be used as an aggregation key. The matched route rule is stable per-route, making it a good key for grouping logs.

## How Has This Been Tested?

Extended `test_flask_request` in `tests/manager/service/logging/test_log.py`:
- Asserts `rule` is omitted when no route matched the request path.
- Registers a route (`/books/<identifier>`) and asserts the matched rule pattern is included in the log while `path` reflects the concrete request path.

Ran the affected tests via `tox -e py312-docker`; all pass.

## Checklist

### DIFF
--- a/src/palace/manager/service/logging/log.py
+++ b/src/palace/manager/service/logging/log.py
@@ -107,10 +107,7 @@ class JSONFormatter(logging.Formatter):
                 "method": flask_request.method,
                 "host": flask_request.host_url,
             }
-            # The matched route rule (e.g. "/<library_short_name>/works/
-            # <identifier_type>/<path:identifier>") -- a stable per-route key for
-            # log aggregation, unlike `path`, which varies per request. None when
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🪵

PR #3400, opened 2026-05-27
SAML per-library and enhanced filtering capabilities (PP-4307)

tdilauro:
## Description

Extends SAML patron filtering with per-library filter expressions, a richer evaluation context, and a free-form "Extra Data" (`extra_data`) configuration field.

Key changes:
- Filtering is moved from `SAMLAuthenticationManager` into `SAMLWebSSOAuthenticationProvider`, which has access to per-library settings and the database. The dedicated `SAMLSubjectFilter` wrapper class is removed.
- A per-library filter expression can now be configured on `SAMLWebSSOAuthLibrarySettings`. Both the integration-level and library-level expressions must evaluate to `True` for a patron to be granted access.
- Filter expressions now receive a richer context: `subject` (SAML subject), `integration` (selected integration settings), and `library` (id, name, short_name).
- A new `extra_data` JSON field on the integration settings lets operators attach arbitrary configuration values (objects, arrays, scalars) that are available to filter expressions via `integration.extra_data`.
- A `filter_context_dump()` method on `BaseSettings` collects only fields explicitly opted in via `patron_auth_filter_context=True` on their `FormMetadata`, keeping credentials and large blobs out of the expression context.

## Motivation and Context

Operators need the ability to restrict patron access based on per-library policy as well as broader integration-level rules. Previously, only a single global filter expression was supported, and it had no access to library identity or operator-supplied metadata. This change enables multi-library deployments to apply different access rules per library and to parameterize those rules through configuration rather than by duplicating expressions.

[Jira PP-4307]

## How Has This Been Tested?

- New unit tests cover `filter_context_dump()` in `BaseSettings`, validation of the `extra_data` field, and validation of the new library-level filter expression.
- Integration tests for `_filter_subject` cover: no expression (pass-through), expression pass, expression reject, evaluation error, library-not-found error, integration/library AND behavior, and access to `integration.*` / `library.*` context variables.
- Existing `SAMLAuthenticationManager` filter tests removed (responsibility moved to provider); provider tests expanded accordingly.
- CI tests and checks pass.

## Checklist


jonathangreen (approved):
Looks good to me!

PR #3403, opened 2026-05-28
Migrate BibliothecaCirculationSweep to Celery tasks

dbernstein:
## Description

Migrates `BibliothecaCirculationSweep` — a legacy `IdentifierSweepMonitor` that previously ran via an external cron job — into a pair of Celery tasks that page through all identifiers in a collection one batch at a time, chaining successive batches via `task.replace()`.

This is the third and final PR in the Bibliotheca → Celery migration series:
- PR 1: event monitor (`BibliothecaEventImporter` / `import_collection`)
- PR 2: purchase record importer (`BibliothecaPurchaseRecordImporter` / `import_purchase_records_by_collection`)
- **PR 3 (this PR)**: circulation sweep (`BibliothecaCirculationUpdater` / `circulation_update_collection`)

---

### Loop implementation

Each sweep run starts from `circulation_update_all_collections` (beat-triggered once a day at 5:00 AM), which fans out one `circulation_update_collection` task per active Bibliotheca collection. Each task invocation:

1. Reads `Timestamp.counter` via `updater.get_offset()` to find the last identifier DB ID processed (0 on first run or after a completed sweep).
2. Queries up to 25 identifiers with `id > offset` scoped to the collection, ordered by `id`.
3. Calls `api.bibliographic_lookup()` on the batch; applies results (hash-based dedup via `needs_apply()`); zeros out availability for any identifier Bibliotheca no longer recognises.
4. If the page is full (25 records), updates `Timestamp.counter` to the last ID seen and calls `task.replace()` with the next offset — the task replaces itself in-flight so no extra queue slot is consumed.
5. If the page is partial (sweep complete), resets `Timestamp.counter` to 0 and stamps `Timestamp.finish` with the current time. The next beat trigger starts from scratch.

Concurrency is bounded by a per-collection Redis workflow lock (`CirculationUpdateWorkflow` key, 2-hour timeout) via `workflow_lock_guard`. The lock's owner token is the Celery task id, which Celery preserves across `task.replace()` and retries, so every batch re-acquires — and renews the TTL of — the same lock. The guard intentionally does **not** release the lock when `task.replace()` (which raises `Ignore`) or a retried exception propagates, so it is held for the whole sweep; it is released only on normal completion of the final batch or on an unexpected error. This ensures at most one sweep chain per collection runs at a time.

### Failure modes

| Scenario | Outcome |
|---|---|
| Task exception during a batch | Celery retries up to 4 times (60 s back-off) for `BadResponseException` / `RequestTimedOut`; non-retriable errors raise `RemoteIntegrationException` and the chain stops at that batch |
| Process crash mid-sweep | `Timestamp.counter` holds the last committed offset; next beat trigger resumes from there |
| Sweep takes longer than 24 hours | Lock prevents the next beat invocation from starting a duplicate chain |
| Bibliotheca returns no data for an ID | Availability is zeroed out unconditionally (cannot use `needs_apply()` without data) |

### First run vs. subsequent runs

**First run:** `Timestamp.counter` is 0 (no prior record), so the sweep starts from the beginning of the collection's identifier list and pages through all identifiers until a partial batch signals completion.

**Subsequent runs:** `Timestamp.counter` is 0 (reset at completion), so every beat trigger starts a fresh full-collection sweep. There is no incremental/delta mode — the entire collection is re-checked each cycle.

### Manual scripts

```bash
# Kick off a sweep for all Bibliotheca collections
bin/bibliotheca_circulation_update --import-all

# Kick off a sweep for a single named collection
bin/bibliotheca_circulation_update --collection "My Library Bibliotheca"
```

Both dispatch Celery tasks asynchronously; the script exits immediately and workers handle the actual processing. The `CirculationUpdateCollection` script class lives in `bibliotheca_scripts.py` alongside the existing `ImportEventCollection` and `ImportPurchaseRecordCollection`.

---

### Key design decisions

- **`Timestamp.counter` (not `Timestamp.finish`)** is used for mid-sweep pagination because the sweep tracks progress by identifier DB ID (an int), not a datetime.
- **`process_identifiers()` for on-demand use**: `BibliothecaAPI.update_availability()` and `AvailabilityRefreshScript` need to process a specific list of identifiers directly. `process_identifiers()` provides this path without touching the Timestamp and applies the same `needs_apply()` hash guard. Unlike the sweep (which queues asynchronous `bibliographic_apply` tasks), this on-demand path applies changes **synchronously** in the caller's session, because callers such as the circulation dispatcher read `LicensePool.licenses_available` immediately after requesting a refresh.
- **Availability bug fixed**: The old `availability.py` code called `sweeper.process_batch(identifiers)` passing `Sequence[Identifier]` where `SweepMonitor.process_batch()` expected an `int` offset. The migration to `process_identifiers()` silently fixes this.
- **Lock key** `"CirculationUpdateWorkflow"` is distinct from `"PurchaseRecordCollectionWorkflow"` and the event-import lock so all three workflows can run concurrently per collection.

## Motivation and Context

The legacy `BibliothecaCirculationSweep` (an `IdentifierSweepMonitor`) ran as a blocking process via an external cron job with no built-in retry logic, no Redis lock to prevent overlap, and no progress checkpointing. Migrating to Celery gives us retries, distributed execution, workflow lock protection, and Celery beat scheduling — consistent with the rest of the Bibliotheca import pipeline modernized in PRs 1 and 2.

## How Has This Been Tested?

- All new and existing tests in the affected suites pass.
- Tests cover: `get_offset`, `update_batch` (first run, mid-sweep, completion, full/partial page), `process_identifiers` (including synchronous apply on the on-demand path), task fan-out, task self-chaining via `replace()`, lock acquisition and hold across `replace()`/retries, and all `CirculationUpdateCollection` script variants.
- Verified no remaining references to `BibliothecaCirculationSweep` in `src/` or `tests/`.
- mypy passes on all modified source files.
- manually tested the updater on a local instance.

## Checklist


dbernstein:
On the `availability.py` local-import finding — I dug into it and the local import actually does need to stay, though the comment was wrong about *why*.

Moving the import to module top breaks with a circular `ImportError` whenever `availability.py` is imported before the Celery tasks have loaded. The real cycle is:

`availability` → `bibliotheca_circulation_updater` → (`from palace.manager.celery.tasks import apply`) → `celery/tasks/__init__.py` eagerly autoloads **every** task module via `import_celery_tasks()` → `celery.tasks.bibliotheca` → `from ...bibliotheca_circulation_updater import BibliothecaCirculationUpdater` (partially-initialized) → 💥

So `availability.py` isn't directly part of the loop — it just triggers the `updater → celery.tasks → bibliotheca → updater` cycle by being the first thing to import the updater. I verified this empirically (importing `palace.manager.scripts.availability` first raises `ImportError: cannot import name 'BibliothecaCirculationUpdater' ... (most likely due to a circular import)`; importing a celery task module first, then `availability`, works).

I kept the local import but rewrote the comment to describe the actual cycle (commit `f50336562`).

I also rebased the branch on the latest `main` and addressed a couple of things found in a follow-up review:
- Removed the obsolete `bibliotheca_circulation_sweep` cron entry + duplicate bin script — the cron invoked the script with no args, which the new `CirculationUpdateCollection` rejects, and the sweep is now beat-scheduled (`3e65f5086`).
- Made `BibliothecaAPI.update_availability()` apply synchronously again on the on-demand path — it had become async via `bibliographic_apply.delay()`, but `dispatcher.enforce_limits()` reads `pool.licenses_available` immediately after calling it, so it needs in-band application (`65a247e46`).

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -325,3 +334,138 @@ def import_purchase_records_by_collection(
...
+) -> RedisLock:
+    """Create a workflow-level lock for the circulation updater.
+
+    Uses a key distinct from the event-import and purchase-record-import locks
+    so all three workflows can run concurrently per collection.
+    """
+    return RedisLock(
+        client,
+        [
+            "CirculationUpdateWorkflow",
```
tdilauro, line 350: Very minor - You could consider using an enum for these distinct keys. That would avoid bare strings (not as much of a problem in this case, since it looks like its only used here in code and below in a single comment) and ensure that key prefixes aren't accidentally duplicates.

src/palace/manager/celery/tasks/bibliotheca.py
```diff
@@ -325,3 +334,138 @@ def import_purchase_records_by_collection(
...
+    )
+
+
+@shared_task(queue=QueueNames.default, bind=True)
+def circulation_update_all_collections(task: Task) -> None:
+    """Queue a ``circulation_update_collection`` task for every Bibliotheca collection.
+
+    This is the beat-scheduled entry point for the circulation sweep.  It runs once a
+    day (05:00), matching the cadence of the legacy ``bibliotheca_circulation_sweep``
+    cron job it replaced.
```
tdilauro: Suggestion - remove this docstring paragraph. It would be useful as a PR comment, but it references legacy code (pretty much immediately out of date) and a schedule that is maintained elsewhere (has to be maintained in more that one place). The top line seems clear enough on its own.

src/palace/manager/integration/license/bibliotheca.py
```diff
@@ -1202,109 +1206,6 @@ def process_one(
...
-            pools = [
-                lp
-                for lp in identifier.licensed_through
-                if lp.data_source.name == DataSource.BIBLIOTHECA
-                and lp.collection == self.collection
-            ]
-            if pools:
-                [pool] = pools
-            else:
-                continue
```
tdilauro, line 1272: `[pool] = pools` is performing an implicit check that there is exactly one and only one `pool` in `pools` here. That check is not included in the new `BibliothecaCirculationUpdater._process_batch` method. I'll add a comment about the requested change there, as well.

src/palace/manager/integration/license/bibliotheca_circulation_updater.py
```diff
@@ -0,0 +1,247 @@
...
+        for identifier in identifiers_not_mentioned:
+            pools = [
+                lp
+                for lp in identifier.licensed_through
+                if lp.data_source.name == DataSourceConstants.BIBLIOTHECA
+                and lp.collection == self._collection
+            ]
+            if not pools:
+                continue
+            pool = pools[0]
```
tdilauro: Change requested - The `pool = pools[0]` here is dropping the implicit check that ensured that there was exactly one and only one `pool` in `pools` that was previously performed in the just-deleted `BibliothecaCirculationSweep.process_items` (see paired comment above).

We could do something like the change suggested below to get back to parity or we could check and handle the case explicitly.

```suggestion
            [pool] = pools
```

tdilauro (commented):
This is looking good. I left a few comments below.

My biggest concerns are general, so are not included below:
- I'm wondering if the main PR description comment is up to date. It mentions that the workflow lock is captured "and forwarded via `lock_value` ...", but I don't see that variable and it looks like the lock is persisting because we purposely not releasing it. That's all fine, but the description should be accurate or at least not misleading. At any rate, would be good to ensure that the entire description is up to date and accurate.
- I did not see an "end-to-end" workflow test with multiple replacement runs that ensures that the lock is held across all of the replacement runs and that the full sweep chain completes successfully. This is a key behavior, so seems important to test.

src/palace/manager/celery/tasks/bibliotheca.py
dbernstein: Good call — removed the paragraph in 7384240e9, leaving just the summary line. The cadence lives in `service/celery/celery.py` as the single source of truth, and this drops the now-stale reference to the removed `bibliotheca_circulation_sweep` script.

dbernstein: You're right on both points re the PR description — there is no `lock_value` in this code (it only survives in some test locals), and the lock persists precisely because the guard deliberately doesn't release it across the `task.replace()`/retry hand-offs.  It was factored out with the arrival of the `workflow_lock_guard`.

To be precise about the mechanism: `workflow_lock_guard` is a new function (introduced by Jonathan) that uses the Celery task id as the lock's owner token, which Celery preserves across `task.replace()` and retries, so each batch re-acquires (and renews the TTL of) the same lock. It passes `ignored_exceptions=(Ignore, *autoretry_for)` to `RedisLock.lock()`, so the `finally` block skips `release()` when `task.replace()` (raises `Ignore`) or a retried exception propagates. The lock *is* released — on normal completion of the final batch, and on any unexpected (non-ignored) error.

I had Claude rewrite that paragraph in the description accordingly, and had him do (another)  full accuracy pass over the rest while I was at it:
- Fixed the prior-PR task names (`import_collection`, `import_purchase_records_by_collection`).
- Noted that the on-demand `process_identifiers()` path now applies **synchronously** (the dispatcher reads availability immediately after).
- Dropped the stale `137 tests pass` count and the "lock acquisition/forwarding" wording in the testing section.

Verified the remaining claims still hold (25-batch size, retry/throws config, the `availability.py` `process_batch`-offset bug, distinct lock keys).

dbernstein:
@tdilauro — on the missing end-to-end test: good call. Added one in `eff94ae82`.

`TestCirculationUpdateCollection::test_full_sweep_chains_through_every_batch_and_releases_lock` drives a real multi-batch sweep through `task.replace()` to completion (`replace` not mocked): 26 identifiers → batch of 25 (full → `replace()`) → batch of 1 (partial → complete). It asserts the chain ran every batch (`bibliographic_lookup` called twice), the sweep completed (`Timestamp.counter` reset to `0`, `finish` stamped), and the workflow lock was held across the `replace()` hand-off and released on completion.

One possibly pedantic caveat which you are probably already aware of but in case it's not clear: a live chain runs straight through in the worker, so it proves "chain completes + lock released at the end" but can't pause between hand-offs to assert "lock held during each intermediate run." That stays covered by the unit tests:
- `RedisLock::test_lock_not_released_on_ignored_exception` — the `Ignore` that `task.replace()` raises doesn't release the lock;
- `workflow_lock_guard::test_continuation_reacquires_own_lock` — a continuation re-acquires the same lock under the preserved task id;
- `test_lock_not_released_on_autoretry` — the lock stays held across repeated runs under one task id.

So between the new end-to-end test and those existing tests (from Jonathan's PR), the new behavior is solidly covered.

tdilauro (approved):
🚀

PR #3405, opened 2026-05-28
Escape regex metacharacters in search query parser

jonathangreen:
## Description

`QueryParser._without_match` interpolates a user-derived match string directly into a regex pattern (`r"\b%s[\w'\-]*\b" % match`). When the matched portion contains regex metacharacters — most commonly an unbalanced `(` pulled from the search query — `re.compile` raises `re.error: missing ), unterminated subpattern`, the search request 500s, and `ErrorHandler` logs the exception.

The fix wraps the match in `re.escape()` so any metacharacters from user input are treated literally. A regression test exercises the exact production query.

## Motivation and Context

Observed in production for the query `robbie faith inspired romance (european romance suspense series book 2)`. `KeywordBasedClassifier.genre_match` returned the matched portion `romance (european romance suspense`, which then broke regex compilation in `_without_match`. The same crash has shown up before in production — see [this Slack thread](https://lyrasis.slack.com/archives/C049AA32E3S/p1779990695142299).

```
File ".../palace/manager/search/query.py", line 1098, in _without_match
    return re.compile(word_boundary_pattern % match.strip(), re.IGNORECASE).sub(
re.error: missing ), unterminated subpattern at position 10
```

## How Has This Been Tested?

Added a regression case to `TestQueryParser.test__without_match` using the failing production query verbatim, and confirmed it fails on `main` and passes with the fix. Ran:

```
tox -e py312-docker -- --no-cov tests/manager/search/test_query.py -k "test__without_match or test_constructor"
```

All tests pass.

## Checklist

### DIFF
--- a/src/palace/manager/search/query.py
+++ b/src/palace/manager/search/query.py
@@ -1095,6 +1095,9 @@ class QueryParser:
         # dash.
         word_boundary_pattern = r"\b%s[\w'\-]*\b"
 
-        return re.compile(word_boundary_pattern % match.strip(), re.IGNORECASE).sub(
-            "", query_string
-        )
+        # The match is taken from the user's query string and may contain
+        # regex metacharacters (e.g. an unbalanced parenthesis), so escape
+        # it before interpolating into the pattern.
+        return re.compile(
+            word_boundary_pattern % re.escape(match.strip()), re.IGNORECASE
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
🐛 🔨

PR #3407, opened 2026-05-28
Add generate_playtime_report collection setting to gate playtime reports (PP-4346)

dbernstein:
## Description

Replaces the current behavior of generating reports for every OPDS2 and OPDSForDistributors protocol with an explicit per-collection opt-in flag that controls whether a collection's playtime data is included in the monthly Google Drive report.

### Changes

**New setting: `generate_playtime_report` (bool, default `False`)**

Added via the `PlaytimeReportSettings` mixin to:
- `OPDS2APISettings` — the settings class used exclusively by `OPDS2API` (OPDS 2.0 Import)
- `OPDSForDistributorsSettings` — the settings class used by `OPDSForDistributorsAPI`

The field is **not** present on `OPDS2ImporterSettings` (the shared base class), which means `OPDS2WithODLSettings` (ODL 2.0) does not inherit it and the setting does not appear on ODL 2.0 collection forms. `OPDS2APISettings` is a thin subclass of `OPDS2ImporterSettings` that adds the mixin.

The field is sysadmin-facing and rendered as a SELECT ("Generate playtime report for audio books"). All other protocols are ineligible by design.

**Migration**

Sets `generate_playtime_report = True` for all existing OPDS 2.0 / OPDS for Distributors collections whose `data_source` name begins with `"Blackstone"` or `"Unlimited"`. All other collections default to `False` and require no update. I verified in Redshift that using the aforementioned constraints will match all Blackstone and Unlimited Listens data sources and no others.

Downgrade removes the key entirely (absence is equivalent to the Pydantic default of `False`).

**`_fetch_distinct_eligible_data_source_names`**

Now returns only data source names for collections where `generate_playtime_report` is explicitly `True`. The `PlaytimeSummary` fallback (which was the root cause of folder proliferation — it included every historical data source regardless of intent) has been removed.

> **Behavioral note — deleted collections:** Because eligibility is now determined entirely by a collection's settings, a deleted collection produces no report regardless of how much historical playtime data it accumulated. Without a live collection there are no collection settings, and therefore no `generate_playtime_report` flag to evaluate. This is intentional: report generation should reflect an active, administrator-approved configuration, not the presence of stale summary rows. If we need to go back and find playtime summaries for deleted collections, the data will still be there.

## Motivation and Context

The Partner Success team observed large numbers of Google Drive folders being generated for every data source in the system. The underlying cause was a combination of:

1. `_fetch_distinct_eligible_data_source_names` returning all data sources that ever appeared in `PlaytimeSummary` records, and
2. A prior change expanding the set of eligible protocols beyond the two Blackstone-platform integrations.

The flag-based approach gives sysadmins explicit control, makes the opt-in visible and auditable in the collection settings UI, and removes the unbounded fallback.

JIRA: PP-4346

## How Has This Been Tested?

- `test_fetch_distinct_eligible_data_source_names` (8 parametrised cases): verifies that only flagged OPDS 2.0/OFD collections are included; unflagged collections, ineligible protocols, and mixed scenarios all behave correctly.
- `test_fetch_distinct_eligible_data_source_names_deleted_collection`: verifies the new no-fallback behaviour — a deleted collection is excluded even if it had prior playtime data.
- `test_generate_playtime_reports`: updated to set `generate_playtime_report=True` on active collections and verify that the deleted `collection4` (ds_d) no longer produces a report (3 files instead of 4).
- `test_generate_playtime_report_folder_lock_contention`: verifies that the task completes and logs a warning when the folder-creation lock cannot be acquired (from PR #3377, now included here after rebase).
- All 14 affected tests pass.

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/opds/opds2/settings.py
+++ b/src/palace/manager/integration/license/opds/opds2/settings.py
@@ -22,7 +22,7 @@ from palace.manager.integration.settings import (
 from palace.manager.sqlalchemy.constants import IdentifierType
 
 
-class OPDS2ImporterSettings(PlaytimeReportSettings, OPDSImporterSettings):
+class OPDS2ImporterSettings(OPDSImporterSettings):
     custom_accept_header: Annotated[
         str,
         FormMetadata(
@@ -99,3 +99,13 @@ class OPDS2ImporterSettings(PlaytimeReportSettings, OPDSImporterSettings):
 
 class OPDS2ImporterLibrarySettings(OPDSImporterLibrarySettings):
     pass
+
+
+class OPDS2APISettings(PlaytimeReportSettings, OPDS2ImporterSettings):
+    """Settings for OPDS2API.
+
+    Extends :class:`OPDS2ImporterSettings` with the ``generate_playtime_report``
+    opt-in flag.  Kept as a separate subclass so that ``OPDS2WithODLSettings``
+    (which also inherits from ``OPDS2ImporterSettings``) does not inadvertently
+    expose the playtime-report setting on ODL 2.0 collection forms.
+    """
--- a/src/palace/manager/integration/license/opds/opds2/settings.py
+++ b/src/palace/manager/integration/license/opds/opds2/settings.py
@@ -20,6 +20,23 @@ from palace.manager.sqlalchemy.constants import IdentifierType
 
 
 class OPDS2ImporterSettings(OPDSImporterSettings):
+    generate_playtime_report: Annotated[
+        bool,
+        FormMetadata(
+            label=_("Generate playtime report for audio books"),
+            description=_(
+                "When enabled, this collection will be included in the monthly "
+                "playtime report uploaded to Google Drive. This is a system "
+                "administrator setting."
+            ),
+            type=FormFieldType.SELECT,
+            options={
+                True: "Yes",
+                False: "(Default) No",
+            },
+        ),
+    ] = False
+
     custom_accept_header: Annotated[
         str,
         FormMetadata(
--- a/src/palace/manager/integration/license/opds/opds2/settings.py
+++ b/src/palace/manager/integration/license/opds/opds2/settings.py
@@ -12,6 +12,9 @@ from palace.manager.integration.license.opds.opds1.settings import (
     OPDSImporterLibrarySettings,
     OPDSImporterSettings,
 )
+from palace.manager.integration.license.opds.settings.playtime_report import (
+    PlaytimeReportSettings,
+)
 from palace.manager.integration.settings import (
     FormFieldType,
     FormMetadata,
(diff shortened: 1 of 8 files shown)


src/palace/manager/integration/license/opds/settings/playtime_report.py
```diff
@@ -0,0 +1,44 @@
...
+                False: "(Default) No",
+            },
+        ),
+    ] = False
+
+
+# The key as it appears in settings_dict; derived from the model field to avoid bare string literals.
+PLAYTIME_REPORT_FLAG_KEY: str = (
+    PlaytimeReportSettings.model_fields["is_generate_playtime_report"].alias
+    or "is_generate_playtime_report"
```
dbernstein: I'm not sure why Claude added the "or" clause.

dbernstein: I'm going to change that.

dbernstein:
@jonathangreen : I reworked this one according to your suggestion re using a configuration setting on the collection rather than hardcoding the data source names in the report generator.

src/palace/manager/integration/license/opds/opds2/settings.py
```diff
@@ -12,14 +12,17 @@
...
+)
 from palace.manager.integration.settings import (
     FormFieldType,
     FormMetadata,
 )
 from palace.manager.sqlalchemy.constants import IdentifierType
 
 
-class OPDS2ImporterSettings(OPDSImporterSettings):
+class OPDS2ImporterSettings(PlaytimeReportSettings, OPDSImporterSettings):
```
jonathangreen: Because `OPDS2WithODLSettings` subclasses `OPDS2ImporterSettings`, adding `PlaytimeReportSettings` here also exposes `is_generate_playtime_report` on ODL 2.0 collection configuration forms. Since OPDS2+ODL collections are filtered out, it doesn't actually enable it for them, but it does create a misleading config setting.

src/palace/manager/integration/license/opds/settings/playtime_report.py
```diff
@@ -0,0 +1,37 @@
...
+
+class PlaytimeReportSettings(BaseSettings):
+    """Mixin that adds the is_generate_playtime_report opt-in flag.
+
+    Intended for OPDS 2.0 and OPDS for Distributors collections only.
+    When True the collection's data source is included in the monthly
+    playtime report uploaded to Google Drive.
+    """
+
+    is_generate_playtime_report: Annotated[
```
jonathangreen: Minor: I really don't like the name `is_generate_playtime_report`. `is_generate_playtime_report` is off in two ways:

1. `is_` prefix: nothing else in the settings models uses it; the convention is bare verbs.
2. `is_` + generate reads as "is generate": grammatically broken. `is_` wants an adjective/state (is_enabled), not a verb.

`opds1/settings.py` has a setting `include_in_inventory_report`, which serves a similar purpose, and follows the conventions we've been using more closely. 

My suggestion would be to rename this to either, in order of preference:
1. `include_in_playtime_report` 
2. `generate_playtime_report`

jonathangreen (approved):
I'm approving this one, but I'd to get the `OPDS2WithODLSettings` issue resolved before merging. Also added a minor comment to consider about variable naming.

PR #3408, opened 2026-05-28
Add OIDC integration- and library-level patron filtering (PP-4387)

tdilauro:
## Description

Adds patron filter expressions to the OIDC authentication provider, matching the capability just added for SAML in #3400. Two filter levels are supported:

- **Integration-level** (`filter_expression` on `OIDCAuthSettings`): applies to all libraries using this OIDC integration.
- **Library-level** (`filter_expression` on `OIDCAuthLibrarySettings`): applies only to a specific library.

Either or both expressions -- when provided -- must evaluate to `True` for a patron to be granted access. Expressions are evaluated during the OIDC callback (before any patron record is created or updated) using the existing sandboxed `FilterExpression` evaluator.

A new `extra_data` field (accepts any JSON value) is also added to `OIDCAuthSettings` to allow admins to supply arbitrary data accessible to filter expressions via `integration.extra_data`.

The evaluation context available to expressions includes:
- `claims` — ID token claims dict from the OIDC provider
- `integration` — selected integration settings (scopes, patron ID claim, session lifetime, extra data)
- `library` — id, name, short_name

## Motivation and Context

Libraries need to restrict OIDC-authenticated patron access based on claims returned in the ID token — for example, limiting access to patrons from a specific email domain or with a particular group membership. The per-library filter allows a single OIDC integration to serve multiple libraries with different access rules.

[Jira PP-4387]

## How Has This Been Tested?

- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/26605010934) and checks pass.

## Checklist

### DIFF
--- a/src/palace/manager/integration/patron_auth/oidc/configuration/model.py
+++ b/src/palace/manager/integration/patron_auth/oidc/configuration/model.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from re import Pattern
-from typing import Annotated
+from typing import Annotated, Any
 
 from flask_babel import lazy_gettext as _
 from pydantic import (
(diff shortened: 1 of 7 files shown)


dbernstein (approved):
Looks good, @tdilauro !

PR #3409, opened 2026-05-28
Fix a very slow test

tdilauro:
## Description

Patches `palace.manager.integration.patron_auth.oidc.util.HTTP.get_with_timeout` in  `tests/manager/api/test_authenticator.py::TestLibraryAuthenticator::test_register_oidc_provider` to return a minimal discovery document rather than making a real HTTP call.

## Motivation and Context

Improves specific and overall test performance by avoiding the wait for the fetch to timeout, which took a couple of minutes.

## How Has This Been Tested?

- All tests continue to pass locally.
- CI tests pass.

## Checklist

### DIFF
--- a/tests/manager/api/test_authenticator.py
+++ b/tests/manager/api/test_authenticator.py
@@ -1333,8 +1333,26 @@ class TestLibraryAuthenticator:
         providers = list(authenticator.providers)
         assert oidc_provider in providers
 
+        # Stub the discovery HTTP call so no real network request is made.
+        # Omitting end_session_endpoint/revocation_endpoint keeps supports_logout()
+        # False, so only the authenticate link appears.
+        mock_discovery = {
+            "issuer": "https://idp.example.com",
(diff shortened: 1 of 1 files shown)


tests/manager/api/test_authenticator.py
```diff
@@ -1333,8 +1333,26 @@ def test_register_oidc_provider(
...
+        mock_http_response.json.return_value = mock_discovery
+
         # Verify provider appears in authentication document
-        with app.test_request_context("/"):
+        with (
+            patch(
+                "palace.manager.integration.patron_auth.oidc.util.HTTP.get_with_timeout",
+                return_value=mock_http_response,
+            ),
+            app.test_request_context("/"),
```
jonathangreen, line 1354: Minor: You could use `MockHttpClientFixture` to do this mocking, but this totally works, and it might be easier here where you don't want the http client mocked for the whole test.

jonathangreen (approved):
Looks good! ⏱️

PR #3412, opened 2026-06-02
Fail over on OpenSearch read timeouts (PP-4472)

jonathangreen:
## Description

Make the user-facing OpenSearch read path resilient to transient node unavailability by giving reads their own client with a short timeout and timeout retries, while leaving indexing/admin behavior unchanged.

`SearchServiceOpensearch1` now takes an explicit `write_client` and `read_client`:

- **`write_client`** — indexing and admin operations. Keeps the full client-wide timeout (`write_timeout`, default **20s**) because these legitimately run longer, and has no timeout retries, so its behavior is unchanged.
- **`read_client`** — the user-facing read/search path. Configured with a short `read_timeout` (default **4s**) plus `retry_on_timeout` / `max_retries=2`, so a read that stalls against a bouncing node fails over fast to the domain's healthy node.

The settings were renamed to describe what they govern, matching the client split:

- `read_timeout` — read path timeout (`PALACE_SEARCH_READ_TIMEOUT`, default 4).
- `write_timeout` — indexing/admin timeout (`PALACE_SEARCH_WRITE_TIMEOUT`, default 20). The old `PALACE_SEARCH_TIMEOUT` is still honored as a deprecated alias and logs a warning when it is the only one set.
- `read_max_retries` / `read_retry_on_timeout` — read-path retry behavior (`PALACE_SEARCH_READ_MAX_RETRIES`, `PALACE_SEARCH_READ_RETRY_ON_TIMEOUT`).

## Motivation and Context

Production was emitting user-facing `500`s on the groups/search endpoints every night, sharply concentrated in the **04:00–07:00 UTC** window (5-day peak: 240 errors at 05:00 UTC). The errors were `opensearchpy.ConnectionTimeout` (`read timeout=20`) from `query_works_multi`.

JIRA: PP-4472

Investigation (tpp-prod):

- The errors span **all four prod ES shards**, not one cluster — a shared event, not a sick node.
- **05:00 UTC is exactly the configured OpenSearch off-peak maintenance window** (`auto_software_update_enabled = true`), and the ES application log confirms shard recovery / node bounces in that window (e.g. kappa `failed recovery … recovery was cancelled` at 05:11 UTC).
- The clusters are **healthy throughout** — 7-day SearchLatency p99.9 ≤ 41 ms (max 218 ms), no node loss, no yellow/red. So a 20 s timeout is a connection-level stall against a bouncing node, **not** a slow query.

The client did not retry on timeout, so a single stalled connection failed the whole request. With two nodes per domain, retrying against the other node recovers transparently.

Latency budget: the read path worst case becomes `4s × (1 + 2 retries) = 12s`, comfortably inside the **60 s ALB idle timeout**, while a single node bounce now recovers in ~4 s instead of failing at 20 s. The 4 s per-attempt value sits ~25× above the observed msearch server-side cost, leaving margin for the degraded (single-node) maintenance window.

## How Has This Been Tested?

- `mypy` clean.
- New `tests/manager/service/search/test_configuration.py` covers the defaults, the new `PALACE_SEARCH_WRITE_TIMEOUT` env var, the deprecated `PALACE_SEARCH_TIMEOUT` fallback (value honored + warning logged), and new-over-deprecated precedence.
- Updated `tests/manager/search/test_service.py` to assert the read search / multi-search clients are built on the dedicated read client.
- Full `tests/manager/search/` suite passes, including `test_filter.py` and `test_external_search.py`, which execute against a live test cluster via msearch and would surface the metadata-header rejection.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,8 +200,11 @@ To let the application know which OpenSearch instance to use, you can set the fo
 - `PALACE_SEARCH_URL`: The URL of the OpenSearch instance (**required**).
 - `PALACE_SEARCH_INDEX_PREFIX`: The prefix to use for the OpenSearch indices. The default is `circulation-works`.
     This is useful if you want to use the same OpenSearch instance for multiple Circulation Manager instances (optional).
-- `PALACE_SEARCH_TIMEOUT`: The timeout in seconds to use when connecting to the OpenSearch instance. The default is `20`
-  (optional).
+- `PALACE_SEARCH_WRITE_TIMEOUT`: The timeout in seconds to use for indexing and admin operations against the OpenSearch
+  instance. The default is `20` (optional). Replaces the deprecated `PALACE_SEARCH_TIMEOUT`, which is still honored for
+  backwards compatibility.
+- `PALACE_SEARCH_READ_TIMEOUT`: The timeout in seconds to use for the user-facing read/search path. Kept short so a node
+  that briefly stalls during OpenSearch maintenance fails over quickly. The default is `4` (optional).
 - `PALACE_SEARCH_MAXSIZE`: The maximum size of the connection pool to use when connecting to the OpenSearch instance
   (optional).
 
--- a/README.md
+++ b/README.md
@@ -205,6 +205,10 @@ To let the application know which OpenSearch instance to use, you can set the fo
   backwards compatibility.
 - `PALACE_SEARCH_READ_TIMEOUT`: The timeout in seconds to use for the user-facing read/search path. Kept short so a node
   that briefly stalls during OpenSearch maintenance fails over quickly. The default is `4` (optional).
+- `PALACE_SEARCH_READ_RETRY_ON_TIMEOUT`: Whether a timed-out read should be retried against another node. The default is
+  `true` (optional).
+- `PALACE_SEARCH_READ_MAX_RETRIES`: The maximum number of retries for a timed-out read before the error is propagated.
+  The default is `2` (optional).
 - `PALACE_SEARCH_MAXSIZE`: The maximum size of the connection pool to use when connecting to the OpenSearch instance
   (optional).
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -200,8 +200,15 @@ To let the application know which OpenSearch instance to use, you can set the fo
 - `PALACE_SEARCH_URL`: The URL of the OpenSearch instance (**required**).
 - `PALACE_SEARCH_INDEX_PREFIX`: The prefix to use for the OpenSearch indices. The default is `circulation-works`.
     This is useful if you want to use the same OpenSearch instance for multiple Circulation Manager instances (optional).
-- `PALACE_SEARCH_TIMEOUT`: The timeout in seconds to use when connecting to the OpenSearch instance. The default is `20`
-  (optional).
+- `PALACE_SEARCH_WRITE_TIMEOUT`: The timeout in seconds to use for indexing and admin operations against the OpenSearch
+  instance. The default is `20` (optional). Replaces the deprecated `PALACE_SEARCH_TIMEOUT`, which is still honored for
+  backwards compatibility.
+- `PALACE_SEARCH_READ_TIMEOUT`: The timeout in seconds to use for the user-facing read/search path. Kept short so a node
```
dbernstein, line 206: I'm not seeing where PALACE_SEARCH_READ_* variables are being passed to the configuration.

dbernstein (changes requested):

README.md
jonathangreen: @dbernstein I'm not sure what you mean here. The variables are loaded via pydantic settings, they are a combination of the env prefix and the attribute name.

dbernstein: Ah okay - that's right.  I had forgotten how they are loaded automatically.

dbernstein (approved):
Looks good.

PR #3413, opened 2026-06-02
Cap search query length to protect OpenSearch heap (PP-4472)

jonathangreen:
## Description

Truncate a search query to `Query.MAX_QUERY_WORDS` (32) words in the `Query` constructor, before any hypotheses are built. This bounds the cost of every hypothesis (fuzzy automata, multi-match, parsed query) so a pathologically long query string can't exhaust an OpenSearch node's per-request search heap.

## Motivation and Context

Production OpenSearch logs show search-backpressure cancelling queries with `cancelled task with reason: heap usage exceeded [...]`. The stack trace is always the fuzzy-query rewrite (`FuzzyTermsEnum` /
`TopTermsRewrite`) building Levenshtein automata over every term in the query. Two incidents in 30 days, both driven by query *length*, not the cluster being unhealthy:

- **kappa, 05-27:** a patron pasted a 656-word job description into `/GA0008/search/`. One search task hit 787MB, OpenSearch cancelled it (`All shards failed`), and the search came back as `search_phase_execution_exception` (handled as an empty result set).
- **theta, 05-22:** `/IL9993/search/` (a QA library) sent a 60-115 word type-ahead query that grew one word per keystroke, triggering a burst of fetch-phase cancellations. Recovered transparently via replica.

Both build expensive fuzzy automata over many terms. Capping the query length addresses both at the source.

JIRA: PP-4472

### Why 32 words

Validated against 30 days of production search traffic (~2.0M searches):

| percentile | words |
|---|---|
| p50 | 2 |
| p90 | 5 |
| p99 | 15 |
| p99.9 | 26 |

Exact count of searches that exceed the cap, across all ~2.0M:

| cap | over cap | % of all | ~/day |
|---|---|---|---|
| 24 words | 1,793 | 0.090% | ~59 |
| **32 words** | **1,119** | **0.056%** | **~37** |
| 48 words | 531 | 0.027% | ~17 |

32 leaves 99.94% of searches untouched and sits above the legitimate p99.9 of 26 words, so no real catalog search is affected -- only the pathological long queries are clipped (and a clipped query still searches on its first 32 words).

## How Has This Been Tested?

- `mypy` clean on `query.py`.
- New `test_constructor_caps_query_length`; updated `test_constructor`
  (`Query(None).contains_stopwords` is now `False` rather than `None` --
  both falsy, downstream only checks truthiness).

## Checklist

### DIFF
--- a/src/palace/manager/search/query.py
+++ b/src/palace/manager/search/query.py
@@ -22,6 +22,7 @@ from opensearchpy.helpers.query import (
 from spellchecker import SpellChecker
 
 from palace.util.exceptions import BasePalaceException
+from palace.util.log import LoggerMixin
 
 from palace.manager.core.classifier.age import AgeClassifier, GradeLevelClassifier
 from palace.manager.core.classifier.keyword import KeywordBasedClassifier
@@ -41,7 +42,7 @@ from palace.manager.util.personal_names import display_name_to_sort_name
 from palace.manager.util.stopwords import ENGLISH_STOPWORDS
 
 
-class Query:
+class Query(LoggerMixin):
     """An attempt to find something in the search index."""
 
     # This dictionary establishes the relative importance of the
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
Good idea.

PR #3414, opened 2026-06-02
Return 404 for permalink of a work with no presentation edition (PP-4472)

jonathangreen:
## Description

Guard the `permalink` controller against works that have no presentation edition. When such a work is encountered we now return `NOT_FOUND_ON_REMOTE` (404) instead of letting `OPDSAcquisitionFeed.single_entry` raise an unhandled `PalaceValueError`. This mirrors the existing behavior of the sibling `related` endpoint.

## Motivation and Context

We were occasionally seeing unhandled 500s in production from the permalink endpoint, e.g. `Work has no associated identifier: <Work #5968341 "None" (by None) lang=None (1 lp)>`.

The work is looked up via a valid identifier, but `single_entry` re-derives the identifier from the *work's own state* rather than the looked-up identifier. For a degenerate work both derivation paths come up empty: `active_license_pool()` returns `None` (the work's single pool is filtered out because it has no titled edition / owned licenses) and `work.presentation_edition` is also `None`. With no identifier derivable, `single_entry` raises `PalaceValueError`, which surfaces as a 500.

Since a work with no presentation edition has no usable metadata to render, treating it as "not found" is the correct, graceful behavior.

Example of error occuring in production: 
https://lyrasis.slack.com/archives/C049AA32E3S/p1780405014686049

JIRA: PP-4472

## How Has This Been Tested?

Added `test_permalink_work_with_no_presentation_edition`, which clears the work's `presentation_edition` and asserts the endpoint returns `NOT_FOUND_ON_REMOTE`. Ran the full permalink test suite (`tests/manager/api/controller/test_work.py -k permalink`, 10 passed) under the docker tox environment, plus mypy on the changed controller.

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/work.py
+++ b/src/palace/manager/api/controller/work.py
@@ -114,6 +114,13 @@ class WorkController(CirculationManagerController):
         if isinstance(work, ProblemDetail):
             return work
 
+        if work.presentation_edition is None:
+            # The work exists but has no presentation edition, so there's no
+            # usable metadata or identifier to build an entry around. Treat it
+            # as not found rather than letting single_entry raise an unhandled
+            # error. This mirrors the behavior of the `related` endpoint.
+            return NOT_FOUND_ON_REMOTE
+
         patron = get_request_patron(default=None)
 
         if patron:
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
Looks good to me. 🔢

PR #3415, opened 2026-06-02
Hold a workflow lock across OPDS2+ODL import pages (PP-4472)

jonathangreen:
## Description

The `opds_odl.import_collection` Celery task now holds a single Redis *workflow* lock across an entire import run instead of acquiring a fresh per-page lock. The lock value is a UUID generated on the first page and forwarded through each `task.replace()` so one lock spans every page and is held across Celery retries. When another run already holds the lock, the task logs a warning and returns (first page) or continues with a warning (lock expired mid-chain) rather than crashing.

This adopts the same `import_workflow_lock` pattern already used by the OverDrive and Bibliotheca importers.

## Motivation and Context

We were getting recurring unhandled-exception alerts: `Task opds_odl.import_collection[...] raised unexpected: LockNotAcquired('Lock ...::ImportCollection::Collection::104 could not be acquired')`.

The task acquired a per-page `import_lock` with the default `raise_when_not_acquired=True`, so any collision with a concurrent run raised `LockNotAcquired`. Because that exception is not in `autoretry_for`, it surfaced as an unhandled Celery exception instead of being handled gracefully. The per-page lock was also released and re-acquired between `task.replace()` pages, leaving a race window for the every-3-hours `import_all` beat job to collide with an in-progress multi-page import.

Holding one workflow lock across all pages and retries closes that race, and `raise_when_not_acquired=False` turns contention into a graceful skip-with-warning instead of a crash.

Note: `boundless` and `opds` (OPDS2 non-ODL) still use the old per-page `import_lock` and have the same latent issue. Those are intentionally left out of this PR and can be migrated as a follow-up, after which `import_lock` can be deleted entirely.

## How Has This Been Tested?

- Added three tests to `TestImportCollection` mirroring the Bibliotheca suite: lock already held (skip + warning, no import), lock expired mid-chain (continue + warning), and lock held across an autoretry.
- `tox -e py312-docker -- --no-cov tests/manager/celery/tasks/test_opds_odl.py` — 32 passed.
- `mypy` — clean.
- pre-commit hooks pass.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/opds_odl.py
+++ b/src/palace/manager/celery/tasks/opds_odl.py
@@ -1,6 +1,8 @@
 import datetime
+from uuid import uuid4
 
 from celery import shared_task
+from celery.exceptions import Ignore
 from sqlalchemy import delete, select
 from sqlalchemy.exc import NoResultFound
 from sqlalchemy.orm import Session
@@ -8,7 +10,7 @@ from sqlalchemy.orm import Session
 from palace.util.datetime_helpers import utc_now
 
 from palace.manager.celery import importer
-from palace.manager.celery.importer import import_lock
+from palace.manager.celery.importer import import_workflow_lock
 from palace.manager.celery.task import Task
 from palace.manager.celery.tasks import apply
 from palace.manager.celery.utils import load_from_id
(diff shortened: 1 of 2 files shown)


jonathangreen:
The codecov gap was pre-existing, and the claude code review minor comment does point to something somewhat real, but I'd like to handle fixing that in a follow up PR. Since its a real issue in all the importers using `import_workflow_lock`.

dbernstein (approved):
Badabang:  Nice to close the gap and align with the behavior similar celery tasks.

PR #3416, opened 2026-06-02
Remove deprecated PALACE_SEARCH_TIMEOUT setting

jonathangreen:
## Description

Removes the backwards-compatibility alias for the search write timeout that #3412 added for the deprecation period:

- Drops the `PALACE_SEARCH_TIMEOUT` env var alias from `write_timeout` (only `PALACE_SEARCH_WRITE_TIMEOUT` remains).
- Removes the deprecation warning and its `model_validator`, along with the now-unused imports.
- Removes the corresponding deprecation tests and the README note.

## Motivation and Context

#3412 renamed `PALACE_SEARCH_TIMEOUT` to `PALACE_SEARCH_WRITE_TIMEOUT` and kept the old name working as a deprecated alias so deployments could migrate without a flag day. This PR completes that deprecation by removing the alias once the migration is done.

## How Has This Been Tested?

- `mypy` clean.
- `tests/manager/service/search/test_configuration.py` updated to drop the deprecated-env cases; the remaining defaults and `PALACE_SEARCH_WRITE_TIMEOUT` env tests pass.

## Checklist


tdilauro (approved):
🔥

PR #3417, opened 2026-06-02
Return 404 instead of 500 for invalid identifiers on the work permalink route

jonathangreen:
## Description

Catch the `ValueError` raised by `Identifier.for_foreign_id` when an identifier is invalid for its type, in `CirculationManagerController.load_licensepools`, and return the `NO_LICENSES` (404) problem detail instead of letting it escape as a 500.

## Motivation and Context

We've been seeing 500s in production like `"axv8gvz9/open_book" is not a valid Bibliotheca ID.` from `GET /<library>/works/Bibliotheca ID/axv8gvz9/open_book`.

The request that triggers this is a misrouted analytics event: `open_book` is a `CirculationEvent` whose valid endpoint is `/<library>/analytics/<identifier_type>/<path:identifier>/<event_type>` (`track_analytics_event`). The Palace iOS app (v476) is instead posting it to the work permalink path (`/works/...`), where the greedy `<path:identifier>` converter swallows `axv8gvz9/open_book` as the identifier. `Identifier.for_foreign_id` then rejects the embedded slash (slashes are forbidden in Bibliotheca IDs) by raising a bare `ValueError`. The permalink route's `@returns_problem_detail` only catches `ProblemError`, so the `ValueError` propagates to the error handler and becomes a 500.

This PR is a server-hardening fix: an identifier that's invalid for its type can't exist in any collection, so we now report it as a missing item (404) rather than crashing. It does **not** address the root cause — the app posting open-book events to the wrong path — which is a separate client-side issue and means those analytics events are still being lost.

Note: `annotation.py` and `playtime_entries.py` also call `Identifier.for_foreign_id` on raw user-supplied identifiers and share this failure mode; they are left out of scope for this targeted bugfix.

## How Has This Been Tested?

- Added a case to `test_load_licensepools` asserting that `Bibliotheca ID` + `"axv8gvz9/open_book"` returns the `NO_LICENSES` problem detail rather than raising.
- Ran `tox -e py312-docker -- --no-cov tests/manager/api/controller/test_base.py -k test_load_licensepools` (passes).
- `mypy` clean on the changed file.

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/circulation_manager.py
+++ b/src/palace/manager/api/controller/circulation_manager.py
@@ -191,9 +191,23 @@ class CirculationManagerController(BaseCirculationManagerController):
             to look up an Identifier.
         """
         _db = Session.object_session(library)
-        identifier_obj, ignore = Identifier.for_foreign_id(
-            _db, identifier_type, identifier, autocreate=False
-        )
+        try:
+            identifier_obj, ignore = Identifier.for_foreign_id(
+                _db, identifier_type, identifier, autocreate=False
+            )
+        except ValueError:
+            # The identifier isn't valid for its type (e.g. a Bibliotheca ID
+            # with an embedded slash, which happens when a malformed request
+            # path is captured by the greedy <path:identifier> route). Such an
+            # identifier can't exist in any collection, so report it as a
+            # missing item rather than letting the ValueError escape as a 500.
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
Badabing 🚀

PR #3418, opened 2026-06-02
Re-run imports on retry in workflow-lock importers, extract workflow_lock_guard (PP-4472)

jonathangreen:
## Description

Two changes to the paginated workflow tasks (OPDS2+ODL, Bibliotheca, Overdrive import + reap):

1. **Re-run the import on retry.** These tasks hold a Redis workflow lock across a run and keep it held when a retryable exception (`BadResponseException`, `RequestTimedOut`) is raised, via the lock's `ignored_exceptions`. But the lock's `lock_value` was minted inside the task body, so it was absent from the original request that `autoretry_for` re-sends. Each retry therefore generated a *fresh* `lock_value`, failed to acquire the lock still held by the failed attempt, hit the first-page guard, logged "another run is already in progress", and **returned SUCCESS** — silently swallowing the failure with no real retry until the lock's TTL expired. Fixed by keying the lock on `task.request.id` instead of a minted value: Celery preserves the task id across both `task.replace()` page/batch hand-offs and `autoretry_for` retries, so every retry and every page re-acquires the *same* lock for free (`RedisLock.acquire()` extends a lock already held by the same value), while a concurrent run — which has a different task id — is locked out. This removes the minted `lock_value`, the `task.request.kwargs` injection, the explicit threading through every `task.replace()`, and the `lock_value` parameter on all four tasks.

2. **Extract `workflow_lock_guard`.** The four tasks each repeated the same locking preamble (resolve the Redis client, build the lock, acquire with the shared `ignored_exceptions` policy, branch on the logging). That `ignored_exceptions` tuple was duplicated in four places and could drift. Pulled it all into a single `workflow_lock_guard` context manager that yields `proceed`. The held-exception set is now derived from the task's own `autoretry_for`, so the lock is held for exactly the exceptions the task retries on.

Because the lock is now keyed on the stable task id, the first-vs-continuation distinction is gone: a run always re-acquires its own lock, so the only `proceed=False` case is a genuinely concurrent run already holding the lock. This carries one intentional behavior change — a continuation whose lock expired mid-run *and was taken over by another run* now skips cleanly rather than proceeding best-effort alongside it, which is strictly safer since it avoids two concurrent runs on the same collection.

## Motivation and Context

Found while investigating the `LockNotAcquired` alerts addressed by #3415. In the previous workflow lock pattern an error on the first page generated retries, but each retry minted a different `lock_value`, so recovery was delayed to the next scheduled beat run instead of the intended fast autoretry. Keying the lock on the task id — which Celery already keeps stable across `replace()` and retries — restores real retry behavior across all four importers/reaper, and removes the duplicated locking boilerplate so the policy can't drift between tasks.

## How Has This Been Tested?

- Verified that retries now re-run the import (`import_feed` / `import_collection` call count goes from 1 → `1 + max_retries`) instead of skipping.
- Confirmed against the Celery 5.6.3 source that `task.request.id` is preserved across both `task.replace()` (`sig.freeze(self.request.id)`) and `autoretry_for` retries (`signature_from_request` re-sends `task_id=self.id`).
- Updated the autoretry tests for all four tasks to assert re-run-on-retry; rewrote the `workflow_lock_guard` unit tests around task-id keying (free lock, concurrent-run skip, continuation re-acquiring its own lock, custom `lock_factory`); added a worker-level test that pins `task_id` via `apply_async` to exercise a continuation end-to-end.
- `tox -e py312-docker -- --no-cov` across the four affected suites — 94 passed.
- `mypy` clean.

## Checklist

### DIFF
--- a/src/palace/manager/celery/importer.py
+++ b/src/palace/manager/celery/importer.py
@@ -1,9 +1,13 @@
+from collections.abc import Callable, Generator
+from contextlib import contextmanager
 from datetime import timedelta
 
 from celery.canvas import Signature
+from celery.exceptions import Ignore
 
 from palace.util.log import LoggerType, pluralize
 
+from palace.manager.celery.task import Task
 from palace.manager.service.redis.models.lock import RedisLock
 from palace.manager.service.redis.redis import Redis
 from palace.manager.sqlalchemy.model.collection import Collection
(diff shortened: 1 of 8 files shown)


dbernstein (approved):
Looks good.

PR #3419, opened 2026-06-02
Treat unencodable SIP2 patron credentials as invalid (PP-4472)

jonathangreen:
## Description

`SIP2AuthenticationProvider.patron_information()` only caught `OSError`, but when a patron's credentials contain characters that can't be represented in the server's configured encoding (CP850 by default), `SIPClient.send()` raises a `UnicodeEncodeError`. That is a `ValueError` subclass, not an `OSError`, so it escaped the provider's error handling and surfaced as an unhandled 500. This change catches `UnicodeEncodeError` in `patron_information()` and returns `INVALID_CREDENTIALS` instead.

## Motivation and Context

Seen in production: a `POST /<library>/patrons/me/token/` request crashed with `UnicodeEncodeError: 'charmap' codec can't encode characters ... encoding with 'cp850' codec failed` while building the SIP2 patron-information message. The offending bytes were the patron-supplied barcode/PIN, which contained characters CP850 can't represent. Credentials that can't even be encoded in the server's charset can't be valid for that server, so this is a failed authentication (401), not a server error (500). 

Part of PP-4472.

## How Has This Been Tested?

Added `test_unencodable_credentials_become_problemdetail`, which authenticates with a CP850-incompatible password and asserts a `ProblemDetail` matching `INVALID_CREDENTIALS` is returned rather than an exception propagating. The test exercises the real `send()`/encode path via `MockSIPClient`. Ran the new test alongside the existing SIP2 error-handling tests (all pass) and confirmed mypy is clean on the changed file.

## Checklist

### DIFF
--- a/src/palace/manager/integration/patron_auth/sip2/provider.py
+++ b/src/palace/manager/integration/patron_auth/sip2/provider.py
@@ -403,7 +403,8 @@ class SIP2AuthenticationProvider(
             # authentication rather than letting it become a 500.
             cls.logger().warning(
                 f"Credentials could not be encoded using the configured "
-                f"SIP2 encoding ({client.encoding}): {e}"
+                f"SIP2 encoding ({client.encoding}): {e}",
+                exc_info=e,
             )
             return INVALID_CREDENTIALS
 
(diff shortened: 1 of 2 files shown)


jonathangreen:
Looks like the codecov issue is pre-existing, so I think this one is ready for review.

dbernstein (approved):
Well done.

PR #3420, opened 2026-06-02
Add signature_with helper to re-queue Celery tasks without re-listing args

jonathangreen:
## Description

Adds `signature_with(task, **overrides)` to `celery/utils.py` — a helper that builds a signature for the currently-running Celery task by reusing its live invocation arguments (`request.args` / `request.kwargs`) and applying only the given overrides. Positional arguments are normalized to keyword arguments using the task's own signature, so an override applies to the correct parameter regardless of how it was originally passed, and the resulting signature never mixes a positional and keyword value for the same parameter.

The self-replacing tasks across `overdrive`, `bibliotheca`, `boundless`, `opds_odl`, `search`, `marc`, `notifications`, `reaper`, and `collection_delete` are migrated to call it, so each site now passes only the values that actually change. Values that are reassigned in the task body (e.g. `start_time`/`lock_value`/`modified_since` resolved from `None`, `last_work_id` advancing in a loop) are kept as explicit overrides, since they genuinely differ from the received arguments.

`marc_export_cleanup` is intentionally left unchanged: its no-arg `.s()` deliberately resets `batch_size` to the default, which the helper would instead preserve, and it has no pass-through arguments to forget.

## Motivation and Context

Self-replacing tasks paginate/batch by calling `task.replace(task.s(...))`, where `task.s(...)` had to re-list **every** parameter the task accepts so the next run received the full set. This is fragile: when a parameter is added to a task, every self-replace site must be updated to thread it through, and forgetting one silently drops the value (it reverts to its default on the next run). This helper removes the need to re-list pass-through parameters at all.

## How Has This Been Tested?

- New `TestSignatureWith` unit tests in `tests/manager/celery/test_utils.py` cover kwargs / positional / mixed originals, no-override, overriding a parameter that used its default, and keyword-only parameters.
- `mypy` passes.
- Full Celery task suite (`tox -e py312-docker -- tests/manager/celery/test_utils.py tests/manager/celery/tasks`) passes (325 tests), exercising the re-queue paths of every migrated task (reaper batching, multi-page import integration, overdrive reaper batching, etc.).

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/bibliotheca.py
+++ b/src/palace/manager/celery/tasks/bibliotheca.py
@@ -20,7 +20,7 @@ from palace.util.datetime_helpers import utc_now
 
 from palace.manager.celery.importer import import_workflow_lock
 from palace.manager.celery.task import Task
-from palace.manager.celery.utils import load_from_id
+from palace.manager.celery.utils import load_from_id, signature_with
 from palace.manager.integration.license.bibliotheca import BibliothecaAPI
 from palace.manager.integration.license.bibliotheca_importer import (
     EVENT_IMPORT_OVERLAP,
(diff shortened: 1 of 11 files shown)


dbernstein (approved):
Very useful.

PR #3423, opened 2026-06-02
Convert equivalent_identifiers_refresh to a Celery task (PP-1466)

dbernstein:
## Description

Replaces the deprecated `BaseCoverageProvider`-based cron script (`bin/equivalent_identifiers_refresh`) with a native Celery task that uses a Redis set as the incremental work queue.

**New components:**
- `DirtyIdentifierIds` — a persistent Redis set (stable well-known key, no expiry) that tracks which identifier IDs need `RecursiveEquivalencyCache` recomputation. Supports `add`, `pop` (atomic SPOP), `count`, and `add_all_from_db` (full-refresh seeding via `yield_per` + pipeline chunking).
- `refresh_equivalents` module — `process_identifier_ids` (chain expansion + cache refresh), `add_identity_equivalents` (self-reference gap fill), and `refresh_equivalent_identifiers` (synchronous test helper replacing `EquivalentIdentifiersCoverageProvider(db.session).run()`).
- `equivalent_identifiers_refresh` Celery task — acquires a `TaskLock` (keyed on task name, owned by `root_id`) to prevent concurrent runs, pops dirty IDs in batches, calls `process_identifier_ids`, then self-replaces via `task.replace()` until the queue is empty. On failure, re-queues the popped IDs before re-raising so no work is silently lost. `full_refresh=True` seeds the queue from the DB before processing and sets `ensure_identity=True`, which is carried through self-replacements so the missing-self-reference sweep (`add_identity_equivalents`) runs exactly once when the queue drains.

**Updated SQLAlchemy listeners:**
- On `Equivalency` create: push `input_id` and `output_id` to the Redis dirty set.
- On `Equivalency` delete: expand to chain members via `RecursiveEquivalencyCache`, push all affected parent IDs.
- Both listeners catch any exception from the Redis interaction — unavailability is non-fatal and the weekly full refresh recovers any missed IDs.

**Beat schedule** (added to `celery.py`):
- Delta run twice an hour at :05 and :35 past the hour — a no-op (lock acquire + one `SCARD`) when the queue is empty, matching the previous cron cadence.
- Weekly full refresh every Sunday at 3:30 AM (`full_refresh=True`), which re-seeds the queue from the DB before draining, recovering from any Redis drift or missed listener events.

**Removed:**
- `EquivalentIdentifiersCoverageProvider` (`core/equivalents_coverage.py`)
- `EquivalencyCoverageQueries` (`core/query/coverage.py`)
- `EquivalencyCoverageRecord` model class (`sqlalchemy/model/coverage.py`)
- `bin/equivalent_identifiers_refresh`
- Stale cron entry from `docker/services/cron/cron.d/circulation`
- Stale mypy lax-override entry for the deleted module

**DB migration:** drops the `equivalentscoveragerecords` table.

## Motivation and Context

JIRA: PP-1466

The `BaseCoverageProvider` pattern is deprecated and this was one of the last remaining uses of it. Moving to a Celery task with Redis as the work queue removes the DB write overhead for change tracking, provides natural deduplication, and aligns with the established pattern for background work in the codebase.

## How Has This Been Tested?

- New unit tests for `DirtyIdentifierIds`, `refresh_equivalents` utilities, and the Celery task (full run, batching, `full_refresh` flag, failure re-queue, lock contention, Redis failure non-fatal on both create and delete listener paths).
- Shared `RecursiveEquivalencyCache` test fixtures extracted to `tests/fixtures/equivalents.py` and registered in `conftest.py`.
- 4 existing test call sites updated from `EquivalentIdentifiersCoverageProvider(db.session).run()` to `refresh_equivalent_identifiers(db.session)`.
- `mypy` passes with no new issues.
- Full test suite for affected modules passes.
- Manuaully tested locally.
## Checklist


dbernstein:
> ### Details
> #### **Minor**: `src/palace/manager/service/redis/models/dirty_identifiers.py:63-77`
> `add_all_from_db` uses `union` (which forces a DB-side `DISTINCT` — a sort/hash-aggregate over the entire equivalents table before the `yield_per` cursor can stream) even though the destination is a Redis set, where `SADD` already deduplicates. On the weekly full-refresh path against a large equivalents table this is redundant work. Switching to `union_all` lets rows stream immediately and leaves dedup to Redis; to keep the logged count accurate (distinct newly-added IDs rather than rows sent), derive `total` from the `SADD` return value instead of `len(ids)`:
> 

add_all_from_db is a one-shot bulk pre-load that fully completes before the first pop. It's a barrier, not an incremental pipeline — nothing downstream consumes the stream as it arrives, so there's no pipelined overlap to preserve. That removes the main reason streaming would matter, which was the strongest part of the UNION ALL case.

Given that, keeping UNION is the better trade:

Accurate count for free — total += len(ids) is correct because the rows are already distinct. UNION ALL would force the sadd-return-sum change just to keep the log honest.
Less to move — fewer rows DB→app and fewer SADD members, with no streaming benefit forgone.

The only residual cost is a one-time server-side hash/sort during seeding, with a possible work_mem spill on a genuinely huge table — but it's a weekly, off-peak (Sun 3:30 AM) job, so probably not going to be a problem.  If it ever showed up in DB metrics we could bump work_mem for that transaction or revisit UNION ALL then, but there's no reason to pay for that complexity now.

So: I think we should leave add_all_from_db on union. No change.

alembic/versions/20260602_7ae5c2b3cbf3_drop_equivalentscoveragerecords_table.py
```diff
@@ -0,0 +1,85 @@
...
+COVERAGE_STATUS_ENUM_VALUES = (
+    "success",
+    "transient failure",
+    "persistent failure",
+    "registered",
+)
+
+
+def upgrade() -> None:
+    op.drop_table("equivalentscoveragerecords")
```
jonathangreen: Since we run migrations online, we can't immediately drop this table. It must be dropped in a follow up PR that goes in after the release.

src/palace/manager/sqlalchemy/listeners.py
```diff
@@ -179,16 +183,87 @@ def last_update_time_change(target, value, oldvalue, initator):
...
     )
+    # Marking dirty IDs is a best-effort side effect: any Redis problem (the
+    # service being down, or even unconfigured) must never break the equivalency
+    # change itself. The weekly full refresh recovers any IDs missed here.
+    try:
+        DirtyIdentifierIds(container_instance().redis().client()).add(
+            input_id, output_id, *parent_ids
+        )
+    except Exception as e:
+        log.warning(f"Failed to mark dirty identifiers on equivalency delete: {e}")
```
jonathangreen: Minor: Lets add a stack trace here, so we can see a stack trace:

```suggestion
        log.warning(f"Failed to mark dirty identifiers on equivalency delete: {e}", exc_info=True)
```

Nit: I never like swallowing all exceptions like this. It catches so much, but maybe it makes sense inside a listener like this.

src/palace/manager/sqlalchemy/listeners.py
```diff
@@ -179,16 +183,87 @@ def last_update_time_change(target, value, oldvalue, initator):
...
+    """
+    input_id, output_id = _equivalency_identifier_ids(target)
+    # Best-effort side effect — see the delete listener above; a Redis problem
+    # must never break the equivalency creation.
+    try:
+        DirtyIdentifierIds(container_instance().redis().client()).add(
+            input_id, output_id
+        )
+    except Exception as e:
+        log.warning(f"Failed to mark dirty identifiers on equivalency create: {e}")
```
jonathangreen: Minor: Same comment about stack trace as above ☝🏻

tests/fixtures/equivalents.py
```diff
@@ -0,0 +1,40 @@
...
+        return set(rows)
+
+    def drop(self) -> None:
+        """Delete every cache row (to isolate what a refresh recomputes)."""
+        self.session.query(RecursiveEquivalencyCache).delete()
+        self.session.commit()
+
+
+@pytest.fixture
+def recursive_equivalency_cache(
```
jonathangreen, line 37: Minor: I generally only add a new fixture if its something I see being used widely by a lot of test files. Since this seems pretty specific to RecursiveEquivalency and its only used there (from what I see anyway). Maybe it should just live inline in the test file.

jonathangreen (approved):
I approved this one, but there is a test failing that needs to be resolved, and the migration comment needs to be resolved before this gets merged.

alembic/versions/20260602_7ae5c2b3cbf3_drop_equivalentscoveragerecords_table.py
jonathangreen: I added a PR https://github.com/ThePalaceProject/circulation/pull/3459, so we have better documentation of this requirement.

tests/fixtures/equivalents.py
dbernstein: Good call to question it — but this fixture is actually shared by two test files: `tests/manager/sqlalchemy/test_refresh_equivalents.py` (for the `refresh_equivalents` utility functions) and `tests/manager/celery/tasks/test_equivalents.py` (for the Celery task). It was extracted into a shared fixture precisely because an earlier review flagged `_cache_for`/`_drop_cache` as duplicated verbatim across those two files.

Inlining it would either re-duplicate the helper in both files, or create a test-to-test import. I considered merging the two test files so it could live inline in one, but they test different source modules in different packages (`sqlalchemy/refresh_equivalents.py` vs `celery/tasks/equivalents.py`), and merging would break the repo convention that test files mirror the `src/palace/manager` structure.

So I've left it as a shared fixture in `tests/fixtures/` — happy to revisit if you'd prefer one of the alternatives.

alembic/versions/20260602_7ae5c2b3cbf3_drop_equivalentscoveragerecords_table.py
dbernstein: Done — removed the drop-table migration so the table stays this release. One wrinkle worth noting: the PR had also removed the `EquivalencyCoverageRecord` model, and the migration test harness seeds the schema from `Base.metadata` (create_all + stamp head), so with the model gone the existing 2024 migration `8dde64eab209` (which alters `equivalentscoveragerecords` on up/down) made `test_up_down_consistency` fail with `relation "equivalentscoveragerecords" does not exist`. So I've restored `EquivalencyCoverageRecord` as a dormant, table-only model (no behavior — the refresh no longer reads or writes it) to keep the table in the schema. A follow-up PR after the release will remove the model and add the drop migration. Thanks for the #3459 doc reference.

src/palace/manager/sqlalchemy/listeners.py
dbernstein: Done — added `exc_info=True` to both the create and delete listener warnings so a Redis failure logs a full stack trace. (Agreed on the broad `except` being unlovely, but inside a before_flush listener it's deliberate: marking dirty IDs is a best-effort side effect that must never break the equivalency write itself, and the weekly full refresh recovers anything missed.)

PR #3430, opened 2026-06-03
Make date-boundary tests deterministic with freezegun

jonathangreen:
## Description

Make the announcement and dashboard tests deterministic so they no longer fail when CI happens to run across the midnight UTC boundary.

`AnnouncementFixture` captured `today`/`yesterday`/`tomorrow`/`a_week_ago`/`in_a_week` as class attributes evaluated at *import* time, while the model (`announcements.py`) and validator (`announcement_list_validator.py`) call `datetime.date.today()` at *runtime*. When the import landed on one date and the test executed on the next, expectations and code-under-test disagreed. These dates are now instance attributes, and the `announcement_fixture` freezes the clock for the duration of each test so both sides read the same date.

`test_bulk_circulation_events` creates a `CirculationEvent` at `now() - 1 minute` and then queries for events on the "current day"; within a minute of midnight the event lands on the previous day. It's now frozen at midday.

## Motivation and Context

The boto3 dependabot PR (#3428) failed CI purely because the run started at `2026-06-03T00:06:14Z` — six minutes past midnight UTC — and these three time-of-day-sensitive tests broke at the date boundary. The change under test was innocent; the tests were flaky. This removes the recurring midnight failure.

## How Has This Been Tested?

Ran the changed tests plus every other consumer of `announcement_fixture` in the docker tox env:

```
tox -e py312-docker -- --no-cov \
  tests/manager/sqlalchemy/model/test_announcements.py \
  tests/manager/api/admin/test_announcement_list_validator.py \
  tests/manager/api/admin/controller/test_dashboard.py \
  tests/manager/api/test_authenticator.py \
  tests/manager/api/admin/controller/test_library.py \
  tests/manager/api/admin/controller/test_announcement_service.py
```

173 passed. `mypy` is clean on the changed files.

## Checklist

### DIFF
--- a/tests/fixtures/announcements.py
+++ b/tests/fixtures/announcements.py
@@ -1,6 +1,8 @@
 import datetime
+from collections.abc import Generator
 
 import pytest
+from freezegun import freeze_time
 from sqlalchemy.orm import Session
 
 from palace.manager.sqlalchemy.model.announcements import Announcement, AnnouncementData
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🥶🕛

PR #3432, opened 2026-06-03
Replace per-page import_lock with the workflow lock

jonathangreen:
## Description

Replaces the per-page `import_lock` with the whole-chain `workflow_lock_guard` at the two remaining call sites, so every paginated importer now shares one locking policy:

- `boundless.import_collection` — wrapped in `workflow_lock_guard`; the `task.replace()` re-queue moved inside the guard so the lock is held across pages and across its `autoretry_for` retries. Dropped the now-unused `redis` local.
- `opds.opds_import_task` (the shared helper behind opds1/opds2/opds_for_distributors) — same migration. The skip path returns `None`, which callers already tolerate (the import-failed path already returned `None`).
- `importer.import_lock` is now unused and removed. `import_key` stays — it still builds the `IdentifierSet` key.

This is a follow-up to #3418, which introduced `workflow_lock_guard` and migrated the Bibliotheca/Overdrive/OPDS2+ODL importers.

## Motivation and Context

`import_lock` was a per-page lock, released between `task.replace()` hand-offs. That left a window for a second run to interleave between pages, and a collision raised `LockNotAcquired` rather than backing off.

`workflow_lock_guard` is keyed on `task.request.id`, which Celery preserves across both `task.replace()` page hand-offs and `autoretry_for` retries. The lock is therefore held across the entire chain (and across retries), and a run that finds the lock already held is skipped with a warning instead of raising — matching the behavior of the other paginated importers.

## How Has This Been Tested?

- Rewrote `test_boundless.py::test_import_lock` → `test_skips_when_lock_held`: asserts the graceful skip + warning instead of expecting `LockNotAcquired`.
- Added `test_opds2.py::test_skips_when_lock_held`: the shared OPDS path had no lock coverage before; it now verifies the skip makes no feed request and logs the warning.
- Ran the affected suites (boundless, opds1/opds2/opds_for_distributors, importer = 70 tests) — all pass. `mypy` clean.

## Checklist

### DIFF
--- a/src/palace/manager/celery/importer.py
+++ b/src/palace/manager/celery/importer.py
@@ -73,16 +73,6 @@ def import_workflow_lock(
     )
 
 
-def import_lock(client: Redis, collection_id: int) -> RedisLock:
-    """
-    Create a lock for the given collection.
-
-    This makes sure only one task is importing data for the collection
-    at a time.
-    """
-    return RedisLock(client, import_key(collection_id), lock_timeout=timedelta(hours=1))
-
-
 def reap_workflow_key(collection_id: int) -> list[str]:
     """
     Generate a Redis key for the reap workflow-level lock for the given collection.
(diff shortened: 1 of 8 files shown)


tdilauro (approved):
🚀

PR #3436, opened 2026-06-04
Give a clear error when uploading an oversized library logo

jonathangreen:
## Description

`scale_and_store_logo` only caught `UnidentifiedImageError` when opening an uploaded library logo. Pillow's decompression-bomb guard raises a separate `DecompressionBombError` *inside* `Image.open` when an image's pixel count exceeds its limit — before our scaling code ever runs. That exception was uncaught and surfaced as a generic 500. This catches it and returns an `INVALID_CONFIGURATION_OPTION` problem detail telling the admin the image has too many pixels and what the limit is.

## Motivation and Context

A OPs team member tried repeatedly to upload a library logo that was huge by pixel dimensions (~326 megapixels) — not necessarily large in file size. Each attempt failed with an opaque 500 and no explanation, so they kept retrying. Rejecting the image is the correct behavior, but the admin had no way to understand why. This gives them an actionable message.

Note this is a pixel-count limit, not a file-size limit, so a small file can still trip it.

See these slack messages: 
https://lyrasis.slack.com/archives/CCY3PH8JD/p1780577648186309?thread_ts=1780489931.830289&cid=CCY3PH8JD
https://lyrasis.slack.com/archives/C049AA32E3S/p1780577271279769

## How Has This Been Tested?

Added a case to `test_libraries_post_errors` that monkeypatches `Image.MAX_IMAGE_PIXELS` down so a small fixture image trips the same guard (avoiding allocating a real multi-hundred-megapixel image), and asserts the new problem-detail message. Ran the full test via the docker tox env; it passes.

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/library_settings.py
+++ b/src/palace/manager/api/admin/controller/library_settings.py
@@ -4,7 +4,7 @@ import base64
 import json
 import uuid
 from io import BytesIO
-from typing import Any
+from typing import Any, cast
 
 import flask
 from flask import Response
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🚀

PR #3437, opened 2026-06-04
Tune OpenSearch read-path defaults: 10s timeout, no retries (PP-4472)

jonathangreen:
## Description

Adjusts the OpenSearch read-path defaults introduced in #3412. The read/write client split, the dedicated read client, and the tunable env vars are all kept; only the defaults change:

- `read_timeout`: `4` → `10`
- `read_max_retries`: `2` → `0`
- `read_retry_on_timeout`: `true` → `false`

Operators can still restore the previous behavior via `PALACE_SEARCH_READ_TIMEOUT` / `PALACE_SEARCH_READ_MAX_RETRIES` / `PALACE_SEARCH_READ_RETRY_ON_TIMEOUT`.

## Motivation and Context

#3412 gave the read path a 4s timeout with 2 timeout retries on the theory that the nightly user-facing `500`s were connection stalls against a node bouncing during the OpenSearch maintenance window, so a fast retry would fail over to the healthy node.

A follow-up investigation after deploying contradicted that premise:

- The failures are not node bounces. Node count held steady at 2 across every night, the ES domain log group had zero recovery/master/transport events, and JVM/GC were flat in-window. Completed searches stayed ~15ms, so the stall is node-busy/IO contention, not query execution or GC.
- The real driver is the daily indexing operation, which produces a 10–20× write surge peaking ~7k docs/s at 01:00–02:00 UTC. 
- The 4s timeout cut into the legitimate slow tail of crawlable reads. The slowest *successful* crawlable read all week was ~7.4s, with zero successful reads between 7.4s and 20s. A read that genuinely needs 6s fails all three 4s attempts, so #3412 converted previously-successful reads into `500`s (crawlable 500s rose from a ~3–19/night baseline to 47 the night after deploy, with a `~12,300ms` = 3×4s latency fingerprint).

`10s` sits above 100% of observed successful reads (~2.6s margin) so it converts none of the legitimate slow tail into failures, while still giving up well before the 20s write timeout, bounding how long a genuinely-stuck read holds a uwsgi worker. Retries are disabled, since they didn't help, and we don't want to hold a worker for 3x10s (client will just timeout).

The remaining root cause is being worked on separately; this PR stops the regression I introduced.

JIRA: PP-4472

## How Has This Been Tested?

- `mypy` clean.
- `tests/manager/service/search/test_configuration.py::test_defaults` updated to assert the new defaults (`read_timeout=10`, `read_max_retries=0`, `read_retry_on_timeout=False`); test fixture defaults updated to match.
- Full `tests/manager/service/search/` and `tests/manager/search/test_service.py` suites pass.

## Checklist

### DIFF
--- a/tests/fixtures/search.py
+++ b/tests/fixtures/search.py
@@ -28,9 +28,9 @@ from tests.mocks.search import SearchServiceFake
 class SearchTestConfiguration(FixtureTestUrlConfiguration):
     url: HttpUrl
     write_timeout: int = 20
-    read_timeout: int = 4
-    read_max_retries: int = 2
-    read_retry_on_timeout: bool = True
+    read_timeout: int = 10
+    read_max_retries: int = 0
+    read_retry_on_timeout: bool = False
     maxsize: int = 25
     model_config = SettingsConfigDict(env_prefix="PALACE_TEST_SEARCH_")
 
--- a/README.md
+++ b/README.md
@@ -205,9 +205,10 @@ To let the application know which OpenSearch instance to use, you can set the fo
   backwards compatibility.
 - `PALACE_SEARCH_READ_TIMEOUT`: The timeout in seconds to use for the user-facing read/search path. The default is `10`
   (optional).
-- `PALACE_SEARCH_READ_RETRY_ON_TIMEOUT`: Whether a timed-out read should be retried. The default is `false` (optional).
-- `PALACE_SEARCH_READ_MAX_RETRIES`: The maximum number of retries for a timed-out read before the error is propagated.
-  The default is `0` (optional).
+- `PALACE_SEARCH_READ_RETRY_ON_TIMEOUT`: Whether a timed-out read should be retried. Must be `true` for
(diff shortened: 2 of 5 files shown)


tests/fixtures/search.py
```diff
@@ -28,9 +28,9 @@
 class SearchTestConfiguration(FixtureTestUrlConfiguration):
     url: HttpUrl
     write_timeout: int = 20
-    read_timeout: int = 4
-    read_max_retries: int = 2
-    read_retry_on_timeout: bool = True
+    read_timeout: int = 10
+    read_max_retries: int = 0
+    read_retry_on_timeout: bool = False
```
tdilauro, line 33: Minor - Should we pull these from mainline constants to ensure that they don't get out of sync?

tdilauro (approved):
Looks good! ⏲️ 

One small suggestion below for test fixture, but could similarly apply to test values and the defaults on the settings.

tests/fixtures/search.py
jonathangreen: Probably, but I'm going to leave this for later. I'd like to get this rolled out today.

PR #3439, opened 2026-06-04
Throttle search_indexing so a backlog can't flood OpenSearch (PP-4472)

jonathangreen:
## Description

`search_indexing` drains the `WaitingForIndexing` queue, but it released its lock every iteration and requeued itself with **no delay**, while the beat schedule also fires it **every minute**. When the queue is deep, this fans out `index_works` tasks, so a large burst causes a concentrated write spike.

This change paces the indexing:

- **Hold the lock across self-replacements** (`release_on_exit=False`) so the every-minute beat can't start a second, concurrent task. A beat run that fires mid-drain fails to acquire the lock and exits via `LockNotAcquired`, now declared in the task's `throws` so Celery logs it as expected rather than an error.
- **Add a short cooldown between batches** (`SEARCH_INDEXING_COOLDOWN`, 1–3s jittered) on the requeue, so a backlog drains as a single paced stream bounded to ~`batch_size / cooldown` works/sec.

## Motivation and Context

While investigating nightly `ConnectionTimeout` errors on `/<library>/crawlable` feeds (after #3412), the root cause was an OpenSearch write surge whose segment-merge tail degraded read latency. Log/metric analysis showed:

- Indexing churn is ~330–1,500 works/min steady, but a backlog drained in a concentrated burst (IndexingRate spiking ~10–20× baseline) because nothing paced the `index_works` fan-out.
- Per-batch indexing is cheap (p50 0.07s, p95 0.45s for a 500-work batch), so the bottleneck was concurrency, not OpenSearch throughput.

With `batch_size=500` and a 1–3s cooldown the drain is ~6–13k works/min — comfortably above the steady dirty rate (indexing stays in sync; bursts smaller than one batch never hit the cooldown path), but far below the unbounded fan-out that caused the flood. The result: **no single dirty-burst can saturate the index again, regardless of size.**

This is a structural safeguard and complements #3438, which removes today's specific trigger (the nightly Lexile reclassification churn). Together they eliminate the surge *and* prevent its recurrence.

JIRA: PP-4472

## How Has This Been Tested?

- New `test_search_indexing_cooldown_between_batches`: with a backlog larger than `batch_size`, each full batch takes a cooldown from the configured range, the backlog still fully drains, and the lock is released at the end.
- Existing `test_search_indexing_lock` (a concurrent run raises `LockNotAcquired`) and `test_search_indexing` (drains correctly across self-replacements, lock released) still pass — the latter confirms the lock re-acquires across replacements now that it's held.
- `tox -e py312-docker -- --no-cov tests/manager/celery/tasks/test_search.py` → 19 passed.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/search.py
+++ b/src/palace/manager/celery/tasks/search.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import random
 from collections.abc import Sequence
+from datetime import timedelta
 from typing import Any
 
 from celery import chain, shared_task
@@ -21,13 +22,6 @@ from palace.manager.service.redis.models.lock import LockNotAcquired, TaskLock
 from palace.manager.service.redis.models.search import WaitingForIndexing
 from palace.manager.sqlalchemy.model.work import Work
 from palace.manager.util.backoff import exponential_backoff
-
-# Cooldown (min, max seconds) between search_indexing batches when draining a
-# backlog. A batch is batch_size works, so the drain rate is bounded to roughly
-# batch_size / cooldown works/sec -- fast enough to stay in sync with the normal
-# dirty rate but slow enough that a large burst can't flood OpenSearch. Kept
-# shorter than the reindex cooldown because indexing must stay roughly current.
-SEARCH_INDEXING_COOLDOWN = (1.0, 3.0)
 
 
 def get_work_search_documents(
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good! 🛠️

PR #3441, opened 2026-06-05
Treat empty Overdrive collection as empty page, not an error

jonathangreen:
## Description

When an Overdrive collection or page contains no titles, the API omits the `products` key entirely and reports `"totalItems": 0`. `OverdriveAPI.fetch_book_info_list` treated any missing `products` key as a malformed response and raised `BadResponseException`. Since the Overdrive import Celery task uses `autoretry`, this turned a normal empty collection into a retry storm that eventually surfaced as a production exception.

This change distinguishes the empty-collection case (`products` missing **and** `totalItems == 0`) and returns an empty page, while still raising `BadResponseException` for genuinely malformed responses so real errors continue to retry.

## Motivation and Context

Production was intermittently raising:

```
IntegrationException: Overdrive response missing 'products' key. Response data: {'limit': 100, 'offset': 0, 'totalItems': 0, 'id': 'v1L1BdnEAAA2v', 'links': {...}}
```

The response is simply an empty collection (`totalItems: 0`), not an error condition, so it should not raise or trigger retries.

## How Has This Been Tested?

- Updated `test_fetch_book_info_list_missing_products_key` to clarify it covers the non-empty malformed case (still raises).
- Added `test_fetch_book_info_list_empty_collection`, which feeds a `totalItems: 0` response and asserts an empty list and `None` next endpoint are returned without raising.
- Ran `tox -e py312-docker -- --no-cov tests/manager/integration/license/overdrive/test_api.py -k fetch_book_info_list` — all 5 tests pass.

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/overdrive/api.py
+++ b/src/palace/manager/integration/license/overdrive/api.py
@@ -712,6 +712,13 @@ class OverdriveAPI(
             async_task_list = list()
             response_products = data.get("products")
             if response_products is None:
+                # Overdrive omits the 'products' key entirely when a collection
+                # (or page) contains no titles. In that case 'totalItems' is 0
+                # and there is simply nothing to import, so we treat it as an
+                # empty page rather than an error.
+                if data.get("totalItems") == 0:
+                    return [], next_endpoint
+
                 self.log.warning(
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
👍 Good catch.

PR #3442, opened 2026-06-05
Treat empty Bibliotheca event response as no events

jonathangreen:
## Description

`EventParser.process_all` now treats a completely empty (or whitespace-only) Bibliotheca response body the same as a response that contained no events, instead of letting the XML parser raise.

## Motivation and Context

Bibliotheca occasionally returns a completely empty response body when fetching circulation events. An empty body has no root element, so lxml raises `Error('Document is empty, line 1, column 1')` — even in recovery mode (`recover=True`), since recovery handles malformed XML but cannot invent a root for an empty document. This surfaced in production as an unhandled exception in the `bibliotheca.import_collection` Celery task:

```
File ".../integration/license/bibliotheca.py", line 1160, in process_all
File ".../util/xmlparser.py", line 90, in _load_xml
lxml.etree.Error: Document is empty, line 1, column 1
```

A blank body is now funneled into the existing `no_events_error` handling: the import path (`no_events_error=False`) returns no events gracefully, while the self-test path (`no_events_error=True`) raises a properly-classified `RemoteInitiatedServerError`. This is distinct from the already-handled "no events" case, where the body is a well-formed `<CloudLibraryEvents/>` document — the crash was specifically a zero-byte body.

## How Has This Been Tested?

Added a parametrized test `test_parse_empty_response_body` covering `b""`, whitespace-only bytes, and `""`, asserting both the graceful-empty result and the `RemoteInitiatedServerError` raised when `no_events_error=True`. Ran the `TestEventParser` suite (`tox -e py312-docker`) — all pass — and confirmed mypy is clean.

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/bibliotheca.py
+++ b/src/palace/manager/integration/license/bibliotheca.py
@@ -1157,9 +1157,15 @@ class EventParser(
         self, string: bytes | str, no_events_error: bool = False
     ) -> Generator[tuple[str, str, str | None, datetime, datetime | None, str]]:
         has_events = False
-        for i in super().process_all(string):
-            yield i
-            has_events = True
+        # Bibliotheca occasionally returns an empty response body. An empty
+        # document cannot be parsed as XML (lxml raises "Document is empty"
+        # even in recovery mode), so treat a blank body the same as a
+        # response that contained no events rather than letting the parse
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
🚀

PR #3452, opened 2026-06-10
Support alternate OIDC auth flow with "prompt" param (PP-4429)

tdilauro:
## Description

Adds support for the OIDC `prompt` query parameter on the `/oidc/authenticate` endpoint. When present, the value is validated against the OIDC-defined set of tokens (`none`, `login`, `consent`, `select_account`) and forwarded as-is to the upstream IdP's authorization URL. Multi-value prompts (e.g. `select_account consent`) are supported per the OIDC spec, which defines the parameter as a space-separated list.

## Motivation and Context

The web patron client (CPW) shows a "Use a different account" button when an OIDC authentication error occurs or the user cancels sign-in. Clicking it retries authentication with `prompt=select_account` appended to the CM's `/oidc/authenticate` URL, which forces the IdP to show its account chooser even when a session already exists. Previously the CM silently ignored any `prompt` parameter; this change wires it through end-to-end.

[Jira PP-4429]

## How Has This Been Tested?

- New updated test cases for new functionality.
- Manual testing in local environment to verify valid values passing through.
- CI tests and checks pass.

## Checklist

### DIFF
--- a/src/palace/manager/integration/patron_auth/oidc/auth.py
+++ b/src/palace/manager/integration/patron_auth/oidc/auth.py
@@ -279,6 +279,7 @@ class OIDCAuthenticationManager(LoggerMixin):
         state: str,
         nonce: str,
         code_challenge: str | None = None,
+        prompt: str | None = None,
     ) -> str:
         """Build authorization URL with parameters.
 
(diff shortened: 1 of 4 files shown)


tdilauro:
Closing because I accidentally opened this one from a fork, which broke CI.

Reopened as #3453.

PR #3459, opened 2026-06-16
Document online migration backwards-compatibility convention

jonathangreen:
## Description

Adds an "Online migrations (backwards compatibility)" subsection to `CLAUDE.md` under Database Migrations. It documents the N-1 compatibility convention: migrations run online while the previous release's webservers are still serving traffic, so a release's database must work with version N-1 of the code. The practical rules captured are that a schema object can only be dropped one release after the code stops using it, that new required columns must be nullable or use a server default so N-1 code keeps working, and the two-stacked-PR workflow (merge the first, leave the second in draft until a release ships) used to carry out the two-release split.

## Motivation and Context

This is an important, non-obvious operational constraint that wasn't written down anywhere. Without it documented, it's easy to write a single migration that drops a still-used object or adds a `NOT NULL` column without a default, which breaks the still-running previous-version webservers during an online deploy. Recording the convention in `CLAUDE.md` keeps it from being forgotten.

## How Has This Been Tested?

Documentation-only change; no code is affected. Verified the Markdown renders correctly and pre-commit hooks (including PyMarkdown) pass.

## Checklist


tdilauro (approved):
🤖

PR #3460, opened 2026-06-16
Add external-schema test configuration and db marker for backwards-compatibility testing

jonathangreen:
## Description

First of two PRs that add a CI check verifying our online-migration requirement: a release's database must keep working with version N-1 of the code. This PR adds only the test-harness building blocks; the CI job that uses them follows in a stacked PR.

- **`external_schema` mode** in the test database fixtures (`tests/fixtures/database.py`). When `PALACE_TEST_DATABASE_EXTERNAL_SCHEMA` is set, `DatabaseCreationFixture` uses the configured database as-is (no create/drop, serial only) and `DatabaseFixture` leaves the schema and seed data untouched, so a run can execute against a schema that was applied externally rather than one built from the current models.
- **A general-purpose `db` marker**, applied automatically (via `pytest_collection_modifyitems`) to any test that uses the `db` fixture, so database-backed tests can be included or excluded with `-m db` / `-m "not db"`. This is the supported pytest pattern for deriving a marker from fixture usage (marks on fixtures themselves have no effect).

By default the same tests run as before; the marker only matters when an `-m` expression is given, and `external_schema` mode is off unless the env var is set.

Related to: https://github.com/ThePalaceProject/circulation/pull/3459

## Motivation and Context

The follow-up PR adds a CI job that builds the new schema with the current image and then runs the previous release's database tests against it. For that to work, this PR must already be present in the previous release's image, so they need to ship one release ahead of the check itself. This PR ships them.

## How Has This Been Tested?

- New unit tests for the config, both fixture modes, and the marker hook (`tests/fixtures/test_database.py`) pass via `tox -e py312-docker`.
- Verified `-m db` selects DB tests and `-m "not db"` excludes them (the auto-applied marker is present before `-m` is evaluated).
- Confirmed a normal run is unaffected (existing `db`-using tests still pass).
- `mypy` is clean on the changed files.

## Checklist


jonathangreen:
I think that writing tests for tests is a bit overkill (who tests the tests tests tests), so I'm not going to add tests for the database fixture change. It will be exercised and used in CI when we add the additional workflow.

tdilauro:
Enjoyed this little casting of shade between our new AI helpers/overlords:
> This resolves the open greptile thread on the same lines. Note that greptile's suggested fix (self.create_database = config.create_database and not config.external_schema) should not be applied: it assumes create_database is silently forced to False, which is exactly the implicit override you deliberately replaced with an explicit "the two settings must agree" guard — only the message wording is worth refining.

tdilauro (approved):
Looks great! And a great idea. 🥇

PR #3461, opened 2026-06-16
Document env var README requirement in CLAUDE.md

jonathangreen:
## Description

Adds an **Environment Variables** subsection to the "Instructions for Claude" section of `CLAUDE.md`. It directs that `README.md` be updated in the same change whenever an application environment variable is added, renamed, removed, or has its default changed, and points to where each kind of variable is documented (the `### Environment variables` section for app/runtime vars, the testing section for `PALACE_TEST_*` vars) and the expected entry format.

## Motivation and Context

New environment variables are frequently added to the code without a corresponding update to `README.md`, leaving the documented configuration surface stale. This codifies "update the README in the same change" as an explicit, hard-to-miss instruction so the docs stay in sync with the variables the application actually reads.

## How Has This Been Tested?

Documentation-only change to `CLAUDE.md`; there is no application code to test. Verified the referenced `README.md` sections and entry format match the file as it exists today.

## Checklist


tdilauro (approved):
🤖 👍🏽

PR #3466, opened 2026-06-17
Harden Redis connection resilience across all three connections (PP-4545)

jonathangreen:
## Description

Hardens connection resilience for all three of our Redis connections — the application Redis client, the Celery broker, and the Celery result backend — so a dropped connection is detected and replaced instead of leaving the pool stuck handing out dead sockets.

The mechanism is `health_check_interval`: redis-py PINGs an idle connection before reuse, and if the PING fails (e.g. the connection was left stale by a Redis reboot) it disconnects and re-establishes the connection transparently within ~30s instead of blocking on a dead socket. The Celery result backend additionally gets `result_backend_always_retry`, since result writes are keyed by task id and therefore idempotent.

## Motivation and Context

A transient network blip between the workers and ElastiCache caused a burst of "Connection reset by peer" errors across all prod shards, killing Celery tasks and spiking web 5xx. None of our three Redis connections were configured to detect or recover from a dropped connection, so they stayed stuck after the blip.

Companion to #3467 (PP-4545), which makes web requests survive a Redis outage while it is being detected. The two are independent and can merge in either order.

## How Has This Been Tested?

- `mypy` passes on the changed files.
- Added unit tests: one asserting the new Celery resilience settings route correctly through `CeleryConfiguration.model_dump()` (the `broker_transport_options_*` keys fold into the options dict while the `redis_backend_*` / `result_backend_*` keys stay top-level), and one for `RedisConfiguration`/`RedisContainer` confirming `health_check_interval` defaults to 30 and is threaded into the connection pool.
- Ran the Redis and Celery configuration suites against a real Redis in the `py312-docker` tox env; all pass and the client constructs/connects with the new settings.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,11 @@ Celery and caching, we recommend that you use a separate database for each purpo
         ```
 - `PALACE_REDIS_KEY_PREFIX`: The prefix to use for keys stored in the Redis instance. The default is `palace`.
     This is useful if you want to use the same Redis database for multiple Circulation Manager instances (optional).
+- `PALACE_REDIS_SOCKET_KEEPALIVE`: Whether to enable TCP keepalive on Redis connections, so the OS can
+    reclaim a connection whose peer has gone away. The default is `true` (optional).
+- `PALACE_REDIS_HEALTH_CHECK_INTERVAL`: How often, in seconds, an idle Redis connection is health-checked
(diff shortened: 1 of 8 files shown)


dbernstein (approved):
Looks good.

PR #3467, opened 2026-06-17
Handle transient Redis/broker outages gracefully in web requests (PP-4545)

jonathangreen:
## Description

When Redis briefly goes away (e.g. an ElastiCache reboot), web requests that touch Redis — either directly or by publishing a Celery task through the Redis broker — previously failed with a raw 500. This handles the two cases at the right altitude:

- **Primary actions** (admin collection import/delete/reap, report generation, and any direct Redis use): the transient errors are mapped centrally in `ErrorHandler`, alongside the existing database `OperationalError` → 503 branch, so the client gets "Service temporarily unavailable. Please try again later." and can retry. No per-call-site wrapping.
- **Best-effort follow-ups** (the patron-activity syncs queued after a loans listing, borrow, or revoke): the error is swallowed and logged in the loan controller. The borrow/revoke has already succeeded, so failing to queue the follow-up sync must not become an error for the patron — they get fresh data on their next refresh.

The transient error set (kombu `OperationalError` for the broker, redis `ConnectionError` / `TimeoutError` for the client) is defined once as `TRANSIENT_REDIS_ERRORS` and shared by both paths.

## Motivation and Context

During the same incident that prompted #3466, requests that touched Redis returned 500s while the connection was down. Hardening the connections (#3466) shrinks that window; this PR makes the requests themselves degrade gracefully — returning a clean 503 where the action could not be performed, and quietly continuing where the Redis work was only a best-effort follow-up to an action that already succeeded.

Companion to #3466 (PP-4545). The two are independent and can merge in either order.

## How Has This Been Tested?

- `mypy` passes on all changed files.
- Added tests: `ErrorHandler` returns a 503 for each transient error type; the loan `revoke` path swallows each transient error type and still returns 200; the loan `sync` path swallows a Redis outage on the direct read and still serves the loans feed.
- Ran the full `test_app_server.py` and `test_loan.py` suites in the `py312-docker` tox env (121 passed), confirming the `borrow`/`revoke` refactor onto the shared helper produces identical task-queue calls (no regressions).

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/loan.py
+++ b/src/palace/manager/api/controller/loan.py
@@ -32,6 +32,7 @@ from palace.manager.feed.acquisition import OPDSAcquisitionFeed
 from palace.manager.integration.license.boundless.constants import (
     BAKER_TAYLOR_KDRM_PARAMS,
 )
+from palace.manager.service.redis.exception import TRANSIENT_REDIS_ERRORS
 from palace.manager.service.redis.models.patron_activity import PatronActivity
 from palace.manager.sqlalchemy.model.library import Library
 from palace.manager.sqlalchemy.model.licensing import (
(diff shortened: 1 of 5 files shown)


dbernstein (approved):
👍

PR #3472, opened 2026-06-17
Remove orphaned static image route and login button image

jonathangreen:
## Description

Removes the dead `GET /images/<filename>` route and the orphaned basic-auth login-button-image mechanism that was its only intended caller.

- `routes.py` — delete the `/images/<filename>` route and its `static_image` handler.
- `basic.py` — delete the `login_button_image` property (hardcoded to `None`), the now-unreachable `if self.login_button_image:` link branch in `_authentication_flow_document`, and the now-unused `url_for` / `Link` imports.
- `test_authenticator.py` — drop the mock provider's `login_button_image` override and the logo-link assertions; the flow document now has no `links`.

Requests to `/images/...` now return a normal `404` instead of a `500`.

## Motivation and Context

The `/images/<filename>` route called `app.manager.static_files.image()`, but the `static_files` controller and its `image()` method were removed in #1642 ("Remove first book auth API"), along with the only image it ever served (`FirstBookLoginButton280.png`) and the `resources/images` directory. The route handler was left behind, so every request to `/images/...` raised `'CirculationManager' object has no attribute 'static_files'` and returned a 500.

This fires in production whenever a client requests anything under `/images/` — for example browsers probing for a favicon at `/images/favicon-196x196.png` ([example from #palace-alerts](https://lyrasis.slack.com/archives/C049AA32E3S/p1781709443829049)).

The route's only intended caller was the basic-auth `login_button_image` mechanism, which is itself dead: the base property is hardcoded to `None`, and the one provider that overrode it (First Book) was removed in #1642. So the whole chain — route, extension point, and `url_for("static_image", ...)` link branch — is orphaned and is removed here.

## How Has This Been Tested?

- `mypy` on the changed source files: clean.
- `tox -e py312-docker -- --no-cov tests/manager/api/test_authenticator.py tests/manager/api/test_routes.py`: 183 passed.

## Checklist

### DIFF
--- a/src/palace/manager/api/authentication/basic.py
+++ b/src/palace/manager/api/authentication/basic.py
@@ -9,7 +9,6 @@ from functools import partial
 from re import Pattern
 from typing import Annotated, Any, cast
 
-from flask import url_for
 from pydantic import PositiveInt, field_validator, model_validator
 from pydantic_core.core_schema import ValidationInfo
 from sqlalchemy.orm import Session
@@ -23,7 +22,6 @@ from palace.opds.authentication.palace import (
     AuthenticationInput,
     AuthenticationInputs,
 )
-from palace.opds.rwpm import Link
 from palace.util.log import elapsed_time_logging
 
 from palace.manager.api.admin.problem_details import (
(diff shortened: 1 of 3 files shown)


dbernstein (approved):
Tally ho!

PR #3473, opened 2026-06-17
Fix duplicate 'status' kwarg when borrow returns an OPDS message

jonathangreen:
## Description

`BaseOPDSFeed.entry_as_response` passes `status=entry.status_code` explicitly when the entry is an `OPDSMessage`. The borrow flow (`LoanController.borrow`) also passes a status (`201` for a new loan, `200` otherwise) through `response_kwargs`. When `single_entry` returns an `OPDSMessage` instead of a `WorkEntry`, both ended up in the same `OPDSEntryResponse(...)` call, raising `TypeError: OPDSEntryResponse() got multiple values for keyword argument 'status'`.

The fix drops the caller-supplied `status` from `response_kwargs` for `OPDSMessage` responses, so the message's own status code wins. This matches the pre-existing intent: the `OPDSMessage` branch has always used `entry.status_code`. An error message status (e.g. `403`) is the correct response when the work cannot be presented, rather than the `200`/`201` from the borrow.

## Motivation and Context

Observed intermittently in production on the borrow endpoint:

```
GET /<library>/works/<identifier_type>/<identifier>/borrow
TypeError: palace.manager.util.flask_util.OPDSEntryResponse() got multiple values for keyword argument 'status'
```

It surfaced as an unhandled web-app exception (HTTP 500). It only triggers when a borrow/hold succeeds but `single_entry` cannot build a work entry and instead returns an `OPDSMessage` (no active edition, or an `UnfulfillableWork`). Without the fix the patron sees a 500 instead of the intended error message.

## How Has This Been Tested?

- Added a regression test in `TestOPDSFeedProtocol::test_entry_as_response` that calls `entry_as_response` with an `OPDSMessage` and a colliding `status=201`, asserting the response uses the message's status code (`403`) and no longer raises.
- `tox -e py312-docker -- --no-cov tests/manager/feed/test_opds_acquisition_feed.py` — 37 passed.
- `mypy` clean on the changed file.

## Checklist

### DIFF
--- a/src/palace/manager/api/controller/loan.py
+++ b/src/palace/manager/api/controller/loan.py
@@ -143,12 +143,11 @@ class LoanController(CirculationManagerController):
         # If we have a loan, serve a feed that tells the patron how to fulfill the loan. If a hold,
         # serve a feed that talks about the hold. We also need to drill in the Accept header, so that
         # it eventually gets sent to core.feed.opds.BaseOPDSFeed.entry_as_response
-        response_kwargs = {
-            "status": 201 if is_new else 200,
-            "mime_types": flask.request.accept_mimetypes,
-        }
         return OPDSAcquisitionFeed.single_entry_loans_feed(
-            self.circulation, loan_or_hold, **response_kwargs  # type: ignore[arg-type]
+            self.circulation,
+            loan_or_hold,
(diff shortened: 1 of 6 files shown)


dbernstein (approved):
LGTM 👍

PR #3474, opened 2026-06-18
Fix RecursionError in LicensePool.calculate_work

jonathangreen:
## Description

Fixes an infinite recursion (`RecursionError: maximum recursion depth exceeded`) in `LicensePool.calculate_work`. The "kick-out" loop that removes mis-grouped license pools from a `Work` cleared `pool.work` before recursively calling `pool.calculate_work()`, but left `pool.presentation_edition.work` set. Because `calculate_work` re-resolves a pool's work from its presentation edition (the `elif presentation_edition.work:` branch), the recursive call re-adopted the pool back into the same `Work` and kicked the original pool out again, ping-ponging between the two pools until the stack was exhausted. The fix also clears `pool.presentation_edition.work` before recursing, mirroring how `Work.make_exclusive_open_access_for_permanent_work_id` already kicks pools out.

## Motivation and Context

This was observed in production as a Celery unhandled exception in `apply.bibliographic_apply` -> `data_layer/bibliographic.apply` -> `circulation.apply` -> `LicensePool.calculate_work`, with the traceback showing `licensing.py:1280` (`pool.calculate_work(...)`) repeated until `RecursionError`. The trigger is a `Work` that erroneously contains more than one commercial `LicensePool` (with different identifiers) whose presentation editions both still point back at that `Work` — which is the normal, internally-consistent state for a `Work`'s license pools, so the bug surfaces on real data. The existing `test_calculate_work_fixes_work_in_invalid_state` did not catch it because it left the kicked-out pools' `presentation_edition.work` as `None`.

A comment dating to 2018 claimed the recursion could not happen "because we're setting pool.work to None before calling pool.calculate_work()" — that guard was incomplete, and the comment has been corrected.

## How Has This Been Tested?

- Added `test_calculate_work_does_not_recurse_when_editions_point_to_work`, which reproduces the production `RecursionError` at `licensing.py:1280` on the unfixed code and passes with the fix.
- Ran the full `TestWorkConsolidation` suite (28 tests) — all pass, confirming the existing kick-out behavior (each kicked-out pool gets its own `Work`) is preserved.
- `mypy` clean on the changed module.

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/model/licensing.py
+++ b/src/palace/manager/sqlalchemy/model/licensing.py
@@ -1268,15 +1268,23 @@ class LicensePool(Base):
             # itself. All other LicensePools need to be kicked out and
             # associated with some other work.
             #
-            # This won't cause an infinite recursion because we're
-            # setting pool.work to None before calling
-            # pool.calculate_work(), and the recursive call only
-            # happens if self.work is set.
+            # We clear both pool.work *and* pool.presentation_edition.work
+            # before the recursive call. Clearing only pool.work is not
+            # enough: calculate_work() re-resolves a pool's work from its
+            # presentation edition (the ``elif presentation_edition.work``
+            # branch below), so if the edition still pointed at this Work the
+            # recursive call would re-adopt it and kick *self* back out,
+            # ping-ponging until the stack was exhausted (a RecursionError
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
🥇

PR #3475, opened 2026-06-18
Retry RemoteInitiatedServerError in Bibliotheca purchase record import

jonathangreen:
## Description

`import_purchase_records_by_collection` was paging on every occurrence of `RemoteInitiatedServerError`, raised by `BibliothecaAPI.marc_request` (via `ErrorParser`) when Bibliotheca returns a non-200 body or a malformed / `"Unknown error"` / `"Authentication failed"` document.

`RemoteInitiatedServerError` lives in `api.circulation.exceptions` (`InternalServerError` → `IntegrationException`) and is a **sibling** of `RemoteIntegrationException`, not a subclass. It therefore matched neither the task's `autoretry_for=(BadResponseException, RequestTimedOut)` nor its `throws=(RemoteIntegrationException,)`, so it was never retried and surfaced as an unhandled Celery task exception.

This change adds `RemoteInitiatedServerError` to both `autoretry_for` and `throws` so it is handled like the other transient remote errors: retried up to `max_retries`, then logged as an expected remote failure rather than an unhandled exception.

## Motivation and Context

Observed in production as a recurring unhandled-exception alert:

```
Task bibliotheca.import_purchase_records_by_collection[...] raised unexpected: RemoteInitiatedServerError('')
  File ".../celery/tasks/bibliotheca.py", line 294, in import_purchase_records_by_collection
    result = importer.import_day(current_day, cutoff, offset)
  File ".../bibliotheca_purchase_record_importer.py", line 135, in import_day
    records = list(
  File ".../bibliotheca.py", line 288, in marc_request
    raise ErrorParser().process_first(response.content)
palace.manager.api.circulation.exceptions.RemoteInitiatedServerError
```

`ErrorParser` itself documents these as transient errors on the Bibliotheca side that "happen relatively frequently," so retrying (and then treating an exhausted retry as expected) matches the existing handling of `BadResponseException` / `RequestTimedOut`. The 4xx/5xx case was already covered — the HTTP client raises `BadResponseException` for non-2xx/3xx responses. Only the `marc_request` `ErrorParser` path slipped through.

Scope is limited to `import_purchase_records_by_collection`; the sibling Bibliotheca tasks (`import_collection`, `circulation_update_collection`) don't route their responses through `ErrorParser`, so they don't surface this exception.

## How Has This Been Tested?

- Added `test_remote_initiated_server_error_retried_and_expected`, which asserts `RemoteInitiatedServerError` is present in both `autoretry_for` and `throws`, and behaviorally verifies the task retries 5× (1 initial + `max_retries=4`) instead of failing immediately.
- Ran the full `TestImportPurchaseRecordsByCollection` class (16 tests) — all pass.
- `mypy` clean on both changed files.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/bibliotheca.py
+++ b/src/palace/manager/celery/tasks/bibliotheca.py
@@ -28,6 +28,7 @@ from celery import shared_task
 
 from palace.util.datetime_helpers import utc_now
 
+from palace.manager.api.circulation.exceptions import RemoteInitiatedServerError
 from palace.manager.celery.importer import workflow_lock_guard
 from palace.manager.celery.task import Task
 from palace.manager.celery.utils import ModelNotFoundError, load_from_id, signature_with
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
👍 👍 💯 🥇

PR #3481, opened 2026-06-18
Make lanes.size safe to drop: defer reads + server_default for writes (PP-4506)

dbernstein:
## Description

Follow-up to #3424 (merged), which removed all lane-size *maintenance* but left `size` / `size_by_entrypoint` eagerly mapped on the `Lane` model. This is release-1 of the online-migration sequence to remove those columns: **stop using them**. The column drop is a stacked follow-up (release-2), gated on this shipping first.

"Stop using" an ORM column has two sides, and both must be handled before the column can be dropped safely:

- **Reads.** SQLAlchemy eagerly `SELECT`s every mapped column, so simply "keeping them mapped" (as #3424 did) does not stop the current code from reading them. Both columns are marked `deferred()`, which excludes them from the default `SELECT`.
- **Writes.** SQLAlchemy also emits any column with a Python-side `default=` in every `INSERT`. `size` is `NOT NULL` with `default=0`, so it was still written on every `Lane` insert — and lanes are created at runtime (`create_default_lanes`: reset lanes / configure a library / create a custom lane). A follow-up migration that dropped the column would then break the still-running previous release's inserts (`column "size" does not exist`). This PR moves the default into the database (`ALTER COLUMN size SET DEFAULT 0`) and removes the Python-side `default=0`. With a `server_default` and no Python default, SQLAlchemy omits the column from `INSERT`s and Postgres fills in `0`, so the eventual drop is write-safe. `size_by_entrypoint` is nullable with no default, so it was already omitted from inserts and needs no write-side change.

The `SET DEFAULT` migration is catalog-only in Postgres (instant, no table rewrite) and backwards-compatible: the previous release still writes `size=0` explicitly, harmlessly overriding the new default, and still reads `size` (the column is untouched). The columns stay in the model and the database, so `test_model_definitions_match_ddl` still passes and the previous release keeps working.

Once a release containing this change has shipped, the follow-up PR can drop both columns without breaking the then-previous release.

## Motivation and Context

JIRA: PP-4506

Completes the "stop using" step of the online-migration sequence for removing the lane-size columns (release-1). The column drop is release-2, gated on this shipping first, and is tracked by the `cleanup migration` label.

## How Has This Been Tested?

- `tests/migration/test_alembic_builtin_tests.py` passes, including `test_model_definitions_match_ddl` (model matches the migrated schema, with `compare_server_default=True`) and the up/down migration round-trip.
- `mypy` clean on the Lane model.
- Verified against a `NOT NULL` column with the same config that, with a `server_default` and no Python default, SQLAlchemy omits the column from `INSERT`s entirely (and the insert still succeeds after the column is dropped), whereas the previous `deferred`-only / Python-`default=0` version emitted it on every insert.

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,23 +152,17 @@ code stops using it. Removing a schema object is therefore a two-step process ac
 **"Stop using" an ORM column means stopping both reads _and_ writes.** SQLAlchemy eagerly includes every
 mapped column in its default `SELECT`, and eagerly emits any column that has a Python-side `default=` in every
 `INSERT`. So leaving a column mapped — even when no application code reads or assigns it — does **not** stop
-using it, and a later drop will break the still-running previous release. `test_model_definitions_match_ddl`
-passing does not catch this; it only checks that the model matches the schema, not that inserts avoid the
-column. To fully stop using a column in release 1, before dropping it in release 2:
+using it, and a later drop will break the still-running previous release. To fully stop using a column in
+release 1, before dropping it in release 2:
 
 - **Reads:** mark the column `deferred()` so it is excluded from the default `SELECT`.
 - **Writes:** remove any Python-side `default=`. If the column is `NOT NULL`, removing the Python default
   alone makes inserts fail the not-null constraint (SQLAlchemy sends no value and the database has none), so
-  first move the default into the database with a `server_default` via an `ALTER COLUMN ... SET DEFAULT`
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
This looks good. 

I pushed a couple commits to clean up some AI code review comments here. And I'm going to merge this one once all the tests pass, since @dbernstein is out on vacation and it would be nice to get this into this release.

PR #3489, opened 2026-06-22
Test OpenSearch 2.19 and 3.5 against the opensearch marker (PP-4546)

jonathangreen:
## Description

> This PR is **stacked on #3490** (which makes the `opensearch` marker actually apply)

We are migrating production OpenSearch **1.x → 2.19 → 3.5** over next few sprints, after that is complete we'll test only on 3.5. Because the OpenSearch *version* is the only thing that varies, we just run the **`opensearch`-marked tests** against the new versions rather than the whole suite three times.

**Removal later:** once production is on 3.5, delete the `opensearch-version` job, the `os219`/`os35` factor lines, the two `[docker:os-circ-*]` tox sections, and `Dockerfile.os2.ci`, and point `Dockerfile.os.ci` at 3.5. The two test fixes stay (correct for 3.5).

## Motivation and Context

Upgrading our production opensearch clusters.

## How Has This Been Tested?

Locally, faithful to how CI runs them: `tox -e py312-docker-os35` and `tox -e py312-docker-os219` each run the 62 `opensearch`-marked tests against 3.5 / 2.19 and pass. Confirmed via `tox config` that the default `py312-docker` env still uses the 1.x `os-circ` container and runs the full `pytest tests` command (verified `tox -e py312-docker -- tests/manager/search/test_service.py` runs all 19 tests, not just the marked one). `mypy` clean on the changed files; pre-commit (incl. workflow validation) green.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,3 +60,54 @@ jobs:
           files: ./coverage.xml
           name: test-${{ matrix.python-version }}
           verbose: true
+
+  # Temporary bandaid for the OpenSearch 1.x -> 2.19 -> 3.5 migration. The `test`
+  # job above still runs the full suite against OpenSearch 1.x (current prod);
+  # this job additionally runs just the OpenSearch-marked tests against the
+  # intermediate (2.19) and target (3.5) versions. These are separate, non-required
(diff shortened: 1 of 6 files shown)


tdilauro (approved):
LGTM! 🎸🤘🏽

PR #3492, opened 2026-06-23
Fix crash when a ProblemDetail reaches cache_control_headers

jonathangreen:
## Description

The `cache_control_headers` decorator is applied *inside* `returns_problem_detail` on the `index`, `authentication_document`, and `groups` routes. Because decorators wrap bottom-up, `cache_control_headers` runs first at request time. When a wrapped controller returns a raw `ProblemDetail` instead of a Flask `Response`, `cache_control_headers` tried to set `response.headers[...]` on the `ProblemDetail` — which has no `headers` attribute — raising `AttributeError` before `returns_problem_detail` could convert it to a real response.

This fix returns the `ProblemDetail` unchanged so the outer `returns_problem_detail` decorator converts it to a `Response`. Error responses also should not be given `Cache-Control` headers.

## Motivation and Context

Observed in production as an unhandled exception:

```
Exception in web app: 'ProblemDetail' object has no attribute 'headers'
GET /<library>/groups/favicon.ico
AttributeError: 'ProblemDetail' object has no attribute 'headers'
  File ".../core/app_server.py", line 175, in decorated
    "Cache-Control" not in response.headers and default_max_age is not None
```

A request like `GET /<library>/groups/favicon.ico` treats `favicon.ico` as a lane identifier. `load_lane` finds no matching lane and returns a `ProblemDetail`, which then triggered the crash. The same latent bug affected the `index` and `authentication_document` routes, which share the decorator stack.

## How Has This Been Tested?

- Added `test_cache_control_headers_problem_detail`, which asserts the decorator passes a `ProblemDetail` through untouched even when a `default_max_age` and a request `Cache-Control` header are present.
- `tox -e py312-docker -- --no-cov tests/manager/core/test_app_server.py -k CacheControl` — all pass.
- `mypy` clean on the changed files.

## Checklist

### DIFF
--- a/src/palace/manager/core/app_server.py
+++ b/src/palace/manager/core/app_server.py
@@ -140,7 +140,9 @@ def _parse_cache_control(cache_control_header: str | None) -> dict[str, int | No
 
 def cache_control_headers[**P](
     default_max_age: int | None = None,
-) -> Callable[[Callable[P, Response]], Callable[P, Response]]:
+) -> Callable[
+    [Callable[P, Response | ProblemDetail]], Callable[P, Response | ProblemDetail]
+]:
     """
     Decorator that manages Cache-Control headers on Flask responses based on request and configuration.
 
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
💂🏽

PR #3498, opened 2026-06-23
Switch CI test and docker compose infrastructure to Valkey (PP-4544)

jonathangreen:
## Description

Switches the test and dev infrastructure from `redis/redis-stack-server` to the `valkey/valkey:9.0` image, in both the tox (`[docker:valkey-circ]`) and `docker-compose` stacks. `9.0` is a rolling tag, so the stacks pick up the latest 9.0.x patch automatically (matching the floating `postgres:16` style already used here). The container/service is renamed to reflect the engine it actually runs (`valkey-circ` in tox; the `valkey` compose service, with its `depends_on` and `redis://valkey:6379` host references updated, and a `valkey-cli` healthcheck). The `redis://` URL scheme and all `PALACE_*REDIS*` config keys are deliberately left unchanged, since they name the protocol/contract rather than the instance — Valkey is protocol- and command-compatible and we still use the `redis-py` client.

The README `### Redis` section becomes `### Valkey`, noting that we run and test against Valkey while Redis remains a supported drop-in.

This change also drops the unused ElastiCache detection (`Redis.elasticache`) and its now-orphaned `cached_property` import. The property had no callers, and we run Valkey directly now.

> Note: I'm going to keep this in draft until we've updated our infrastructure to valkey, but its ready for code review any time.

## Motivation and Context

We want to evaluate (and move to) Valkey instead of Redis. Our usage is limited to core Redis commands (strings, sets, expiry, `WATCH`/pipelines, and Lua scripts via `EVALSHA`) plus Celery as a broker/result backend — none of the Redis Stack modules our old image bundled — so Valkey is a true drop-in. We pin to the `9.0` line because that is the version AWS ElastiCache for Valkey currently runs, which is our intended upgrade target; testing against it directly keeps the test stack aligned with production.

JIRA: PP-4544

## How Has This Been Tested?

- Tests run in CI

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,12 +175,15 @@ CREATE USER palace with password 'test';
 grant all privileges on database circ to palace;
 ```
 
-### Redis
+### Valkey
 
-Redis is used as the broker for Celery and the caching layer. You can run Redis with docker using the following command:
+[Valkey](https://valkey.io/) is used as the broker for Celery and the caching layer. We run and test
+against Valkey, but Redis works as well — Valkey is a protocol- and command-compatible fork of Redis, so
+the same `redis://` connection URLs point at either. You can run Valkey with docker using the following
+command:
 
 ```sh
-docker run -d --name redis -p 6379:6379 redis/redis-stack-server
+docker run -d --name valkey -p 6379:6379 valkey/valkey:8.1.8
 ```
 
 ### Environment variables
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
Looks good! 🔑

jonathangreen:
Going to merge this one since we are updating to valkey now.

PR #3502, opened 2026-06-24
Make test coverage opt-in, collect it only on the Python 3.14 CI job

jonathangreen:
## Description

Make test coverage **opt-in** (it was the pytest default) and collect it on **only the Python 3.14 job** to speed up our test runs.

- `pyproject.toml`: drop `--cov`/`--cov-report=xml` from `addopts`, so the default `pytest`/`tox` run measures no coverage and is fast (local dev included).
- `test.yml`: the 3.14 matrix entry opts in with `--cov --cov-report=xml` and is the only job that uploads to Codecov; 3.12/3.13 run the default coverage-free invocation.
- `build.yml`: drop the now-redundant `--no-cov` from the Docker Build unit-test job (coverage is off by default now).
- `CLAUDE.md`: update the test commands to match (coverage is opt-in).

## Motivation and Context

Coverage measurement is the dominant cost in the test jobs — roughly an 80% overhead (the same ~6021-test suite runs in ~488s with `--cov` vs ~268s with `--no-cov` on 3.12). 

The matrix runs in parallel, so PR feedback time is gated by the slowest job — which stays 3.14 (~6.5min, the version that must run coverage), while 3.12/3.13 drop to the coverage-free runtime. Net: ~9m41s baseline → ~6.5min, while **keeping real branch coverage** in Codecov.

## How Has This Been Tested?

- Confirmed against coverage 7.13.5 source that `sys.monitoring` is the default core only on Python 3.14+ and only measures branches there.
- The real evaluation is CI on this PR: compare the `Test (Py 3.12/3.13/3.14)` durations and review the Codecov report (branch coverage should be present, sourced from the 3.14 upload).

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -366,7 +366,7 @@ jobs:
             cat src/palace/manager/_version.py &&
             source env/bin/activate &&
             uv sync --frozen --active &&
-            pytest --no-cov tests
+            pytest tests
           "
 
       - name: Stop container
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.12", "3.13", "3.14"]
-        # Coverage is collected only on the Python 3.14 job. 3.14 is the first version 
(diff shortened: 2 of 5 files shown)


dbernstein (approved):
Looks good.

PR #3503, opened 2026-06-24
Retire the CoverageProvider machinery (PP-4468)

dbernstein:
## Description

Retires the legacy `CoverageProvider` machinery. Overdrive bibliographic coverage now runs as a direct metadata-apply step inside `OverdriveAPI.update_licensepool` instead of routing through `OverdriveBibliographicCoverageProvider.ensure_coverage`. The base `CoverageProvider` classes, the Overdrive/Bibliotheca providers, the coverage-provider runner scripts, and the now-dead readers (`Identifier.missing_coverage_from`, the `Explain` script's coverage block, the always-failing admin `refresh_metadata` provider path) are deleted.

`BibliographicData.apply` no longer writes `CoverageRecord` rows (the `create_coverage_record` parameter is removed; the Celery apply task and the Bibliotheca updater already passed `create_coverage_record=False`).

The `CoverageRecord` model and `coveragerecords` table are intentionally left in place (marked dormant) for one release, per the online-migration / N-1 rule. A stacked follow-up PR drops the `coveragerecords` and dormant `equivalentscoveragerecords` tables.

## Motivation and Context
JIRA (PP-4468)
The `CoverageProvider` batch-processing pattern is being retired in favor of Celery. Investigation showed nothing live reads coverage rows:

- The only live caller of `ensure_coverage` (Overdrive's `update_licensepool`) passes `force=True` and ignores the result.
- The admin `refresh_metadata` endpoint is wired with no provider, so it always returns `METADATA_REFRESH_FAILURE` without invoking a provider.
- `Identifier.missing_coverage_from` / `should_update` / the `Explain` coverage block are only reachable through the already-dead runner scripts.

The Celery `bibliographic_apply` task and the Bibliotheca updater already skip coverage-record writes, so Overdrive was the last writer.

## How Has This Been Tested?

- `mypy` and `ruff` pass on all changed files.
- Targeted suites pass under `tox -e py312-docker` (325 tests): Overdrive API including the rewritten `update_licensepool` bibliographic-coverage path, the Overdrive Celery `reap_collection` task, data-layer `apply`, Bibliotheca, admin work editor, the `Explain` script, and the `CoverageRecord`/`Collection` model tests.

## Checklist

- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3504, opened 2026-06-24
Drop coveragerecords and equivalentscoveragerecords tables (PP-4653)

dbernstein:
## Description

Stacked follow-up to #3503 ("Retire the CoverageProvider machinery"). Now that no running code reads or writes the coverage tables, this PR removes the `CoverageRecord` and (already-dormant) `EquivalencyCoverageRecord` models, the `coverage_records` relationships on `Identifier`/`DataSource`/`Collection`, and drops the `coveragerecords` and `equivalentscoveragerecords` tables along with their shared `coverage_status` enum.

`Timestamp`, its separate `service_type` enum, and the `BaseCoverageRecord` mixin (still imported by the `workcoveragerecords`-removal migration for `status_enum`) are retained.

## Motivation and Context
JIRA (PP-4653)
This is the "release 2" half of the CoverageProvider retirement. Per the online-migration / N-1 rule, the tables could not be dropped in the same release that stopped using them (#3503) — N-1 app servers still wrote to them during a rolling deploy.

> ⚠️ **Draft until the release containing #3503 has shipped.** Merging earlier would break N-1 webservers that still write to these tables.

## How Has This Been Tested?

- `mypy` and `ruff` pass on all changed files.
- The migration round-trips on a real Postgres: `upgrade` drops both tables **and** the `coverage_status` enum (leaving `timestamps` / `service_type` intact); `downgrade` recreates both tables with identical indexes, constraints, and foreign keys; a second `upgrade` drops them again.
- The full `tests/migration/` suite passes under `tox -e py312-docker` (incl. pytest-alembic single-head / up-down-consistency / model-matches-DDL checks and a new drop test), along with the `Timestamp`, collection-deletion, identifier, and data-layer `apply` tests.

## Checklist


dbernstein:
Superseded by #3521, which moves this branch to an upstream branch stacked on #3520 (and additionally removes the now-orphaned `BaseCoverageRecord` mixin).

PR #3505, opened 2026-06-25
Speed up slow tests by removing avoidable real waits

jonathangreen:
## Description

Removes three avoidable real-time waits from the test suite that don't exercise any behaviour under test:

- Mock `random` in `test_search_reindex` so the requeue cooldown countdown (`random.uniform(5, 15)`, honoured by the real Celery worker as an ETA) doesn't add ~32s of real sleep across its parameters. Pagination is still exercised — this mirrors the existing `test_search_indexing` pattern.
- Neutralize the tenacity exponential backoff in `test_retries` (1+2+4s of real `time.sleep`) with `wait_none()`; the test asserts retry behaviour, not timing.
- Drop an inert `time.sleep(1)` from an OIDC `max_age` test — the assertion passes with or without it.

## Motivation and Context

Try to get our tests running a little faster then they were. Nice to keep an eye on slow tests.

## How Has This Been Tested?

`tox -e py312-docker` — the full suite passes (6022 tests), and the three affected tests pass individually. The ablation confirmed these changes remove the real waits without changing any assertions or test outcomes.

## Checklist

### DIFF
--- a/tests/manager/celery/tasks/test_search.py
+++ b/tests/manager/celery/tasks/test_search.py
@@ -64,13 +64,20 @@ def test_get_work_search_documents(db: DatabaseTransactionFixture) -> None:
 
 
 @pytest.mark.parametrize("batch_size", [2, 3, 500])
+@patch("palace.manager.celery.tasks.search.random")
 def test_search_reindex(
+    mock_random: MagicMock,
     batch_size: int,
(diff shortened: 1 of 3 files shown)


tdilauro (approved):
⏲️🔧

PR #3509, opened 2026-06-25
Treat empty Bibliotheca bibliographic_lookup response as a retryable error (PP-4665)

dbernstein:
## Description

`BibliothecaAPI.bibliographic_lookup` now treats an empty (or whitespace-only) HTTP 200 response body as a transient remote error: it raises `RemoteInitiatedServerError` instead of letting lxml's `Document is empty` error propagate. The two Celery tasks whose `bibliographic_lookup` call sites lacked it — `circulation_update_collection` and `import_collection` — now retry `RemoteInitiatedServerError` and declare it expected, matching `import_purchase_records_by_collection` (#3475).

## Motivation and Context

JIRA (PP-4665)
Observed in production as an unhandled exception in `circulation_update_collection`:

```
lxml.etree.Error: Document is empty, line 1, column 1
  .../bibliotheca.py", line 335, in bibliographic_lookup
  .../util/xmlparser.py", line 90, in _load_xml
    return etree.parse(BytesIO(xml), parser)
```

Bibliotheca occasionally returns an HTTP 200 with an empty body — the same quirk already handled for MARC (#3476) and events (#3442). `bibliographic_lookup` was the last unguarded XML-parse path, so an empty body crashed the circulation sweep.

Crucially, an empty body must **not** be collapsed into "no items returned". `BibliothecaCirculationUpdater._process_batch` treats any identifier *absent* from the response as removed and zeroes out its availability. Returning `[]` on an empty body would therefore wipe out the availability of every identifier in the batch (or a single title on the on-demand `update_availability` path) on every transient glitch. Raising a transient error keeps "empty body" distinct from "a valid document that omits an identifier", so the work is retried rather than corrupting data.

`bibliographic_lookup` is shared by three tasks. `import_purchase_records_by_collection` already retried `RemoteInitiatedServerError` (#3475); this PR adds the same handling to `circulation_update_collection` and to `import_collection` (which calls `bibliographic_lookup` once per event). `RemoteInitiatedServerError` is a sibling of `RemoteIntegrationException` (both derive from `IntegrationException`), not a subclass, so it must be listed explicitly in both `autoretry_for` and `throws`.

Note on behavior: a *persistently* empty response will, after exhausting the 4 retries, fail that task run (logged as expected via `throws`, not as an unhandled exception) without advancing the chain — failing safe rather than corrupting availability. Transient empties are absorbed by the retries.

## How Has This Been Tested?

New tests (all passing):

- `test_bibliographic_lookup_empty_response_raises` — parametrized over empty and whitespace-only bodies; asserts `RemoteInitiatedServerError`.
- `test_empty_api_response_does_not_zero_out_availability` — asserts an empty response leaves a pool's `licenses_owned`/`licenses_available` untouched (the safety property).
- `test_remote_initiated_server_error_retried_and_expected` added to both `TestCirculationUpdateCollection` and `TestBibliothecaImportCollection` — assert membership in `autoretry_for`/`throws` and behavioral retry (1 initial + 4 retries).

Ran the full test files for every `bibliographic_lookup` caller and all Bibliotheca tasks:

```
tox -e py312-docker -- --no-cov \
  tests/manager/integration/license/test_bibliotheca.py \
  tests/manager/integration/license/test_bibliotheca_circulation_updater.py \
  tests/manager/integration/license/test_bibliotheca_importer.py \
  tests/manager/integration/license/test_bibliotheca_purchase_record_importer.py \
  tests/manager/celery/tasks/test_bibliotheca.py
# 170 passed
```

`mypy` clean on all changed files.

- I will test on minotaur before releasing in hotfix release to production.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/bibliotheca.py
+++ b/src/palace/manager/celery/tasks/bibliotheca.py
@@ -76,12 +76,19 @@ def import_all_collections(task: Task) -> None:
     )
 
 
+# BibliothecaAPI.bibliographic_lookup, called per event by BibliothecaEventImporter,
+# raises RemoteInitiatedServerError when Bibliotheca returns an empty response body
+# (HTTP 200 with no XML document) -- a transient Bibliotheca-side condition. As with the
+# purchase-record and circulation-update tasks, it must be retried rather than surfacing
+# as an unhandled task exception. RemoteInitiatedServerError is a sibling of
+# RemoteIntegrationException (both derive from IntegrationException), not a subclass, so it
+# must be listed explicitly in both autoretry_for and throws.
 @shared_task(
     queue=QueueNames.default,
     bind=True,
     max_retries=4,
-    autoretry_for=(BadResponseException, RequestTimedOut),
-    throws=(RemoteIntegrationException,),
+    autoretry_for=(BadResponseException, RequestTimedOut, RemoteInitiatedServerError),
+    throws=(RemoteIntegrationException, RemoteInitiatedServerError),
     retry_backoff=60,
 )
 def import_collection(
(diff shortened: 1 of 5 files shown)


tdilauro (approved):
Not as thorough a review as I would like, but looks good! 🚀

PR #3510, opened 2026-06-26
Apply circulation-only updates during the Bibliotheca circulation sweep (PP-4666)

dbernstein:
## Description

The Bibliotheca circulation sweep (`BibliothecaCirculationUpdater._process_batch`) gated its apply on `BibliographicData.needs_apply()`. That method hashes the bibliographic data, and `BibliographicData.fields_excluded_from_hash()` **deliberately excludes `circulation`**. So for any title whose metadata (title/author/cover/etc.) was unchanged — i.e. nearly every title on nearly every sweep — `needs_apply()` returned `False`, the `bibliographic_apply` task was never queued, and the fresh availability counts read from Bibliotheca were silently discarded. The sweep logged `handled N identifier(s)` with no failures while updating nothing.

The fix extends the gate to also consult `CirculationData.needs_apply(session, collection)`, which hashes the `LicensePool` (availability included). When only circulation changed, `BibliographicData.apply()` takes its existing circulation-only fallback and applies the availability alone.

```python
needs_apply = bibliographic.needs_apply(self._session) or (
    bibliographic.circulation is not None
    and bibliographic.circulation.needs_apply(self._session, self._collection)
)
```

## Motivation and Context

JIRA (PP-4666)
Bibliotheca availability stopped updating from the periodic circulation sweep: identifiers were processed 25 at a time with no errors, but copy/hold counts never changed in the database.

This is a regression introduced when the sweep was migrated to Celery in #3403. The previous `BibliothecaCirculationSweep` (an `IdentifierSweepMonitor`) called `bibliographic.apply()` **unconditionally** for every title the API returned, so `apply()`'s circulation-only fallback always ran. The Celery migration added a `needs_apply()` short-circuit as a dedup optimization, but keyed it on the bibliographic hash while it was guarding the circulation apply — the one field that hash excludes.

The fix preserves the skip-genuinely-unchanged optimization (it does not blindly fan out an apply task per identifier) while no longer being blind to availability changes. Both the async sweep path and the synchronous on-demand path (`process_identifiers` → `BibliothecaAPI.update_availability`) sat behind the same gate and are both fixed.

## How Has This Been Tested?

- Added `test_circulation_applied_when_metadata_unchanged` — an end-to-end regression test: it applies metadata once, then feeds a second lookup with identical metadata but fewer available copies, asserts `needs_apply()` is `False` for it (pinning the exact bug scenario), and asserts the pool availability still updates.
- Added `test_circulation_change_queues_bibliographic_apply` — covers the async/sweep path dispatching when only circulation changed.
- Updated the two pre-existing tests that encoded the broken behavior so they hold circulation unchanged too.
- Verified the regression tests **fail on the pre-fix code** (the end-to-end one left the pool stuck at `available=5` instead of `0`, matching the production symptom) and **pass with the fix**.
- Full local runs (against throwaway Postgres/Redis): all tests in `test_bibliotheca_circulation_updater.py` and `test_bibliotheca.py` (95) and the bibliotheca Celery task tests (47) pass. `mypy` clean; pre-commit hooks pass on the changed files.
- I will test on minotaur once merged.

## Checklist

### DIFF
--- a/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
+++ b/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
@@ -159,8 +159,9 @@ class BibliothecaCirculationUpdater(LoggerMixin):
 
         Used by :meth:`~palace.manager.integration.license.bibliotheca.BibliothecaAPI.update_availability`
         and :class:`~palace.manager.scripts.availability.AvailabilityRefreshScript` for
-        on-demand single-title refreshes.  Applies the same hash-based deduplication
-        as :meth:`update_batch` so unchanged titles produce no database writes.
+        on-demand single-title refreshes.  Applies the same bibliographic- and
+        circulation-hash deduplication as :meth:`update_batch`, so titles whose
+        metadata and availability are both unchanged produce no database writes.
 
         Unlike the sweep (:meth:`update_batch`), changes are applied **synchronously**
         in the caller's session rather than queued as ``bibliographic_apply`` tasks, so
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):

jonathangreen:
Before merging you should fix the doc comment as the claude code review suggests.

PR #3511, opened 2026-06-26
Reconcile Bibliotheca circulation against live pool columns instead of the data hash (PP-4691)

dbernstein:
## Description

The Bibliotheca circulation sweep decided whether to write a `LicensePool`'s availability by
comparing the incoming snapshot's hash against the pool's stored `updated_at_data_hash`. That
hash is **not** a reliable proxy for the live availability columns, so the sweep skipped real
updates and Bibliotheca availability stopped syncing.

This change makes the sweep reconcile **circulation against the pool's live column values**
(`licenses_owned/available/reserved`, `patrons_in_hold_queue`, `status`) rather than the hash,
and forces the apply past the (possibly stale) hash with `even_if_not_apparently_updated=True`.
Bibliographic **metadata** keeps its hash-based dedup (that hash *is* reliable — an `Edition` is
only ever written through `apply()`), dispatched separately so unchanged metadata is not
re-applied every sweep.

This is a follow-up to #3510 (PP-4666), which was necessary but not sufficient — see below.

## Motivation and Context
JIRA (PP-4691)
### Symptom

Bibliotheca availability was not updating from the circulation sweep. Identifiers were processed
25 at a time with no errors, but copy/hold counts never changed. The only workaround was manually
setting `license_pools.updated_at_data_hash = NULL` for Bibliotheca pools before running the sweep.

### Root cause: the hash drifts away from the columns

`LicensePool.updated_at_data_hash` is written in exactly **one** place — `CirculationData.apply()`.
But the availability **columns** are also mutated out of band, without touching the hash, by:

- the **Bibliotheca event importer** (runs every few minutes), which applies checkout/checkin/hold
  events via `LicensePool.update_availability_from_delta()` → `update_availability()`, and
- **loan/hold** operations via `update_availability()`.

So the hash freezes at the last full-snapshot apply while the columns keep moving. When
Bibliotheca's current snapshot hashes equal to that frozen value — common — `should_apply_to(pool)`
returns `False` and the sweep skips, even though the columns have drifted. Nulling the hash is
exactly the escape hatch, which is why the manual workaround worked.

Note this is *by design* for the deltas: the event importer's own docstring says its data is
"unlikely to be completely accurate." The sweep's whole purpose is to reconcile that drift against
Bibliotheca's authoritative snapshot — and the stale hash was blinding it to the very drift it
exists to fix.

### Why #3510 didn't fully fix it

#3510 added a circulation gate to the sweep, but keyed it on `CirculationData.needs_apply()` —
which *is* that same hash comparison. It only helped when the hash differed; it did not help the
common case where the hash matches but the columns have drifted.

### Why Bibliotheca specifically (and not OPDS/ODL)

The bug needs three ingredients together: (1) an out-of-band writer of the availability columns
that doesn't stamp the hash, (2) that writer being *approximate* so a correction pass is genuinely
needed, and (3) the correction being a hash-gated reconciliation against a remote snapshot. Only
Bibliotheca has all three.

- **Pure OPDS** lacks (1): `OPDSAPI.update_availability` is a no-op; only the feed import writes
  availability, and it stamps the hash. One writer ⟹ the hash can't desync.
- **ODL** has out-of-band writers (loans/holds/expiration via `update_availability_from_licenses`)
  but lacks (2)/(3): those writes compute the *authoritative* counts from local license/hold state,
  ODL counts never come from a remote snapshot (the feed sets `licenses_owned/available = None`),
  and the OPDS importer already refuses to trust the hash for ODL circulation
  ("license expiry is time-dependent and undetectable by hashing").

In other words, the codebase's existing answer to "availability can change out of band" is *don't
gate it on the hash* — ODL already does this. This PR brings Bibliotheca's sweep in line with that
stance.

## Alternatives considered

**A. Reconcile against the live columns (this PR).** Compare the snapshot to the actual columns;
apply (forced) only when they differ. Localized to the Bibliotheca updater; needs no migration
(it ignores the stale hash entirely, so the next sweep self-heals); precise (skips net-zero churn).

**B. Recompute the hash whenever `update_availability(_from_delta)` changes the columns.** This
would also be correct, but it computes the same "did availability change?" decision indirectly, and
the recompute is fragile: the stored hash is a hash of the incoming `CirculationData` *wire object*
(counts **plus** `formats`, `status`, `type`, `default_rights_uri`, identifier, …), not of the pool.
To produce a *matching* hash, the delta path would have to rebuild that exact object from DB state
and serialize it byte-identically — there is no `CirculationData.from_license_pool()`, the delta
path doesn't have the source's `formats`, and `update_availability` is a generic method shared by
every distributor (each with a differently-shaped `CirculationData`). Getting any field wrong flips
the failure from "skips too much" (visible) to "re-applies everything every run" (silent).

**C. Invalidate (null) the hash on any out-of-band column change.** The tractable cousin of B — no
reconstruction needed. But it touches the shared hot path for *all* distributors, re-applies every
active title every sweep (even net-zero churn that A skips), and still requires the one-time
"null all existing hashes" migration to fix already-drifted pools.

**D. Canonical-state dedup refactor.** Compute the circulation dedup hash from the pool's canonical
state instead of the wire object, uniformly across distributors. This is the "right" long-term shape
if we want one mechanism everywhere, but it's a deliberate, separately-tested refactor with a
migration — not a hotfix.

### Rationale for choosing A

A and B/D make the **same** apply/skip decision (apply ⟺ the snapshot differs from the columns); A
just computes it directly instead of through a maintained-hash detour. Given that, A wins on every
axis that matters for this fix:

- **Blast radius**: touches only the Bibliotheca updater, not `update_availability`'s ~20 call
  sites across every distributor.
- **No migration**: it never consults the stale hash, so existing drift self-heals on the next
  sweep — nothing to reset.
- **Precision**: skips titles whose net availability is unchanged, which C would re-apply.
- **Consistency**: matches ODL's existing "don't trust the hash for availability" stance.

The cost is a narrow completeness gap — A compares counts + status, not the other hashed fields
(`formats`, `type`, …) — which is effectively moot for Bibliotheca (no circulation links/licenses,
`type` always METERED, `status` is compared, EPUB/EPUB3 formats collapse to identical `FormatData`)
and can be widened later if needed.

We are deliberately **deferring** the broader canonical-state refactor (D). It's the cleaner uniform
solution, but it's a larger, migration-bearing change that should be done on its own, not folded into
this fix.

## How Has This Been Tested?

- Added `test_sweep_reconciles_drift_when_hash_is_stale`, an end-to-end regression for the exact
  reported scenario: apply (5,5), drift the columns to (5,2) via `update_availability` (no hash
  change), then feed the authoritative (5,5) snapshot whose hash *matches* the stored one. It
  asserts the precondition (`needs_apply(...) is False`) and that the sweep still restores the
  columns. **Fails on the hash-gate, passes now.**
- Converted the dispatch tests to real data and added coverage for the two-track behaviour
  (metadata-only → `bibliographic_apply`; availability-only → `circulation_apply` with the
  force-apply policy; no-change → nothing) and for `_availability_matches_pool`.
- Full local runs (throwaway Postgres/Redis): the updater suite plus `test_bibliotheca.py` and the
  Bibliotheca Celery task tests pass. `mypy` and pre-commit clean.
- On minotaur I updated the Biblotheca licensepools.licenses_available to 0 and then ran the bibliotheca_circulation_update script.  I verified that the count went from 0 to the expected 71.
## Checklist

### DIFF
--- a/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
+++ b/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
@@ -312,11 +312,29 @@ class BibliothecaCirculationUpdater(LoggerMixin):
         :attr:`~palace.manager.sqlalchemy.model.licensing.LicensePool.updated_at_data_hash`,
         which is not a reliable proxy for the current column values (see
         :meth:`_process_batch`).
+
+        A field the snapshot leaves unset (``None``) is treated as "no assertion" and
+        does not, on its own, count as a mismatch — mirroring
+        :meth:`~palace.manager.sqlalchemy.model.licensing.LicensePool.update_availability`,
+        which skips ``None`` values rather than writing them. (Bibliotheca always
+        populates every count and ``status``, so this only matters defensively.)
         """
         return (
-            circulation.licenses_owned == pool.licenses_owned
-            and circulation.licenses_available == pool.licenses_available
-            and circulation.licenses_reserved == pool.licenses_reserved
-            and circulation.patrons_in_hold_queue == pool.patrons_in_hold_queue
+            (
+                circulation.licenses_owned is None
+                or circulation.licenses_owned == pool.licenses_owned
+            )
+            and (
+                circulation.licenses_available is None
+                or circulation.licenses_available == pool.licenses_available
+            )
+            and (
+                circulation.licenses_reserved is None
+                or circulation.licenses_reserved == pool.licenses_reserved
+            )
+            and (
+                circulation.patrons_in_hold_queue is None
+                or circulation.patrons_in_hold_queue == pool.patrons_in_hold_queue
+            )
             and (circulation.status is None or circulation.status == pool.status)
         )
--- a/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
+++ b/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
@@ -166,8 +166,9 @@ class BibliothecaCirculationUpdater(LoggerMixin):
         metadata and availability are both unchanged produces no database writes.
 
         Unlike the sweep (:meth:`update_batch`), changes are applied **synchronously**
-        in the caller's session rather than queued as ``bibliographic_apply`` tasks, so
-        the updated availability is visible as soon as this method returns.  Callers such
+        in the caller's session rather than queued as ``bibliographic_apply`` /
+        ``circulation_apply`` tasks, so the updated availability is visible as soon as
+        this method returns.  Callers such
         as :meth:`~palace.manager.integration.license.bibliotheca.BibliothecaAPI.update_availability`
         rely on this — e.g. the circulation dispatcher reads ``LicensePool.licenses_available``
         immediately after requesting an availability refresh.
--- a/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
+++ b/src/palace/manager/integration/license/bibliotheca_circulation_updater.py
@@ -12,6 +12,7 @@ from palace.util.datetime_helpers import utc_now
 from palace.util.log import LoggerMixin
 
 from palace.manager.celery.tasks import apply
+from palace.manager.data_layer.circulation import CirculationData
 from palace.manager.data_layer.policy.replacement import ReplacementPolicy
 from palace.manager.integration.license.bibliotheca import BibliothecaAPI
 from palace.manager.sqlalchemy.constants import DataSourceConstants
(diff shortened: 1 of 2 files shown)


dbernstein:
Thanks for the reviews — addressed the minor items in 1aa3bf2d4:

- **`_availability_matches_pool` None-handling** (Greptile): made it consistent — a `None` field (count *or* status) is now treated uniformly as "no assertion," mirroring `update_availability` which skips `None` values, and documented the convention.
- **Status coverage** (Claude): `_make_bib` now sets a production-shaped `status` (`ACTIVE` when copies are owned, else `EXHAUSTED`, matching `ItemListParser._make_circulation_data`), so the embedded circulation carries a status like real Bibliotheca data.
- **Test placement** (Greptile): moved the matcher unit test out of `...UpdateBatch` into a dedicated `TestBibliothecaCirculationUpdaterAvailabilityMatchesPool` class, made it self-contained, and added a status-mismatch case so the status leg is actually exercised.

`mypy`, pre-commit, and the updater/bibliotheca/Celery suites are green.

jonathangreen (approved):
Looks good for a hot fix. I think we need to either do option `B` or option `D` as a long term fix, but @dbernstein is going to make a ticket for that as we discussed in the sprint planning meeting (PP-4703).

I think there is a gap here with format changes can still get skipped, since the new check ignores them and only checks counts, but we can address that in the follow up (it probably will be automatically addressed when we start using the hash again).

PR #3512, opened 2026-06-29
Support parallel test runs in external-schema mode

jonathangreen:
## Description

Makes the test suite's external-schema mode parallel-capable. When database creation is enabled (the default), each xdist worker now provisions its own database by cloning the configured database as a Postgres template (`CREATE DATABASE ... TEMPLATE`), instead of all workers sharing the one configured database. The previous release's tests still run against the schema the current code produced, but each worker gets an isolated database, so the suite can run under xdist.

Admin (`CREATE`/`DROP DATABASE`) operations now connect to the `postgres` maintenance database rather than the configured one, since you cannot clone a template you are connected to; these are cluster-level commands that work from any database, and this path is exercised by the normal suite on every run.

The single-shared-database serial mode (external schema with database creation disabled) is preserved for local debugging.

## Motivation and Context

The backwards-compatibility CI check runs a previous release's `-m db` test suite against a schema built by the current code, using external-schema mode. Until now that mode used the one configured database as-is and forbade xdist, so the suite ran serially (`-n0`) and took ~13 minutes. Its not really feasible for this check to take so long, so I'm adding this option to speed it up. 

This will have to be merged and go into a release, before I can test it in the workflow in https://github.com/ThePalaceProject/circulation/pull/3506

## How Has This Been Tested?

Because the backwards-compatibility job won't exercise this for real until the next release, it was proven locally:

- Parallel external clone mode against a real Postgres: a db-test slice ran green under `-n auto` with `PALACE_TEST_DATABASE_EXTERNAL_SCHEMA=true` / `PALACE_TEST_DATABASE_CREATE_DATABASE=true`; per-worker `session_gw*` clones were created from the template and dropped at teardown.
- Negative control: dropping a table from the template made the same parallel run fail with `UndefinedTable`, confirming the check still catches backwards-incompatible schemas under parallelism.
- Normal-mode regression: the standard parallel suite (which now routes admin DDL through the `postgres` maintenance database on every run) passed via `tox -e py312-docker`.
- Full end-to-end against a locally built image using the real backwards-compatibility compose overlay (`-n auto`, clone-per-worker): 3,269 db tests passed in 3:47 on a 4-worker runner, versus ~13 minutes serial.
- `mypy` clean.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -778,11 +778,13 @@ to create and drop databases, the tests will fail. You can disable this behavior
 `PALACE_TEST_DATABASE_CREATE_DATABASE` environment variable to `false`.
 
 Setting `PALACE_TEST_DATABASE_EXTERNAL_SCHEMA` to `true` tells the tests that the schema has already
-been applied to the provided database, so the fixtures use it as-is, without dropping and recreating the
-schema from the current models. Because the database is used as-is, you must also set
-`PALACE_TEST_DATABASE_CREATE_DATABASE` to `false` — the two settings must agree, or the tests will raise
-an error — which also means the tests must be run serially. This is used by the backwards-compatibility
-CI check, which runs a previous release's tests against a schema built by the current code.
+been applied to the provided database (by the current code), so the fixtures use that schema as-is,
+without dropping and recreating it from the current models. With database creation left enabled (the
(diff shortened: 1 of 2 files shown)


dbernstein (approved):
🦞 looks good.

PR #3519, opened 2026-06-29
Reconcile circulation against live pool columns in the shared data layer (PP-4703)

dbernstein:
## Description

Generalizes the Bibliotheca live-column reconciliation from #3510/#3511 (PP-4666/PP-4691) into the
**shared data layer**, so circulation availability dedup is correct for **every** distributor, and
deletes the Bibliotheca-local band-aid. Also folds in the missing Boundless circulation gate.

The fix lives in `CirculationData`:

- **`fields_excluded_from_hash()`** now excludes the four availability count fields
  (`licenses_owned/available/reserved`, `patrons_in_hold_queue`). The dedup hash therefore covers only
  the *stable* fields (`formats`, `status`, `type`, `licenses`, …), which are only ever written by
  `apply()` and so remain reliable.
- **`should_apply_to(pool)`** compares those counts against the pool's **live columns** (in addition to
  the stable-field hash), while preserving the base `as_of_timestamp < pool.updated_at` staleness
  short-circuit. Both `needs_apply()` (the importers' gate) and `apply()`'s own internal check route
  through this method, so no `even_if_not_apparently_updated` is needed and an async apply that arrives
  after the columns already match simply no-ops.

Every importer that already gates circulation on `circulation.needs_apply()` (OPDS / OPDS2 /
for-distributors / ODL via the shared base, Overdrive, the Bibliotheca sweep) inherits the fix with no
logic change. The Bibliotheca sweep is simplified back to the OPDS-style two-track gate
(`bibliographic_apply` / `circulation_apply`), dropping `_availability_matches_pool` and the `even_if`
force. Boundless gets its missing circulation gate so it routes through the corrected dedup too.

## Motivation and Context
JIRA (PP-4703)
`LicensePool.updated_at_data_hash` is written only by `CirculationData.apply()`, but the availability
**columns** are mutated **out of band** — by the Bibliotheca event importer
(`update_availability_from_delta`), ODL loans/holds/expiration (`update_availability_from_licenses`), and
the circulation dispatcher (`update_availability`) — none of which restamp the hash. So the hash drifts
from the columns, and a hash-gated reconciliation skips real updates. This caused the Bibliotheca
availability-sync bug fixed locally in #3511; this PR removes the underlying drift class for all
distributors instead of per-distributor.

**ODL/aggregated pools are intentionally unaffected.** Their CirculationData carries `licenses` with the
counts as `None`, so the live-column compare is a no-op and they keep using the hash — matching the
existing `test_needs_apply_aggregated_pool_uses_hash` ("the hash check applies uniformly; license expiry
is handled by a dedicated Celery task"). License expiry remains the `license_expiration` task's job.

### Why this shape (and not the literal "canonical-state" refactor)

A *pure* canonical-state compare (and dropping `updated_at_data_hash`) is impractical: the stable fields
(`formats` ↔ delivery mechanisms, `links`, ODL `licenses`) don't round-trip from the pool and have no
reverse constructor. But the drift is confined to the **count columns**, and the hash is already reliable
for the stable fields — so we exclude the counts from the hash and compare them live, keeping the hash
for everything else. No migration is required: excluding the counts changes the hash value, so each pool
re-applies once on its next sweep/import and self-heals.

Considered and rejected: **dropping the column** (semantic comparators for the hard fields + an N-1
migration — much more code for marginal benefit) and **nulling the hash inside `update_availability`**
(touches a shared hot path for all distributors, re-applies every active title each run, and needs a
resync migration).

## How Has This Been Tested?

- **`tests/manager/data_layer/test_circulation.py`** (the core): added `test_calculate_hash_excludes_availability_counts`,
  `test_needs_apply_compares_live_columns` (the drift regression — a count change is detected while the
  stable-field hash matches), and `test_needs_apply_skips_older_snapshot_even_when_counts_differ`
  (staleness preserved). Existing `test_needs_apply` and `test_needs_apply_aggregated_pool_uses_hash`
  (ODL) continue to pass unchanged.
- **Bibliotheca updater tests** simplified to the two-track shape; the `_availability_matches_pool` unit
  tests are removed (logic now lives in `test_circulation.py`); the end-to-end drift regression is kept
  and its precondition updated (`needs_apply` now correctly reports the drift).
- **Boundless** importer gets the staged circulation-gate change and its tests.
- Full local runs (throwaway Postgres/Redis): **770** `integration/license/` + `data_layer/` tests and
  the apply / per-distributor Celery task tests pass. `mypy` and pre-commit clean. (The only non-passing
  Celery tests are pre-existing `test_search`/`test_marc` cases that require OpenSearch/S3, unrelated to
  this change.)

## Checklist

### DIFF
--- a/src/palace/manager/data_layer/circulation.py
+++ b/src/palace/manager/data_layer/circulation.py
@@ -11,6 +11,7 @@ from palace.manager.data_layer.license import LicenseData
 from palace.manager.data_layer.link import LinkData
 from palace.manager.data_layer.policy.replacement import ReplacementPolicy
 from palace.manager.sqlalchemy.model.collection import Collection
+from palace.manager.sqlalchemy.model.edition import Edition
 from palace.manager.sqlalchemy.model.licensing import (
     DeliveryMechanism,
     LicensePool,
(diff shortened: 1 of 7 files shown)


src/palace/manager/data_layer/circulation.py
jonathangreen: This comment is now wrong, a pool with an unchanged hash but drifted columns now reaches this point.  The conclusion is still right; the stated reason is false and should be fixed.

jonathangreen: This comment is no longer correct, the call now resolves to the new `CirculationData.should_apply_to` override. Either update the reference, or just drop it entirely, since this sort of thing tends to get out of date like this very easily.

src/palace/manager/data_layer/circulation.py
```diff
@@ -402,3 +403,76 @@ def needs_apply(self, session: Session, collection: Collection) -> bool:
...
+
+        ``licenses_owned``/``licenses_available``/``licenses_reserved`` and
+        ``patrons_in_hold_queue`` are mutated **out of band** — by the event
+        importer, loan/hold operations, and the circulation dispatcher via
+        :meth:`~palace.manager.sqlalchemy.model.licensing.LicensePool.update_availability`
+        (and ``update_availability_from_delta`` / ``update_availability_from_licenses``)
+        — which never restamp ``updated_at_data_hash``.  The hash therefore cannot
+        track them, so they are excluded from it and compared against the pool's live
+        columns in :meth:`should_apply_to` instead.  Every other field is only ever
+        written by :meth:`apply`, so the hash stays a reliable change-signal for it.
```
jonathangreen: This spends a lot of time describing the current state, but I think it would be more useful to have a comment that lays out the general rule or reasoning here, so that new fields added to circulation have a test that they can use to decide if they should be excluded or not. Something like: 

```
Rule for any field added to CirculationData: it may ride the dedup hash only if
it is *exclusively* written by :meth:`apply` — then the stored hash (the data as
of the last apply) still reflects the live row, so "incoming hash != stored hash"
is a sound change-signal. A field that the database can mutate independently of
:meth:`apply` will drift from the hash; exclude it here and live-compare it in
:meth:`should_apply_to` instead (as the availability counts are).
```

src/palace/manager/integration/license/bibliotheca_circulation_updater.py
```diff
@@ -238,15 +226,20 @@ def _process_batch(
             if identifier is not None:
                 identifiers_not_mentioned.discard(identifier)
 
-            # Track 1 -- bibliographic metadata. Hash-based dedup is reliable
-            # here (an Edition is only ever written through apply()). Circulation
-            # is stripped off and handled separately below so that an unchanged
-            # title does not force a metadata re-apply every sweep.
+            # Two-track reconciliation (same shape as the OPDS importer): apply
```
jonathangreen: I don't think we should reference the OPDS importer here. Its so easy for this to get out of date. Lets make a comment that will stand on its own.

jonathangreen (commented):
This looks good and is a good improvement. 

I'm approving this, but I added some comments that I'd like to get resolved about out of date code comments before this gets merged. 

I can't add a comment on the file because it wasn't changed in this PR, but I'd like to see this comment updated as well: 
https://github.com/ThePalaceProject/circulation/blob/cd797a5db4bb857fed25870d0f1c77be65efe14d/src/palace/manager/sqlalchemy/model/licensing.py#L333-L334

This is the definition of the field whose meaning this PR changed, and after these changes the comment there is out of date.  It is no longer a hash of the CirculationData, it's a hash of only the stable fields. This should probably point to `fields_excluded_from_hash` so that people know where to go to see what exactly is included in this hash.

jonathangreen (approved):

dbernstein:
Thank you @jonathangreen :  I'll make those changes before merging.

PR #3520, opened 2026-06-29
Retire the CoverageProvider machinery (PP-4468)

dbernstein:
## Description

Retires the legacy `CoverageProvider` machinery. Overdrive bibliographic coverage now runs as a direct metadata-apply step inside `OverdriveAPI.update_licensepool` instead of routing through `OverdriveBibliographicCoverageProvider.ensure_coverage`. The base `CoverageProvider` classes, the Overdrive/Bibliotheca providers, the coverage-provider runner scripts, and the now-dead readers (`Identifier.missing_coverage_from`, the `Explain` script's coverage block, the always-failing admin `refresh_metadata` provider path) are deleted.

`BibliographicData.apply` no longer writes `CoverageRecord` rows (the `create_coverage_record` parameter is removed; the Celery apply task and the Bibliotheca updater already passed `create_coverage_record=False`).

The new `_ensure_bibliographic_coverage` helper wraps `bibliographic.apply(...)` in a broad `except Exception` that logs a warning and returns, preserving the resilience of the retired `BibliographicCoverageProvider.set_bibliographic` so a database error or unexpected data shape during apply does not propagate out of `update_licensepool`.

The `CoverageRecord` model and `coveragerecords` table are intentionally left in place (marked dormant) for one release, per the online-migration / N-1 rule. A stacked follow-up PR drops the `coveragerecords` and dormant `equivalentscoveragerecords` tables.

> Note: this PR supersedes #3503; it is opened from a branch on the upstream repository, rebased on the latest `main`, and addresses the review feedback there (the `bibliographic.apply` exception guard).

## Motivation and Context
JIRA (PP-4468)
The `CoverageProvider` batch-processing pattern is being retired in favor of Celery. Investigation showed nothing live reads coverage rows:

- The only live caller of `ensure_coverage` (Overdrive's `update_licensepool`) passes `force=True` and ignores the result.
- The admin `refresh_metadata` endpoint is wired with no provider, so it always returns `METADATA_REFRESH_FAILURE` without invoking a provider.
- `Identifier.missing_coverage_from` / `should_update` / the `Explain` coverage block are only reachable through the already-dead runner scripts.

The Celery `bibliographic_apply` task and the Bibliotheca updater already skip coverage-record writes, so Overdrive was the last writer.

## How Has This Been Tested?

- `mypy` and `ruff` pass on all changed files.
- Targeted suites pass under `tox -e py312-docker`: Overdrive API including the rewritten `update_licensepool` bibliographic-coverage path and a new test covering the `bibliographic.apply` error guard, the Bibliotheca circulation updater, data-layer `apply`, and the affected model tests.

## Checklist


jonathangreen:
@dbernstein it looks like tests are failing on this one, and greptile has a valid review comment about unconditionally calling `self.load_work`. 

I'll let you resolve those issues before reviewing this one. When its ready you can request a review from @ThePalaceProject/backend again.

src/palace/manager/api/admin/controller/work_editor.py
```diff
@@ -448,39 +446,23 @@ def unsuppress(
...
-        self, identifier_type: str, identifier: str, provider: Any | None = None
+        self, identifier_type: str, identifier: str
     ) -> Response | ProblemDetail:
-        """Refresh the metadata for a book from the content server"""
+        """Refresh the metadata for a book from the content server.
+
+        Metadata refresh used to run through the per-source CoverageProvider
+        machinery, which has been retired. No provider is wired up, so this
+        endpoint is retained for API compatibility but no longer performs a
+        refresh; it always reports failure.
```
jonathangreen: Since this does nothing can we deprecate it, or just remove it entirely? What calls this endpoint? I think we need a plan to deal with this endpoint since its no longer functional.

src/palace/manager/integration/license/overdrive/api.py
```diff
@@ -1806,6 +1799,62 @@ def _edition(self, licensepool: LicensePool) -> tuple[Edition, bool]:
...
         )
 
+    def _ensure_bibliographic_coverage(self, license_pool: LicensePool) -> None:
+        """Fetch Overdrive bibliographic metadata, apply it to the Edition, and
+        make the LicensePool's Work presentation-ready.
+
+        This replaces the former ``OverdriveBibliographicCoverageProvider``
+        path that ran inside :meth:`update_licensepool`. A bad or unrecognized
+        Overdrive ID simply leaves the pool without a Work, matching the prior
+        behavior (``update_licensepool`` ignored the coverage result).
```
jonathangreen: I find referring to old removed code in a new code like this does not add much value. Claude likes to do this a lot.

If anything it causes confusion as you have to go back in git history looking for `OverdriveBibliographicCoverageProvider` to try to understand the comment. This should just tell us what we are currently doing not give a history lesson.

src/palace/manager/integration/license/overdrive/api.py
```diff
@@ -1806,6 +1799,62 @@ def _edition(self, licensepool: LicensePool) -> tuple[Edition, bool]:
...
+                edition,
+                self.collection,
+                replace=ReplacementPolicy.from_license_source(),
+            )
+        except Exception as e:
+            # Mirror the resilience of the former
+            # ``BibliographicCoverageProvider.set_bibliographic``: a database
+            # error or unexpected data shape during apply should not crash
+            # ``update_licensepool``. Log and leave the pool without updated
+            # coverage, matching the prior "silently continue" behavior.
```
jonathangreen: Again we should explain why we are doing this, not talk about old removed code in the comment.

src/palace/manager/integration/license/overdrive/api.py
```diff
@@ -1806,6 +1799,62 @@ def _edition(self, licensepool: LicensePool) -> tuple[Edition, bool]:
...
+
+        edition, _ = self._edition(license_pool)
+        try:
+            bibliographic.apply(
+                self._db,
+                edition,
+                self.collection,
+                replace=ReplacementPolicy.from_license_source(),
+            )
+        except Exception as e:
```
jonathangreen: I really really don't like broad exceptions like this. We should avoid them as much as we possibly can. They catch and mask so much, like this would catch and hide a key name typo in the code.

tests/manager/integration/license/overdrive/test_api.py
```diff
@@ -1783,6 +1774,41 @@ def test_update_licensepool_provides_bibliographic_coverage(
...
 
+    def test_ensure_bibliographic_coverage_apply_error(
+        self,
+        overdrive_api_fixture: OverdriveAPIFixture,
+        db: DatabaseTransactionFixture,
+        caplog: pytest.LogCaptureFixture,
+    ):
+        # An exception raised while applying bibliographic data is caught and
+        # logged rather than propagating out of update_licensepool, mirroring
+        # the resilience of the retired BibliographicCoverageProvider.
```
jonathangreen: Comment referring to removed code

jonathangreen (commented):
Some comments on this for consideration while you are resolving the test issues

src/palace/manager/api/admin/controller/work_editor.py
jonathangreen: There a 👍🏻 on this, but do we have a ticket or plan for it?

jonathangreen (commented):
Looks like there are some valid AI code review comments that still should be resolved on this one, so I'm going to leave it until you get back from vacation @dbernstein

src/palace/manager/integration/license/overdrive/api.py
```diff
@@ -1806,6 +1800,59 @@ def _edition(self, licensepool: LicensePool) -> tuple[Edition, bool]:
...
+            # A failure while applying the bibliographic data must not abort the
+            # surrounding availability update, so log it and leave the pool's
+            # presentation edition unchanged.
+            self.log.warning(
+                "Error applying Overdrive bibliographic data to edition %s: %s",
+                edition.id,
+                e,
+                exc_info=e,
+            )
+            return
```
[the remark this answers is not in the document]
dbernstein: Took the first option: kept the narrow guard and tightened the docs to match.

Widening back to `except Exception` isn't available — that's what @jonathangreen pushed back on above ("they catch and mask so much, like this would catch and hide a key name typo in the code"), and he's right. So the split is now deliberate and spelled out in the docstring:

- `BasePalaceException` from `apply` means bad or incomplete data for this one title. Logged, skipped, batch continues.
- A SQLAlchemy error means a broken session or a genuine data-integrity problem, and a `KeyError` means a bug. Both propagate, because neither is specific to one title and both should stop the run loudly rather than quietly leaving thousands of pools without Works.

Good catch on the PR description — it still claimed a "broad `except Exception`" from an earlier revision. Corrected.

src/palace/manager/integration/license/overdrive/api.py
```diff
@@ -1806,6 +1799,62 @@ def _edition(self, licensepool: LicensePool) -> tuple[Edition, bool]:
...
+            # error or unexpected data shape during apply should not crash
+            # ``update_licensepool``. Log and leave the pool without updated
+            # coverage, matching the prior "silently continue" behavior.
+            self.log.warning(
+                "Error applying Overdrive bibliographic data to edition %s: %s",
+                edition.id,
+                e,
+                exc_info=e,
+            )
+            return
```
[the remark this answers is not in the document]
dbernstein: Leaving `metadata_lookup` outside the guard on purpose, and the docstring now says so instead of implying blanket resilience.

The asymmetry is the point. A failed `apply` is usually about one title's data. A failed fetch usually isn't: Overdrive is unreachable, or the credentials are wrong, or the collection token is stale — conditions that will hit every identifier in the batch. Callers run this in a loop over batches of identifiers, so swallowing a fetch failure would walk the whole batch logging warnings and finish looking like a successful run, having left every pool without a Work. Letting it propagate fails the task, which is the honest outcome and the one that gets retried.

So the guard's intent isn't "bibliographic failures must not crash availability updates" — it's narrower than that, and the docstring is now explicit about which failures are skipped and which are allowed to stop the run.

src/palace/manager/api/admin/controller/work_editor.py
dbernstein: No ticket needed — I checked what calls it, and the answer is nothing, so it's removed outright in this PR rather than deprecated.

`circulation-admin` only renders its "Refresh Metadata" button when the admin OPDS entry carries a `http://librarysimplified.org/terms/rel/refresh` link (`editorAdapter.ts` looks for that rel, and `BookDetailsEditor` gates the button on `bookData.refreshLink`). Nothing in the manager has emitted that rel since the metadata wrangler came out in #792 — the current admin annotators emit only suppress / unsuppress / edit / search. `ios-core` has no reference to the rel or the endpoint, and doesn't use the admin API at all.

So the button never appears and the route was unreachable; anything that did reach it got an unconditional 400. Removed the route, the controller method, its tests, and the two orphaned `METADATA_REFRESH_*` problem details. Also dropped `test_feed_includes_refresh_link`, which despite the name only asserted the *absence* of that rel — a leftover from the wrangler removal that no longer has an endpoint behind it.

Tagged the PR `incompatible changes` since it drops a registered admin route, though in practice no client could reach it.

jonathangreen (approved):
Looks good

PR #3523, opened 2026-06-30
Remove the superseded v5/v6/v7 search schema revisions

jonathangreen:
## Description

Stacked on top of #3522 (`feature/search-v8`). Removes the superseded v5/v6/v7 search schema revisions now that v8 is self-contained and is the only revision production creates indexes from.

- Deletes `src/palace/manager/search/v5.py`, `v6.py`, and `v7.py`, along with their dedicated tests (`test_search_v6.py`, `test_search_v7.py`).
- Drops the old revisions from `SearchRevisionDirectory`, leaving `REVISIONS = [SearchV8()]`.
- Repoints the remaining references at v8: the `MockSearchSchemaRevisionLatest` mapping document and a `name_for_index` docstring example.
- Moves the author char-filter test (previously written against v5 in `test_external_search.py`) into `TestSearchV8`, and drops v8's v7-equivalence test now that v7 no longer exists to compare against.

## Motivation and Context

#3522 made v8 self-contained specifically so that older revisions could be deleted once they fell out of use, instead of accumulating indefinitely behind the inheritance chain. This PR is that cleanup. Opening it now (in draft) keeps the follow-up visible so the removal isn't forgotten after the v8 release goes out.

## How Has This Been Tested?

Ran the affected suites under the docker tox environment — `tests/manager/search/`, `tests/manager/scripts/test_initialization.py`, and `tests/manager/celery/tasks/test_search.py` (the latter two exercise `MockSearchSchemaRevisionLatest`, whose backing document now comes from v8). All 171 tests pass. `mypy` is clean on the changed source.

## Checklist


jonathangreen:
I'm going to leave this in draft until its safe to merge, but its ready for a code review now, then it can just wait to get merged once approved.

tdilauro (approved):
Looks good! 🏁

PR #3524, opened 2026-06-30
Remove the orphaned v5/v6/v7 search indexes

jonathangreen:
## Description

Stacked on top of #3523. Cleans up the orphaned `circulation-works-v5`, `-v6` and `-v7` OpenSearch indexes that earlier schema revisions left behind in the cluster.

When the search schema migrates to a new revision, the reindex flow creates the new index and re-points the read/write aliases at it, but it never deletes the index it replaced — so every retired revision leaves an orphaned `{base}-v{n}` index sitting in the cluster, consuming shards and disk. Nothing in the codebase has ever removed them. #3523 removes the v5–v7 revision *code*; this PR removes the leftover *indexes*.

This adds:

- **`SearchService.index_remove(name)`** — deletes an index if it exists and reports whether it did. The service had no delete-index capability before.
- **`remove_search_indices(service, versions, *, log)` helper** — deletes the named old versions, with a safety guard that **never deletes an index a read or write alias still points at**, so it cannot drop the live index even if asked to. It's idempotent (delete-if-exists) and returns the names it removed.
- **A one-shot startup task** (`startup_tasks/2026_06_30_remove_old_search_indexes.py`) that calls the helper inline for versions `[5, 6, 7]` on the next boot.

Removing a handful of indexes is a fast metadata operation, so this runs inline in the startup task rather than via a background job. If OpenSearch is briefly unavailable the helper raises, the startup runner declines to record the task as done, and it simply retries on the next boot.

## Motivation and Context

Removing the v5–v7 revision modules (#3523) stops the code from referencing those schemas, but the actual indexes remain in the cluster forever because the migration flow only swaps aliases and never drops the old index. This reclaims that cluster state (shards, disk) as the final step of retiring those revisions.

## How Has This Been Tested?

New tests, all passing under the docker tox environment:

- `tests/manager/search/test_service.py::TestService::test_index_remove` — `index_remove` deletes a real index and is a no-op (returns `False`) when it's absent.
- `tests/manager/search/test_service.py::TestRemoveSearchIndices` — three cases against the real OpenSearch fixture: the happy path (old indexes removed, the live index preserved), the alias guard (an index a read alias points at is skipped), and a no-op when the indexes don't exist.

Also ran the startup-task and initialization suites to confirm the new task file is discovered and the boot wiring is unaffected.

## Checklist

### DIFF
--- a/src/palace/manager/search/service.py
+++ b/src/palace/manager/search/service.py
@@ -363,7 +363,10 @@ class SearchServiceOpensearch1(SearchService, LoggerMixin):
         if not self._write_client.indices.exists(index=name):
             return False
         self.log.info(f"removing index {name}")
-        self._write_client.indices.delete(index=name)
+        # ignore=[404] covers the small window where the index is dropped between
+        # the exists() check and here, so a concurrent delete is a no-op rather
+        # than a NotFoundError.
+        self._write_client.indices.delete(index=name, ignore=[404])
         return True
 
 
--- a/src/palace/manager/search/service.py
+++ b/src/palace/manager/search/service.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import re
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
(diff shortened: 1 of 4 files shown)


jonathangreen:
This is ready for review, but I'll keep it in draft until its ready to be merged

tdilauro (approved):
This looks good! 🏁🔥

PR #3532, opened 2026-07-05
Fix SAML SP metadata model form field type

tdilauro:
## Description

Sets the form field type for the SAML Service Provider XML Metadata setting to `TEXTAREA`, so the admin UI renders it as a multi-line input instead of a single-line text box.

## Motivation and Context

The SP XML metadata field holds a multi-line XML document. Without an explicit `type`, `FormMetadata` defaults to `FormFieldType.TEXT` (single-line input), which makes the field impractical and annoying to edit. The adjacent `service_provider_private_key` and `non_federated_identity_provider_xml_metadata` fields already use `TEXTAREA`; this brings the SP metadata field into line with them.

## How Has This Been Tested?

This is a UI-metadata-only change with no logic impact. No new tests were added. The existing test suite covers the field's validation behavior and continues to pass.

## Checklist


dbernstein (approved):
🚀

PR #3543, opened 2026-07-08
Convert the Bibliotheca integration into a package

jonathangreen:
## Description

Restructures the Bibliotheca integration from five flat `bibliotheca_*` modules into a `bibliotheca/` package, matching the layout of every other integration.

```
bibliotheca.py (1295)                    bibliotheca/
bibliotheca_importer.py            →       api.py                      596
bibliotheca_purchase_record_...py          parser.py                   621
bibliotheca_circulation_updater.py         coverage.py                  65
bibliotheca_scripts.py                     settings.py                  57
                                           constants.py                 13   (new)
                                           importer.py                 212
                                           purchase_record_importer.py 217
                                           circulation_updater.py      288
                                           scripts.py                  158
```

The 1295-line `bibliotheca.py` is split along the seams the siblings already use: `settings.py` (as in `overdrive/` and `boundless/`), `parser.py` (as in `boundless/`), and `coverage.py` (as in `overdrive/`).

**A cycle had to be broken first.** `BibliothecaAPI` constructs the parsers, but `ErrorParser`, `EventParser` and `PatronCirculationParser` reached back for `BibliothecaAPI.SERVICE_NAME` and `.ARGUMENT_TIME_FORMAT`. The new `constants.py` holds those values and the parsers read them from there — the same way `boundless/parser.py` and `overdrive/coverage.py` already avoid this. `BibliothecaAPI` keeps both class attributes (tests and `license_providers.py` read them off the class), so no caller changes.

**Two dead attributes removed.** `EventParser.EVENT_SOURCE` has no readers anywhere in the repo. `EventParser.SET_DELIVERY_MECHANISM_AT` also has none: the only two read sites (`api/circulation/dispatcher.py` and `feed/annotator/circulation.py`) access it off a `BaseCirculationAPI`, and `EventParser` is an `XMLProcessor`. Dropping them lets `parser.py` stop importing the circulation API layer entirely. `BibliothecaAPI.SET_DELIVERY_MECHANISM_AT` is a separate attribute on a separate class and is unchanged.

**Tests mirror the new layout**, with the shared fixture moved into a `conftest.py` as `boundless/` does. `TestBibliographicCoverageProvider` subclassed `TestBibliothecaAPI` while using nothing from it, so pytest was silently re-running all 22 API tests a second time — that inheritance is gone. All 38 declared tests survive; `test_internal_formats` moves to `TestItemListParser`, the class it actually exercises.

## Motivation and Context

`src/palace/manager/integration/license/` held `bibliotheca_api.py`-style flat modules for Bibliotheca while `overdrive/`, `boundless/` and `opds/` were all packages. This is a structural cleanup only — no behavior change.

## How Has This Been Tested?

- `mypy` clean across all 1171 source files.
- All pre-commit hooks pass.
- Bibliotheca + Celery task suites: 167 passed.
- Every external test file importing Bibliotheca (`test_loan`, `test_dispatcher`, `test_collection`, `test_monitor`, `test_configuration`, `service/`, …): 522 passed.
- The two files that actually consume `SET_DELIVERY_MECHANISM_AT` (`test_dispatcher.py`, `test_library_annotator.py`) run green after the attribute removal.
- All six `bin/` scripts had their imports statically resolved against the new module paths.
- Verified mechanically that the split moved code rather than changed it: the class inventory and the member inventory (defs, decorators, constants) diff empty against the original file apart from the four intended constant substitutions. Likewise the set of declared `Class::method` test pairs diffs empty apart from the intentional `test_internal_formats` relocation.

## Checklist


jonathangreen:
The coverage issues here are preexisting, and the overall coverage doesn't change. So I think this one is ready for a review. I tagged you @dbernstein since you did the recent Bibliotheca refactoring.

tdilauro:
@jonathangreen Did you still want to wait for @dbernstein to review?

jonathangreen:
It's a pretty straight forward moving around of the files, so if you want to review it @tdilauro that would be great since @dbernstein is out. It would be nice to get it merged.

tdilauro (approved):
LGTM 🎸🤘🏽

PR #3546, opened 2026-07-09
Use monospace font for some settings in admin UI

tdilauro:
## Description

Enable monospace font rendering for form fields in the OIDC and SAML patron authentication configuration models that contain code-like content — regular expressions, filter expressions, JSON, and XML/PEM blobs.

## Motivation and Context

Fields holding structured or code-like values (regex patterns, Python expressions, XML metadata, JSON objects, private keys) are easier to read and edit in a monospace font. This sets `use_monospace_font=True` on those fields in the admin UI so their content renders as expected.

## How Has This Been Tested?

- Manual testing in local dev environment to verify that fields display with correct font family.
- No logic changes, so existing tests should continue to pass.

## Checklist

- N/A - All new and existing tests passed.


jonathangreen (approved):
`Looks good to me!`

PR #3551, opened 2026-07-10
Validate auto_update_query when a custom list is saved (PP-4506)

dbernstein:
## Description

> [!NOTE]
> Companion to #3550, but independent — the two touch disjoint files (this one is the admin controller; #3550 is the Celery task), so this targets `main` and can merge in any order. #3550 makes the entry sweep resilient to a bad query at runtime; this stops a bad query being saved in the first place.

A custom list's `auto_update_query` was only checked for JSON *serializability* before being stored — never for whether the search layer could actually parse it. `CustomListsController._create_or_update_list` now builds a `JSONQuery` from the submitted query and returns `INVALID_INPUT` (400) if it can't be parsed, passing along the parser's own explanation of what's wrong.

Also corrects the `auto_update_query` parameter annotation, which claimed `dict[str, str]` even though a query's `value` is a nested dict. That's why the existing tests were able to pass placeholders like `{"query": "foo"}` without mypy complaining.

## Motivation and Context

Follow-up to #3550, which fixed the *symptom*: a list whose stored query contained a `published` value of `2025>01>01` failed its Celery task every sweep and (before that PR) wedged the sweep lock for every library.

This PR fixes how the bad value got in. Two gaps combined:

- **Creating** a list incidentally exercised its query, via the `populate_query_pages` call on the `is_new` path — but a `QueryParseException` there surfaced as an unhandled 500, not a validation error.
- **Editing** a list never parsed the query at all. So an invalid query could only ever be introduced — and never surfaced — through an edit.

Either way the result was a list that silently stopped updating, since the entry-update task can do nothing with an unparseable query but log it and skip. Rejecting the query at save time turns a silent, indefinitely-stalled list into an immediate 400 for the librarian who fat-fingered the date.

**Behavior change worth flagging in review:** a librarian editing an existing list whose stored query is *already* invalid will now get a 400 rather than a successful save, even if they only meant to change the list's name — the admin UI resubmits `auto_update_query` unchanged. That seems right to me (it forces the broken query to be fixed), but it is a real change, and it means any already-corrupted rows in production should be corrected before this ships. (At this time as far as I know there are no corrupted rows in production - I manually fixed the one issue that alerted us to the problem.)

## How Has This Been Tested?

`tox -e py312-docker -- tests/manager/api/admin/controller/test_custom_lists.py tests/manager/core/test_customlist_queries.py tests/manager/celery/tasks/test_custom_lists.py` (64 passed).

- `test_auto_update_query_must_be_a_valid_search_query` is parametrized over four ways to be invalid — the `2025>01>01` value from production, an unknown key, an unknown operator, and a missing `query` root — and asserts the parser's reason is surfaced in the problem detail and that nothing is persisted.
- `test_auto_update_query_validated_on_edit` covers the gap specifically: an existing list rejects a bad query on edit and keeps its previous good one.
- Three existing tests had to be updated. They passed placeholder queries (`{"query": "foo"}`, `{"query": "...changed"}`) that the new validation correctly rejects. They test facets serialization, entry handling, and the repopulate-on-query-change path, so they now pass a valid query. That they needed changing at all is a decent illustration of how little the old code cared what was in this column.

## Checklist

### DIFF
--- a/src/palace/manager/api/admin/controller/custom_lists.py
+++ b/src/palace/manager/api/admin/controller/custom_lists.py
@@ -34,6 +34,7 @@ from palace.manager.core.problem_details import INVALID_INPUT, METHOD_NOT_ALLOWE
 from palace.manager.core.query.customlist import CustomListQueries
 from palace.manager.feed.acquisition import OPDSAcquisitionFeed
 from palace.manager.feed.worklist.base import WorkList
+from palace.manager.search.query import JSONQuery, QueryParseException
 from palace.manager.sqlalchemy.model.collection import Collection
 from palace.manager.sqlalchemy.model.customlist import CustomList
 from palace.manager.sqlalchemy.model.datasource import DataSource
@@ -129,7 +130,7 @@ class CustomListsController(
         deleted_entries: list[dict[str, Any]] | None = None,
         id: int | None = None,
         auto_update: bool | None = None,
-        auto_update_query: dict[str, str] | None = None,
+        auto_update_query: dict[str, Any] | None = None,
         auto_update_facets: dict[str, str] | None = None,
     ) -> ProblemDetail | Response:
         data_source = DataSource.lookup(self._db, DataSource.LIBRARY_STAFF)
@@ -167,6 +168,21 @@ class CustomListsController(
                     raise ProblemDetailException(
                         INVALID_INPUT.detailed(
                             "auto_update_query is not JSON serializable"
+                        )
+                    )
+
+                # A query that is valid JSON may still be a query our search layer
+                # cannot parse (an unknown key, a bad operator, a `published` value
+                # that isn't YYYY-MM-DD, ...). Reject it here rather than storing it:
+                # the nightly entry-update task can only log and skip such a list, so
+                # an unvalidated query means a list that silently stops updating.
+                try:
+                    # Building the query is what validates it; the result is discarded.
+                    _ = JSONQuery(auto_update_query).search_query
+                except QueryParseException as exc:
+                    raise ProblemDetailException(
+                        INVALID_INPUT.detailed(
+                            f"auto_update_query is not a valid search query: {exc.detail}"
                         )
                     )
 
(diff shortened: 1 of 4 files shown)


dbernstein:
Good catch — this was real. Fixed in 59c57e44f.

The parser wasn't type-safe against JSON-serializable-but-wrongly-typed input. Confirmed these all raised a bare `AttributeError`/`TypeError` deep in `JSONQuery`, escaping the `except QueryParseException` in the controller as a 500:

| input | was |
|---|---|
| `{"query": "foo"}` | `AttributeError: 'str' object has no attribute 'keys'` |
| `{"query": {"key": "language", "value": 5}}` | `AttributeError: 'int' object has no attribute 'lower'` |
| `{"query": {"key": ["title"], "value": "x"}}` | `TypeError: unhashable type: 'list'` |
| `{"query": {"and": "notalist"}}` | `AttributeError` (string walked char-by-char) |
| `{"query": {"key": "title", "value": 5, "op": "contains"}}` | `AttributeError: 'int' object has no attribute 'translate'` |

Rather than broaden the catch in the controller — which would risk swallowing a genuine `SQLAlchemyError` from the `data_source` transform's cache lookup and mislabeling it "invalid query" — I hardened `JSONQuery` at the four chokepoints so malformed input raises `QueryParseException` (→ 400), while infra errors still propagate. Bonus: this also protects the runtime search path and the entry-update task, not just the admin save path.

One deliberate non-change: **numeric values stay legal** for fields without a value transform. A range filter over `licensepools.availability_time` uses a float timestamp — it's exactly what the custom-list entry sweep injects — so the guards only constrain the spots that genuinely require a string (the transforms and `contains`/`regex`).

Coverage: parser-level cases added to `TestJSONQuery::test_errors` plus a `test_numeric_values_are_not_rejected` guard, and the non-string-value/-key/-object shapes added to the admin controller's `test_auto_update_query_must_be_a_valid_search_query`.

jonathangreen (approved):
Looks good to me.

PR #3553, opened 2026-07-11
Fix UniqueViolation in equivalent_identifiers_refresh cache rebuild (PP-4786)

dbernstein:
## Description

Routes all `recursiveequivalentscache` inserts in the `equivalent_identifiers_refresh` task through a shared `_insert_cache_rows` helper that uses PostgreSQL `INSERT ... ON CONFLICT (parent_identifier_id, identifier_id) DO NOTHING`, so a concurrently-committed cache row no longer aborts the task. Also collapses the per-parent deletes into a single sorted `WHERE parent_identifier_id IN (...)` statement and chunks/streams the inserts.

## Motivation and Context

JIRA: PP-4786

Production is throwing:

```
IntegrityError: duplicate key value violates unique constraint
"recursiveequivalentscache_parent_identifier_id_identifier_i_key"
DETAIL: Key (parent_identifier_id, identifier_id)=(188174, 2678719) already exists.
```

The refresh task rebuilds the cache with a delete-then-insert per parent identifier, implicitly assuming it is the only writer of the table. It is not:

1. The `Identifier` creation listener inserts `(id, id)` self-reference rows from ordinary webapp/import transactions, independent of this task.
2. The task runs under `task_acks_late=True` / `task_reject_on_worker_lost=True` and re-queues itself via `task.replace()` across thousands of batches (`replaced_task_nesting: 5348` in the failing log). A lost worker or a lapsed visibility timeout lets the broker redeliver, and a second refresh chain can run concurrently on overlapping identifier chains. The Redis `TaskLock` only guards against the *beat schedule* starting a second run — a redelivery carries the same `root_id`, so it re-acquires the same lock and slips through.

When either writer commits a conflicting row between our delete and our flush, the batched `INSERT ... RETURNING` hits the unique constraint and the whole task aborts. Since every parent's rows are recomputed from the same source equivalencies, a row slipped in under us holds the same value we were about to write, so `ON CONFLICT DO NOTHING` converges on the correct chain instead of failing.

## How Has This Been Tested?

- Added `TestInsertCacheRows` and `TestProcessIdentifierIds::test_tolerates_rows_the_delete_did_not_remove`. Confirmed both reproduce the production `IntegrityError` when the `ON CONFLICT` clause is removed, and pass with it in place.
- Full run of `test_refresh_equivalents.py`, `test_equivalents.py`, and `test_listeners.py` passes.
- `mypy` and pre-commit checks are clean.

## Checklist

### DIFF
--- a/src/palace/manager/sqlalchemy/refresh_equivalents.py
+++ b/src/palace/manager/sqlalchemy/refresh_equivalents.py
@@ -12,7 +12,11 @@ use in tests.
 
 from __future__ import annotations
 
+from collections.abc import Collection, Iterable
+from itertools import batched
+
 from sqlalchemy import and_, delete, select, union
+from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.orm import Session
 
 from palace.manager.sqlalchemy.model.identifier import (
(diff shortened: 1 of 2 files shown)


dbernstein:
### Note: why we did *not* pursue "align the broker/lock timeouts"

While reviewing this fix we considered whether raising `broker_transport_options_visibility_timeout` (currently 1h) above the equivalents `lock_timeout` (2h) — or otherwise "aligning" the two — would be a better/additional defense than making the inserts idempotent. Conclusion: **no, and it should not be resurrected later without this context.**

1. **`visibility_timeout` is a single global broker knob**, applied to every queue and task. Putting it above 2h raises it for *everything*, and the observable side effect is a system-wide regression on the failure path: on a total connection loss (the broker-side redelivery path) a dropped message stays invisible for the full new window before any worker re-picks it. Broad cost for a narrow gain.

2. **It doesn't target the redelivery path that actually causes the race.** `task_time_limit = 1800` (30 min, hard) means no single task execution can exceed 30 min, which is already under the 1h visibility window — so visibility-timeout-triggered redelivery of an in-flight hop is already structurally prevented. The concurrent chain comes from `task_reject_on_worker_lost=True` (worker OOM/time-limit-kill → immediate requeue), which is **independent of `visibility_timeout`**. Raising it does essentially nothing for this bug.

3. **"Aligning" by lowering the locks instead would break a deliberate invariant.** The 2h locks are sweep-guards, not per-hop guards; the invariant the codebase actually maintains is `lock_timeout >= task_time_limit` (see the rationale in `custom_lists.py`, `_SWEEP_LOCK_TTL`). Lowering those under the visibility window would reopen duplicate-sweep launches in importer/bibliotheca/custom_lists — a regression in unrelated tasks.

Net: `reject_on_worker_lost` redelivery — the real culprit — can't be locked or timed out of existence, which is exactly why idempotent (`ON CONFLICT DO NOTHING`) inserts are the correct primary defense. The per-false-redelivery wasted work is already bounded to a single 200-row batch, which these inserts now make harmless, so there is no timeout/config change here worth the blast radius.

jonathangreen:
Superseded by #3564, which is the same change recreated from a branch in the main repo (`bugfix/recursiveequivalentscache-unique-violation`) rather than a fork, so the full set of CI workflows runs correctly. Closing this one in favor of #3564.

I would have asked @dbernstein to do this himself, but he is out on vacation, so I'm going to handle recreating it.

PR #3554, opened 2026-07-12
Reset MARC files on settings update (PP-4637)

tdilauro:
## Description

- Adds a `marc_export_reset` Celery task that sets up the next MARC export as a first run by removing existing files (database records and S3 objects). The records are deleted in one transaction, and S3 deletion is best-effort, so a failure won't leave the reset half-applied.
- Queues the reset when a library's MARC exporter settings change in the admin interface and when a library registry pushes a new web client URL (only if the change would result in a new web client URL).
- First-run full exports now also record a full-content delta pointing at the same S3 object, so (1) consumers that only apply delta files receive the complete refreshed dataset without a second upload and (2) we avoid an extra upload and the doubled storage utilization.
- Export finalization now detects a configuration change that raced an in-flight export (records reset, settings drift, or the library no longer MARC-enabled), discards the stale upload, and immediately re-queues a fresh full export.
- Cleanup and reset only delete a shared S3 object when the last record referencing it is removed, and upload aborts are best-effort so cleanup failures cannot cascade.
- The MARC download page labels the full-content delta as "Full content as of <date>" instead of a date range starting at the full delta sentinel epoch.

## Motivation and Context

MARC record exports run periodically, but a configuration change would not trigger an update. Record updates based on those changes might not become available until long after the change was made. Further, some MARC consumers apply only delta files rather than reprocessing full exports. When a library's MARC export configuration changed (organization code, summary/genre inclusion, web client URLs), records exported before the change were never corrected for those consumers, leaving stale data in their catalogs indefinitely. Resetting the library's export state on configuration changes forces a full refresh that reaches both full-file and delta-only consumers.

[Jira PP-4637]

## How Has This Been Tested?

- All tests/checks pass locally.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/29181152088)/checks pass.

## Checklist


src/palace/manager/integration/discovery/opds_registration.py
```diff
@@ -514,6 +515,38 @@ def _process_registration_result(
...
+                    old_url=registration.web_client,
+                    new_url=web_client_url,
+                ):
+                    # The task is dispatched before this transaction commits. If
+                    # the transaction later rolls back, the reset still runs: the
+                    # library's MARC files are deleted and only regenerated by
+                    # the next scheduled export run. We accept that here because
+                    # this classmethod doesn't own the transaction, so it can't
+                    # defer dispatch until after the commit.
+                    marc_export_reset.delay(registration.library_id)
```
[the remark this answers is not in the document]
tdilauro: This is already noted in the comment and was a design decision. The harm in this case is no worse than that seen in the design of the expected behavior ()i.e., the old files go away and don't reappear until they are recreated on the next scheduled run). More are disproportionately heavy considering both the low likelihood of conditions described above arising and the low impact of the harm.

src/palace/manager/celery/tasks/marc.py
```diff
@@ -104,6 +124,126 @@ def marc_export_collection_lock(
...
+            if not delta and library.last_updated is None:
+                create(
+                    session,
+                    MarcFile,
+                    library_id=library.library_id,
+                    collection_id=collection_id,
+                    created=start_time,
+                    key=upload.context.s3_key,
+                    since=MARC_EPOCH,
+                )
```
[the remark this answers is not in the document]
tdilauro: This is exactly how the deltas are intended to work.

src/palace/manager/celery/tasks/marc.py
```diff
@@ -104,6 +124,126 @@ def marc_export_collection_lock(
...
+    }
+    drifted_libraries: list[LibraryInfo] = []
+    for library, upload in uploads.items():
+        # Recording an upload after the library's configuration changed would
+        # leave stale content behind, so discard it and let the caller re-queue
+        # a fresh full export.
+        # Note: There is a race condition window between the check here and the
+        # commit for the export.
+        current = current_infos.get(library.library_id)
+        if _export_config_changed(session, collection_id, library, current):
```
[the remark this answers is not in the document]
tdilauro: Commit b5b0d1621 should close the rest of the race condition window, unlikely as it is.

src/palace/manager/api/admin/controller/catalog_services.py
```diff
@@ -87,19 +89,50 @@ def process_post(self) -> Response | ProblemDetail:
...
 
+        # Dispatch the resets before committing: a failed enqueue then rolls the
+        # settings change back, and a retried save re-queues the reset. The
+        # reverse order could commit the settings and then lose a reset, leaving
+        # delta-only consumers with stale records indefinitely. If the commit
+        # fails after dispatch, the cost is only a spurious reset: the files are
+        # regenerated from unchanged settings by the next scheduled export run.
+        if reset_library_ids:
+            for library_id in reset_library_ids:
+                marc_export_reset.delay(library_id)
```
jonathangreen: You might want to add a `cooldown` here. I've been bitten before where, surprisingly (to me at least), the background task starts before the transaction commits, so it gets old settings.

src/palace/manager/api/admin/controller/catalog_services.py
```diff
@@ -87,19 +89,50 @@ def process_post(self) -> Response | ProblemDetail:
...
+            self._db.commit()
```
jonathangreen: Do we want to add an explicit commit here? I think you can rely on the commit done by the controller instead of adding another here.

src/palace/manager/celery/tasks/marc.py
```diff
@@ -104,6 +124,165 @@ def marc_export_collection_lock(
...
+    and compares the tuple fields order-independently.
+    """
+
+    def comparable(info: LibraryInfo) -> dict[str, Any]:
+        dumped = info.model_dump(exclude=set(_VOLATILE_LIBRARY_INFO_FIELDS))
+        for field in _ORDER_INSENSITIVE_LIBRARY_INFO_FIELDS:
+            dumped[field] = sorted(dumped[field])
+        return dumped
+
+    return comparable(snapshot) != comparable(current)
```
jonathangreen, line 138: Minor: Might want to consider using `json_canonical` util (or maybe some generalized version) for this.

jonathangreen (approved):
Looks good! A couple comments to think about, but this could be merged as is.

src/palace/manager/api/admin/controller/catalog_services.py
tdilauro: I wasn't too worried about the settings here, since this reset is just deleting the records. The reset task itself doesn't read settings, so running early wouldn't feed it stale data. But the cooldown does fix another problem. We queue the reset just before saving the settings change, so it was possible for the reset to run before the save landed. We could (if we were really unlucky) have an export that was finishing at that exact moment that could have its files survive our checks and keep content generated with the old settings available (I went back and forth with the AI reviewers about this). So, the cooldown will help ensure that the reset always runs after the settings are saved, which is what our checks assume.

src/palace/manager/api/admin/controller/catalog_services.py
tdilauro: I'm adopting the cooldown from your previous comment, so it should be safe to remove this commit.

src/palace/manager/celery/tasks/marc.py
tdilauro: Thanks for the tip! 🎉

PR #3563, opened 2026-07-15
Drop deprecated lanes.size columns (PP-4506)

jonathangreen:
## Description

Release-2 (the drop) of the online-migration sequence for removing the deprecated lane-size cache columns. Stacked on the release-1 PR #3619, which stops all reads of `size` / `size_by_entrypoint` (they were already made write-safe via `server_default` in #3481). This PR removes the columns from the `Lane` model and drops them from the schema.

**Do not merge until a release containing #3619 has shipped.** Until then the still-running previous release reads `lanes.size` and dropping the column would break it — so this is left in draft. Note that the CI `Backwards compatibility test` on this PR is *expected to fail* until then: that job runs the latest published release's suite, which still reads the columns; it will pass once a release containing #3619 becomes the baseline.

## Motivation and Context

JIRA: PP-4506

Completes the two-release removal. The read-side fix in #3619 was validated against this exact drop locally with the backwards-compatibility check's new local mode (#3618): the release-1 branch's `pytest -m db` suite passes against this dropped schema.

## How Has This Been Tested?

- `tests/migration/test_alembic_builtin_tests.py` passes, including `test_model_definitions_match_ddl` (model matches the migrated schema after the drop) and the up/down migration round-trip.
- Backwards-compatibility check (local mode) with the #3619 branch as the previous side: **3341 passed, 0 failed** against this dropped schema (vs. 20 failures before #3619).

## Checklist

### DIFF
--- a/alembic/versions/20260715_de6ae4bbf4a5_drop_deprecated_lanes_size_columns.py
+++ b/alembic/versions/20260715_de6ae4bbf4a5_drop_deprecated_lanes_size_columns.py
@@ -1,0 +1,49 @@
+"""Drop deprecated lanes.size columns
+
+Revision ID: de6ae4bbf4a5
+Revises: dbeb42224c1b
+Create Date: 2026-07-15 14:08:25.692742+00:00
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🔥

PR #3571, opened 2026-07-15
Replace textblob/nltk summary scoring with pure-Python heuristic

jonathangreen:
## Description

Removes the `textblob` dependency (and its transitive `nltk`, `joblib`, `regex`, and `tqdm` packages) by reimplementing the one place they were used: `SummaryEvaluator`, which ranks a work's candidate descriptions during presentation calculation.

The evaluator now scores each summary on how well it covers the work's *consensus vocabulary* — content words (non-stopwords, via the repo's existing `ENGLISH_STOPWORDS`) that recur across at least two of the work's summaries — instead of TextBlob noun phrases. Sentence counting uses a regex instead of the NLTK tokenizer. The other scoring terms (bad-phrase penalties, the home-grown English-bigram language penalty) are unchanged. "All else being equal, shorter is better" is now an explicit tie-breaker rather than an accident of TextBlob's part-of-speech tagging.

Also removes the now-unneeded corpus-download steps from `docker/Dockerfile.baseimage` and `tox.ini`.

## Motivation and Context

Both `textblob` and `nltk` are barely maintained, and `nltk` has an unpatched path-traversal vulnerability (`nltk.data.load()` decode-after-check via the `nltk:` URL scheme). They were the source of recurring dependency friction for a single, soft display-quality feature, so removing them outright is preferable to waiting on upstream fixes.

The noun-phrase signal being replaced only ever affected *which* description is displayed when a work has multiple, and only ~27%+ of described works have more than one candidate. The consensus-vocabulary replacement preserves the behavior that actually matters — an on-topic description still outranks generic marketing copy, which shares little vocabulary with the other descriptions — with no heavy dependencies.

## How Has This Been Tested?

- Reworked unit tests in `tests/manager/core/test_summary_evaluator.py` (7 passing), including a new case asserting generic marketing copy loses to an on-topic summary.
- The `evaluate_summary_quality` integration test (`tests/manager/sqlalchemy/model/test_edition.py`) passes against Postgres/OpenSearch via `tox -e py312-docker`.
- `mypy` is clean.

## Checklist

### DIFF
--- a/docker/Dockerfile.baseimage
+++ b/docker/Dockerfile.baseimage
@@ -74,11 +74,6 @@ COPY --chown=palace:palace packages/palace-opds/pyproject.toml /var/www/circulat
 RUN uv sync --frozen --no-dev --no-install-workspace && \
     SIMPLIFIED_ENVIRONMENT=/var/www/circulation/environment.sh && \
     echo "if [ -f $SIMPLIFIED_ENVIRONMENT ]; then source $SIMPLIFIED_ENVIRONMENT; fi" >> env/bin/activate && \
-    . env/bin/activate && \
-    python3 -m textblob.download_corpora lite && \
-    mv /root/nltk_data /usr/lib/ && \
-    find /usr/lib/nltk_data -name *.zip -delete && \
-    rm -Rf /root/.cache && \
     rm pyproject.toml && \
     rm uv.lock && \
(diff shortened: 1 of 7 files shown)


tdilauro (approved):
Welcome change! 🔥

PR #3574, opened 2026-07-16
Support alternate SAML auth flow with `force_authn` query param (PP-4792)

tdilauro:
## Description

- Adds support for an optional `force_authn` query parameter (`"true"` or `"false"`) on the SAML authentication redirect endpoint.
- When `force_authn=true` is requested, the generated SAML `AuthnRequest` sets `ForceAuthn="true"`, instructing the identity provider to re-authenticate the patron even if the patron already has an active IdP session ([SAML 2.0 Core 3.4.1](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf)).
- Invalid values are rejected before any IdP round-trip, in which case the patron is redirected back to the client with a problem detail error.

## Motivation and Context

A patron with an existing IdP session for the wrong account can get stuck at sign-in. For example, when a patron is signed into a Microsoft account from a different tenant, Microsoft may silently reuse that session and fail with an unrecoverable `AADSTS50020` error instead of prompting for credentials. This gives clients (e.g., Palace Web Catalog) an escape hatch: an alternate sign-in path can be used to force the IdP to present a fresh login prompt. `ForceAuthn` is part of the SAML 2.0 Core specification, so it is supported across the identity providers we integrate with, unlike vendor-specific alternatives.

[Jira PP-4792]

## How Has This Been Tested?

- New and updated tests.
- [CI tests](https://github.com/ThePalaceProject/circulation/actions/runs/29525392506) and checks pass.

## Checklist

### DIFF
--- a/src/palace/manager/integration/patron_auth/saml/auth.py
+++ b/src/palace/manager/integration/patron_auth/saml/auth.py
@@ -19,7 +19,7 @@ from palace.manager.integration.patron_auth.saml.configuration.service_provider 
 from palace.manager.integration.patron_auth.saml.metadata.parser import (
     SAMLSubjectParser,
 )
-from palace.manager.util.problem_detail import ProblemDetail as pd
+from palace.manager.util.problem_detail import ProblemDetail, ProblemDetail as pd
 
 if TYPE_CHECKING:
     import sqlalchemy.orm.session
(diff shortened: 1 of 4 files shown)


tdilauro:
Both Claude and greptile don't like my extra problem detail alias (great that they agree on something 🙄), but I didn't like `pd` for the type annotation and I didn't want to change the other uses and pollute the PR.

jonathangreen (approved):
I support this revolution against the robots! 🤖

PR #3618, opened 2026-08-03
Run the backwards-compatibility check locally against any git ref (PP-4506)

jonathangreen:
## Description

The `Backwards compatibility test` CI job (`docker/ci/test_backwards_compatibility.sh`) runs the previous *release's* database test suite against the schema built from the current code. Because it resolves the previous release from the latest published GitHub release, it can only catch a broken online-migration change *after* that release has shipped — a full release cycle after the mistake was written.

This adds an opt-in **local mode**: pass a git ref as the first argument (or set `PREV_RELEASE_REF`) and the script builds both images from git instead of pulling a release — the previous side from the ref's committed tree (via `git archive | docker build`), the current side from the working tree — so a "stop using" change can be validated against the eventual drop before either is released:

```
./docker/ci/test_backwards_compatibility.sh <release-1-commit-or-HEAD~1>
```

CI leaves `PREV_RELEASE_REF` unset and continues to test against the real previous release; that path is unchanged.

This also hardens the shared step-2 startup with `--wait`. `initialize_instance` runs via `compose run --no-deps`, which bypasses the `webapp` service's `depends_on` health gating, so without `--wait` it can race a still-starting OpenSearch/Postgres and fail — a latent flake in the CI job as well, not just locally.

For that `--wait` to actually gate OpenSearch, this also fixes the `os` healthcheck in `docker-compose.yml`, which was effectively a no-op. Its test command ended in `if [[ ... ]]; then echo 0; else echo 1; fi`; Docker decides health from the command's **exit code**, not its output, and an `if` block exits with its taken branch's status — both branches are `echo`, which always exits `0` — so the container reported healthy on its first check regardless of OpenSearch's state. It is now a real `curl --fail` check with a `start_period` grace for OpenSearch's boot. This healthcheck is shared by every job that uses this compose file.

The workflow is documented in `CLAUDE.md`.

## Motivation and Context

JIRA: PP-4506

While completing the lane-size column removal we hit exactly this gap: #3481 shipped in v45.1.0 but did not fully stop reading the columns, and nothing surfaced it until the #3563's backcompat gate failed — a release cycle later. This gives that signal locally in minutes. This is helpful, since there are many edge cases to take into account when dropping a column, as we are seeing.

## How Has This Been Tested?

- Ran `./docker/ci/test_backwards_compatibility.sh <ref>` locally end-to-end against the stacked lane-size work: it builds the previous side from the ref, builds the current side from the working tree, and runs the previous side's `pytest -m db` suite against the freshly-built schema.
- Exercised both directions as a control: against an incomplete "stop using" commit it fails (6 failing db tests), and against the completed one it passes (3341 passed, 0 failed).
- Confirmed the CI path is unaffected: with `PREV_RELEASE_REF` unset the script still resolves and pulls the latest release image.
- Re-verified after the `os` healthcheck fix: `os` reaches `healthy` under `compose up --wait`, and a full local run stays green (3341 passed).

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -180,6 +180,28 @@ the removal isn't forgotten later:
 Pushing both PRs up front is deliberate: it makes the follow-up cleanup visible and prevents the second step from
 being lost.
 
+**Verifying backwards compatibility.** CI runs a `Backwards compatibility test`
+(`docker/ci/test_backwards_compatibility.sh`): it builds the schema from the current code, then runs the **latest**
+**published release's** database test suite (`pytest -m db`, external-schema mode) against it. A failure there means
+the migration would break the still-running previous release. It most often means a release-1 "stop using" change
+did not actually stop using the object — e.g. a `deferred()` column still re-selected by `Query.count()` /
+`from_self()`, or a Python-side `default=` still emitted on `INSERT`.
+
+Because that gate tests against a *published release*, it can only catch such a mistake **after** release-1 has
+shipped — a full release cycle after the bug was written. To get the same signal locally in minutes, pass a **git**
+**ref** to the script; it then builds both sides from git (previous side from the ref, current side from your working
+tree) instead of pulling a release:
+
+```bash
+# Prove that <release-1 commit> still works against the schema in your working tree
+# (e.g. the stacked release-2 drop). Green here means the "stop using" change is real.
+./docker/ci/test_backwards_compatibility.sh <release-1-commit-or-HEAD~1>
+```
+
+Run this before merging any release-1 "stop using" PR: check out the drop in your working tree, point the script at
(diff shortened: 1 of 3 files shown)


jonathangreen:
> passing `--no-build` on the `webapp-prev` run would also turn any future missing-image case into a hard failure instead of a vacuous pass

I tried this, but it fails immediately with unknown flag: `--no-build` (compose v2.40.3). So I didn't apply this part of the suggestion. I believe it is unnecessary anyway because `compose run` doesn't build by default.

jonathangreen:
Claude still has some nits about this, but I'm happy with where the script is at and its ready for review.

tdilauro (approved):
Looks good! ◀️ ▶️

PR #3619, opened 2026-08-03
Stop reading deprecated lanes.size columns via Query.count() (PP-4506)

jonathangreen:
## Description

#3481 (shipped in v45.1.0) marked `lanes.size` / `size_by_entrypoint` as `deferred()` to stop reading them before a follow-up drop. That turns out to be insufficient: SQLAlchemy's `Query.count()` wraps its source in `SELECT count(*) FROM (<source>)`, and that inner subquery re-selects **every** mapped column, including deferred ones. So every `query(Lane).count()` still emitted the deprecated columns, and the previous release kept reading `lanes.size` — which is why the stacked drop PR's backwards-compatibility gate fails (#3563).

This completes the read-side "stop using" step by counting a single column instead of the whole entity everywhere `Lane` is counted:

- **Production:** `Library.has_root_lanes` and `CustomListsController.share_locally_DELETE` now count `query(Lane.id)`. These were the only two production paths that selected the deprecated columns — every other `query(Lane)` use is `.all()` / `.one()` / iteration, which `deferred()` already excludes from the SELECT.
- **Tests:** the lane-creation, admin lane delete/reset, and `affected_by_customlist` tests counted the `Lane` entity; they now follow the pattern used elsewhere and assert that `.all()` is `[]`.

After this, no code in this release — production or test — selects `lanes.size` / `size_by_entrypoint`, so the follow-up drop PR is backwards compatible.

## Motivation and Context

JIRA: PP-4506

First PR for the online-migration sequence to remove the deprecated lane-size columns; the column drop is #3563, stacked on this and held until this ships. Fixes the previous PR #3481, whose `deferred()`-only approach did not cover `Query.count()`. The columns were already made write-safe (`server_default`) in #3481.

## How Has This Been Tested?

- `mypy` clean; the full existing test suite passes.
- Ran the backwards-compatibility check locally with the new local mode (`docker/ci/test_backwards_compatibility.sh <this branch>`) against the stacked drop: the previous side (this branch) runs its full `pytest -m db` suite against the dropped schema with **0 failures** (3341 passed), where the previous, `deferred()`-only release fails 20. This new local mode was added in https://github.com/ThePalaceProject/circulation/pull/3618.
- Confirmed via the emitted SQL that `query(Lane.id).count()` no longer references the deprecated columns.

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,7 +155,10 @@ mapped column in its default `SELECT`, and eagerly emits any column that has a P
 using it, and a later drop will break the still-running previous release. To fully stop using a column in
 release 1, before dropping it in release 2:
 
-- **Reads:** mark the column `deferred()` so it is excluded from the default `SELECT`.
+- **Reads:** mark the column `deferred()` so it is excluded from the default `SELECT`. Note that
+  `deferred()` alone is not always sufficient:
+  - `Query.count()` wraps its source in `SELECT count(*) FROM (<source>)`, and that inner subquery re-selects
+    every mapped column, deferred ones included.
 - **Writes:** remove any Python-side `default=`. If the column is `NOT NULL`, removing the Python default
   alone makes inserts fail the not-null constraint (SQLAlchemy sends no value and the database has none), so
   first move the default into the database with a `server_default`. A **nullable** column with no default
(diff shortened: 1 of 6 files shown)


jonathangreen:
This is ready for review. I've fixed the PR description as noted by the claude code review bot, and the changes in https://github.com/ThePalaceProject/circulation/pull/3618 obviate the need for the CLAUDE.md changes suggested.

tdilauro (approved):
✅

PR #3621, opened 2026-08-04
Advance the search read pointer from the reindex instead of a chain (PP-4908)

jonathangreen:
## Description

Folds the search read-pointer advance into the end of `search_reindex`, and deletes both `get_migrate_search_chain()` and the `update_read_pointer` task.

A schema migration used to create the new index, point writes at it, and queue `chain(search_reindex.si(), update_read_pointer.si())`. Now `search_reindex` advances the read pointer itself once it has completed a full pass, so whichever run first fills the latest index publishes it.

To make that safe, `search_reindex` records the index it started filling (`target_index`, resolved from the write pointer at offset 0) and carries it across requeues via `signature_with`. The pointer only advances when that index is both the latest revision *and* still the write pointer at the end. If the write pointer moved partway through — a deploy landing mid-run — the documents are split across two indexes and neither received a complete pass, so the pointer is left alone.

**A running search reindex that happens during a deploy will error out due to this added kwarg**, but this is fine, its a one time thing during deploy and a new reindex will be rescheduled the next night anyway. I think this is better trade-off then dealing with the complexity of a task that can handle this kwarg or not.

- `InstanceInitializationScript.migrate_search` and `RebuildSearchIndexScript` now queue a bare `search_reindex`.
- `RebuildSearchIndexScript` loses its `--migration` flag, which no longer distinguishes anything.
- `update_read_pointer` is deleted. It was the chain's second link and had no other caller — no beat entry, no CLI — and the only thing it could still do was publish an index that no single run filled end to end, which means serving reads from an index nobody verified is complete. The repair for that is a reindex that finishes, and #3623 makes `RebuildSearchIndexScript --blocking` run a real pass in-process, in hours rather than the days a queued pass takes.
- Retries moved out of the helpers and into the task bodies. `task.retry` reschedules the whole task, so a retry raised from inside a helper reads as if only that helper is retried; the helpers now raise and each task decides what to retry. `search_reindex` retries inside its lock, so a run that is waiting to retry keeps holding it.
- The helpers take a `LoggerType` rather than a `Task`, so nothing outside a task body touches `Task`. Passing `task.log` keeps the per-task logger name that `LoggerMixin` derives from the task class.
- A pass now pages by works rather than by the documents built from them. `Work.to_search_documents` omits any work whose document it cannot build, so a full batch of works can come back as a short batch of documents — which the reindex read as the end of the works, stopping the pass there. That was survivable when a short pass just meant an incomplete index; now that a completed pass publishes the index, one work with a broken presentation edition would have published a truncated one. `get_work_search_documents` is split into `get_presentation_ready_work_ids` plus a `Work.to_search_documents` call, so the batch that decides whether there is more to do is the batch of works we asked for.
- The startup message for a read pointer behind the write pointer no longer tells operators they may need to repair the index by hand. That was right when nothing advanced the pointer on its own; now the first reindex to complete a full pass publishes it.

## Motivation and Context

JIRA: PP-4908

Three production circulation managers — `ca-california`, `ct-connecticut`, `nj-newjersey` — have been serving search from the v7 index since v45.1.0 rolled out on 2026-07-20, two weeks ago. Writes go to v8, reads come from v7, and nothing is repairing it.

A chain only reaches its second link if the first one succeeds. `search_reindex` raises `LockNotAcquired` when a reindex is already running, and on these managers a full pass takes 3-5 days against a daily `full_search_reindex` beat, so a run is almost always already in flight. At deploy time the migration chain lost that race and died at the first link:

```
2026-07-20 23:33:35 ca-california ERROR Task search.search_reindex[a8a683da…] raised unexpected:
  LockNotAcquired('Lock ca-california::TaskLock::search_reindex could not be acquired')
2026-07-20 23:36:27 nj-newjersey ERROR Task search.search_reindex[3d6005b2…] raised unexpected:
  LockNotAcquired('Lock nj-newjersey::TaskLock::search_reindex could not be acquired')
```

The chain's second link was never received for any of the three (it ran and succeeded for `ks-kansas` and `wa-washington`, whose nightly reindex had already finished and so left the lock free).

Nothing recovers from that. The nightly beat queues a bare `search_reindex` with no second link, so the reindex has since completed several times — `nj-newjersey` 07-21, `ct-connecticut` 07-23 / 07-29 / 08-02, `ca-california` 07-27 — with zero read-pointer updates in two weeks of logs. The `ca-california` and `ct-connecticut` v8 indexes are now fully populated and still unpublished.

With this change, any one of those completed runs would have finished the migration on its own.

The failure also degrades in the right direction: a run that dies partway through leaves the pointer where it was, so a stalled migration means "reads stay on the old index" rather than a half-migrated one.

## How Has This Been Tested?

Eleven new tests in `tests/manager/celery/tasks/test_search.py`, most built on a new `SearchMigrationFixture` (new revision created, write pointer moved, read pointer still on the old index). Two of them cover states nothing else reaches: a batch whose documents come back short of its works still pages to the end, and an index with no read pointer at all gets published rather than treated as current.

## Checklist

### DIFF
--- a/src/palace/manager/celery/tasks/search.py
+++ b/src/palace/manager/celery/tasks/search.py
@@ -120,6 +120,8 @@ def advance_read_pointer(task: Task, target_index: str | None) -> None:
     pass over that index publishes it. No caller has to remember to advance the pointer,
     and a run that dies partway through simply leaves the pointer where it was.
 
+    :param task: The task that completed the reindex, retried with a backoff if the
+        pointers cannot be read.
     :param target_index: The index the completed run was filling.
     """
     service = task.services.search.service()
@@ -161,7 +163,9 @@ def advance_read_pointer(task: Task, target_index: str | None) -> None:
     set_read_pointer(task, service, revision)
 
 
-@shared_task(queue=QueueNames.default, bind=True, max_retries=4)
+@shared_task(
+    queue=QueueNames.default, bind=True, max_retries=4, throws=(LockNotAcquired,)
+)
 def search_reindex(
     task: Task, offset: int = 0, batch_size: int = 500, target_index: str | None = None
 ) -> None:
--- a/src/palace/manager/celery/tasks/search.py
+++ b/src/palace/manager/celery/tasks/search.py
@@ -26,13 +26,18 @@ from palace.manager.sqlalchemy.model.work import Work
 from palace.manager.util.backoff import exponential_backoff
 
 
-def get_work_search_documents(
+def get_presentation_ready_work_ids(
     session: Session, batch_size: int, offset: int
-) -> Sequence[dict[str, Any]]:
+) -> Sequence[int]:
     """
-    Get a batch of search documents for works that are presentation ready.
+    Get a batch of ids for works that are presentation ready.
+
+    A pass over the works pages by these ids rather than by the documents built from
+    them. `Work.to_search_documents` omits any work whose document it cannot build, so a
+    full batch of works can produce a short batch of documents, and a caller that treats
+    a short batch as the end of the works would stop there.
     """
-    works = [
+    return [
         w.id
         for w in session.execute(
             select(Work.id)
@@ -42,7 +47,6 @@ def get_work_search_documents(
             .offset(offset)
         )
     ]
-    return Work.to_search_documents(session, works)
 
 
 def add_documents_to_index(
@@ -193,11 +197,12 @@ def search_reindex(
                     skip_start=True,
                 ),
             ):
-                documents = get_work_search_documents(session, batch_size, offset)
+                work_ids = get_presentation_ready_work_ids(session, batch_size, offset)
+                documents = Work.to_search_documents(session, work_ids)
 
             add_documents_to_index(task.log, index, documents)
 
-            if len(documents) == batch_size:
+            if len(work_ids) == batch_size:
                 # This task is complete, but there are more works waiting to be indexed. Requeue ourselves
                 # to process the next batch. We add a random delay to avoid hammering the search service
                 # when this task is running in parallel on multiple workers.
--- a/src/palace/manager/celery/tasks/search.py
+++ b/src/palace/manager/celery/tasks/search.py
@@ -241,11 +241,12 @@ def update_read_pointer(task: Task) -> None:
     an index that no single reindex run filled end to end, or one filled by a run that
     started before the revision it ended up writing to existed.
 
-    It has no CLI entry point, since the usual repair is another full reindex
-    (bin/repair/search_index), which publishes the index when it finishes. To publish an
-    already-filled index without reindexing, queue this task directly:
+    It has no CLI entry point, because the repair is normally just a completed reindex:
+    whichever run finishes a full pass publishes the index itself. To publish an index that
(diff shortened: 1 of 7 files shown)


jonathangreen:
Pairs with #3622, which raises the `search_reindex` lock timeout. **This one should land first** — #3622 on its own makes the deploy-time chain failure deterministic rather than ~41% likely, because a lock that no longer lapses means the incumbent nightly run always wins the race against a migration chain queued at deploy time.

src/palace/manager/celery/tasks/search.py
```diff
@@ -64,18 +66,88 @@ def add_documents_to_index(
...
+        # The write pointer moved partway through, so our documents are split across the
+        # old and new indexes and neither one received a complete pass.
+        task.log.warning(
+            f"Not advancing read pointer: the write pointer moved to "
+            f"{write_pointer.index if write_pointer is not None else None} "
+            f"while this reindex was filling {target_index}."
+        )
+        return
+
+    set_read_pointer(task, service, revision)
```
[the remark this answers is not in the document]
jonathangreen: The window is real, but I don't think it's worth closing, because losing the race leaves reads better off than declining would.

If the write pointer moves to a newer index between the check and `read_pointer_set`, we publish `target_index` — an index this run filled end to end, missing only the writes since the pointer moved seconds ago. The alternative is what happens if we decline: reads stay on the *older* index, which is missing those same recent writes plus everything since its own fill. Neither serves the newest index, so the race costs nothing relative to the guard firing, and it self-heals — the next reindex on the new deployment sees `read_pointer.version` behind the highest revision, fills the newest index, and publishes it.

Closing it properly would need a compare-and-swap on the alias update, which the aliases API doesn't offer; the alternatives (re-read and revert after setting) trade a harmless stale read for a flapping read pointer.

Leaving the guard as a best-effort check for the common case, where the pointer moved long before the run ended.

src/palace/manager/celery/tasks/search.py
```diff
@@ -44,93 +46,188 @@ def get_work_search_documents(
 
 
 def add_documents_to_index(
-    task: Task, index: ExternalSearchIndex, documents: Sequence[dict[str, Any]]
+    log: LoggerType, index: ExternalSearchIndex, documents: Sequence[dict[str, Any]]
```
jonathangreen, line 53: This is just a bit of refactoring to make `add_documents_to_index` take a `log` parameter instead of `task`, so its aligned with the rest of the helper functions defined in this PR. The `retry` logic here really shouldn't have lived in this function, as it applies to the whole task.

src/palace/manager/celery/tasks/search.py
```diff
@@ -44,93 +46,188 @@ def get_work_search_documents(
...
+                    ).set(countdown=delay)
+                )
+
+            advance_read_pointer(
+                task.log,
+                task.services.search.service(),
+                task.services.search.revision_directory().highest(),
+                target_index,
+            )
+        except (FailedToIndex, OpenSearchException) as e:
```
jonathangreen, line 222: This `except` was moved here from where it was buried in `add_documents_to_index` before.

tdilauro (approved):
This looks good! 🕵🏽

PR #3623, opened 2026-08-04
Make the blocking search rebuild actually run in this process (PP-4908)

jonathangreen:
> **Stacked on #3621** — Review that one first.

## Description

Gives `RebuildSearchIndexScript --blocking` its own batching loop, so it indexes everything in the calling process instead of handing the work to Celery.

- `--blocking` pages over `get_presentation_ready_work_ids`, building the documents and calling `add_documents_to_index` itself. Like the task, it pages by the works it asked for rather than the documents it managed to build, so a work `Work.to_search_documents` drops cannot end the pass early.
- A batch that fails to index is retried, with the attempts and the backoff the task uses. Losing a batch to a transient search failure costs the task one requeue; it would cost the rebuild everything, since there is no offset to resume from and a blip two hours in would send it back to the first work.
- It takes `search_reindex`'s own Redis lock for the duration, extending it each batch, so the scheduled reindex cannot run against the index at the same time. `TaskLock` grows an optional `task` argument to make that possible — callers outside Celery can now contend for the same key, as long as they supply `lock_name` and `redis_client`, and its two call shapes are declared as `@overload`s so that optionality doesn't cost the static check that one or the other was supplied. If an extension ever fails the rebuild stops, rather than carry on indexing and publishing beside whatever took the lock next.
- **`--delete` now waits for that lock too, on both paths.** It used to empty the index before anything took the lock, so a rebuild that then couldn't get the lock had already cleared it, and a reindex that was running had the index pulled out from under it — leaving reads on an empty index until the queued pass refilled it, which on these managers is days. `bin/repair/search_index --delete` will now fail with `LockNotAcquired` while a reindex is running, rather than clearing the index anyway. The queued path gives the lock back before queueing, since the task it queues has to take it.
- New `--force`, for when the reindex holding the lock is the days-long pass we are running the rebuild to get ahead of. `RedisLock.acquire_force` takes the key over atomically and logs what was holding it. The running reindex is not interrupted: it takes the lock once a batch, so it stops when it comes back for the next one, and until then it is indexing the same works into the same index we are. It is an error to ask for `--force` without `--blocking`, since the queued path has no lock of its own to take.
- On completing a pass it advances the read pointer by calling `advance_read_pointer`, so it publishes under exactly the conditions the task uses.
- New `--batch-size`, which a local run wants for tuning. It has to be at least 1: a batch of zero pages forever, in the loop here *and* in the task's requeue, holding the reindex lock the whole time.
- Each batch ends its transaction, so a pass that runs for hours doesn't hold one snapshot open for all of it and keep autovacuum off the tables the rest of the application is writing to. The task gets this for free by taking a session per batch; the script has one session, so it commits.

## Motivation and Context

JIRA: PP-4908

`--blocking` did not really block and run in the local process as expected. It called the task signature directly, which runs the task body in-process but `raise task.replace(...)`, still queues the next round into celery. So `--blocking` indexed the first batch locally, published the second to the broker.

That left no way to reindex a manager without going through the queue — which matters because **the task's throughput is set by queue latency, not by indexing**. Measured on `nj-newjersey`: a batch is ~4.6s of work (3.4s building documents, 1.1s indexing), but the next batch lands a median of **328 seconds** later, because the `default` queue carries a persistent ~120-deep backlog. A full pass there needs 1,343 batches.

| | per batch | full pass (nj-newjersey) |
|---|---|---|
| via the queue | ~328s | **4–5 days** |
| in process | ~6s | **~2 hours** |

This backlog is its own issue I want to fix, but I think having a script can run off-queue is also important, as a backup in case this situation arises in the future.

## How Has This Been Tested?

- Added a number of new tests.

## Checklist

### DIFF
--- a/src/palace/manager/scripts/search.py
+++ b/src/palace/manager/scripts/search.py
@@ -12,7 +12,7 @@ from palace.manager.celery.tasks.search import (
 from palace.manager.scripts.base import Script
 from palace.manager.search.external_search import ExternalSearchIndex
 from palace.manager.service.container import Services
-from palace.manager.service.redis.models.lock import TaskLock
+from palace.manager.service.redis.models.lock import LockError, TaskLock
 
 
 def _batch_size(value: str) -> int:
@@ -94,13 +94,17 @@ class RebuildSearchIndexScript(Script):
 
         `search_reindex` requeues itself once per batch, so its throughput is set by how
         long a trip through the `default` queue takes rather than by how long indexing
-        takes. On a manager whose queue carries a backlog that is minutes per batch, and
-        a full pass can run for days. Doing the same work in one process costs a couple
-        of hours instead, at the price of tying up whoever ran the script.
+        takes. On a manager whose queue carries a backlog that is minutes per batch, a
+        full pass can run for days. Doing the same work in one process costs a couple of
+        hours instead, at the price of tying up whoever ran the script.
 
         We take the task's own lock so the scheduled reindex can't run against the index
         at the same time, and extend it every batch, since nothing else will while we
         hold it.
+
+        Each batch ends its transaction, so a pass that runs for hours doesn't hold one
+        snapshot open for all of it and keep autovacuum off the tables the rest of the
+        application is writing to.
         """
         service = self.services.search.service()
         target_index = resolve_target_index(service)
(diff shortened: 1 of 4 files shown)


tdilauro (approved):
🥇

PR #3631, opened 2026-08-06
Add logging and setting overrides for active SAML ACS URL(s) (PP-4899)

tdilauro:
## Description

SAML Service Provider metadata may declare more than one callback (Assertion Consumer Service)
endpoint. Until now, the endpoint named in our authentication requests was fixed, so moving to a different one was not possible without a code or per-integration SP metadata change.

- Each SAML integration can now choose how that endpoint is selected:
  - **Lowest index** (default) — reproduces current behavior
  - **Metadata default** — honors the `isDefault` attribute in the metadata
  - **Defer to the identity provider** — name no endpoint, letting the identity provider use whichever one it has registered (not all IdPs may support this).
  - A new environment variable may override the site-wide default / fallback for integrations that do not choose one, and clearing the setting in the Admin UI returns an integration to the active site-wide default.

- Adds our own logging that identifies when a response arrives at an endpoint other than one we expected, which previously could only be diagnosed from an (external) IdPs error message.

- Improves endpoint selection, which previously ignored the metadata's own default attribute and compared endpoint indexes as text rather than as numbers.

Existing integrations are unaffected, as the default reproduces current behavior.

## Motivation and Context

Some patron auth integrations need to move to a different SAML callback endpoint, but that change has to be coordinated with each identity provider partner individually. Making the choice a per-integration setting allows endpoints to be migrated incrementally, rather than requiring every partner to change at the same time.

[Jira PP-4899]

## How Has This Been Tested?

- Manual testing of admin UI interaction in local dev environment.
- All tests / checks pass locally.

## Checklist

### DIFF
--- a/src/palace/manager/integration/patron_auth/saml/configuration/model.py
+++ b/src/palace/manager/integration/patron_auth/saml/configuration/model.py
@@ -171,6 +171,9 @@ class SAMLWebSSOAuthSettings(AuthProviderSettings, LoggerMixin):
                 SAMLACSSelectionPolicy.METADATA_DEFAULT.value: (
                     "Metadata default (isDefault, otherwise lowest index)"
                 ),
+                SAMLACSSelectionPolicy.DEFER_TO_IDP.value: (
+                    "Defer to identity provider (name no endpoint in the request)"
+                ),
             },
         ),
     ] = SAMLACSSelectionPolicy.FIRST_INDEX
--- a/src/palace/manager/integration/patron_auth/saml/configuration/model.py
+++ b/src/palace/manager/integration/patron_auth/saml/configuration/model.py
@@ -30,6 +30,7 @@ from palace.manager.integration.patron_auth.saml.configuration.service_provider 
 )
 from palace.manager.integration.patron_auth.saml.metadata.federations import incommon
 from palace.manager.integration.patron_auth.saml.metadata.model import (
+    SAMLACSSelectionPolicy,
     SAMLAttributeType,
     SAMLBinding,
     SAMLIdentityProviderMetadata,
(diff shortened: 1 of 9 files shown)


src/palace/manager/integration/patron_auth/saml/configuration/model.py
```diff
@@ -125,6 +130,17 @@ def _validate_filter_expression(cls: type[BaseSettings], v: str | None) -> str |
     return v
 
 
+# Names shown in the administrative interface. Also used in the field's description, so
+# that the stated default follows DEFAULT_ACS_SELECTION_POLICY rather than being restated.
+ACS_SELECTION_POLICY_LABELS: Final[Mapping[SAMLACSSelectionPolicy, str]] = frozendict(
```
jonathangreen, line 135: I like the type hint here! I usually annotate these as `FrozenDict`, but `Final[Mapping[` is really nice.

src/palace/manager/integration/patron_auth/saml/metadata/model.py
```diff
@@ -328,6 +328,32 @@ class SAMLBinding(Enum):
...
+    # Lowest index, ignoring isDefault. Identity providers are registered against the
+    # endpoint this picks, which is why it is the built-in fallback.
+    FIRST_INDEX = "first_index"
+
+    # The SAML rule: isDefault="true" wins, falling back to the lowest index.
+    METADATA_DEFAULT = "metadata_default"
+
+    # Name no endpoint at all, leaving the identity provider to use whichever one it
+    # has registered. A mismatch between the two sides then cannot arise.
+    DEFER_TO_IDP = "defer_to_idp"
```
jonathangreen, line 354: Nit: You could probably use `StrEnum` and `auto()` to avoid repeating these.

jonathangreen (approved):
Looks good! I didn't go through the test cases in detail, but they looked good at a high level and the non-test code looks great.

src/palace/manager/integration/patron_auth/saml/metadata/model.py
tdilauro: I was going to use `strEnum`, but noticed that just about everything else there was `enum`s. I'll got ahead and do that.

tdilauro: Actually, I don't think I want to use `auto()` here, since the values might be used by the env var override. So someone changing the member name would break that. I'll add a comment about it, though, so it's more evident.

tdilauro:
Claude's assertion in its latest round
> FIRST_INDEX is not equivalent to the pre-PR ACS selection whenever any AssertionConsumerService node carries an isDefault attribute, so the claim that "existing integrations are unaffected" holds only for SP metadata that has no isDefault anywhere. 

is simply wrong. That broken behavior was the main impetus for this PR. It is implemented as a settings feature only because simply fixing the bug would've broken any active integration that was affected by the bug, which would've been most all of them.

PR #3640, opened 2026-08-11
Pin kombu to unreleased commit to fix worker hang after Redis connection reset

jonathangreen:
## Description

Pins `kombu` to an unreleased upstream commit ([`2c8372c`](https://github.com/celery/kombu/commit/2c8372c6b35b6c61bb6cf93470c57d9eab8836b8), the merge commit of [celery/kombu#2498](https://github.com/celery/kombu/pull/2498)) via a direct dependency and a `[tool.uv.sources]` entry pointing at GitHub's commit archive tarball. Using the tarball URL rather than a git source lets uv fetch it over plain HTTPS, so no git executable is needed in the docker images. kombu still identifies as 5.6.2 at this commit, so celery's own dependency constraint remains satisfied.

This pin is a temporary bandaid: it should be removed once a kombu release containing the fix (5.7.0 milestone) is published. The comment in `pyproject.toml` documents the removal condition.

## Motivation and Context

When a celery worker's Redis broker connection is reset (e.g. by a Redis failover or restart), the worker logs `consumer: Connection to broker lost. Trying to re-establish the connection...` and then permanently stops consuming from all queues, while beat and already-running tasks continue. We hit this in production on 2026-08-11 (~10:54 UTC) on `fl-florida-scripts` and at least one other instance; the only recovery is restarting the worker.

This is [celery/celery#10205](https://github.com/celery/celery/issues/10205): kombu's redis transport timer callbacks (`maybe_restore_messages`, `maybe_check_subclient_health`) raise `ConnectionError` through the event loop while the connection is down, and each reconnect attempt registers new timers without cancelling the old ones, so stale timers accumulate and repeatedly crash the loop. The celery half of the fix ([celery/celery#10218](https://github.com/celery/celery/pull/10218)) is already in celery 5.6.3, which we run — our incident shows it is insufficient on its own. The kombu half ([celery/kombu#2498](https://github.com/celery/kombu/pull/2498)) is merged but unreleased.

Opened as a **draft**: we may prefer to wait for the kombu release. This PR is ready to merge if the hang recurs before that release ships.

## How Has This Been Tested?

- Verified the installed kombu at the pinned commit contains both halves of the fix: the `connection_errors` guards inside `maybe_restore_messages` / `maybe_check_subclient_health`, and the timer references (`_restore_messages_tref` / `_subclient_health_tref`) that are cancelled on reconnect.
- `tox -e py312-docker -- tests/manager/celery/ tests/manager/service/celery/` — 445 passed.
- Built the `common` stage of `docker/Dockerfile` locally (both `uv sync` steps succeed with no git in the image) and confirmed the installed kombu inside the image contains the fix.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,12 @@ dependencies = [
     "joserfc>=1.7.4,<2",
     "jsonschema>=4.22.0,<5",
     "jwcrypto>=1.5.8,<2",
+    # kombu is a dependency of celery. We pin it to an unreleased commit here, because we
+    # need the fix from https://github.com/celery/kombu/pull/2498, so that celery workers
+    # recover when their redis broker connection is reset. This pin should be removed
+    # (see the git source in [tool.uv.sources]) once a kombu release containing that fix
+    # is published and picked up by our celery dependency.
+    "kombu",
     "levenshtein>=0.27,<0.28",
     "lxml[html-clean]==6.1.1",
     "nameparser>=2.0.0,<3",  # nameparser is for author name manipulations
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🤕

